PR #8, opened 2015-06-10
avocado-vt: Fail to load plugins on uninitialized virt-test

lmr:
Solves #7

This avoids the error:

list available virt-tests.


lmr:
Ok, closing in favor of #9.

PR #58, opened 2015-08-06
Ensure bootstrap always updates the local backend directory

### LINKED ISSUE #56: problem in case of avocado vt-bootstrap --vt-type libvirt
Hello,
I installed avocado/avocado-vt according to the manual described here http://avocado-vt.readthedocs.org/en/latest/GetStartedGuide.html and wanted to try it with libvirt.
First I performed the bootstrap step as it's described: avocado vt-bootstrap --vt-type libvirt  
Then checked with avocado plugins -> no errors.
But when I wanted to list the tests I got the following error:
avocado list --verbose
** Message: pygobject_register_sinkfunc is deprecated (GstObject)
Test discovery plugin <avocado.core.plugins.virt_test.VirtTestLoader object at 0x7f1f3f8c5590> failed: 'include base.cfg' (/usr/share/avocado/data/avocado-vt/backends/qemu/cfg/tests-shared.cfg:13): file does not exist or it's not a regular file

looks like the config file is taken from the qemu dir and not from the libvirt one (and there is no base.cfg in qemu subdir, but there is one in libvirt subdir)

ls -la /usr/share/avocado/data/avocado-vt/backends/qemu/cfg/
total 16
drwxr-xr-x. 2 root root 4096 Aug  5 15:20 .
drwxr-xr-x. 3 root root 4096 Aug  5 15:03 ..
-rw-r--r--. 1 root root  373 Aug  4 03:58 tests.cfg
-rw-r--r--. 1 root root 1312 Aug  4 03:58 tests-shared.cfg

ls -la /usr/share/avocado/data/avocado-vt/backends/libvirt/cfg/
total 1304
drwxr-xr-x. 2 root root   4096 Aug  5 15:20 .
drwxr-xr-x. 3 root root   4096 Aug  5 15:03 ..
-rw-r--r--. 1 root root  17202 Aug  5 15:04 base.cfg
-rw-r--r--. 1 root root    526 Aug  5 15:04 cdkeys.cfg
-rw-r--r--. 1 root root   8472 Aug  4 03:58 default_tests
-rw-r--r--. 1 root root    329 Aug  4 03:58 drivers.cfg
-rw-r--r--. 1 root root  13303 Aug  5 15:04 guest-hw.cfg
-rw-r--r--. 1 root root 295070 Aug  5 15:52 guest-os.cfg
-rw-r--r--. 1 root root   1008 Aug  5 15:04 machines.cfg
-rw-r--r--. 1 root root 947043 Aug  5 15:52 subtests.cfg
-rw-r--r--. 1 root root    688 Aug  4 03:58 tests.cfg
-rw-r--r--. 1 root root    989 Aug  4 03:58 tests-shared.cfg
-rw-r--r--. 1 root root   7595 Aug  5 15:04 virtio-win.cfg

Looks like I don't do anything wrong in the steps above, so probably it's an issue related to libvirt type in avocado?

Thank you

lmr:
Fixes #56
Basically, in some conditions [1], the base backend dir and the local backend dir might be out of sync and some files might be missing, causing the plugins to fail. Let's play safe and use rsync to synchronize the backend directories and make sure all files are there after a vt-bootstrap.

[1] For example, user interrupts the bootstrap and starts it again

### DIFF
--- a/avocado-plugins-vt.spec
+++ b/avocado-plugins-vt.spec
@@ -8,7 +8,7 @@ URL: http://avocado-framework.readthedocs.org/
 Source: avocado-plugins-vt-%{version}.tar.gz
 BuildRequires: python2-devel
 BuildArch: noarch
-Requires: python, avocado, autotest-framework, p7zip, tcpdump, iproute, iputils, gcc, glibc-headers, python-devel, nc
+Requires: python, avocado, autotest-framework, p7zip, tcpdump, iproute, iputils, gcc, glibc-headers, python-devel, nc, rsync
 
 %description
 Avocado Virt Test is a plugin that lets you execute virt-tests
--- a/virttest/bootstrap.py
+++ b/virttest/bootstrap.py
@@ -738,6 +738,13 @@ def bootstrap(options, interactive=False):
             logging.debug("Dir %s exists, not creating",
                           sub_dir_path)
 
+    base_backend_dir = data_dir.get_base_backend_dir()
+    local_backend_dir = data_dir.get_local_backend_dir()
+    logging.info("Syncing backend dirs %s -> %s", base_backend_dir,
+                 local_backend_dir)
+    process.run('rsync -rvv %s/* %s' %
+                (base_backend_dir, local_backend_dir), shell=True)
+
     test_dir = data_dir.get_backend_dir(options.vt_type)
     if options.vt_type == 'libvirt':
         step = create_config_files(test_dir, shared_dir, interactive, step,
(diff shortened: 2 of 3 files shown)


clebergnu:
Couldn't we create a lock file that signals that the bootstrap process has
started can only remove it if the bootstrap process finishes successfully?
If, in a second run, after an interrupted execution, we find the lock file,
we instruct the user to clean up their env and start again. This would
avoid the added dependency on rsync.

lmr:
@clebergnu, I suppose we could. I'd prefer to use some python implementation that does roughly what rsync does, though -> It has the added benefit of keeping the user files up to date upon package updates, for example.

lmr:
@clebergnu There might be conditions that are hit by other situations, not just the 'user hits Ctrl+C at the wrong moment' though.

lmr:
I'm closing this in favor of #67

PR #71, opened 2015-08-11
Specify backend's cartesian configs in tp-ini file

Andrei-Stepanov:
Up to this time, avocado-vt repo should hold cartesian configs for each known
backend at backends/<backend>/cfg/*.cfg

Each modification to above files should become a pull-req to avocado-vt tree.
This commit allows store cartesian configs at test-provider git-repo. As a
result escape pull-reqs.

Cartesian files specified at tp's ini will be copied to backends/<backend>/cfg/
at `avocado vt-bootstrap' stage.

It is possible to specify few cartesian configs using next technique:

[spice]
subdir: spice/
configs: tests.cfg, tests-shared.cfg, spice-vm-install.cfg, tests-spice.cfg

### DIFF
--- a/virttest/asset.py
+++ b/virttest/asset.py
@@ -264,6 +264,14 @@ def get_test_provider_info(provider):
 
     for backend in get_known_backends():
         subdir = provider_cfg.get(backend, 'subdir')
+        cart_cfgs = provider_cfg.get(backend, 'configs')
+
+        backend_dic = {}
+        if cart_cfgs is not None:
+            # Give ability to specify few required configs separated with comma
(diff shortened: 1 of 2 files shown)


lmr:
I'm going to explain why I'm not applying this patch for another couple of weeks:

1) It needs a second reviewer. @ldoktor, if you could help out before KVM Forum, I'd appreciate it.
2) This is a change that is more involved and we have LinuxCon and KVM Forum next week. Our goal is to have a bugfix release only by tomorrow.

We appreciate your understanding of the issue, and we'll do the possible to get back to it after we are done with next week's tasks. We'll be get back to it as soon as we can.

PR #86, opened 2015-09-03
Remove virttest/utils_spice.py

Andrei-Stepanov:
This file is used by spice tests
Spice tests residence is tp-qemu
No one runs spice test using avocado-vt
So, this file can be safely removed from avocado-vt
As well as, removing this file allows avoid conflicts for developing
spice tests for avocado-vt


lmr:
Sure.

PR #88, opened 2015-09-03
Print test name for spice tests as sub-test variant

Andrei-Stepanov:
Currently you suppose that 1 test == 1 entry at subtests.cfg

SpiceQE team uses one script wich can take many options at input.
Actual test name will be quite different from script name.
Test name and it meaning can be generated on second step,
depending on actual run requests.
I ask to take test-name from from different config.


Andrei-Stepanov:
You take listed `test name' and put it as `only' filter to cartesian config input.
Now, SpiceQE tests can take huge amount of input values.
This input values define actual test behavior.
Now you print test name as a part subtests.cfg
But for us (SpiceQE) tests name is combination of variants that are defined at another place.
Basic test set persist the same. But actual test set can grow in time, even the base scripts are the same.

lmr:
Oh well, OK.

PR #89, opened 2015-09-03
print implicit filters for `list' cmd

Andrei-Stepanov:
We have main config file:

backends/qemu/cfg/tests.cfg

Command print tons of possible tests:
./virttest/cartesian_config.py
/var/lib/avocado/data/avocado-vt/backends/qemu/cfg/tests.cfg (1*)

It would be fine to know which filters were applied to 1* to get actual
tests list (avocado list --vt-type qemu)


lmr:
Same syntax problem. I'm not taking this one too.

PR #90, opened 2015-09-03
Fixes for join, suffix operators

Andrei-Stepanov:
`join': process correct dic name when merge two copies of the same sub-tree
`suffix': drop all suffixes if same variable with different suffuxes have same
value
Improve performance for `join'


lmr:
Please fix the check errors. And also the typos I did mention in the first review.

PR #114, opened 2015-09-11
utils_v2v: Add windows VM init function

will-Do:
Move this function from test case as it's more like a common function,
so other cases which need boot windows guest can reuse it.


will-Do:
Corresponding tp-libvirt patch: autotest/tp-libvirt#639

lmr:
/home/travis/build/avocado-framework/avocado-vt/virttest/utils_v2v.py:576:5: E303 too many blank lines (2)

Please fix the problem, thanks!

lmr:
Please fix the typos mentioned and re-submit, we should be good from there, thanks!

will-Do:
@lmr 
Thanks, update these sentences according to your comments.

PR #118, opened 2015-09-12
utils_misc: Deprecate find_command to avoid duplicate

Hao-Liu:
The find_command in utils_misc is already implemented in
avocado.utils.path.

This command deprecate this function and show a warning logging
if some command still using it.
We can't change the existing reference of this function in
virt-test providers, since it will fail if someone still using
virt-test.

Also some reference to path.command is changed to path.find_command
since the command function no longer exists.

### DIFF
--- a/avocado/core/plugins/vt.py
+++ b/avocado/core/plugins/vt.py
@@ -31,6 +31,7 @@ from avocado.core import multiplexer
 from avocado.core import test
 from avocado.core.settings import settings
 from avocado.core.plugins import plugin
+from avocado.utils import path as utils_path
 from avocado.utils import stacktrace
 from avocado.utils import genio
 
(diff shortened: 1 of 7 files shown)


avocado/core/plugins/vt.py
```diff
@@ -31,6 +31,7 @@
 from avocado.core import test
 from avocado.core.settings import settings
 from avocado.core.plugins import plugin
+from avocado.utils import path
 from avocado.utils import stacktrace
```
PandaWei: It's better to keep consistent with the following codes and import it as utils_path.

Hao-Liu: Changed. Thanks :+1:

PandaWei:
ACK

will-Do:
ACK

PandaWei:
Thanks @Hao-Liu  @will-Do 
Merge now

PR #127, opened 2015-09-15
virttest/qemu_qtree.py: Get the disks error log for debugging

spcui:

ID: 1231011


spcui:
@CongSmile 
like in multi_disk.py, will invoke it.

lmr:
I didn't understand. Please give more details.

spcui:
@lmr 
Sometimes, there are many error logs, and cannot easy find the log that we need, so we want to return the  logs and print them in related place.

PR #140, opened 2015-09-17
virttest/qemu_qtree.py: Update guess_type() for QtreeDev

spcui:
Make the get method can get the correct QtreeDisk device
on different platform.

id: 1211151

### DIFF
--- a/virttest/qemu_qtree.py
+++ b/virttest/qemu_qtree.py
@@ -163,11 +163,8 @@ class QtreeDev(QtreeNode):
         super(QtreeDev, self).add_child(child)
 
     def guess_type(self):
-        if self.qtree['type'] == 'virtio-blk-device':
-            return QtreeDisk
-        elif ('drive' in self.qtree and
-              self.qtree['type'] != 'usb-storage' and
-              self.qtree['type'] != 'virtio-blk-pci'):
+        if ('drive' in self.qtree and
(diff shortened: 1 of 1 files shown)


ypu:
This update makes the function guess_type meaningless. So please check the qemu code carefully and find another solution for cover all situations we met.

virttest/qemu_qtree.py
```diff
@@ -163,15 +163,7 @@ def add_child(self, child):
         super(QtreeDev, self).add_child(child)
 
     def guess_type(self):
-        if self.qtree['type'] == 'virtio-blk-device':
-            return QtreeDisk
-        elif ('drive' in self.qtree and
-              self.qtree['type'] != 'usb-storage' and
-              self.qtree['type'] != 'virtio-blk-pci'):
```
xutian: @spcui,

I think this logical looks more clearly,  expect for usb-strorage, other devices if it has backend driver, it should be a disk.

if 'driver' in self.qtree and self.qtree['type'] != 'usb-storage':
    return QtreeDisk
else:
    return QtreeDev

spcui: Yes, agree your comment, will update it.

spcui:
@xutian updated, help review again, thx?

suqinhuang:
Acked-by: Suqin Huang

xutian:
looks ok to me.

PR #145, opened 2015-09-18
virttest/libvirt_xml/vm_xml: refactor vm_rename()

PandaWei:
Main modification as following,
* Polish the docstring and comments
* Keep consistent for VM status
* No catch CmdError as same as other calls for define
* If define or undefine failed,
  No change for original VM and raise LibvirtXMLError
* Update name and uuid if rename succeeded
* Check UUID if it's expected

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -8,6 +8,7 @@ import logging
 from avocado.utils import process
 
 from .. import xml_utils
+from .. import utils_misc
 from ..libvirt_xml import base, accessors, xcepts
 from ..libvirt_xml.devices import librarian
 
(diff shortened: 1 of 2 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -613,41 +613,44 @@ def vm_rename(vm, new_name, uuid=None, virsh_instance=base.virsh):
         :param vm: VM class type instance
         :param new_name: new name of vm
         :param uuid: new_vm's uuid, if None libvirt will generate.
-        :return: a new VM instance
+        :return: a new VM instance or raise LibvirtXMLError
         """
+        if vm.name == new_name:
```
Hao-Liu: What if the uuid is specified but the new vm name is not changed? Do you think it'll be more reasonable to redefine the vm using new uuid?

virttest/libvirt_xml/vm_xml.py
```diff
@@ -613,41 +613,44 @@ def vm_rename(vm, new_name, uuid=None, virsh_instance=base.virsh):
...
         if vm.is_alive():
             vm.destroy(gracefully=True)
-        vmxml = VMXML.new_from_dumpxml(vm_name=vm.name,
-                                       virsh_instance=virsh_instance)
+            start_vm = True
+        vmxml = VMXML.new_from_dumpxml(vm.name, virsh_instance=virsh_instance)
         backup = vmxml.copy()
-        # can't do in-place rename, must operate on XML
+
+        def _cleanup(details=""):
```
Hao-Liu: I'd prefer not putting helper function in the middle of a code block. This will interrupt code reading. But it's just a style nitpick. You can ignore it if you insist.

virttest/libvirt_xml/vm_xml.py
```diff
@@ -613,41 +613,44 @@ def vm_rename(vm, new_name, uuid=None, virsh_instance=base.virsh):
...
+            backup.define()
+            if start_vm:
+                vm.start()
+            raise xcepts.LibvirtXMLError("%s" % details)
```
Hao-Liu: Exception messages will be automatically converted to string. So `raise xcepts.LibvirtXMLError(details)` is safe.

virttest/libvirt_xml/vm_xml.py
```diff
@@ -613,41 +613,44 @@ def vm_rename(vm, new_name, uuid=None, virsh_instance=base.virsh):
...
         if uuid is None:
-            # invalidate uuid so libvirt will regenerate
+            # UUID will be regenerated automatically
             del vmxml.uuid
-            vm.uuid = None
         else:
             vmxml.uuid = uuid
-            vm.uuid = uuid
-        # Re-define XML to libvirt
-        logging.debug("Rename %s to %s.", vm.name, new_name)
```
Hao-Liu: I don't think we should remove this logging message.

virttest/libvirt_xml/vm_xml.py
```diff
@@ -613,41 +613,44 @@ def vm_rename(vm, new_name, uuid=None, virsh_instance=base.virsh):
...
-            raise xcepts.LibvirtXMLError(error_msg + "%s" % detail)
-        # Keep names uniform
+        if not vmxml.define():
+            _cleanup(details="Define VM %s failed" % new_name)
+        # Update the name and uuid property for VM object
         vm.name = new_name
+        vm.uuid = VMXML.new_from_dumpxml(new_name,
+                                         virsh_instance=virsh_instance).uuid
+        if uuid is not None and vm.uuid != uuid:
+            _cleanup(details="UUID %s is not excepted %s" % (vm.uuid, uuid))
```
Hao-Liu: Is it right to fail out if the UUID is not exactly same? libvirt supports a lot of different formats of UUID. with or without hyphen(`-`), lower or upper case. I think we can't check the validity just by string comparison.

Hao-Liu: s/excepted/expected

virttest/libvirt_xml/vm_xml.py
PandaWei: Thanks for your comment.
But Just alter UUID without changing VM name, which will leads error.
[root@localhost avocado-vt]# virsh edit win7
error: operation failed: domain 'win7' already exists with uuid 8f01e012-015c-4622-993b-e5df6fa1f7b7
Failed. Try again? [y,n,f,?]:

virttest/libvirt_xml/vm_xml.py
PandaWei: Yes, Indeed helper function in the middle leads bad influence on reading.
But if we don't define this helper function,
It will need three copy of the helper function when operationes fail.

PandaWei: I will update this PR as your other comments later

virttest/libvirt_xml/vm_xml.py
Hao-Liu: But we are not changing VM xml using `virsh edit`, do we? :)

virttest/libvirt_xml/vm_xml.py
Hao-Liu: Sorry for my vague description. I didn't mean to remove helper functions. I love them. I just prefer not putting it in the __middle__ of the code block.

virttest/libvirt_xml/vm_xml.py
PandaWei: Yes, you are right.

virttest/libvirt_xml/vm_xml.py
PandaWei: I will move it to the beginning.

PandaWei:
Updated

Hao-Liu:
ACK. Thanks.

PR #149, opened 2015-09-21
Lookup for unattended_file at test provider dir

Andrei-Stepanov:
General answer files locate at: avocado-vt/shared/unattended

SpiceQA asked few times to add special answer files

It would be fine to have ability keep such files in test-provider repo
And after this patch
shared/unattended/RHEL-6-spice.ks
shared/unattended/RHEL-7-spice.ks
can be removed from avocado-vt


lmr:
Thanks @Andrei-Stepanov !

PR #154, opened 2015-09-24
libvirt_vm: check if OS is suported before unattended install

PandaWei:


PandaWei:
Update.
Log supported os variant.

lmr:
Looks good.

PR #158, opened 2015-09-24
Path to Windows ISOs should have same prefix

Andrei-Stepanov:
Most Windows ISO images have prefix: isos/windows.
Only 3 fixed above ISOs had strange prefix.
It would be fine to have all ISOs at the same dir.

It is necessary to notify: Xiaoqing Wei <xwei@redhat.com>
He was a author who introduced the strange prefix.

### DIFF
--- a/shared/cfg/guest-os/Windows/Win2012/x86_64.cfg
+++ b/shared/cfg/guest-os/Windows/Win2012/x86_64.cfg
@@ -26,7 +26,7 @@
         - r2:
             image_name += r2
             unattended_install.cdrom, whql.support_vm_install, svirt_install:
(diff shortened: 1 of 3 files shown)


lmr:
@xiaoqing-wei, consider yourself notified. I understand you've contributed the files with these prefixes due to your particular internal repo organization. We're changing it to make it uniform in upstream files. Thanks.

xiaoqing-wei:
@Andrei-Stepanov Thanks for the patch, the path is strange to me as well. I was following the internal file path. just like @lmr 's comment.
Thanks for the explain, @lmr

PR #160, opened 2015-09-25
add grub_file parameter for SLES and Ubuntu

PyLearner:
add grub_file parameter for SLES and ubuntu,
ID:1122759


lmr:
Looks reasonable, merged.

PR #164, opened 2015-09-29
Miscellaneous fixes 2015-09-29

lmr:
Fix a bunch of issues reported by @ehabkost.


avocado/core/plugins/vt_bootstrap.py
```diff
@@ -102,5 +104,9 @@ def run(self, args):
         args.show_job_log = False
         args.test_lister = True
 
-        bootstrap.bootstrap(options=args, interactive=True)
-        sys.exit(0)
+        try:
+            bootstrap.bootstrap(options=args, interactive=True)
+            sys.exit(0)
+        except (KeyboardInterrupt, process.CmdError):
```
ehabkost: What if CmdError is raised for other reasons, not because of user interruption?

virttest/asset.py
```diff
@@ -306,20 +307,27 @@ def download_test_provider(provider, update=False):
...
-        os.chdir(download_dst)
-        process.system('git log -1')
-        os.chdir(original_dir)
+            process.system('git log -1')
+            os.chdir(original_dir)
+        except (KeyboardInterrupt, process.CmdError):
+            logging.error('Cleaning up provider %s download dir %s', provider,
+                          download_dst)
+            if os.path.isdir(download_dst):
+                shutil.rmtree(download_dst)
```
ehabkost: I have important data in my test provider download dir, because that's where I develop new test code. I would be very upset if you deleted it because something went wrong while running `git log -1`.

IMO the most important thing to do here is to explain which git repository looks corrupted (because `git log -1` failed), and let the user delete it if they think it is OK. Later we may have very specific code that tries to recover from an interrupted git-clone, but I would never try to recover by deleting data if the user hasn't asked for it explicitly.

ehabkost: Maybe this can be solved easily by only deleting the directory if `repo_downloaded` is False. This way we will clean up only a directory we know we have just created, not directories created by the user.

lmr: OK, I see.

PR #169, opened 2015-09-30
git clone + user interruption safety

ehabkost:
This makes the git clone and bootstrap code more robust against corrupted git repositories and user interruption (issues I commented about on pull request #164). Most importantly, it won't try to delete the test provider directory unless we know it didn't exist yet and was just created by vt-boostrap.


lmr:
This tested well, and the code looks good. I'm merging it.

PR #188, opened 2015-10-14
virttest/utils_test/qemu.py: Add random time before migration

spcui:
Sometimes, we need wait random time before migration.

ID: 1256109

### DIFF
--- a/virttest/utils_test/qemu.py
+++ b/virttest/utils_test/qemu.py
@@ -24,6 +24,7 @@ import os
 import socket
 import threading
 import time
+import random
 
 from avocado.core import exceptions
 from avocado.utils import crypto
@@ -813,6 +814,8 @@ class MultihostMigration(object):
                         for vm in mig_data.vms:
                             vm.resume()
                         wait = self.params.get("start_migration_timeout", 0)
+                        if wait == "random":
+                            wait = random.randint(5, 20)
                         logging.debug("Wait for migration %s seconds." %
                                       (wait))
                         time.sleep(int(wait))


virttest/utils_test/qemu.py
```diff
@@ -813,6 +814,8 @@ def migrate_wrap(vms_name, srchost, dsthost, start_work=None,
                         for vm in mig_data.vms:
                             vm.resume()
                         wait = self.params.get("start_migration_timeout", 0)
+                        if wait == "random":
```
CongSmile: Why not if self.params.get() == "***":  ?

CongSmile:
Acked.

PandaWei:
sometimes?
which case we need to wait *random* time?
No 'random' is found in cfg file.
can you explain it clearly？

spcui:
Hi @PandaWei 
https://github.com/autotest/tp-qemu/pull/450/files this one

vipmike007:
Acked-By: Mike Cao <bcao@redhat.com>

PandaWei:
@spcui 
Thanks for your response firstly.
But I doubt the rationality of the range from 5 to 20.
And It's bad idea that you set the range value to be fixed value in API function.


But i think it's a bad idea add random time here.
avocado-vt,

PandaWei:
Maybe you can make range as a variable in cfg file
- timeout_random:
  start_migration_timeout = random
  min_boundary = XXX
  max_boundary = YYY

spcui:
Hi @PandaWei 
OK, thanks your comments, will update later.

spcui:
Close it, will set random time in config file.

PR #199, opened 2015-10-16
virttest.gluster: echo the confirm infomation for gluster volume stop/delete command.

rbian:
gluster volume stop/delete command promote user to confirm the
operation, the code should handle stdin and input the confirmation
string. Otherwise, the command will timeout.


rbian:
The patch seems can not work, I will submit another later.

rbian:
It has passed my test.

PandaWei:
Looks good

will-Do:
ACK and merge this pr.

PR #200, opened 2015-10-19
virttest: Add 'shell' parameter if using process.run function.

rbian:
process.run use subprocess.Popen to run a command, the command
string will be splitted to a list if shell parameter is False.
If command string includes some special characters such as
'|, &&, ||', the command wouldn't be executed correctly. It need
to set shell to 'True' in parameter for this kind of commands.


lmr:
Excellent.

PR #210, opened 2015-10-23
Fix process run failure

kylazhang:
1. Add missing parameter for ConnPrivKeyError
2. Avoid commands split by process.run as some have special characters


kylazhang:
Add comments and as part of patch same with https://github.com/avocado-framework/avocado-vt/pull/200 so remove it.

PR #213, opened 2015-10-23
virttest.utils_test.libvirt: update vm disk source functions

kylazhang:


virttest/utils_test/libvirt.py
```diff
@@ -2589,29 +2591,33 @@ def update_vm_disk_source(vm_name, disk_source_path, source_type="file"):
...
     disks = devices[disk_index]
-    disk_source = disks.source.get_attrs().get(source_type)
-    logging.debug("The disk source file of the VM: %s", disk_source)
-    if not os.path.exists(disk_source):
-        logging.error("The disk source doesn't exist!!")
-        return False
+    # Generate a disk image name if it doesn't exist
+    if not disk_image_name:
+        disk_source = disks.source.get_attrs().get(source_type)
+        logging.debug("The disk source file of the VM: %s", disk_source)
```
PandaWei: Why remove the check for disk_source.
I think it's better to reserve it.

virttest/utils_test/libvirt.py
```diff
@@ -2580,7 +2580,9 @@ def do_migration(vm_name, uri, extra, auth_pwd, auth_user="root",
         return False
 
 
-def update_vm_disk_source(vm_name, disk_source_path, source_type="file"):
+def update_vm_disk_source(vm_name, disk_source_path,
+                          disk_image_name="",
+                          source_type="file"):
```
PandaWei: Where's the caller of update_vm_disk_source?
Due to you add new param, It's better to check if it will affect all of the callers.

will-Do: Seems no caller exist, so it's safe to add this param

virttest/utils_test/libvirt.py
will-Do: @PandaWei 
Yes, check the disk exist is useful. But as the author is taking a long time leave, I think we can merge this patch first then add the check part if you want. What do you think?

PandaWei: I'm OK for your proposal.

will-Do:
Okay then, I will merge this PR.

PR #218, opened 2015-10-26
add copy_from_keys support 

xutian:


xutian:
sometime, we don't need whole test params dict,  we only need some of them if them exists. This commit provider function to copy a sub-dict by keys from origin test params dict.

Thanks,
Xu

PR #222, opened 2015-10-29
virttest/libvirt_vm: Fix unused prompt for ShellSession

PandaWei:
The prompt variable isn't used by aexpect.ShellSession.
In fact,
Different distros have declared different shell_prompt
in ./shared/cfg/guest-os/Linux/*.

so it's better pass it to aexpect.ShellSession.


will-Do:
LGTM

PR #223, opened 2015-10-30
virttest/ovs_utils: Do not distinguish between IPv4 and IPv6

PandaWei:
This commit refers to the codes of utils_net.ping.


will-Do:
LGTM

PandaWei:
Thanks @will-Do

PR #225, opened 2015-10-30
remove parameter "verbose" as it may access RemoteRunner

kylazhang:
On RemoveRunner.run there is no parameter named "verbose"


lmr:
Alternatively we could add the missing `verbose` param to the `RemoteRunner` method.

kylazhang:
Fix the commits comments error, thanks

lmr:
Merged, thanks!

PR #238, opened 2015-11-06
utils_test/libvirt: cleanup for iothread option

PandaWei:
The 1b86901 added iothread option to create_disk_xml, but not used.
later, the fb6704d introduced iothread option for disk paramters
in set_vm_disk.

In fact, set_vm_disk calls create_disk_xml and pass the iothread
option to it by disk paramters.

The code in create_disk_xml becomes meaningless and redundant,
so remove it.


will-Do:
Yes, it is. ACK

lmr:
Merged, thanks @PandaWei !

PR #240, opened 2015-11-09
avocado-vt: fix static content in cd1

vipmike007:
cd1 always been the installation iso in current setup
This patch is removing it for installation case 
ID: 1279315
Signed-off-by: Mike Cao <bcao@redhat.com>


CongSmile:

xutian:
Acked. thanks @vipmike007  and @CongSmile

PR #242, opened 2015-11-09
utils_misc: Add new function to get service status in VM

PandaWei:
Extract get_guest_service_status from tp-qemu as API function,
which is called by both sr_iov_irqbalance.py and
migration_with_dst_problem.py.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -851,6 +851,44 @@ def archive_as_tarball(source_dir, dest_dir, tarball_name=None,
     tarball.close()
 
 
+def get_guest_service_status(session, service, service_former=None):
+    """
+    Get service's status in guest.
+    Return 'active' for 'running' and 'active' case.
+    Return 'inactive' for the other cases.
+
+    :param session: An Expect or ShellSession instance to operate on
+    :type devices: Expect or ShellSession class
+    :param service: Service name that we want to check status.
+    :type service: String
+    :param service_former: Service's former name
+    :type service: String
+    :return: 'active' or 'inactive'
+    :rtype: String
+    """
+    output = ""
+    cmd = ("service %(key)s status || "
+           "systemctl %(key)s.service status || "
+           "status %(key)s")
+    try:
+        output = session.cmd_output(cmd % {'key': service})
+    except Exception:
+        if not service_former:
+            raise exceptions.TestError("Fail to get %s status.\n%s" %
+                                       (service, output))
+        output = session.cmd_output(cmd % {'key': service_former})
+
+    status = "inactive"
+    if (re.search(r"Loaded: loaded", output, re.M) and
+            output.count("Active: active") > 0):
+        status = "active"
+    elif (re.search(r"running", output.lower(), re.M) and
+            not re.search(r"not running", output.lower(), re.M)):
+        status = "active"
+
+    return status
+
+
 def parallel(targets):
     """
     Run multiple functions in parallel.


virttest/utils_misc.py
```diff
@@ -851,6 +851,43 @@ def archive_as_tarball(source_dir, dest_dir, tarball_name=None,
...
+        session.cmd("systemctl --version")
+        cmd = "systemctl status %s.service" % service
+        output = session.cmd_output(cmd)
+    except Exception:
+        if service_former is not None:
+            cmd = "service %s status" % service_former
+            output = session.cmd_output(cmd)
+        else:
+            raise exceptions.TestError("Fail to get '%s' service status.\n%s" %
+                                       (service, output))
```
will-Do: Define cmd 'output' before the try.except block, then you can avoid arise TestError here

PandaWei: Yes, Thanks for your remind.
Updated

virttest/utils_misc.py
```diff
@@ -851,6 +851,44 @@ def archive_as_tarball(source_dir, dest_dir, tarball_name=None,
...
+    :return: 'active' or 'inactive'
+    :rtype: String
+    """
+    output = ""
+    try:
+        session.cmd("systemctl --version")
+        cmd = "systemctl status %s.service" % service
+        output = session.cmd_output(cmd)
+    except Exception:
+        if service_former is not None:
```
xutian: if not  service_former:
    xxx

Looks more better.

PandaWei: Indeed, it's more brief.
Updated.

lmr: In general, you use `is not None` when you are specifically testing for `None`, and absolutely does not want an expression that evaluates to a boolean `False`. Otherwise `is not [variable]` is the preferred way of doing things.

PandaWei: Thanks @lmr,
I really appreciate your kindness and explanation.

virttest/utils_misc.py
xutian: let's use below command to make this function work on OS which use both  upstart or systemd.
 "{service %s status  || systemctl %s.service status} ||  status %s"

PandaWei:
Update as @xutian 's comments,
Good idea, thanks

PandaWei:
Thanks @xutian 
But If i remove '\',  pyflakes fails.
[root@localhost avocado-vt]# pyflakes virttest/utils_misc.py
virttest/utils_misc.py:870: EOL while scanning string literal
    cmd = ("service %(key)s status || systemctl %(key)s.service status ||

xutian:
cmd = ("service %(key)s status || "
             "systemctl %(key)s.service status || "
             "status %(key)s")

PandaWei:
Update and Thanks @xutian

virttest/utils_misc.py
```diff
@@ -851,6 +851,44 @@ def archive_as_tarball(source_dir, dest_dir, tarball_name=None,
...
+    :rtype: String
+    """
+    output = ""
+    cmd = ("service %(key)s status || "
+           "systemctl %(key)s.service status || "
+           "status %(key)s")
+    try:
+        output = session.cmd_output(cmd % {'key': service})
+    except Exception:
+        if not service_former:
```
xutian: Does this mean if service has former name, try former name again? if I'm  right. I think it should be:

if not service_former:
     raise TestError 
session.cmd_output(cmd)

PandaWei: why raise TestError before running session.cmd_output(cmd) ?

xutian: If not service_former, that mean service_former should be a false values, view from your code, it should be None or "",  so that next line should be execute without any replacement. so what you want to get?

PandaWei: ah, Thanks
Update as your comment

virttest/utils_misc.py
```diff
@@ -851,6 +851,43 @@ def archive_as_tarball(source_dir, dest_dir, tarball_name=None,
...
+            raise exceptions.TestError("Fail to get %s status.\n%s" %
+                                       (service, output))
+        output = session.cmd_output(cmd % {'key': service_former})
+
+    if re.search("running|active", output, re.I):
```
xutian: It looks we need to search multi-lines, output of systemctl status httpd on my host likes below:
[xutian@dhcp-10-17]$ systemctl status httpd 
● httpd.service - The Apache HTTP Server
   Loaded: loaded (/usr/lib/systemd/system/httpd.service; disabled; vendor preset: disabled)
   Active: active (running) since Tue 2015-12-15 13:32:57 CST; 1 day 21h ago
  Process: 15807 ExecReload=/usr/sbin/httpd $OPTIONS -k graceful (code=exited, status=0/SUCCESS)
 Main PID: 9002 (httpd)
   Status: "Total requests: 6; Idle/Busy workers 100/0;Requests/sec: 3.68e-05; Bytes served/sec: 9.0KB/sec"
   CGroup: /system.slice/httpd.service
           ├─ 9002 /usr/sbin/httpd -DFOREGROUND
           ├─15809 /usr/sbin/httpd -DFOREGROUND
           ├─15810 /usr/sbin/httpd -DFOREGROUND
           ├─15811 /usr/sbin/httpd -DFOREGROUND
           ├─15813 /usr/sbin/httpd -DFOREGROUND
           └─15815 /usr/sbin/httpd -DFOREGROUND

Dec 15 13:32:57 dhcp-10-17.nay.redhat.com systemd[1]: Starting The Apache HTTP Server...
Dec 15 13:32:57 dhcp-10-17.nay.redhat.com systemd[1]: Started The Apache HTTP Server.
Dec 16 03:08:02 dhcp-10-17.nay.redhat.com systemd[1]: Reloaded The Apache HTTP Serve

virttest/utils_misc.py
```diff
@@ -851,6 +851,43 @@ def archive_as_tarball(source_dir, dest_dir, tarball_name=None,
...
+        status = "active"
+    elif re.search("stopped|inactive", output, re.I):
```
xutian: same as above line,  re.M is need too.

PandaWei:
@xutian Appreciate wery much
I rewrite the judgement of output, May i trouble you to check it again.
referred the code of ./virttest/staging/service.py

xutian:
LGTM

xutian:
Merged. thanks @PandaWei

will-Do:
@xutian 
I does not find the commit.
Did you merge this from command line or just leave the merge comment here without merge it?

xutian:
Merged, thanks @will-Do .

PR #252, opened 2015-11-12
Switch default keys for Win8.1, XP

Andrei-Stepanov:
This keys better work for specified ISO images at
shared/unattended/*.xml


lmr:
Sure

PR #255, opened 2015-11-13
data_dir.py: get_tmp_dir()  public argument is added

svirt:
As discussed in PR #846 a new function get_public_tmp_dir() is
added. Provides a temporary directory with less restrictive permissions.
Will be used in libvirt tests.

### DIFF
--- a/virttest/data_dir.py
+++ b/virttest/data_dir.py
@@ -198,8 +198,11 @@ def get_deps_dir(target=None):
                               nesting_limit)
 
 
-def get_tmp_dir():
-    return data_dir.get_tmp_dir()
+def get_tmp_dir(public=False):
+    tmp_dir = data_dir.get_tmp_dir()
+    if public:
+        os.chmod(tmp_dir, 0755)
+    return tmp_dir
 
 
 def get_download_dir():


PandaWei:
#846 ?  I don't find it.

svirt:
sorry forgot to mention, its from avocado repo: https://github.com/avocado-framework/avocado/pull/876
I will review now the travis error and recommit again

virttest/data_dir.py
```diff
@@ -201,6 +201,10 @@ def get_deps_dir(target=None):
 def get_tmp_dir():
     return data_dir.get_tmp_dir()
 
+def get_public_tmp_dir():
+    tmp_dir = data_dir.get_tmp_dir()
+    os.chmod(tmp_dir, 0755)
+    return tmp_dir
```
PandaWei: In my humble opinion, 
There is no need to add a function in case of get_tmp_dir has similar functions with it.
We can add new param to allows users to set permissions or select which level they need.

PandaWei: I have read your discussion.
I suggest you that it's better to add bool param named public to give user permission or not.
The default of public maybe False.

svirt:
Actually locally first I implemented with a boolean parameter public but then I tried to keep the apprach found in the discussion. Although me personally I am for the parameter solution. I will now change it that way here.

will-Do:
@svirt Please check the following PR
https://github.com/avocado-framework/avocado-vt/pull/209

svirt:
Aha.. I can see that #209 is a similar PR. So one of these two PRs  will be selected for merging I assume?

will-Do:
@svirt 
Yes, we already using #209 for awhile, and it's ready for merging

svirt:
**@will-Do**
ok, then after it's merged I will adjust the tests according to it

lmr:
I'm closing this since the nearly identical #209 was merged. Thanks @svirt !

PR #257, opened 2015-11-13
Switch cdkey for Win10

Andrei-Stepanov:
Use key for version specified at ./shared/cfg/guest-os/Windows/Win10/*


lmr:
Merged, thanks!

PR #258, opened 2015-11-13
Update test-provider ini file for tp-spice

Andrei-Stepanov:
Configs for tp-spice have been reorganised.


Andrei-Stepanov:
Yes, it is supposed to be like that )

lmr:
Merged, thanks!

PR #264, opened 2015-11-24
data_dir.py: public permissions read flags added

svirt:
Adding +r permissions to get_tmp_dir() public=True case.
Otherwise currently the newly created temporary
directory's contents will not be readable by public.


lmr:
Merged, thanks!

svirt:
Thanks

PR #265, opened 2015-11-26
virttest/test_setup: cleanup unused variable

PandaWei:
48feee4 updated the way used to update kvm module param,
No longer need the kvm_re_probe in sr_iov_cleanup.


lmr:
Sure

PR #267, opened 2015-11-27
guest-os/Windows.cfg: update all windows tools path.

CongSmile:
For windows guests, there is a problem that driver letter will be changed when bootup VM.
Copy those useful tools to 'C:' after OS installation to avoid such driver letter problems.

id: 1273263

### DIFF
--- a/shared/cfg/guest-os/Windows.cfg
+++ b/shared/cfg/guest-os/Windows.cfg
@@ -179,8 +179,8 @@
         set_online_cmd += " diskpart /s cmd"
         testfile_name = "I:\\format_disk-test.txt"
         writefile_cmd = "echo %s > %s &&"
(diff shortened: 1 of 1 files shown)


CongSmile:
Hi @xutian and @suqinhuang 

Could you help review this PR?

Thanks.

xutian:
@lmr For windows guest there has a problem that driver letter will changed when bootup a VM, so we copy some small but always used tool in 'C:' after OS installation finished to avoid file not found issue.

Thanks,
Xu

lmr:
@xutian thanks for the explanation. Could you please update the patch description with that explanation?

CongSmile:
Hi @lmr @xutian @suqinhuang 

I have updated the commit log, could you help review it again?

Thanks.

PR #268, opened 2015-11-27
guest-os.Linux: remove "-no-kvm-pit-reinjection" for ppc as the param…

Zhengtong:
'Option no-kvm-pit-reinjection not supported' on powerpc, so do not set it  

ID:1285216


lmr:
Sure.

PR #269, opened 2015-11-30
increase vm_create_timeout value

xutian:
socket connect timeout in qemu_monitor is 30s bigger
that wait_for_create_monitor timeout default value(20s),
it will cause create_monitor loop only run once.


xutian:
ID:  1267186

suqinhuang:
Acked-by: Suqin Huang

xutian:
Thanks @suqinhuang. Merged.

PR #270, opened 2015-11-30
virttest.libvirt_vm: Perform package update if package was already installed

rbian:
Latest package should be used for testing, we need to update packages if it's
installed, such as qemu-guest-agent, selinux-policy.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2373,14 +2373,22 @@ class VM(virt_vm.BaseVM):
         """
         session = self.wait_for_login()
         try:
-            # Install the package if it does not exists
-            cmd = "rpm -q %s || yum install -y %s" % (name, name)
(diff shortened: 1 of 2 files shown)


will-Do:
Maybe you can calling 'utils_misc.yum_install' directly

lmr:
@rbian the failures are not your fault, you can simply rebase against the latest master and then the problem will go away. Please do that, thanks!

rbian:
@will-Do , "utils_misc.yum_install" has same problem. It just install the package if it's not installed, it can't update the package.  I think this is right place to do it.

will-Do:
Okay, I'm good with this pr, so ACK

waynesun09:
Found remove_numa_cells function be used in another PR:
https://github.com/autotest/tp-libvirt/pull/657

waynesun09:
ACK

PR #273, opened 2015-12-01
Fix undetected unattended install success 

olavphibm:
Due to a race condition, when virtual machine was detected as shut down, there may not have been time for the post finish string to be read from serial log.
unattended_install then reported "ERROR" even though installation ended successfully.
The actual fix is in the third commit, the other two just do some refactoring.


lmr:
Looks good, apart from the comment in fd924e9. Please address it so I can merge this PR.

lmr:
@olavphibm, good work, merged, thanks!

olavphibm:
That was quick, thanks!

PR #281, opened 2015-12-03
virttest/utils_net.py: Update setup() and cleanup() for IPv6 address

dzhengfy:
When IPv6Manager sets up IPv6 environment, the setup() does not check if the specified address exists already on the host which sometimes may cause IfChangeAddrError exception.
When IPv6Manager cleans up IPv6 environment, the cleanup() does not check if the specified address is added by avocado-vt which also may cause IfChangeAddrError.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -1861,7 +1861,8 @@ class IPv6Manager(propcan.PropCanBase):
     __slots__ = ('server_ip', 'server_user', 'server_pwd', 'server_ifname',
                  'client_ifname', 'client_ipv6_addr', 'server_ipv6_addr',
                  'client', 'port', 'runner', 'prompt', 'session',
-                 'auto_recover', 'check_ipv6_connectivity')
+                 'auto_recover', 'check_ipv6_connectivity', 'client_ipv6_added',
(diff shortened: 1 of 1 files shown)


will-Do:
The codes are LGTM, just two suggestions:
1) It's better to update your commit summary, an example:
```
utils_net: Check IPv6 address before add it

Before add a specific address, it's necessary to check if it exist to
avoid IfChangeAddrError exception.
```

2) For the two flags you added, I suggest to name them consistent(server/client vs src/dest)

dzhengfy:
Running 'inspekt lint'

Pylint disabled: C,E,R,W,F0401,I0011

Pylint enabled : W0611

************* Module avocado-vt.virttest.qemu_devices.qcontainer

W0611: 21,0: : Unused import utils_misc

Pylint check fail: /home/travis/build/avocado-framework/avocado-vt/virttest/qemu_devices/qcontainer.py

Syntax check FAIL

will-Do:
@dzhengfy 
https://github.com/avocado-framework/avocado-vt/pull/292 fix this issue and already merged, you can rebase your patch

will-Do:
ACK and merge this fix

PR #286, opened 2015-12-04
utils_test.libvirt: Fix setup_or_cleanup_gluster parameter problem.

rbian:
setup_or_cleanup_gluster should always use "pool_name" parameter
for cleanup purpose, otherwise it will cleanup whole tmp dir.


will-Do:
ACK and merge the trivial fix.

PR #296, opened 2015-12-11
Request to add OpenSUSE Leap 42.1.x86_64.cfg and its xml file for unattended install

nflondo:
I should have mentioned I did test unattended install with the OpenSUSE-Leap-42.1.xml file being submitted here.  Thank you !


lmr:
Looks good.

PR #298, opened 2015-12-15
virttest.test_setup: Fix system call to use shell

Hao-Liu:
This command containing stdout redirection requires shell to be enabled
or the call will be useless.


PandaWei:
Good catch,
I notice there're the same problems in avocado-vt.
You can fix all of them

lmr:
@PandaWei, I'd love to review the entire code to see if there are more places where I failed to notice that the conversion is necessary, but I'm short in time right now.

PandaWei:
@lmr  I created a new PR avocado-framework/avocado-vt#304 for it

PR #299, opened 2015-12-16
Update driver_id_cmd in driver_load case

suqinhuang:
instead of devcon.exe find * | find "details driver name"
as the driver name maybe updated, we can get all virtio driver
info when test failed

### DIFF
--- a/shared/cfg/guest-os/Windows.cfg
+++ b/shared/cfg/guest-os/Windows.cfg
@@ -315,22 +315,22 @@
         i386,i686:
             install_path = "C:\Program Files\JAM Software\HeavyLoad"
     driver_load:
-        driver_id_pattern = "(.*?):"
         driver_load_cmd = C:\devcon.exe enable @DRIVER_ID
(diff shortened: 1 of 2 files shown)


shared/cfg/guest-os/Windows.cfg
```diff
@@ -315,22 +315,22 @@
         i386,i686:
             install_path = "C:\Program Files\JAM Software\HeavyLoad"
     driver_load:
-        driver_id_pattern = "(.*?):"
         driver_load_cmd = C:\devcon.exe enable @DRIVER_ID
         driver_unload_cmd = C:\devcon.exe disable @DRIVER_ID
+        driver_id_cmd = C:\devcon.exe find * | find "VirtIO"
```
CongSmile: If do not use devcon.exe, why add this line?

suqinhuang: use devcon to find all virtio driver instead of detail info "VirtIO Balloon Driver",
then we can check the logs when it failed.

spcui:
LGTM
Acked-by: scui@redhat.com

PR #302, opened 2015-12-17
virttest/utils_conn.py: Fix ssh login timeout error on remote server

dzhengfy:
Add one parameter in conn_setup() with default value to specify the
    timeout period when waiting for the prompt on remote server.

### DIFF
--- a/virttest/utils_conn.py
+++ b/virttest/utils_conn.py
@@ -582,7 +582,7 @@ class SSHConnection(ConnectionBase):
 
         logging.debug("SSH authentication recover successfully.")
 
-    def conn_setup(self):
+    def conn_setup(self, timeout=10):
         """
         Setup of SSH connection.
 
@@ -591,6 +591,8 @@ class SSHConnection(ConnectionBase):
         (3).Initialization of id_rsa.
         (4).set a ssh_agent.
         (5).copy pub key to server.
+
+        :param timeout: The time duration (in seconds) to wait for prompts
         """
         client_session = self.client_session
         ssh_rsa_pub_path = self.ssh_rsa_pub_path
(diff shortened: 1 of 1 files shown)


virttest/utils_conn.py
```diff
@@ -582,7 +582,7 @@ def conn_recover(self):
 
         logging.debug("SSH authentication recover successfully.")
 
-    def conn_setup(self):
+    def conn_setup(self, rTimeout=10):
```
xutian: Please use intelligible var name and update doc string, thanks.

virttest/utils_conn.py
```diff
@@ -603,6 +603,8 @@ def conn_setup(self):
         ssh_copy_id = self.SSH_COPY_ID
         ssh_agent = self.SSH_AGENT
         shell = self.SHELL
+        if rTimeout < 0:
+            rTimeout = 10
```
will-Do: This judgement make no sense, timeout could be None. 
Just pass it to handle_prompts is enough.

will-Do: If you want to check the value,  I think you should deal with any invalid value. Or just pass it to handle_prompts

virttest/utils_conn.py
lmr: Yes, we use all lower case with underscores instead of camel case, so I'd ask you to convert this to `r_timeout` or similar.

virttest/utils_conn.py
lmr: Like `assert r_timeout >= 0, 'Invalid timeout value: %s' % r_timeout`

dzhengfy:
@lmr , @xutian , @will-Do , I have updated the codes based on your comments. Thanks.

waynesun09:
ACK

PR #305, opened 2015-12-18
iscsi: Using targetcli in a high priority

will-Do:
Current iSCSI target standard is the LIO target, so we try to use it in
a high priority. And, check the iscsi-initiator-utils separately and
print the prompt error.


will-Do:
Will update the unittest later.

PandaWei:
Yes, It's reasonable to give LIO a high priority.

PandaWei:
@will-Do 
If you update the unitest case as soon as possible, we can pass it

will-Do:
@PandaWei 
Just update the unittest.

PandaWei:
Thanks @will-Do and Merge

PR #314, opened 2015-12-30
Windows.cfg: add the cfg for case balloon_hotplug.

CongSmile:

id: 1285979

### DIFF
--- a/shared/cfg/guest-os/Windows.cfg
+++ b/shared/cfg/guest-os/Windows.cfg
@@ -276,7 +276,7 @@
         analyze_cmd = E:\autoit3.exe E:\dump_control.au3
         sub_type = nmi_bsod_catch
         rename_error_image = yes
-    balloon_check:
+    balloon_check, balloon_hotplug:
         free_mem_cmd = wmic os get FreePhysicalMemory
     live_snapshot_chain:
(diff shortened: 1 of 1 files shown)


CongSmile:
Based on: https://github.com/autotest/tp-qemu/pull/519

CongSmile:
Close this PR.

As using the Qdevice to implement balloon hotplug, instead of using case pci_hotplug.

suqinhuang:
Acked-by: Suqin Huang

shared/cfg/guest-os/Windows.cfg
```diff
@@ -276,7 +276,7 @@
         analyze_cmd = E:\autoit3.exe E:\dump_control.au3
         sub_type = nmi_bsod_catch
         rename_error_image = yes
-    balloon_check:
+    balloon_check, balloon_hotplug:
         free_mem_cmd = wmic os get FreePhysicalMemory
```
suqinhuang: need double check the cmd

CongSmile: cmd = 'systeminfo | findstr /C:"Available Physical Memory"'

CongSmile: There is no problem for the cmd, as there is a ratio to handle the free mem difference for above 2 cases.

xutian:
that's ok for me.

PR #318, opened 2016-01-05
fix logical error in add numa node

xutian:


CongSmile:

PR #323, opened 2016-01-07
virttest/virt_vm.py: Update get_memory_size() for windows guest

spcui:
Update the method to get windows guest TotalMemory.

ID: 1295311

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -1283,16 +1283,9 @@ class BaseVM(object):
             if not cmd:
                 cmd = self.params.get("mem_chk_cmd")
             mem_str = session.cmd_output(cmd, timeout=timeout)
-            mem = re.findall(re_str, mem_str)
-            mem_size = 0
-            for m in mem:
-                mem_size += int(m)
-            if "GB" in mem_str:
-                mem_size *= 1024
-            elif "MB" in mem_str:
-                pass
-            else:
-                mem_size /= 1024
+            mem = re.findall(r"\d+\s*\w?", mem_str, re.M)
+            mem = map(lambda x: utils_misc.normalize_data_size(x), mem)
+            mem_size = sum(map(float, mem))
             return int(mem_size)
         finally:
             session.close()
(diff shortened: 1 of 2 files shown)


xutian:
Nack,  MaxCapacity mean OS support max memory size.  We are not always assign VM max support memory.

CongSmile:
Agree w/ xu.

Maybe you can boot the guest with the limited memory for testing for this case.

Thanks.

virttest/virt_vm.py
```diff
@@ -1287,13 +1287,16 @@ def get_memory_size(self, cmd=None, timeout=60, re_str=None):
             mem_size = 0
             for m in mem:
                 mem_size += int(m)
-            if "GB" in mem_str:
-                mem_size *= 1024
-            elif "MB" in mem_str:
-                pass
+            if self.params.get("os_type") == "windows":
```
xutian: we  have framework function utils_misc.normalize_data_size to convert it

CongSmile:
@spcui 

Could you give an explanation about the way / idea to fix this bug? I'm a little confused.

Thanks.

virttest/virt_vm.py
```diff
@@ -1287,13 +1287,19 @@ def get_memory_size(self, cmd=None, timeout=60, re_str=None):
             mem_size = 0
             for m in mem:
                 mem_size += int(m)
-            if "GB" in mem_str:
```
xutian: it looks 'normalize_data_size' can handle size string,  please test.

spcui: @xutian mem_str is return by "dmidecode -t 17 | awk -F: '/Size/ {print $2}'", not for "cat /proc/meminfo", so mem_str like this:
"
 4096 MB
 4096 MB
"
so, cannot use mem_str,  need to catch the unit.

spcui:
PASS smp_8.4096m.repeat1.run_test.Host_RHEL.m7.u2.qcow2.virtio_blk.up.virtio_net.Win10.x86_64.io-github-autotest-qemu.physical_resources_check
PASS smp_8.4096m.repeat1.run_test.Host_RHEL.m7.u2.qcow2.virtio_blk.up.virtio_net.RHEL.6.7.x86_64.io-github-autotest-qemu.physical_resources_check

spcui:
@xutian updated with your comment, pls review it again.

PR #324, opened 2016-01-08
utils_misc: add get_linux_drive_path function

luckyh:
This function provides a method to get target drive path in guest by serial or wwn.

ID: 1279721 1279719 1272397

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2749,6 +2749,29 @@ def format_guest_disk(session, did, mountpoint, size, fstype, ostype):
     return format_linux_disk(session, did, mountpoint, size, fstype)
 
 
+def get_linux_drive_path(session, did, timeout=120):
+    """
+    Get drive path in guest by drive serial or wwn
+
+    :param session: session object to guest.
+    :param did: drive serial or wwn.
+    :return String: drive path
+    """
+    cmd = 'for dev_path in `ls -d /sys/block/*`; do '
+    cmd += 'echo `udevadm info -q property -p $dev_path`; done'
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -2749,6 +2749,29 @@ def format_guest_disk(session, did, mountpoint, size, fstype, ostype):
...
+def get_linux_drive_path(session, did):
+    """
+    Get drive path in guest by drive serial or wwn
+
+    :param session: session object to guest.
+    :param did: drive serial or wwn.
+    :return String: drive path
+    """
+    cmd = 'for dev_path in `ls -d /sys/block/*`; do '
+    cmd += 'echo `udevadm info -q property -p $dev_path`; done'
```
xutian: is default timeout  enough for multi-disk test ?

luckyh: Seems so.

multi_disk.max_disk(~25 disks):
```
18:48:49 client           L1042 DEBUG| Sending command: for dev_path in `ls -d /sys/block/*`; do echo `udevadm info -q property -p $dev_path`; done
18:48:50 client           L1042 DEBUG| Sending command: echo $?
```

xutian:
better to pass a timeout args and set default values for that args. if default timeout is long enough, we can set it when we call it.

luckyh:
@xutian have modified the function, would you please help review it, thanks!

spcui:
LGTM, Acked-by: scui@redhat.com

xutian:
Merged, thanks @luckyh  and @spcui .

PR #325, opened 2016-01-13
virttest.ssh_key: Add two functions from autotest

dzhengfy:
Add setup_remote_ssh_key() and get_remote_public_key() which exist
originally in autotest for bi_directional migration test


lmr:
OK, merged, thanks @dzhengfy !

PR #336, opened 2016-01-20
virttest/qemu_vm.py: Add new func get_qemu_threads()

spcui:
Add new function to get the qemu threads.

ID: 1297288

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -3045,6 +3045,18 @@ class VM(virt_vm.BaseVM):
         except (TypeError, IndexError, ValueError):
             return None
 
+    def get_qemu_threads(self):
+        """
+        Return the list of qemu SPIDs
+
+        :return: the list of qemu SPIDs
+        """
+        cmd = "ls /proc/%d/task" % self.get_pid()
+        status, output = commands.getstatusoutput(cmd)
+        if status:
+            return []
+        return output.split()
+
     def get_shell_pid(self):
         """
         Return the PID of the parent shell process.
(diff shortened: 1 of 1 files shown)


virttest/qemu_vm.py
```diff
@@ -3045,6 +3045,20 @@ def get_pid(self):
         except (TypeError, IndexError, ValueError):
             return None
 
+    def get_qemu_threads(self):
+        """
```
xutian: you can list tasks launched by qemu by command
"ls /proc/%d/task"  % self.get_pid()

spcui:
@xutian updated, please help review again?

virttest/qemu_vm.py
```diff
@@ -3045,6 +3045,19 @@ def get_pid(self):
...
+        Return the list of qemu SPIDs
+
+        :return: the list of qemu SPIDs
+        """
+        try:
+            cmd = "ls /proc/%d/task" % self.get_pid()
+            threads_list = commands.getoutput(cmd).split()
+            return threads_list
+        except (TypeError, IndexError, ValueError):
```
xutian: useless catch!!!

virttest/qemu_vm.py
```diff
@@ -3045,6 +3045,16 @@ def get_pid(self):
...
             return None
 
+    def get_qemu_threads(self):
+        """
+        Return the list of qemu SPIDs
+
+        :return: the list of qemu SPIDs
+        """
+        cmd = "ls /proc/%d/task" % self.get_pid()
+        threads_list = commands.getoutput(cmd).split()
```
xutian: if status not equal 0, output are error message,  output looks like:  
>>> z = commands.getoutput("kk")
>>> z
'sh: kk: command not found'
so, I think you should check status of command before return it.

virttest/qemu_vm.py
```diff
@@ -3656,11 +3668,18 @@ def migrate(self, timeout=virt_vm.BaseVM.MIGRATE_TIMEOUT, protocol="tcp",
...
 
+            threads_list1 = self.get_qemu_threads()
+
             logging.info("Migrating to %s", uri)
             self.monitor.migrate(uri)
             if not_wait_for_migration:
                 return clone
 
+            threads_list2 = self.get_qemu_threads()
+            if threads_list1 == threads_list2:
```
xutian: It should be  if len(thread_list2) <= len(thread_list1) test failed.

xutian:
looks ok to me. Merged. thanks @spcui

PR #337, opened 2016-01-20
virttest.utils_test.libvirt: Create volume by xml file.

rbian:
Add a function in PoolVolumeTest to create volume by xml file.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -53,6 +53,7 @@ from ..libvirt_xml import NetworkXML
 from ..libvirt_xml import IPXML
 from ..libvirt_xml import pool_xml
 from ..libvirt_xml import nwfilter_xml
+from ..libvirt_xml import vol_xml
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1044,6 +1045,16 @@ def pre_vol(self, vol_name, vol_format, capacity, allocation, pool_name):
         if not pv.volume_exists(vol_name):
             raise exceptions.TestFail("Can't find volume: %s" % vol_name)
 
+    def pre_vol_xml(self, pool_name, **vol_params):
```
will-Do: This function create a volume, but from the name, it more like create a XML file, so I suggest you can change the name to 'pre_vol_by_xml'

rbian:
Thanks @will-Do , Fixed.

PandaWei:
ACK. I'm OK for this PR.

Just curious, why give check_exit_status the second paramter false?
It's default is false.

will-Do:
ACK

PR #339, opened 2016-01-21
utils_v2v: Fix the screenshot path on a remote host

will-Do:
In commit bda251d8, it replace all '/tmp' to data_dir.get_tmp_dir(). But
for this case, we need do screenshot on a remote host where does not
have that dir, so change it back to use '/tmp'.

### DIFF
--- a/virttest/utils_v2v.py
+++ b/virttest/utils_v2v.py
@@ -462,7 +462,8 @@ class WindowsVMCheck(VMCheck):
         """
         sshot_file = os.path.join(data_dir.get_tmp_dir(), "vm_screenshot.ppm")
         if self.target == "ovirt":
-            vm_sshot = os.path.join(data_dir.get_tmp_dir(), "vm_screenshot.ppm")
+            # Note: This is a screenshot path on a remote host
+            vm_sshot = os.path.join("/tmp", "vm_screenshot.ppm")
         else:
             vm_sshot = sshot_file
         virsh.screenshot(self.name, vm_sshot, session_id=self.virsh_session_id)


PandaWei:
ACK.

PandaWei:
Maybe there're other cases also need to make  temp dir under /tmp.
I think we should do some modification for get_tmp_dir(), like below,

def get_tmp_dir(tmp_base=None,public=True):
    if tmp_base:
          tmp_dir = tmp_base
    else:
          tmp_dir = data_dir.get_tmp_dir()
    if public:
        tmp_dir_st = os.stat(tmp_dir)
        os.chmod(tmp_dir, tmp_dir_st.st_mode | stat.S_IXUSR |
                 stat.S_IXGRP | stat.S_IXOTH | stat.S_IRGRP | stat.S_IROTH)
    return tmp_dir

And we can call it like data_dir.get_tmp_dir(tmp_base="/tmp")

what about your opinion?

will-Do:
@PandaWei 
Not necessary I think, if someone want to use a specific dir, just use it directly without pass a parameter and call a func is more reasonable.

will-Do:
Thx @PandaWei 
Merge this fix.

lmr:
@will-Do, sorry about that.

@PandaWei, we already have `tempfile` that already has this option, I think adding the tmp_base parameter to get_tmp_dir goes beyond its scope. Besides, the problem in question was that the tmpdir needed to exist in the remote machine, so the only one we could reasonably expect there was /tmp or /var/tmp. I think a simple comment above that line would suffice to prevent future mass changes to touch it like that.

PR #343, opened 2016-01-27
qemu_vm: add debug only monitor support

xutian:
User can't connection to VM's monitor during test running to do debug if VM status is incorrect.
Add below configure in your tests.cfg to add a debug-only monitor.
        monitors += " monn"
        debugonly_monn = yes


suqinhuang:
ID: 1302600

spcui:
LGTM
Acked-by: scui@redhat.com

PR #357, opened 2016-02-01
virttest/utils_net.py: Remove viobr0 in get_linux_ifname

spcui:
viobr0 is added by libvirt which not device added by vm,
so remove it.

ID: 1283531


xutian:
looks good to me.

PR #355, opened 2016-01-29
virttest/test_setup: Catch TypeError when convert expected_hugepage_size

will-Do:
Parameter 'expected_hugepage_size' is not always exist, and the value
also need check when convert it to an integer.

Signed-off-by: Yanbing Du <ydu@redhat.com>

PR #358, opened 2016-02-01
virttest/test_setup: Skip check if parameters are not given or invalid

will-Do:
1. Not all cases which calling HugePageConfig() have these parameters,
or the 'expected_hugepage_size' is invalid, then so we shoule skip these
two check steps .
2. Fix missing decorator bug.

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -304,21 +304,20 @@ class HugePageConfig(object):
         self.pool_path = "/sys/kernel/mm/hugepages"
         self.sys_node_path = "/sys/devices/system/node"
         # Unit is KB as default for hugepage size.
-        default_hugepage_size = 2048
         try:
             self.expected_hugepage_size = int(
-                params.get("expected_hugepage_size", default_hugepage_size))
+                params.get("expected_hugepage_size", 0))
         except TypeError:
-            logging.warn("Using default value %s as "
-                         "'expected_hugepage_size=%s' is invalid",
-                         default_hugepage_size,
+            logging.warn("Invalid value 'expected_hugepage_size=%s'",
                          params.get("expected_hugepage_size"))
-            self.expected_hugepage_size = default_hugepage_size
+            self.expected_hugepage_size = 0
         self.hugepage_cpu_flag = params.get("hugepage_cpu_flag")
         self.hugepage_match_str = params.get("hugepage_match_str")
-        self.check_hugepage_support()
+        if self.hugepage_cpu_flag and self.hugepage_match_str:
+            self.check_hugepage_support()
         self.hugepage_size = self.get_hugepage_size()
-        self.check_hugepage_size_as_expected()
+        if self.expected_hugepage_size:
+            self.check_hugepage_size_as_expected()
         self.hugepage_force_allocate = params.get("hugepage_force_allocate",
                                                   "no")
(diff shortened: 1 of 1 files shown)


will-Do:
@xutian 
This should be a continuation fix of https://github.com/avocado-framework/avocado-vt/pull/355, please take a look, thx

xutian:
@will-Do  on x86_64 host  2048K huge-page is supported by default, so I think set default huge-page size to 2048K is reasonable for x86_64. And if you have test don't need huge-page,  I suggest you use "setup_hugepages = no" to disable it.

will-Do:
@xutian 
Here we just want to skip these check steps.
And for the default hugepage size, I added it in the last patch, that is for ```expected_hugepage_size```, not for setup. In this patch, set it to 0 just means skip ```check_hugepage_size_as_expected()```, that's no influence to the normal cases.

will-Do:
The problem was introduced by https://github.com/avocado-framework/avocado-vt/commit/dcda66d19e1691824be97702d4f12bc1bb3cfa74, which break the existing cases.

xutian:
@will-Do  why not disable  setup hugepage by "setup_hugepages = no" ?

will-Do:
@xutian 
Thanks for your reply, maybe we have discrepancy about this problem, that's two different issues:
1. From program logical, these parameters may not be present, then calling the check functions will raise error.
2. About the ```setup_hugepages=no``` sentence, yes we can add it when write new cases, but for the exist cases, we don't want to update them for this kind of issue. 
If just for updating or enhancement, one important rule is the new patch should not break the existing cases, so here I just want to fix this issue.

virttest/test_setup.py
```diff
@@ -304,21 +304,20 @@ def __init__(self, params):
...
-                         default_hugepage_size,
+            logging.warn("Invalid value 'expected_hugepage_size=%s'",
                          params.get("expected_hugepage_size"))
-            self.expected_hugepage_size = default_hugepage_size
+            self.expected_hugepage_size = 0
         self.hugepage_cpu_flag = params.get("hugepage_cpu_flag")
         self.hugepage_match_str = params.get("hugepage_match_str")
-        self.check_hugepage_support()
+        if self.hugepage_cpu_flag and self.hugepage_match_str:
+            self.check_hugepage_support()
```
xutian: Let's move this check in check_hugepage_support().  Because 'hugepage_match_str'  is a optional param for 2M  hugepage, but cpu feature 'pse' is required. 
For 1G hugepage on x86 platform, cpu feature 'pdpe1gb'  and kernel param 'hugepagesz=1GB' are required.

xutian:
Original default hugepage size is 2048KB and  it was support by default in RHEL host. so I suggest to keep default value to 2048.

xutian:
I'm so sorry for blocking your test and thanks for your quick fix.

will-Do:
@xutian 
Here the ```default_hugepage_size=2048``` has meaningless but do one more check step, which may cause trouble if the host not use the default setting(and test case does not care about it).
In this patch, it use 0 as the default value(if expected_hugepage_size not given) to skip the check step, no effect to any existing cases or new add cases.

xutian:
thanks for your explaining, that's OK for me.

xutian:
Merged, thanks @will-Do

PR #371, opened 2016-02-22
shared.unattended: update telnet server installation cmd

xutian:
Telnet server is not supported by win10, so remove related
configuration

### DIFF
--- a/shared/unattended/win10-32-autounattend.xml
+++ b/shared/unattended/win10-32-autounattend.xml
@@ -151,10 +151,6 @@
 					<Order>3</Order>
 				</SynchronousCommand>
 				<SynchronousCommand wcm:action="add">
-					<CommandLine>%WINDIR%\System32\cmd /c start /w pkgmgr /iu:"TelnetServer"</CommandLine>
-					<Order>4</Order>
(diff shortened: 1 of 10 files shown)


shared/unattended/win10-32-autounattend.xml
```diff
@@ -151,10 +151,6 @@
 					<Order>3</Order>
 				</SynchronousCommand>
 				<SynchronousCommand wcm:action="add">
-					<CommandLine>%WINDIR%\System32\cmd /c start /w pkgmgr /iu:"TelnetServer"</CommandLine>
-					<Order>4</Order>
-				</SynchronousCommand>
-				<SynchronousCommand wcm:action="add">
 					<CommandLine>%WINDIR%\System32\cmd /c sc config TlntSvr start= auto</CommandLine>
 					<Order>5</Order>
```
spcui: need update the Order number?

xutian: It will not block installation, so no need to re-order the number.

xutian:
ID:  1288028

xutian:
Merged  thanks for @CongSmile  and @spcui comments.

PR #375, opened 2016-02-23
libvirt_xml: Implement filesystem XML

will-Do:


waynesun09:
ACK

will-Do:
The case which dependent on this patch is running good, and we can merge this XML update patch.

PR #379, opened 2016-02-29
utils_test: refactor memory test related functions

xutian:


xutian:
We have another pullreq has same update, so closed this one.

PR #383, opened 2016-03-04
Extend test_timeout

suqinhuang:
Some test need long time, espical for the tests repeat hundreds times.

ID: 1244178

### DIFF
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -411,7 +411,8 @@ Ubuntu:
 
 # Timeouts
 login_timeout = 360
-test_timeout = 14400
+# 99 hours
+test_timeout = 356400
 
 # libvirt (virt-install optional arguments)
 # TODO: Rename these with 'libvirt_' prefix


CongSmile:
I think it's better to set the different timeout for different cases.

suqinhuang:
I don't think it's necessary,

Generally, test_timeout is useless to the test.
the test_timeout is set to avoid the env issue at first.
job may hang because of the machine management before, which will make the job keep running long time.
we never met the env issue now

PandaWei:
I am very curious about 356400.
How do you determine this value？ 
Do we need to expand the recommended value 14400 to 356400？

suqinhuang:
99d is the limit time in beaker

CongSmile:
@suqinhuang 

Could you check the timeout defined in run_test() in avocado/avocado/core/runner.py?
        # At this point, the test is already initialized and we know
        # for sure if there's a timeout set.
        timeout = test_status.early_status.get('timeout')
        timeout = float(timeout or self.DEFAULT_TIMEOUT)

According to the test, should be the above timeout works.

Thanks.

suqinhuang:
Hi coli,
You are right
run_tests in utils_misc.py is useless now, 
will send a patch internal to set timeout.

Thanks
Suqin

suqinhuang:
timeout will be covered with --job-timeout

args.job_timeout = self._validate_job_timeout(args.job_timeout)
        job_instance = job.Job(args)
        return job_instance.run()

PR #387, opened 2016-03-09
nfs utils_misc: Fix ssh connection timeout

dzhengfy:
Default timeout 10 seconds is short for NFSClient and SELinuxBoolean to
create SSH connection sometimes. So new implementation will use
'ssh_timeout' to set a timeout if any.

### DIFF
--- a/virttest/nfs.py
+++ b/virttest/nfs.py
@@ -240,7 +240,8 @@ class NFSClient(object):
     def __init__(self, params):
         # Setup SSH connection
         self.ssh_obj = SSHConnection(params)
-        self.ssh_obj.conn_setup()
+        ssh_timeout = int(params.get("ssh_timeout", 10))
+        self.ssh_obj.conn_setup(timeout=ssh_timeout)
 
         self.mkdir_mount_remote = False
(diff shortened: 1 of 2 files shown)


virttest/nfs.py
```diff
@@ -240,7 +240,8 @@ class NFSClient(object):
     def __init__(self, params):
         # Setup SSH connection
         self.ssh_obj = SSHConnection(params)
-        self.ssh_obj.conn_setup()
+        ssh_timeout = int(params.get("ssh_timeout", 10))
+        self.ssh_obj.conn_setup(timeout=ssh_timeout)
```
PandaWei: Where do you intend to configure the value of “ssh_timeout” ？

dzhengfy: @PandaWei , for migration in tp-libvirt, 'ssh_timeout' will be configured in libvirt/tests/cfg/migration/migrate_vm.cfg, like below:

- virsh.migrate_vm:
    type = migrate_vm
    take_regular_screendumps = no
...
    ssh_timeout = 60

will-Do:
Let it configurable is good, ACK and merge this patch.

PR #390, opened 2016-03-15
utils_v2v: modify to adapt to vm whose windows os is not installed on c:\WINDOWS

chloerh:


virttest/utils_v2v.py
```diff
@@ -187,6 +187,7 @@ def __init__(self, test, params, env):
         self.target_path = params.get("target_path", "pool_path")
         self.emulated_img = params.get("emulated_image_path",
                                        "v2v-emulated-img")
+        self.windows_path = params.get("windows_path", "WINDOWS")
```
will-Do: If the dir is case sensitive?

virttest/utils_v2v.py
```diff
@@ -606,7 +607,8 @@ def get_windows_event_info(self):
         """
         Get windows event log info about WSH.
         """
-        cmd = "CSCRIPT C:\WINDOWS\system32\eventquery.vbs /l application /Fi \"Source eq WSH\""
+        cmd = "CSCRIPT C:\%s\system32\eventquery.vbs /l application /Fi \"Source eq WSH\"" % \
+              self.windows_path
```
will-Do: It's better to avoid use  '\' to split a long.  You can using the following method:
```
cmd = "..."
cmd += "..."
```

virttest/utils_v2v.py
chloerh: Verified, it's not case sensitive

will-Do:
@chloerh Please help to resolved the conflict.
And one more suggestion for the commit summary: ```utils_v2v: Let the Windows root configurable```

chloerh:
@will-Do Modified, Please review

will-Do:
ACK

virttest/utils_v2v.py
```diff
@@ -183,6 +183,7 @@ def __init__(self, test, params, env):
         self.export_name = params.get('export_name')
         self.delete_vm = 'yes' == params.get('vm_cleanup', 'yes')
         self.virsh_session_id = params.get("virsh_session_id")
+        self.windows_path = params.get("windows_path", "C:\WINDOWS")
```
will-Do: Maybe name to 'windows_root' is better

chloerh:
@will-Do updated, please review, thanks again :+1:

PR #395, opened 2016-03-16
virttest.utils_misc: Fix the bug of is_mounted function.

rbian:
For some network FS such as glusterfs, it may use fuse.glusterfs
FS type in /etc/mtab, it should use wildcard matching in the function.


will-Do:
ACK

Hao-Liu:
ACK and merge this.

PR #397, opened 2016-03-18
utils_test.libvirt: Strip output before use

Hao-Liu:
The output of this command could possible contains an EOL so we have
to strip it before use.

Or the generated XML will be wrong and pool creation will fail.


will-Do:
ACK

PR #402, opened 2016-03-22
virttest.virt_vm: fix TypeError in verify_kernel_crash function

xutian:
get output of serial console not always return data, it caused
syntax issue if return is None. fix it by return driectly if no
data get.


xutian:
ID: 1319971

spcui:
LGTM
Acked-by: scui@redhat.com

PR #403, opened 2016-03-23
virttest/qemu_monitor.py: Small fix in migrate_set_downtime() for qmp

spcui:
For qmp monitor, the unit of set_downtim  must be in seconds,
like as: 1 is 1s, 0.1 is 100ms.

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -1831,8 +1831,7 @@ class QMPMonitor(Monitor):
 
         :return: The command's output
         """
-        val = value * 10 ** 9
-        args = {"value": val}
+        args = {"value": value}
         return self.cmd("migrate_set_downtime", args)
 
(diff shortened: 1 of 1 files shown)


spcui:
ID: 1266332

mazhang:
Acked-by: Maosheng Zhang <mazhang@redhat.com>

virttest/qemu_monitor.py
```diff
@@ -1831,8 +1831,7 @@ def migrate_set_downtime(self, value):
 
         :return: The command's output
         """
-        val = value * 10 ** 9
```
xutian: view from original code, it looks unit is nanosecond, right?

spcui: yes, but for qmp monitor, ths unit should be second.

xutian:
Acked-by: Xu Tian <xutian@redhat.com>

PR #406, opened 2016-03-24
qemu_monitor: fix debug arg not work in function  human_monitor_cmd

xutian:
fix human_monitor_cmd call _log_command function with default debug
argument issue.

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -1612,7 +1612,7 @@ class QMPMonitor(Monitor):
 
         :return: The response to the command
         """
-        self._log_command(cmd, extra_str="(via Human Monitor)")
+        self._log_command(cmd, debug, extra_str="(via Human Monitor)")
 
         args = {"command-line": cmd}
         ret = self.cmd("human-monitor-command", args, timeout, False, fd)


xutian:
ID: 1320849

xutian:
@suqinhuang  @CongSmile  would help to review this PR when free. thanks

CongSmile:

CongSmile:
def _log_command(self, cmd, debug=True, extra_str=""):
 
The default debug is True.

xutian:
@CongSmile  right. because it always True, so  any command send via human_monitor_cmd will be logged, that not expect and it make our debug looks very promiscuous. (eg.  info registers always logging in debug log). That why I do this hotfix.

Thanks,
Xu

virttest/qemu_monitor.py
```diff
@@ -1612,7 +1612,7 @@ def human_monitor_cmd(self, cmd="", timeout=CMD_TIMEOUT,
 
         :return: The response to the command
         """
-        self._log_command(cmd, extra_str="(via Human Monitor)")
+        self._log_command(cmd, debug, extra_str="(via Human Monitor)")
```
CongSmile: But debug here == "debug=True", right?

CongSmile:

PR #408, opened 2016-03-24
Add guest 7.2 related cfg files

ngu-niny:
Add guest 7.2 related cfg files under dir ..avocado-vt/shared/cfg/guest-os/Linux/RHEL

### DIFF
--- a/shared/cfg/guest-os/Linux/RHEL/7.2.cfg
+++ b/shared/cfg/guest-os/Linux/RHEL/7.2.cfg
@@ -1,0 +1,35 @@
+- 7.2:
+    no setup
+    image_name = images/rhel72
+    os_variant = rhel7
+    unattended_install, check_block_size.4096_512, check_block_size.512_512, svirt_install:
+        unattended_file = unattended/RHEL-7-series.ks
+        syslog_server_proto = udp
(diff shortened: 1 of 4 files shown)


xutian:
Configuration looks good to me, but not test. @Zhengtong  can you have a try about ppc64 part in your test environment. 

Thanks,
Xu

xutian:
Tested PR in x86 platform, it work for me.

Zhengtong:
Hi, Niny, I found two problems in my testing with this pr.
1. The CD name:  My be it is not a problem because it could be customized.  in my test , I changed the name: "isos/linux/RHEL-7.2-Server-ppc64.iso" to "isos/linux/RHEL7.2-Server-ppc64.iso" temporarily to get it running. And it also like this for ppc64le. 
2. In my test , It shows the  value of  md5sum_1m_cd1 is not match.  I tried getting the value manually. In my check , the value is .
  md5sum_1m_cd1 = 8c3e5a70edb328f58887c99fc4c4654b  (ppc64)
  md5sum_1m_cd1 = ad122d234b83bf34c55e7df405bd8f22  (ppc64le)

xutian:
Updated and merged into master

PR #416, opened 2016-03-31
virttest.qemu_vm: remove duplicate call verify_kernel_crash

xutian:
verify_alive has call verify_kernel_crash, so no need to call
again.


spcui:
LGTM
Acked-by: scui@redhat.com

PR #417, opened 2016-03-31
Fix syntax error, when guest runlevel is already 3 as expected.

XueqiangWei:


XueqiangWei:
ID 1321493

PR #418, opened 2016-04-01
utils_misc: new function to check qemu capability

dzhengfy:
Sometimes there is a need to check if qemu has a certain capability in
order to determine if some features are supported. This new function
implements this.


Hao-Liu:
LGTM. Merge this directly since a new function won't affect existing tests at all

PR #419, opened 2016-04-01
fixing logging error in qemu.py

yashkmankad:
Fixed minor error on line 198 in qemu.py 
logging.info("Guest runlevel is already %s as expected") % ori_runlevel

Error was causing an avocado test to fail with "TypeError: unsupported operand type(s) for %: 'NoneType' and 'str'"

Signed-off-by: Yash Mankad <ymankad@redhat.com>


yashkmankad:
Looks like a duplicate of #417

xutian:
thanks @yashkmankad , PR#417 has really fixed this bug.

Xu

yashkmankad:
Thanks @xutian 
Closing the PR.

Issue #420, opened 2016-04-01
TODO: Set defaults based on arch


sathnaga:
Right now DEFAULT_MACHINE_TYPE, DEFAULT_GUEST_OS is always set to intel, working on a patch to handle the same in virttest/defaults.py for other arch, suggestions/commands welcome

sathnaga:
Done.

PR #426, opened 2016-04-02
virttest.defaults: Added support for default machine type based on arch

sathnaga:
Added support for choosing default machine type
based on machine architecture

Addresses https://github.com/avocado-framework/avocado-vt/issues/420

### DIFF
--- a/virttest/defaults.py
+++ b/virttest/defaults.py
@@ -1,9 +1,22 @@
+import platform
+ARCH = platform.machine()
+# Set the default machine to i440fx for unidentifed arch also
+# Below line can be removed once support for all possible arch is covered
 DEFAULT_MACHINE_TYPE = "i440fx"
+
+if ARCH in ('ppc64', 'ppc64le'):
+    DEFAULT_MACHINE_TYPE = "pseries"
+elif ARCH in ('x86_64', 'i386'):
+    DEFAULT_MACHINE_TYPE = "i440fx"
+else:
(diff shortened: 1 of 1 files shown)


will-Do:
LGTM

virttest/defaults.py
```diff
@@ -1,9 +1,22 @@
+import platform
+ARCH = platform.machine()
+# Set the default machine to i440fx for unidentifed arch also
+# Below line can be removed once support for all possible arch is covered
 DEFAULT_MACHINE_TYPE = "i440fx"
 
+if ARCH in ('ppc64', 'ppc64le'):
+    DEFAULT_MACHINE_TYPE = "pseries"
+elif ARCH in ('x86_64', 'i386'):
+    DEFAULT_MACHINE_TYPE = "i440fx"
```
xutian: Q35 machine type maybe the default machine type in future.

xutian: how about to set it to PC?

sathnaga: @will-Do Thanks :-)
@xutian : Seeting it to PC might not be common for all, we can have Q35 aswell, This is for enabling the arch dependent defaults, we can decide/change values later aswell.

xutian:
Ok,  Acked-by: Xu Tian <xutian@redhat.com>

xutian:
Merged, thanks @sathnaga and @will-Do

PR #438, opened 2016-04-06
Also add vt-only-filter to vt_list

Hao-Liu:
In a recently merged commit vt-only-filter is introduced in options
but no option is added in vt_list module. This will making avocado
failed to list any of virt test cases.


sathnaga:
ACK

Hao-Liu:
Thanks @sathnaga for confirming this. I'll merge this patch quickly to avoid further impact.

PR #442, opened 2016-04-07
ssh_key: Fix remote known_hosts file setup

dzhengfy:
Searching local known_hosts for the local host key is not a good
solution as it maybe does not exist or does not include the host key
required. The new implementation is to use ssh-keyscan to search the
host key using given host ip, which is much general solution.

### DIFF
--- a/virttest/ssh_key.py
+++ b/virttest/ssh_key.py
@@ -1,9 +1,9 @@
 import os
-import re
 import logging
 import aexpect
 
 from avocado.utils import process
+from avocado.utils import path
 
 from . import remote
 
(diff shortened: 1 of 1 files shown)


virttest/ssh_key.py
```diff
@@ -201,11 +201,19 @@ def setup_remote_known_hosts_file(client_ip, server_ip,
...
-    home = os.path.expanduser('~')
-    cmd = 'cat %s/.ssh/known_hosts' % home
-    output = process.system_output(cmd, verbose=False)
-    host_key = re.findall('(.*%s[, ].*)' % client_ip, output)[0]
+    logging.debug('Performing known_hosts file setup on %s from %s.' %
+                  (server_ip, client_ip))
+    abs_path = ""
+    try:
+        abs_path = path.find_command("ssh-keyscan")
+    except Exception as err:
```
waynesun09: Exception is too general for this, use path.CmdNotFoundError instead

waynesun09:
ACK

will-Do:
OK

PR #445, opened 2016-04-08
utils_test.libvirt: add a parameter ignore_status

dzhengfy:
MigrationTest.do_migration function always raises an exception when
migration fails. But sometimes it is not needed. The fix is to provide a
way to avoid of raising an exception and just set the boolean result.


will-Do:
ACK

PandaWei:
LGTM, 
You just change the API, Please remember to make modification for the callers.

will-Do:
The new added parameter has the default value, so it's should be safe for other callers.

PR #446, opened 2016-04-11
virttest.qemu_monitor: increase socket connect timeout

xutian:
If VM memory big than 16G, original connect timeout 30s
is not enough. 60s timeout test pass with 32G VM,
so increase default timeout to 60s to avoid
MonitorConnectTimeout exception.

And original default timeout of wait_for_create_monitor
less than default socket connect timeout, it cause wait for
try to create connection only once, so increase default
create monitor timeout too.


xutian:
ID: 1314603

mazhang:
Acked-by: Maosheng Zhang

spcui:
OK, LGTM
Acked-by: scui@redhat.com

PR #447, opened 2016-04-11
Correct command and directory in rh_perf_log_hostinfo_script.sh

mazhang:
ID: 1325758

Signed-off-by: Maosheng Zhang <mazhang@redhat.com>

### DIFF
--- a/shared/scripts/rh_perf_log_hostinfo_script.sh
+++ b/shared/scripts/rh_perf_log_hostinfo_script.sh
@@ -26,22 +26,21 @@ ports=`brctl show|grep -v "bridge.*name.*bridge.*id"|awk {'print $4'}`
 for i in $bridges;do
     call echo "ethtool -k $i"
     call ethtool -k $i
+    call brctl showstp $i
 done
 for i in $ports;do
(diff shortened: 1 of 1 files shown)


shared/scripts/rh_perf_log_hostinfo_script.sh
```diff
@@ -26,22 +26,21 @@ ports=`brctl show|grep -v "bridge.*name.*bridge.*id"|awk {'print $4'}`
 for i in $bridges;do
     call echo "ethtool -k $i"
     call ethtool -k $i
+    call brctl showstp $i
```
xutian: do you need to redirect stderr to /dev/null, do it  no error print if $i is not a bridge.

xutian: Other parts looks good to me.

xutian: Oh, sorry.  $bridges only contain bridge type interface. so ignore comment above comment.

xutian:
LGTM

PR #457, opened 2016-04-15
Fix __getinitargs__() in qemu_monitor.py

Andrei-Stepanov:
Reconciliation __getinitargs__() with __init__().

https://docs.python.org/2/library/pickle.html#object.__getinitargs__

    .... should return a tuple containing the arguments to be passed to the
    class constructor (init() for example).


PandaWei:
ACK

will-Do:
Yes, the ```Ture``` here is redundant
ACK

PR #458, opened 2016-04-16
virttest.storage_migration.py: add new module for storage vm migration.

CongSmile:

id: 1321820

### DIFF
--- a/virttest/storage_migration.py
+++ b/virttest/storage_migration.py
@@ -1,0 +1,74 @@
+"""
+Classes and functions to handle storage vm migration.
+
+Include features:
+1. live snapshot
+2. block mirror
+3. block stream
+4. live commit
+5. transaction
+"""
+import logging
+
+from avocado.core import exceptions
+
+
+class StorageMigrationTest(object):
+
+    """
+    KVM classes for handling common operations of storage vm migration.
+    """
+
+    def __init__(self, params, env):
+        """
+        Init the default values for live snapshot object.
+
+        :param params: A dict containing VM preprocessing parameters.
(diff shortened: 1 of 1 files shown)


virttest/storage_migration.py
```diff
@@ -0,0 +1,74 @@
...
+        :param device: the name of the device to generate the snapshot from.
+        :param snapshot_file: the target of the new image. A new file will be created.
+        :param kwargs: optional keyword arguments to pass to func
+        """
+        StorageMigrationTest.__init__(self, params, env)
+        self.device = device
+        self.snapshot_file = snapshot_file
+        self.kwargs = kwargs
+
+    def create_snapshot(self):
```
waynesun09: At the top of the module file you specify this module will include all storage migration functions, and we see your first step of create_snapshot here.

This function only use qemu_vm function, and a check_snapshot function is introduced. 
As the class name LiveSnapshotTest specified, this is for snapshot related test, and this is only for qemu kvm test.

Base on this I'd suggest you put you test and check part into virttest.utils_test.qemu other than directly under virttest. For most modules under virttest are common modules for generous virt related testing.

@PandaWei any thoughts?

PandaWei: MultihostMigration had been declared at /virttest/utils_test/qemu.py.
In contrast with libvirt/virsh, MigrationTest had been declared at ./virttest/utils_test/libvirt.py.

So, Agree with you @ waynesun09.

CongSmile:
Fine, close this one at first.
Will update it later.

Thanks.

PR #460, opened 2016-04-18
utils_test.libvirt: Support to return CmdResult

dzhengfy:
Pass through the parameter 'ignore_status' to migration process in order
to get the CmdResult object even when the migration fails instead of
getting a CmdError exception object. The CmdResult including the stderr
will be useful for the caller to parse and search specific messages.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1096,28 +1096,46 @@ class MigrationTest(object):
         # The time spent when migrating vms
         # format: vm_name -> time(seconds)
         self.mig_time = {}
+        # The CmdResult returned from virsh migrate command
+        self.ret = None
 
-    def thread_func_migration(self, vm, desturi, options=None):
+    def thread_func_migration(self, vm, desturi, options=None,
+                              ignore_status=False):
         """
         Thread for virsh migrate command.
 
         :param vm: A libvirt vm instance(local or remote).
-        :param desturi: remote host uri.
+        :param desturi: Remote host uri.
+        :param options: The options for migration command.
+        :param ignore_status: True, means no CmdError will be caught
+                              for the failure.
+                              False, means an CmdError will be caught
+                              for the failure.
         """
         # Migrate the domain.
+        is_error = False
+
         try:
             if options is None:
                 options = "--live --timeout=60"
             stime = int(time.time())
-            vm.migrate(desturi, option=options, ignore_status=False,
-                       debug=True)
+            self.ret = vm.migrate(desturi, option=options,
+                                  ignore_status=ignore_status,
+                                  debug=True)
             etime = int(time.time())
             self.mig_time[vm.name] = etime - stime
+            if self.ret.exit_status != 0:
+                logging.debug("Migration to %s returns failed exit status %d",
+                              desturi, self.ret.exit_status)
+                is_error = True
         except process.CmdError, detail:
             logging.error("Migration to %s failed:\n%s", desturi, detail)
-            self.RET_LOCK.acquire()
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1089,10 +1091,15 @@ def thread_func_migration(self, vm, desturi, options=None):
             if options is None:
                 options = "--live --timeout=60"
             stime = int(time.time())
-            vm.migrate(desturi, option=options, ignore_status=False,
-                       debug=True)
+            self.ret = vm.migrate(desturi, option=options, ignore_status=ignore_status,
+                                  debug=True)
             etime = int(time.time())
             self.mig_time[vm.name] = etime - stime
+            if ignore_status is True and self.ret is not None:
```
waynesun09: Is this check necessary here? Since you ignored the failure and what' following is in do_migration function which will release lock if the thread is alive.

dzhengfy: @waynesun09 , you are right. Updated already.

virttest/utils_test/libvirt.py
```diff
@@ -1076,23 +1076,33 @@ def __init__(self):
         # The time spent when migrating vms
         # format: vm_name -> time(seconds)
         self.mig_time = {}
+        # The CmdResult returned from virsh migrate command
+        self.ret = None
 
-    def thread_func_migration(self, vm, desturi, options=None):
+    def thread_func_migration(self, vm, desturi, options=None, ignore_status=False):
```
waynesun09: please keep line length in 80

dzhengfy: @waynesun09 , updated.

waynesun09:
ACK

virttest/utils_test/libvirt.py
```diff
@@ -1076,23 +1076,38 @@ def __init__(self):
...
+                                  ignore_status=ignore_status,
+                                  debug=True)
             etime = int(time.time())
             self.mig_time[vm.name] = etime - stime
+            if self.ret.exit_status is not None:
+                logging.debug("Migration to %s returns failed exit status",
+                              desturi)
+                self.RET_LOCK.acquire()
+                self.RET_MIGRATION = False
+                self.RET_LOCK.release()
```
PandaWei: Sorry, I have some objections... 
Here you check the return and set the self.RET_MIGRATION to Flase,
but all of these is same as the following of except.
And the detail including specific error messages.

waynesun09: @PandaWei Yeah, the behavior is the same as fail and caught in except. 
The goal of this change is to get the CmdResult and analyse it in test case to auto process it with detail log mark as a known bug or else.

PandaWei: OK,  I got the intention and have no objections.

virttest/utils_test/libvirt.py
```diff
@@ -1076,23 +1076,38 @@ def __init__(self):
...
         except process.CmdError, detail:
             logging.error("Migration to %s failed:\n%s", desturi, detail)
             self.RET_LOCK.acquire()
```
will-Do: The except here just log the error and then release the lock, so why not use ```try...finally``` block instead

waynesun09: @will-Do the lock release is only for failed situation, which caught in exception if ignore_status as False or ignore_status as True after migrate got error and done. 
If put in finally then need to handle the succeed branch in finally.

dzhengfy:
@will-Do updated for your last comment.

will-Do:
Thx, think we can merge this patch now.

PR #462, opened 2016-04-19
utils_test.libvirt: support return output messages

dzhengfy:
Change the migration function to also return command
output messages which may be useful for the caller to parse.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -2616,14 +2616,15 @@ def do_migration(vm_name, uri, extra, auth_pwd, auth_user="root",
             else:
                 logging.error("The real prompt text: <%s>", text)
                 break
-
+        log = session.get_output()
         session.close()
-        return True
+        return (True, log)
+
     except (aexpect.ShellError, aexpect.ExpectError), details:
         log = session.get_output()
         session.close()
         logging.error("Failed to migrate %s: %s\n%s", vm_name, details, log)
-        return False
+        return (False, log)
 
 
 def update_vm_disk_source(vm_name, disk_source_path,


virttest/utils_test/libvirt.py
```diff
@@ -2576,9 +2576,16 @@ def get_all_vol_paths():
...
-                 su_user="", timeout=30):
+                 su_user="", timeout=30, return_ret_output=False):
     """
     Migrate VM to target host.
+
+    :param  return_ret_output: True to return both result and output,
+            otherwise return only result
+
+    :return: True if migration ok, otherwise False
+    :return: The output message of migration session
```
waynesun09: 2 returns contradict to each other?

since you updated the docstring, pls add all other params together.

virttest/utils_test/libvirt.py
```diff
@@ -2616,13 +2623,18 @@ def do_migration(vm_name, uri, extra, auth_pwd, auth_user="root",
             else:
                 logging.error("The real prompt text: <%s>", text)
                 break
-
+        log = session.get_output()
         session.close()
+        if return_ret_output is True:
+            return (True, log)
         return True
```
waynesun09: I think you don't need add a new param to decide which kind of return you want, only update the function to return a tuple with status and log, then update all tp-libvirt cases with this update. If log not needed in some cases, just accepted it to _.

waynesun09:
s/utils_test\/libvirt:/utils_test.libvirt:/ in commit title

virttest/utils_test/libvirt.py
dzhengfy: @waynesun09 , ok. I searched tp-libvirt, qemu and tp-spice, and no more use of this function. So I will change the return directly.

dzhengfy:
@waynesun09 , updated and thanks for re-checking.

waynesun09:
LGTM now, you need to update tp-libvirt PR to include updates of other cases which be affected by this

dzhengfy:
@waynesun09 ,  as one reply to your comment above, I searched tp-libvirt, qemu and tp-spice, and no more use of this function besides tp-libvirt pr 745 (https://github.com/autotest/tp-libvirt/pull/745).

waynesun09:
@dzhengfy ok, I see other usage of this function, turns out it's not the same function.

ACK

will-Do:
ACK

PR #464, opened 2016-04-20
Check the existence of setup_modules.py before loading it

PandaWei:
If AUTOTEST_PATH has been declared and setup_modules.py is missing,
avocado will throw IOError for both vt_list and vt,
and without any useful details to user.
So, It's better to check if it exists.


PandaWei:
The problem has been discussed at avocado-devel email list.

will-Do:
Good improvement

waynesun09:
ACK

PR #466, opened 2016-04-21
virttest.storage: Correct the log out of backup_image_file

PandaWei:
We'd better to check the existence of src-file first, and then,
output the appropriate log.
Otherwise, the following is likely to mislead user.

Copying ../jeos-21-64.qcow2.backup -> ../jeos-21-64.qcow2.backup
No source file ../jeos-21-64.qcow2.backup, skipping


xutian:
good catch

PandaWei:
Update commit message

will-Do:
ACK

PR #468, opened 2016-04-21
Try more than once to logout a iscsi session.

rbian:
"iscsiadm --logout" may fail, try more times.


waynesun09:
ACK

will-Do:
@rbian 
I just post a new patch for blockcopy, which should fix the logout bug, please check 
https://github.com/autotest/tp-libvirt/pull/765

rbian:
Thanks @will-Do , your patch seems work well. Abandon this patch.

PR #474, opened 2016-04-23
Fix ImportError and AttributeError for both  cmd_hash.py and download_manager.py

PandaWei:
1) virttest.utils_misc: add param for __init__ of LoggingConfig
2) avocado.core.log had been merged into avocado.core.output,
       and there's no configure() any more.
       so replace it with virttest.utils_misc.LoggingConfig.configure().
    
3) "View" had been removed for logging,
     so let the code to work with the new non-View logging approach.

### DIFF
--- a/scripts/cd_hash.py
+++ b/scripts/cd_hash.py
@@ -7,18 +7,20 @@ Program that calculates several hashes for a given CD image.
 
 import os
 import sys
+import logging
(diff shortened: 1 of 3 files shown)


will-Do:
ACK

xutian:
looks good to me, thanks @PandaWei

PR #475, opened 2016-04-25
utils_libvirtd: Add one more kwarg to pass arguments

Hao-Liu:
Current logging_handler don't support passing any parameters. This
would limit the function of handler to send out any informations.

Adding a method to pass arguments to handler will make this possible.

### DIFF
--- a/virttest/utils_libvirtd.py
+++ b/virttest/utils_libvirtd.py
@@ -92,6 +92,7 @@ class LibvirtdSession(object):
 
     def __init__(self, gdb=False,
                  logging_handler=None,
+                 logging_params=(),
                  logging_pattern=r'.*'):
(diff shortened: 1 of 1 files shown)


PandaWei:
LGTM,   I prefer logging_params=None

Hao-Liu:
Hi @PandaWei.
Using `None` as default argument is a measure to avoid using mutable like list or dict as a default value. Since tuple is not mutable, I think there won't be any trouble from this point of view. 

On the contrary, using a tuple as default argument also relieved the effort to testing this argument against `None` before using it, thus making the code more clear.

waynesun09:
ACK

will-Do:
ACK

PR #477, opened 2016-04-27
utils_misc: Bug fix for data size convert

will-Do:
1. As the value of 'scale' here is always positive，so the if...else
    block is unnecessary.
    2. Using 'str' to format a long float number may output a scientific
    notaion number, which is not acceptable here. So control the precision
    by ourself when format it.
    

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2565,11 +2565,12 @@ def normalize_data_size(value_str, order_magnitude="M", factor="1024"):
     from_index = __get_unit_index(unit)
     to_index = __get_unit_index(order_magnitude)
     scale = int(factor) ** (to_index - from_index)
-    if scale > 0:
-        data_size = float(value) / abs(scale)
+    data_size = float(value) / scale
+    # Control precision to avoid scientific notaion
+    if data_size.is_integer():
+        return "%.1f" % data_size
     else:
-        data_size = float(value) * abs(scale)
-    return str(data_size)
+        return ("%.20f" % data_size).rstrip('0')
 
 
 def get_free_disk(session, mount):


waynesun09:
ACK

virttest/utils_misc.py
```diff
@@ -2564,12 +2564,18 @@ def __get_unit_index(M):
         raise ValueError("Invalid data size format 'value_str=%s'" % value_str)
     from_index = __get_unit_index(unit)
     to_index = __get_unit_index(order_magnitude)
-    scale = int(factor) ** (to_index - from_index)
-    if scale > 0:
-        data_size = float(value) / abs(scale)
+    scale = int(factor) ** abs(to_index - from_index)
+    if to_index == from_index:
+        data_size = float(value)
```
xutian: if factor 1000 not is default 1024 ,  results is unexpected.

xutian: And I'm not see anything wrong when convert from small unit to big unit.

will-Do: Yes, your are right, seems I didn't catch the whole meaning here. I will update this patch.

will-Do: @xutian 
>if factor 1000 not is default 1024 , results is unexpected.

This should be the same result as ```(to_index - from_index) = 0```
>And I'm not see anything wrong when convert from small unit to big unit.

Yes, you are right. The problem here is ```scale = int(factor) ** (to_index - from_index)``` will always greater than 0, so it will always go to the following ```if``` block.

virttest/utils_misc.py
```diff
@@ -2565,11 +2565,12 @@ def __get_unit_index(M):
     from_index = __get_unit_index(unit)
     to_index = __get_unit_index(order_magnitude)
     scale = int(factor) ** (to_index - from_index)
-    if scale > 0:
-        data_size = float(value) / abs(scale)
+    data_size = float(value) / scale
+    # Control precision to avoid scientific notaion
+    if data_size.is_integer():
```
xutian: if you want to keep precision why convert it to int here? it's hard to understand

xutian: I think return "%.20f" % data_size is enough.

xutian: And it's not a good idea to update output of this function, because it used by many test, if updated is indispensable, please ensure your update not impact other test.

Thanks,
Xu

xutian:
Agree with drop if .. else .. part.

virttest/utils_misc.py
will-Do: >I think return "%.20f" % data_size is enough.

Not really, %.20 will keep the precision to .20. 
I will use %.1f for integer to return the consistent result/

PandaWei: >>> data = float(1.232323)
>>> print "%.20f" % data
1.23232300000000005724

so the format output is not precise.
Maybe we should consider adopt decimal module here?

will-Do: I tried several ways for fixing this problem an turned out decimal module it not the right way. For now, we have to control the precise by ourselves.

will-Do:
Seems no more comment, I'm gonna to merge this patch

PR #479, opened 2016-04-29
qemu_qtree: add function 'traverse'

luckyh:
Sometimes, we need to get device properties from qtree, but a property
could be put either in device node or its sub-nodes. So add the
'traverse' function to iterate those nodes.


PandaWei:
ACK.

Zhengtong:
Acked-by : Zhengtong <zhengtli@redhat.com>
I think this is a good tree traverse method.

xutian:
Merged, thanks @luckyh @PandaWei and @Zhengtong

PR #485, opened 2016-05-03
scripts.scan_results: introduce '-s' to disable timestamp output

PandaWei:
The timestamp information causes the output exceed the screen line,
and looks a bit messy.
In most cases, we only care about whether it passed or not.
'-s' will leave 'status' and 'result' only.


xutian:
looks good to me, thanks @PandaWei

PR #487, opened 2016-05-04
Bug fix and code update for check_blockjob

will-Do:
For 'bandwidth' checking, the output is different betweent RHEL6 and
RHEl7, so we need normalize the date before compare.
And update the checking logic to make it more clear.


will-Do:
This patch dependent on https://github.com/avocado-framework/avocado-vt/pull/477

will-Do:
Cases are running normally after apply this patch, I'm gonna merge it.

PR #489, opened 2016-05-05
utils_test.libvirt: two iscsi related bugs fix

will-Do:
1. Replace 'localhost' by '127.0.0.1' 
    When ipv6 enabled on the host, iscsiadm discovery target will fail
    with localhost portal, so set the IP address explicitly to avoid this
    problem.
2. Set ip address when create iscsi target
    We need set the ip address explicitly to enable IPv6 when create iscsi target.


Hao-Liu:
LGTM, Ack

waynesun09:
ACK

PR #491, opened 2016-05-06
Fix log message with not enough arguments in utils_net.py

olavphibm:
Regression introduced in commit b0679df699289180287e379834815dcfdbac0f67


will-Do:
ACK

PandaWei:
ACK and Merge, Thanks @olavphibm @will-Do

PR #495, opened 2016-05-09
Update lvm cache timely

will-Do:
In some cases, the new created VG/PV may not clean up completely, which
may lead to a failure when create VG next time:
    # vgcreate vg_test_0 /dev/sdd
    Incorrect metadata area header checksum on /dev/sdd at offset 4096
    Failed to write VG vg_test_0.
In fact, these  commit 2aadb5b tried to fix this problem in an inappropriate
place, and it was removed by commit 03d92e5b without a new fix.

### DIFF
--- a/virttest/staging/lv_utils.py
+++ b/virttest/staging/lv_utils.py
@@ -202,6 +202,9 @@ def vg_create(vg_name, pv_list):
 
     if vg_check(vg_name):
         raise exceptions.TestError("Volume group '%s' already exist" % vg_name)
+    # Update cached state before create VG
+    cmd = "vgscan --cache"
+    result = process.run(cmd)
     cmd = "vgcreate %s %s" % (vg_name, pv_list)
     result = process.run(cmd)
     logging.info(result.stdout.rstrip())
@@ -220,6 +223,9 @@ def vg_remove(vg_name):
     cmd = "vgremove -f %s" % vg_name
     result = process.run(cmd)
     logging.info(result.stdout.rstrip())
+    # Update cached state after remove VG
+    cmd = "vgscane --cache"
+    result = process.run(cmd)
 
 
 def lv_check(vg_name, lv_name):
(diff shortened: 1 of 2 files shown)


virttest/staging/lv_utils.py
```diff
@@ -202,6 +202,9 @@ def vg_create(vg_name, pv_list):
 
     if vg_check(vg_name):
         raise exceptions.TestError("Volume group '%s' already exist" % vg_name)
+    # Update cached state before create VG
+    cmd = "vgscane --cache"
```
Hao-Liu: Is this a typo? vgscan?

virttest/staging/lv_utils.py
```diff
@@ -220,6 +223,9 @@ def vg_remove(vg_name):
     cmd = "vgremove -f %s" % vg_name
     result = process.run(cmd)
     logging.info(result.stdout.rstrip())
+    # Update cached state before create VG
```
Hao-Liu: Change this comment

virttest/utils_test/libvirt.py
```diff
@@ -949,16 +949,12 @@ def pre_pool(self, pool_name, pool_type, pool_target, emulated_image,
             extra = " --source-dev %s" % device_name
         elif pool_type == "logical":
             logical_device = device_name
-            cmd_pv = "pvcreate %s" % logical_device
```
Hao-Liu: Don't we need to do pvcreate any more?

virttest/staging/lv_utils.py
will-Do: Yes, seems I'm completely confusing when compose this patch. Thanks.

virttest/utils_test/libvirt.py
will-Do: Confirmed that the pv will be created automatically when do vgcreate

virttest/staging/lv_utils.py
will-Do: Updated

Hao-Liu:
LGTM now. Ack.

waynesun09:
ACK

will-Do:
Thx all

PR #496, opened 2016-05-10
updated more functions that should be supported by qemu guest agent

Zhengtong:
add guest-sync-delimited guest-set-user-password guest-get-vcpus guest-set-vcpus command  support

id:1328001 1327995 1327990

### DIFF
--- a/virttest/guest_agent.py
+++ b/virttest/guest_agent.py
@@ -8,6 +8,7 @@ import socket
 import time
 import logging
 import random
+import base64
 try:
     import json
 except ImportError:
@@ -16,6 +17,7 @@ except ImportError:
 
 from .qemu_monitor import Monitor, MonitorError
 from . import error_context
+from avocado.utils import process
 
 
 class VAgentError(MonitorError):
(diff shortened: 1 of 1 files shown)


PandaWei:
LGTM

virttest/guest_agent.py
```diff
@@ -497,19 +502,56 @@ def shutdown(self, mode=SHUTDOWN_MODE_POWERDOWN):
...
+        return self.cmd(cmd=cmd, args=args)
+
+    @error_context.context_aware
+    def get_vcpus(self):
+        """
+        Get the vcpus infomation
+        """
+        cmd = "guest-get-vcpus"
+        if not self._has_command(cmd):
+            return False
```
CongLi: I think it's better to raise error here.

Zhengtong: I handle the error in the upper layer (in tp-qemu).  In my opinion, for API, it's better to just give a result.   I add the error handler in https://github.com/autotest/tp-qemu/pull/623

will-Do: Agree

will-Do:
LGTM

CongLi:
Acked-by: Cong Li

PR #506, opened 2016-05-13
libvirt_xml.vm_xml: Use self.vm_name instead of self

PandaWei:
VM's name is better than the object address of VMXMLBase.


will-Do:
Ack

waynesun09:
ACK

PR #516, opened 2016-05-17
qemu_devices.qcontainer: aio=native requried cache.direct=on

xutian:
so fallback aio to 'threads' if snapshot is on and
aio is not 'threads'.

Signed-off-by: XuTian <xutian@redhat.com>

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -1395,6 +1395,9 @@ class DevContainer(object):
         devices[-1].set_param('snapshot', snapshot, bool)
         devices[-1].set_param('readonly', readonly, bool)
         if 'aio' in self.get_help_text():
+            if aio == 'native' and snapshot == 'yes':
+                logging.warn('snapshot is on, fallback aio to threads.')
+                aio = 'threads'
             devices[-1].set_param('aio', aio)
         devices[-1].set_param('media', media)
         devices[-1].set_param('format', imgfmt)


virttest/qemu_devices/qcontainer.py
```diff
@@ -1395,6 +1395,9 @@ def define_hbas(qtype, atype, bus, unit, port, qbus, pci_bus,
         devices[-1].set_param('snapshot', snapshot, bool)
         devices[-1].set_param('readonly', readonly, bool)
         if 'aio' in self.get_help_text():
+            if aio == 'native' and snapshot == 'on':
```
PyLearner: Hi xu,
i apply the patch and try the case , find that it didn't works ,so i check the cfg and find this variable "image_snapshot = yes",so i think it should be snapshot == 'yes', thanks.

xutian: You are right, thanks your comments. and PR updated

PyLearner:
Acked-by : meyang meyang@redhat.com

CongLi:
@xutian 
Maybe your push is not successful.
Could you double check it?
Thanks.

xutian:
really updated, please  click "view full outdated diff" to see difference between old PR  and current PR.

MiriamDeng:
At first,I find the PR doesn't work becasue "image_snapshot = no" in my cfg file.After change it to "yes" the PR works fine.@xutian

pingl2016:
After apply the patch, guests boot up correctly and pass the case at last.
Acked-by: Ping Li pingl@redhat.com

PR #528, opened 2016-05-24
utils_misc: Fix the cleanup for one side setup

dzhengfy:
If we only setup SELinux boolean on one side, we should not clean up the
other side when done. So fix it.


waynesun09:
ACK

will-Do:
ACK

PR #529, opened 2016-05-24
utils_test.libvirt: Some parameters fix

dzhengfy:
1. 'export_dir' should be passed through for nfs setup
2. Add one parameter of the disk image path converted to for the caller
to use.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -2022,6 +2022,7 @@ def set_vm_disk(vm, params, tmp_dir=None, test=None):
     driver_iothread = params.get("driver_iothread", "")
     mnt_path_name = params.get("mnt_path_name", "nfs-mount")
     exp_opt = params.get("export_options", "rw,no_root_squash,fsid=0")
+    exp_dir = params.get("export_dir", "nfs-export")
     first_disk = vm.get_first_disk_devices()
     blk_source = first_disk['source']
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -2022,6 +2022,7 @@ def set_vm_disk(vm, params, tmp_dir=None, test=None):
     driver_iothread = params.get("driver_iothread", "")
     mnt_path_name = params.get("mnt_path_name", "nfs-mount")
     exp_opt = params.get("export_options", "rw,no_root_squash,fsid=0")
+    exp_dir = params.get("export_dir")
```
waynesun09: The exp_dir as None here will break other cases without given this and use default path.

virttest/utils_test/libvirt.py
```diff
@@ -2162,6 +2164,7 @@ def set_vm_disk(vm, params, tmp_dir=None, test=None):
 
         src_file_path = "%s/%s" % (mnt_path, dist_img)
         disk_params_src = {'source_file': src_file_path}
+        params["source_file"] = src_file_path
```
waynesun09: what's this for, update params and nowhere used?

waynesun09: Ok, the params dict is updated and you could use it in the caller.

waynesun09:
ACK

will-Do:
ACK

PR #530, opened 2016-05-25
define the value of 'skip_hash'

PandaWei:
Use 'skip_hash_xx' together with 'md5sum_xx' and 'sha1sum_xx',
and it's value must be 'yes' or 'no', default is 'no'.


will-Do:
ACK

PandaWei:
Thanks @will-Do

PR #535, opened 2016-05-27
Update params for drive-mirror

vivianQizhu:
Signed-off-by: QianqianZhu qizhu@redhat.com
id:1330414

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -959,33 +959,36 @@ class HumanMonitor(Monitor):
             pass
         return backing_file
 
-    def block_mirror(self, device, target, speed, sync, format, mode,
-                     cmd="drive_mirror", correct=True):
+    def block_mirror(self, device, target, sync, cmd="drive_mirror",
+                     correct=True, **kwargs):
         """
         Start mirror type block device copy job
 
-        :param device: device ID
-        :param target: target image
-        :param speed: limited speed, unit is B/s
-        :param sync: full copy to target image(unsupport in human monitor)
-        :param mode: target image create mode, 'absolute-paths' or 'existing'
-        :param format: target image format
+        :param device: device name to operate on
+        :param target: name of new image file
+        :param sync: what parts of the disk image should be copied to the
+                     destination
         :param cmd: block mirror command
         :param correct: auto correct command, correct by default
+        :param kwargs: optional keyword arguments including but not limited to below
+        :keyword Args:
+            format (str): format of target image file
+            mode (str): target image create mode, 'absolute-paths' or 'existing'
+            speed (int): maximum speed of the streaming job, in bytes per second
 
         :return: The command's output
         """
         if correct:
             cmd = self.correct(cmd)
         self.verify_supported_cmd(cmd)
-        args = " %s %s %s" % (device, target, format)
+        args = " %s %s %s" % (device, target, kwargs.get("format", "qcow2"))
         info = str(self.cmd("help %s" % cmd))
-        if (mode == "existing") and "-n" in info:
+        if (kwargs.get("mode", "absolute-paths") == "existing") and "-n" in info:
             args = "-n %s" % args
         if (sync == "full") and "-f" in info:
             args = "-f %s" % args
-        if (speed is not None) and ("speed" in info):
-            args = "%s %s" % (args, speed)
+        if "speed" in info:
+            args = "%s %s" % (args, kwargs.get("speed", ""))
         cmd = "%s %s" % (cmd, args)
         return self.cmd(cmd)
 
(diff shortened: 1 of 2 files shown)


xutian:
Please ID format, please

virttest/qemu_monitor.py
```diff
@@ -1939,20 +1938,18 @@ def get_backingfile(self, device):
...
         :param sync: what parts of the disk image should be copied to the
-                     destination;
-        :param mode: 'absolute-paths' or 'existing'
-        :param format: target image format
+                     destination
         :param cmd: block mirror command
         :param correct: auto correct command, correct by default
+        :param kwargs: optional keyword arguments
 
         :return: The command's output
```
PandaWei: Frankly speaking,
I personally don't like to use **kwargs,  especially for API function to tp-qemu.
It's hard to understand what and how many params needed.
In contrast, The old code and its' doc string  looks more clear for the caller.

vivianQizhu: @PandaWei , since qemu live snapshot command has some option params, and we don't specify each of them every time in our test, so it's no quite suitable to state them all here, however, I agree that it would be better if keep those description here about the optional params. I will modify it later. Thanks.

xutian: There are some many configurable args support for drive_mirror, and these arguments many be changed in feature, so I think kwargs is suitable to use at here, but we need more clear docstring  help caller to uderstand what should be contain in kwargs dict.

PandaWei: Of course,
Kwargs with detailed docstring is acceptable for me.

virttest/qemu_vm.py
```diff
@@ -4127,23 +4127,20 @@ def block_stream(self, device, speed, base=None, correct=True):
...
         :param device: device ID
         :param target: destination image file name;
-        :param speed: max limited speed, default unit is B/s;
         :param sync: what parts of the disk image should be copied to the
                      destination;
-        :param mode: new image open mode
-        :param format: target image format
         :param correct: auto correct cmd, correct by default
         """
         cmd = self.params.get("block_mirror_cmd", "drive-mirror")
```
PandaWei: The default should be 'drive-mirror'.
It's not your fault.
But I think you can fix it.

virttest/qemu_monitor.py
```diff
@@ -1939,20 +1942,27 @@ def get_backingfile(self, device):
             pass
         return backing_file
 
-    def block_mirror(self, device, target, speed, sync, format, mode,
-                     cmd="drive-mirror", correct=True):
+    def block_mirror(self, device, target, sync, cmd="drive-mirror",
```
PandaWei: cmd="drive_mirror"

virttest/qemu_monitor.py
```diff
@@ -1965,13 +1975,12 @@ def block_mirror(self, device, target, speed, sync, format, mode,
...
-            args["format"] = format
-        if speed:
-            args["speed"] = speed
-        return self.cmd(cmd, args)
+        kwargs.update(args)
+        for key, value in kwargs.items():
+            new_key = key.replace('_', '-')
+            del kwargs[key]
+            kwargs.update({new_key: value})
+        return self.cmd(cmd, kwargs)
```
PandaWei: If drop the keys with '_', we don't need to replace them. right?

PandaWei: If drop the keys with '_', we don't need to replace them. right?

vivianQizhu:
@PandaWei @xutian , Thanks. Would you help review my updates please?

PandaWei:
Good enough for me. 
Thanks for you work @vivianQizhu

virttest/qemu_monitor.py
PandaWei: @vivianQizhu 
Do you run any testcase for this PR.
I check this comment again.
Maybe i mislead you.
**hmp uses "drive_mirror",  "drive-mirror" is right for qmp.**

vivianQizhu: @PandaWei ,Thanks for your reminder.
According to my test, both 'drive_mirror' and 'drive-mirror' passed for qmp, I think 'cmd = self.correct(cmd)' line correct the command. Anyway, I should correct it back to 'drive-mirror'.

virttest/qemu_vm.py
```diff
@@ -4127,23 +4127,30 @@ def block_stream(self, device, speed, base=None, correct=True):
...
+                format (str): format of target image file
+                mode (str): target image create mode, 'absolute-paths' or 'existing'
+                speed (int): maximum speed of the streaming job, in bytes per second
+                replaces (str): the block driver node name to replace when finished
+                granularity (int): granularity of the dirty bitmap, in bytes
+                buf_size (int): maximum amount of data in flight from source to target, in bytes
+                on-source-error (str): the action to take on an error on the source
+                on-target-error (str): the action to take on an error on the target
+        """
+        cmd = self.params.get("block_mirror_cmd", "drive_mirror")
```
PandaWei: It's also better to keep 'drive-mirror'.

PandaWei:
I want to merge your PR,
But It seem that you mix  f707d06 and 76346c0 together.

vivianQizhu:
@PandaWei , Thanks. Updated.

PandaWei:
Thanks @vivianQizhu @xutian

PR #539, opened 2016-05-27
qemu_devices.qcontainer: set cache=none while aio=native

morecache:
Since qemu commit d657c0c, aio=native will error out without
cache.direct=on or cache=none.

Signed-off-by: Lin Ma <lma@suse.com>

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -1387,7 +1387,6 @@ class DevContainer(object):
         else:       # very old qemu without 'addr' support
             devices.append(qdevices.QOldDrive(name, use_device))
         devices[-1].set_param('if', 'none')
-        devices[-1].set_param('cache', cache)
         devices[-1].set_param('rerror', rerror)
         devices[-1].set_param('werror', werror)
         devices[-1].set_param('serial', serial)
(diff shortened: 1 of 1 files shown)


xutian:
Will not to fallback to aio=threads, if cache not in ['none', 'directsync']?

morecache:
After commit d657c0c, qemu wont fallback any more in this case.

xutian:
if applied your fixed, aio=native and cache=driectsync will can be tested.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1399,6 +1399,12 @@ def define_hbas(qtype, atype, bus, unit, port, qbus, pci_bus,
                 logging.warn('snapshot is on, fallback aio to threads.')
                 aio = 'threads'
             devices[-1].set_param('aio', aio)
+            if aio == 'native':
+                # Since qemu 2.6, aio=native has no effect without
+                # cache.direct=on or cache=none, It will be error out.
+                # Please refer to qemu commit d657c0c.
+                logging.warn('set cache=none while aio=native')
+                devices[-1].set_param('cache', 'none')
```
xutian: drop line#1390 and add below lines:
```
if aio == 'native':
    cache not in ['none', 'directsync'] and 'none' or cache
devices[-1].set_param('cache', cache)
```
will fix your issue.

Thanks,
Xu

xutian:
looks good to me, thanks @morecache .

PR #542, opened 2016-05-27
Add guest RHEL.6.8 related cfg files

ngu-niny:
Add guest RHEL.6.8 related cfg files after its release.

ID: 1340343


xutian:
looks good to me, thanks @ngu-niny .

PR #548, opened 2016-06-01
ovirt: make get_mac_address adapt to multiple network interfaces

chloerh:
for single network/bridge, nothing changes
for multiple network/bridge, return the MAC address of the first interface to get ip address

### DIFF
--- a/virttest/ovirt.py
+++ b/virttest/ovirt.py
@@ -168,8 +168,15 @@ class VMManager(virt_vm.BaseVM):
         """
         Return MAC address of a VM.
         """
+        vnet_list = self.instance.nics.list()
         try:
-            return self.instance.nics.get(name=net_name).get_mac().get_address()
+            if len(vnet_list) == 1:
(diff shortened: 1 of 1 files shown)


virttest/ovirt.py
```diff
@@ -168,8 +168,15 @@ def get_mac_address(self, net_name='*'):
...
+        vnet_list = self.instance.nics.list()
         try:
-            return self.instance.nics.get(name=net_name).get_mac().get_address()
+            if len(vnet_list) == 1:
+                return self.instance.nics.get(name=net_name).get_mac().get_address()
+            # Multiple network interfaces
+            elif len(vnet_list) > 1:
+                for vnet in vnet_list:
+                    if self.address_cache.get(vnet.get_mac().get_address()):
+                        return vnet.get_mac().get_address()
```
will-Do: This still return one address, right?

chloerh: Yes.Only one address is needed to connect to the guest.

will-Do: Okay

PR #550, opened 2016-06-01
virttest.qemu_monitor: fix sytax error when pick monitor from env

xutian:

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -165,7 +165,7 @@ class Monitor:
     DATA_AVAILABLE_TIMEOUT = 0
     CONNECT_TIMEOUT = 60
 
-    def __init__(self, vm, name, filename):
+    def __init__(self, vm, name, filename, suppress_exceptions=False):
         """
         Initialize the instance.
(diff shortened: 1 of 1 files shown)


virttest/qemu_monitor.py
```diff
@@ -222,7 +222,7 @@ def __setstate__(self, state):
     def __getinitargs__(self):
         # Save some information when pickling -- will be passed to the
         # constructor upon unpickling
-        return self.vm, self.name, self.filename
+        return self.vm, self.name, self.filename, True
```
PandaWei: You set the default of suppress_exception to False.
but the getinitargs return True.
Is it OK? I'm not sure ... Just looks odd.

xutian: yes, it looks really a little odd. let me update it. thanks

xutian:
@PandaWei  updated, thanks.

PandaWei:
ACK

CongLi:

xutian:
thanks @PandaWei and @CongLi , merged.

PR #552, opened 2016-06-02
utils_net: Fix referenced before assignment error

dzhengfy:


xutian:
looks ok to me, thanks @dzhengfy

virttest/utils_net.py
```diff
@@ -1420,8 +1422,8 @@ def set_guest_ip_addr(session, mac, ip_addr,
 
     :return: True if set up guest ip successfully.
     """
+    info_cmd = ""
     try:
```
PandaWei: Does it suffer from "referenced before assignment" here?

xutian: this one no impact.  move this line out of try block is fine, because this line can't raise any exception.

PR #553, opened 2016-06-02
qemu_storage: fix CmdError for snapshot_del

pingl2016:
As the function is called during negative test and command will not return is expected. So set the ignore_status to avoid raising a exception for process.run().

Signed-off-by: Ping Li pingl@redhat.com

### DIFF
--- a/virttest/qemu_storage.py
+++ b/virttest/qemu_storage.py
@@ -335,20 +335,20 @@ class QemuImg(storage.QemuImg):
 
         :note: params should contain:
                snapshot_image_name -- the name of snapshot image file
+        :return: class: CmdResult object.
         """
 
-        cmd = self.image_cmd
-        if self.snapshot_tag:
-            cmd += " snapshot -d %s" % self.snapshot_image_filename
-        else:
+        if not self.snapshot_tag:
             raise exceptions.TestError("Can not find the snapshot image"
                                        " parameters")
         if blkdebug_cfg:
-            cmd += " blkdebug:%s:%s" % (blkdebug_cfg, self.image_filename)
+            cmd = ("%s snapshot -d %s blkdebug:%s:%s" % (self.image_cmd,
+                   self.snapshot_image_filename, blkdebug_cfg,
+                   self.image_filename))
         else:
-            cmd += " %s" % self.image_filename
-
-        process.system_output(cmd)
+            cmd = ("%s snapshot -d %s %s" % (self.image_cmd,
+                   self.snapshot_image_filename, self.image_filename))
+        return process.run(cmd, shell=True)
 
     def snapshot_list(self):
         """


pingl2016:
ID: 1341920

virttest/qemu_storage.py
```diff
@@ -335,6 +335,7 @@ def snapshot_del(self, blkdebug_cfg=""):
 
         :note: params should contain:
                snapshot_image_name -- the name of snapshot image file
+        :return: An :class:`CmdResult` object.
```
xutian: :return: class:`CmdResult` object.

pingl2016: @xutian correct it.

virttest/qemu_storage.py
```diff
@@ -348,7 +349,7 @@ def snapshot_del(self, blkdebug_cfg=""):
         else:
             cmd += " %s" % self.image_filename
 
-        process.system_output(cmd)
+        return process.run(cmd, ignore_status=True)
```
PandaWei: You set ignore_status inside of snapshot_del.
It's better to check no bad influence on all of it's caller.

pingl2016: @PandaWei Thanks for reminding. update the patch to reduce the impact for positive test.

virttest/qemu_storage.py
```diff
@@ -345,10 +346,10 @@ def snapshot_del(self, blkdebug_cfg=""):
                                        " parameters")
         if blkdebug_cfg:
             cmd += " blkdebug:%s:%s" % (blkdebug_cfg, self.image_filename)
+            return process.run(cmd, ignore_status=True)
```
PandaWei: You set ignore_status=True, which makes process.run() *won't* raise Exception any more.
I know it's your intention.

PandaWei: But you try to catch CmdError in autotest/tp-qemu#654
try:
     cmd_result = image_io.snapshot_del(blkdebug_cfg=blkdebug_cfg)
except error.CmdError, err:
     ....

PandaWei: I'm confused...
You want to supress Exception here.
But you catch it in test case.

If I'm wrong, Please point it out to me. Thanks in advance.

pingl2016: as i commented in the  autotest/tp-qemu#654, you misunderstand the function of this patch.

PandaWei: OK, I was completely confused this PR and autotest/tp-qemu#654.
IMO,
image_io.snapshot_del(blkdebug_cfg=blkdebug_cfg) won't trigger any CmdErrror **with this PR**.
It's meaningless that catch Exception **in testcase** for it.

Maybe some guys in redhat understand all of it and can review this.
Let's wait for their's response.

xutian: logical looks  ugly, make a small path to fix it:
```
diff --git a/virttest/qemu_storage.py b/virttest/qemu_storage.py
index 8225fad..885cd9f 100644
--- a/virttest/qemu_storage.py
+++ b/virttest/qemu_storage.py
@@ -336,19 +336,16 @@ class QemuImg(storage.QemuImg):
         :note: params should contain:
                snapshot_image_name -- the name of snapshot image file
         """
+        if not (bool(self.snapshot_tag) ^ bool(self.blkdebug)):
+            raise exceptions.TestError("Can not find the snapshot image parameters")
 
-        cmd = self.image_cmd
-        if self.snapshot_tag:
-            cmd += " snapshot -d %s" % self.snapshot_image_filename
-        else:
-            raise exceptions.TestError("Can not find the snapshot image"
-                                       " parameters")
         if blkdebug_cfg:
-            cmd += " blkdebug:%s:%s" % (blkdebug_cfg, self.image_filename)
+            cmd = "%s blkdebug:%s:%s" % (self.image_cmd,
+                blkdebug_cfg, self.snapshot_image_filename)
         else:
-            cmd += " %s" % self.image_filename
-
-        process.system_output(cmd)
+            cmd = "%s snapshot -d %s %s" % (self.snapshot_image_filename,
+                self.snapshot_image_filename)
+        return process.run(cmd, shell=True)
 
     def snapshot_list(self):
         """
```

PR #554, opened 2016-06-02
virttest.autotest: Fix timeout when running autotest

gogoxiaoxiao:
The parameter timeout is defined as function argument.
Here, it alway change timeout to 600s, and will cause some
jobs ended prematurely in error:

INFO | Running autotest control file ctcs.control on guest, timeout 600s
...
ERROR| ERROR ... TestError: Timeout elapsed while waiting for job to complete

Signed-off-by: Haishuang Yan <yanhaishuang@cmss.chinamobile.com>

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -960,9 +960,9 @@ def run_autotest(vm, session, control_path, timeout,
             tarball_url = ("https://codeload.github.com/autotest/"
                            "autotest-client-tests/tar.gz/master")
             tarball_url = params.get("client_test_url", tarball_url)
-            timeout = int(params.get("download_tests_timeout", "600"))
+            tests_timeout = int(params.get("download_tests_timeout", "600"))
             tests_tarball = os.path.join(data_dir.get_tmp_dir(), "tests.tgz")
-            download.url_download(tarball_url, tests_tarball, timeout=timeout)
+            download.url_download(tarball_url, tests_tarball, timeout=tests_timeout)
             copy_if_hash_differs(vm, tests_tarball, tests_tarball)
             extract(vm, tests_tarball, tests_dir)
             os.remove(tests_tarball)


PandaWei:
Are you sure the TestError introduced by 'timeout=timeout' ?
I doubt it....

gogoxiaoxiao:
Yes, I'm sure.

See following code:
def run_autotest(vm, session, control_path, timeout,
                 outputdir, params, copy_only=False, control_args=None,
                 ignore_session_terminated=False):
...
            timeout = int(params.get("download_tests_timeout", "600"))
            tests_tarball = os.path.join(data_dir.get_tmp_dir(), "tests.tgz")
            download.url_download(tarball_url, tests_tarball, timeout=timeout)
...

I think the argument "timeout" in the function run_autotest doesn't take effect, because it's reassigned value by "download_tests_timeout".

PandaWei:
I noticed the timeout is reassigned,
Indeed, It's a problem.  you fix is good in this point.

But my doubt is that the reassignment just increases the timeout,
It's not the cause of the TestError you mentioned in commit message.

xutian:
It's really a bug. how about make download timeout less that 1/2 of test timeout.

gogoxiaoxiao:
Hi Panda Wei

I run autotest case with following config, so the reassignment decrease test timeout(3900-600):
 
        - ctcs:
            # If you think this is too lengthy, please change the cerberus
            # control file and set this timeout appropriately.
            test_timeout = 3900
            test_control_file = ctcs.control

PandaWei:
OK,  It seems test_timeout is decreased.
Can you show the full name of case failed?
Let me have a try to reproduce it.

PandaWei:
does the case work well with your fix now?

gogoxiaoxiao:
Yes, it works now in my side.
You can reproduce the issue with the special test case:

avocado run --vt-type libvirt io-github-autotest-qemu.autotest.ctcs --vt-guest-os Linux.CentOS.7.0.x86_64.i440fx

PandaWei:
ACK. 
Good catch，Thanks @gogoxiaoxiao

PandaWei:
Just for advice.
Generally we choose 'module name: blah blah blah ' as title of commit.
There's no virttest/autotest . so your commit looks a little odd.
Just for an example,


[root@localhost avocado-vt]# git log ./virttest/utils_test/__init__.py
commit 0313a9e5b840cae5eeef8346fbbfc68834833348
Author: Xu Tian <xutian@redhat.com>
Date:   Fri Apr 15 15:44:08 2016 +0800

    virttest.utils_test: fix setting error in run_autotest
    
    generate default mini-configuration for autotest
    client tests in guest OS to avoid setting error.
    
gogoxiaoxiao:
@PandaWei Thanks for your reviewing.
I will change the commit message in v2.

gogoxiaoxiao:
@PandaWei 
I have modified the title, please review, thanks.

PandaWei:
Good enough for me

PandaWei:
@xutian do you have any other comments ?

xutian:
looks ok to me, thanks @gogoxiaoxiao and @PandaWei . merged

PR #558, opened 2016-06-07
add new functions for guest agent to use by tp-qemu

Zhengtong:
These functions are about : set-time, get-time, network-interface, fstrim

ID:1327985 1327988 1327992 1327998

### DIFF
--- a/virttest/guest_agent.py
+++ b/virttest/guest_agent.py
@@ -64,6 +64,17 @@ class VAgentCmdError(VAgentError):
                 "error message: %r)" % (self.ecmd, self.eargs, self.edata))
 
 
+class VAgentCmdNotSupportedError(VAgentError):
+
+    def __init__(self, cmd):
+        VAgentError.__init__(self)
+        self.ecmd = cmd
+
+    def __str__(self):
+        return("The command %s is not supported by the current version qga"
+               % self.ecmd)
+
+
 class VAgentSyncError(VAgentError):
 
     def __init__(self, vm_name):
@@ -309,7 +320,7 @@ class QemuAgent(Monitor):
             self._supported_cmds = [None]
             logging.warn("Could not get supported guest agent cmds list")
 
-    def _has_command(self, cmd):
+    def check_has_command(self, cmd):
         """
         Check wheter guest agent support 'cmd'.
 
@@ -319,7 +330,7 @@ class QemuAgent(Monitor):
         """
         # Initiate supported cmds list if it's empty.
         if not self._supported_cmds:
-            self.get_supported_cmds()
+            self._get_supported_cmds()
 
         # If the first element in supported cmd list is 'None', it means
         # autotest fails to get the cmd list, so bypass cmd checking.
@@ -328,7 +339,7 @@ class QemuAgent(Monitor):
 
         if cmd and cmd in self._supported_cmds:
             return True
-        return False
+        raise VAgentCmdNotSupportedError(cmd)
 
     def _log_command(self, cmd, debug=True, extra_str=""):
         """
@@ -477,7 +488,7 @@ class QemuAgent(Monitor):
         Make sure the guest agent is responsive by sending a command.
         """
         cmd = "guest-ping"
-        if self._has_command(cmd):
+        if self.check_has_command(cmd):
             self.cmd(cmd=cmd, debug=False)
 
     @error_context.context_aware
@@ -491,9 +502,7 @@ class QemuAgent(Monitor):
                  'shutdown' is unsupported.
         """
         cmd = "guest-shutdown"
-        if not self._has_command(cmd):
-            return False
-
+        self.check_has_command(cmd)
         args = None
         if mode in [self.SHUTDOWN_MODE_POWERDOWN, self.SHUTDOWN_MODE_REBOOT,
                     self.SHUTDOWN_MODE_HALT]:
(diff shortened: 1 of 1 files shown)


virttest/guest_agent.py
```diff
@@ -552,6 +552,50 @@ def set_vcpus(self, action):
         return self.cmd(cmd=cmd, args=action)
 
     @error_context.context_aware
+    def get_time(self):
```
xutian: it looks you need a function like verify_support commands and call it before really execute the command. verify_supported should raise CommandNotSupport error if command not support by current guest agent.

Thanks,
Xu

Zhengtong: Thanks for comment. I think this implementation may still be ok , because "__has_command()"in these  APIs already packaged functions which determine if the command is supported. if add a new function to wraps these again, it won't save so much codes, and bring more call relationships.

virttest/utils_misc.py
```diff
@@ -3470,6 +3470,30 @@ def check_device_driver(pci_id, driver_type):
     return driver == driver_type
 
 
+def get_guest_os_datetime(vm, os_type):
```
xutian: view from you function name, this function should be return a float or a datetime type values.

virttest/guest_agent.py
xutian: __has_command is a private function, if some day we want to add some case which base on your test class, it will not callable.

virttest/utils_misc.py
Zhengtong: In the __doc__ of the function , it "returns the seconds since the epoch", I think int type is enough, may be the name is confusing as you said, but I think that's ok.  Any suggestions about the function name?

xutian: function name is good, but return type looks not so good.
I think if output is invalid please raise test.Error not return False directly. 
And please follow avocado doc string style to add your doc-string.

virttest/guest_agent.py
xutian: If command not support, how tester know it, because your always return False when command not exists.
Why not raise exception to told tester command not really support by current guest-agent package.
And better to raise exception when exception occur, don't  delay to resolve  or hidden in test. if did not it, will wast tester's time in analysis.

Zhengtong: If the command is not supported for this guest-agent version ,it will raise the error in test case (tp-qemu). the error was handled there.

virttest/utils_misc.py
Zhengtong: will update later

Zhengtong: updated

virttest/guest_agent.py
```diff
@@ -552,6 +552,50 @@ def set_vcpus(self, action):
...
+        """
+        cmd = "guest-fstrim"
+        if not self._has_command(cmd):
+            return False
+        return self.cmd(cmd)
+
+    @error_context.context_aware
+    def get_network_interface(self):
+        """
+        Issue the cmd "guest-network-get-interfaces" to the guest agen
```
PandaWei: I like doc to be more specific ,just like "Get list of guest IP addresses, MAC addresses ......",
You can refer the help info of the command.

virttest/guest_agent.py
```diff
@@ -552,6 +552,50 @@ def set_vcpus(self, action):
...
+            return False
+        if nanoseconds:
+            args = {"time": nanoseconds}
+        return self.cmd(cmd=cmd, args=args)
+
+    @error_context.context_aware
+    def fstrim(self):
+        """
+        Issue the cmd "guest-fstrim" to the guest agent and get the result
+        """
```
PandaWei: The same as below

virttest/guest_agent.py
```diff
@@ -552,6 +552,50 @@ def set_vcpus(self, action):
...
+        cmd = "guest-get-time"
+        if not self._has_command(cmd):
+            return False
+        return self.cmd(cmd=cmd)
+
+    @error_context.context_aware
+    def set_time(self, nanoseconds=None):
+        """
+        set the time of guest, the params passed in is in nanoseconds
+        """
```
PandaWei: This one is good

PandaWei: This one is good

virttest/guest_agent.py
Zhengtong: Updated. thanks

PandaWei:
LTGM

PandaWei:
LGTM

XueqiangWei:
LGTM,
Acked-by XueqiangWei <xuwei@redhat.com>

virttest/guest_agent.py
xutian: if that, all test which call your function need to test is the function return is False or a Dict, it's a bad design.

Zhengtong: Updated . please have a review , thanks

XueqiangWei:
LGTM,
Acked-by XueqiangWei xuwei@redhat.com

virttest/utils_misc.py
```diff
@@ -3474,6 +3474,30 @@ def check_device_driver(pci_id, driver_type):
     return driver == driver_type
 
 
+def get_guest_os_datetime(session, os_type):
```
xutian: easy way to get seconds since epoch:

python -c "import datetime; print (datetime.datetime(2012,04,01,0,0) - datetime.datetime(1970,1,1)).total_seconds()"

Zhengtong:
Thanks for advice , updated

xutian:
Merged, thanks

PR #561, opened 2016-06-10
Fix passing args & kargs for remote commander.

Andrei-Stepanov:


xutian:
looks good to me, thanks @Andrei-Stepanov

PR #560, opened 2016-06-08
virsh: add parameter 'safe' for cmd_status_output

chloerh:
After aexpect's update to 1.2.0

PR #563, opened 2016-06-12
virttest.virsh: Fix syntax error

dzhengfy:
In function cmd_status of  aexpect/client.py, a new parameter 'safe' is
added. The function invokes function cmd_status_output in
virttest/virsh.sh which does not accept the newly added parameter, so it
causes TypeError: cmd_status_output() takes at most 5 arguments (6
given). The fix is to add this parameter to function cmd_status_output
with default value certainly.


will-Do:
Same with https://github.com/avocado-framework/avocado-vt/pull/560, merge this one as this's a more complete fix

PR #569, opened 2016-06-14
virttest.virt_vm: optimize wait_for_serial_login function

xutian:

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -6,6 +6,8 @@ import re
 import socket
 import traceback
 
+from aexpect.exceptions import ShellError
+from aexpect.exceptions import ExpectError
 from avocado.core import exceptions
 
 from . import utils_misc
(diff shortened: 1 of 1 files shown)


xutian:
@PandaWei  would you like to help review this PR. 

Thanks,
Xu

PandaWei:
You want to distinguish LoginError and xxxError introduced by restarting network , and interrupt and raise error if the latter occurs?

xutian:
@PandaWei , yes, I think current logical looks more clearly and more easy to understand and after this update there will no such many useless debug information recording in log-file during guest bootup.

virttest/virt_vm.py
```diff
@@ -1206,33 +1208,28 @@ def wait_for_serial_login(self, timeout=LOGIN_WAIT_TIMEOUT,
...
+                    utils_net.restart_guest_network(session, os_type=os_type)
                 return session
             except remote.LoginError, e:
+                exception = e
                 self.verify_alive()
-                e = str(e)
-                if e not in error_messages:
-                    logging.debug(e)
-                    error_messages.append(e)
-            time.sleep(2)
```
PandaWei: Here I want keep to output the error messages, and sleep 2.
error message is useful and 0.5 is short.

wait_for_login has the same code with wait_for_serial_login.
It's necessary to keep consistence.

virttest/virt_vm.py
```diff
@@ -1206,33 +1208,28 @@ def wait_for_serial_login(self, timeout=LOGIN_WAIT_TIMEOUT,
...
-                    os_type = self.params.get("os_type")
-                    try:
-                        logging.debug("Attempting to restart guest network")
-                        utils_net.restart_guest_network(session,
-                                                        os_type=os_type)
-                    except Exception:
-                        pass
+                    logging.debug("Attempting to restart guest network")
+                    os_type = self.params.get('os_type')
+                    utils_net.restart_guest_network(session, os_type=os_type)
```
PandaWei: Here catching shellError and ExceptError  looks more clear

virttest/virt_vm.py
```diff
@@ -1206,33 +1208,28 @@ def wait_for_serial_login(self, timeout=LOGIN_WAIT_TIMEOUT,
         :params virtio: is a virtio console.
         :return: A ShellSession object.
         """
-        error_messages = []
         logging.debug("Attempting to log into '%s' via serial console "
                       "(timeout %ds)", self.name, timeout)
-        end_time = time.time() + timeout
+        end_time, exception = time.time() + timeout, None
         while time.time() < end_time:
             try:
```
PandaWei: Does need to reset exception to None?

xutian: var exception just used to keep the latest Exception during login VM.

virttest/virt_vm.py
xutian: do you mean use nested try...except ?

virttest/virt_vm.py
xutian: Agree with keep sleep time to 2s. keep error message looks useless I think information we only need to care is exception raised in latest.
Will open a new PR to update wait_for_login function to keep consistence.

Thanks,
Xu

virttest/virt_vm.py
PandaWei: Yes, That's why I ask to reset exception to None.
If the "N-1" time failed and set exception, but "N" time success,
the exception is not correct.
 I'm right?

virttest/virt_vm.py
PandaWei: Yes

virttest/virt_vm.py
xutian: If N time sucess, session will return and exception the last line will not be execute.

PandaWei:
I'm OK for this PR.

xutian:
Merged, thanks @PandaWei

PR #571, opened 2016-06-15
virttest: Use result instead of result_obj

PandaWei:
process.CmdError has a result attribute, not result_obj.


xutian:
Acked

xutian:
Merged, thanks @PandaWei

PR #572, opened 2016-06-15
utils_v2v: remove operations of booting windows guests

chloerh:
Since all windows guests are activated, remove code of comparing
screenshots and other operations on booting windows guests except
for win2003 guests.

### DIFF
--- a/virttest/utils_v2v.py
+++ b/virttest/utils_v2v.py
@@ -531,7 +531,7 @@ class WindowsVMCheck(VMCheck):
         to boot windows.
         """
         logging.info("Booting Windows in %s seconds", timeout)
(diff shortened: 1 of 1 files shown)


virttest/utils_v2v.py
```diff
@@ -531,11 +531,10 @@ def boot_windows(self, timeout=300):
         to boot windows.
         """
         logging.info("Booting Windows in %s seconds", timeout)
-        compare_screenshot_vms = ["win2003", "win2008", "win2008r2", "win7"]
```
will-Do: Maybe just update the list is enough, then we can keep the following codes

chloerh:
@will-Do updated, please review, thanks

PR #581, opened 2016-06-20
Return for remote commander

Andrei-Stepanov:


virttest/remote_commander/remote_runner.py
```diff
@@ -26,6 +26,13 @@
 import messenger as ms
 
 
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+handler = logging.FileHandler("/tmp/remote_runner.log", "w", encoding=None,
```
xutian: How about get temp log file like below:
tempfile.mktemp(suffix='.log', prefix='remote_runner', dir=data_dir.get_tmp_dir()),
so that your log will not be overwrite.

Andrei-Stepanov:
@xutian : cannot use data_dir module in remote_runner as it is copied to remote VM. And it runs in its own, doesn't knowing anything about Avocado.

+ I updated to use tempfile.mktemp

xutian:
thanks your explain, update looks good to me. thanks @Andrei-Stepanov

Andrei-Stepanov:
May I ask what are rules for acceptance of pull-requests?
This pull request is almost month old.
Who is responsible to accept pull requests?
Inconvenience about using your own repo for running tests is very upsetting.

virttest/remote_commander/remote_runner.py
```diff
@@ -768,19 +791,31 @@ def remote_agent(in_stream_cls, out_stream_cls):
...
-        fd_stdout = sys.stdout.fileno()
-        fd_stderr = sys.stderr.fileno()
-        fd_stdin = sys.stdin.fileno()
+        sys.stdout.flush()
+        sys.stderr.flush()
+        fd_stdin = sys.stdin.fileno()    # == 0
+        fd_stdout = sys.stdout.fileno()  # == 1
+        fd_stderr = sys.stderr.fileno()  # == 2
+        orig_stdout = os.dup(fd_stdout)  # Original stdout to master.
+        orig_stderr = os.dup(fd_stderr)  # Original srderr to master.
```
will-Do: Seems this orig_stderr not used
I'm not familiar  with remote runner, so just check this patch code here, no more comment from me.

Andrei-Stepanov: Yes.
All SSH encoded base64 traffic goes through stdout.
Executed program/script has modified stdout/stderr.
stderr/stdout from remote_runner are endcoded and passed through original SSH stdout.
So.. in general stdout from SSH encapsulates stdout/stderr from program.
What you have pointed is not a bug. We can safely remove this variable. But...
I am sure that this code is a bit "tricky", so I just left this variable to represent what is really going here.
I can remove it. But it is safe to keep it for demonstrativeness. Do you want me to remove this variable ?

PandaWei:
In fact, I have read your PR a few weeks ago.
No found any problem.
But you know It's not enough to just use my eyes to examine the code, unless it's a trivial fix.

Only VM.commander and VM.remote_commander will use remote_runner.py.
It's used not quite often and I didn't find any existing case to verify your codes.
That's why I didn't reply to you.  and leave this PR to someone else who is more familiar with remote runner.

so if you can provide some test result or validation method to us next time, maybe you will get quick response.

virttest/remote_commander/remote_runner.py
```diff
@@ -726,7 +741,7 @@ def python_file_run_with_helper(self, test_path):
         assert hasattr(module, "run")
         run = getattr(module, "run")
         helper = Helper(self)
-        run(helper)
+        return run(helper)
```
PandaWei: return code is OK,
ACK commit 983d58985a34b31a1a0ca843ff548782cb3125c5

xutian:
This PR got two ACK,  so merged it. thanks @Andrei-Stepanov , @PandaWei

virttest/remote_commander/remote_runner.py
will-Do: Yest, it's really a tricky point, keep it is fine.

PR #584, opened 2016-06-21
Two iscsi related bugs fix

will-Do:
1.
    utils_test.libvirt: Try to get volume repeatedly
    
    For iscsi type pool, after it start, the volume may not appear
    immediately, and the volume path(cooresponding to a scsi disk) also
    need verify.

2
    iscsi: Get the right attached disk name
    
    For now, the regex pattern may match to another target disk if the
    corresponding target disk not attached yet.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -362,7 +362,7 @@ def check_blockjob(vm_name, target, check_point="none", value="0"):
 
     try:
         cmd_result = virsh.blockjob(
-            vm_name, target, "--info", ignore_status=True)
+            vm_name, target, "--info", debug=True, ignore_status=True)
         output = cmd_result.stdout.strip()
         err = cmd_result.stderr.strip()
         status = cmd_result.exit_status
@@ -2081,15 +2081,9 @@ def set_vm_disk(vm, params, tmp_dir=None, test=None):
                          emulated_image=emu_image,
                          image_size=image_size)
             # Get volume name
-            cmd_result = virsh.vol_list(pool_name)
-            try:
-                vol_name = re.findall(r"(\S+)\ +(\S+)[\ +\n]",
-                                      str(cmd_result.stdout))[1][0]
-            except IndexError:
-                raise exceptions.TestError("Fail to get volume name in "
-                                           "pool %s" % pool_name)
-            emulated_path = virsh.vol_path(vol_name, pool_name,
-                                           debug=True).stdout.strip()
+            vols = get_vol_list(pool_name)
+            vol_name = vols.keys()[0]
+            emulated_path = vols[vol_name]
         else:
             # Setup iscsi target
             if is_login:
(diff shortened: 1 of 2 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -2814,3 +2808,33 @@ def update_polkit_rule(params, pattern, new_value):
...
+    :return: A dict include volumes' name(key) and path(value).
+    """
+    poolvol = libvirt_storage.PoolVolume(pool_name=pool_name)
+    vols = utils_misc.wait_for(poolvol.list_volumes, timeout,
+                               text='Waitting for volume show up')
+    if not vol_check:
+        return vols
+
+    # Check volume name
+    if not vols:
```
xutian: It's better to move this check before return vols. so that exception will raise if wait_for timeout, else TypeError will raise if wait_for timeout occur.

Thanks,
Xu

will-Do: Sorry, I didn't get your point, wait_for timeout will only return ```None```, and that's what we want

virttest/utils_test/libvirt.py
```diff
@@ -2081,15 +2081,9 @@ def set_vm_disk(vm, params, tmp_dir=None, test=None):
...
-            try:
-                vol_name = re.findall(r"(\S+)\ +(\S+)[\ +\n]",
-                                      str(cmd_result.stdout))[1][0]
-            except IndexError:
-                raise exceptions.TestError("Fail to get volume name in "
-                                           "pool %s" % pool_name)
-            emulated_path = virsh.vol_path(vol_name, pool_name,
-                                           debug=True).stdout.strip()
+            vols = get_vol_list(pool_name)
+            vol_name = vols.keys()[0]
```
xutian: if get_vol_list return None, TypeError will raise in this line.

virttest/utils_test/libvirt.py
xutian: see comments in line#2085.

virttest/utils_test/libvirt.py
will-Do: The logic here is: if vol is None, then get_vol_list() will raise TestError inside, unless vol_check is False, but here we use the default value, and it will never trigger TypeError

xutian: oh, right. thanks your explain.

virttest/utils_test/libvirt.py
kylazhang: @will-Do xutian means line2831 and line2827 should exchange, and I think it is reasonable.

waynesun09:
ACK

virttest/utils_test/libvirt.py
will-Do: @kylazhang 
No, here 'vol_check' is a flag to indicate if we need raise Exception. If vol_check if False, then the function should return directly without any check.

kylazhang:
Others looks good to me.

virttest/utils_test/libvirt.py
kylazhang: @will-Do OK, then it will get error when you get the keys. Is that expected?

kylazhang: @will-Do Get your points, you allow the function return null for other applications judgement. thanks and no problems here.

will-Do:
Thanks all, I'm gonna merge this

PR #587, opened 2016-06-23
utils_test.libvirt: Add extra option for virsh

dzhengfy:
Add an extra option used after virsh, like 'virsh -c xxx migrate xxx'.


waynesun09:
ACK

will-Do:
ACK

PR #600, opened 2016-06-29
Clean dup params

suqinhuang:
Both session and vm params are used in setup_win_driver_verifier and clear_win_driver_verifier,
drop session, as it can be get with vm

### DIFF
--- a/virttest/utils_test/qemu.py
+++ b/virttest/utils_test/qemu.py
@@ -133,13 +133,13 @@ def pin_vm_threads(vm, node):
         logging.info("Skip pinning, no enough nodes")
 
 
-def check_driver_verifier(session, driver, timeout=300):
+def _check_driver_verifier(session, driver, timeout=300):
     """
     Check driver verifier status
-    : param session: VM session
-    : param driver: The driver need to query
-    : param timeout: Timeout in seconds
-    : return: driver verifier status
+
+    :param session: VM session.
+    :param driver: The driver need to query
+    :param timeout: Timeout in seconds
     """
     logging.info("Check %s driver verifier status" % driver)
     query_cmd = "verifier /querysettings"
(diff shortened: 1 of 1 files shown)


suqinhuang:
@xutian could you review this patch, thanks

virttest/utils_test/qemu.py
```diff
@@ -148,46 +149,57 @@ def check_driver_verifier(session, driver, timeout=300):
...
-        error_context.context("Enable %s driver verifier" % driver,
-                              logging.info)
-        verifier_setup_cmd = "verifier /standard /driver %s.sys" % driver
-        session.cmd(verifier_setup_cmd,
-                    timeout=timeout,
-                    ignore_all_errors=True)
-        session = vm.reboot(session)
-        verifier_status, output = check_driver_verifier(session, driver)
+    try:
+        session = vm.wait_for_login(timeout=timeout)
```
CongLi: How about serial login?
As there are several nic related cases.

PandaWei: It's really bad idea to get session in the internal of both setup_win_driver_verifier  and  clear_win_driver_verifier. 
Either wait_for_login or serial log takes a long time.

I think Just let theirs caller call wait_for_login one time is better than this.

suqinhuang: boot up guest take long time, login should be quick.

the session of setup_win_driver_verifier and clear_win_driver_verifier may be changed during the test, could not use the same session.

PandaWei: If it's true that the session may be changed during test, I think the PR can be accepted.
If not, It's really unnecessary to login seperately

suqinhuang: setup_win_driver_verifier and clear_win_driver_verifier is used for any virtio-win test, it's normal that session is changed during the test

CongLi:
Could you please update the commit log which is not clear ?

virttest/utils_test/qemu.py
```diff
@@ -148,46 +149,57 @@ def check_driver_verifier(session, driver, timeout=300):
...
+            verifier_setup_cmd = "verifier /standard /driver %s.sys" % driver
+            session.cmd(verifier_setup_cmd,
+                        timeout=timeout,
+                        ignore_all_errors=True)
+            session = vm.reboot(session)
+            verifier_status, output = check_driver_verifier(session, driver)
+            if not verifier_status:
+                msg = "%s verifier is not enabled, details: %s" % (driver,
+                                                                   output)
             raise exceptions.TestFail(msg)
```
xutian: syntax issue.

virttest/utils_test/qemu.py
```diff
@@ -148,46 +149,57 @@ def check_driver_verifier(session, driver, timeout=300):
...
-    logging.info("%s verifier is enabled already" % driver)
+        logging.info("%s verifier is enabled already" % driver)
+    finally:
+        session.close()
```
xutian: syntax issue will caused when wait_for_login raise exception.

suqinhuang:
PASS 1-smp_1.4096m.repeat1.run_test.Host_RHEL.m7.u2.rng_random.qcow2.virtio_blk.up.virtio_net.Win7.x86_64.sp1.io-github-autotest-qemu.rng_bat

virttest/utils_test/qemu.py
```diff
@@ -148,46 +149,57 @@ def check_driver_verifier(session, driver, timeout=300):
...
+    try:
+        verifier_status = check_driver_verifier(session, driver)[1]
+        if verifier_status:
+            logging.info("Clear driver verifier")
+            verifier_clear_cmd = "verifier /reset"
+            session.cmd(verifier_clear_cmd,
+                        timeout=timeout,
+                        ignore_all_errors=True)
+            session = vm.reboot(session)
+    except Exception:
```
xutian: why ignore above exception?

suqinhuang:
@xutian updated, could you review again

virttest/utils_test/qemu.py
```diff
@@ -136,10 +136,11 @@ def pin_vm_threads(vm, node):
...
     Check driver verifier status
-    : param session: VM session
-    : param driver: The driver need to query
-    : param timeout: Timeout in seconds
-    : return: driver verifier status
+
+    :param session: VM session.
+    :param driver: The driver need to query
+    :param timeout: Timeout in seconds
+    :return: driver verifier status
```
pingl2016: I think it's better "return: Tuple (driver verifier status, output)"

virttest/utils_test/qemu.py
```diff
@@ -148,46 +149,55 @@ def check_driver_verifier(session, driver, timeout=300):
...
-            msg = "%s verifier is not enabled, details: %s" % (driver, output)
-            raise exceptions.TestFail(msg)
-    logging.info("%s verifier is enabled already" % driver)
+            error_context.context("Enable %s driver verifier" % driver,
+                                  logging.info)
+            verifier_setup_cmd = "verifier /standard /driver %s.sys" % driver
+            session.cmd(verifier_setup_cmd,
+                        timeout=timeout,
+                        ignore_all_errors=True)
+            session = vm.reboot(session)
```
CongLi: vm.reboot() is better, which will get session by itself.

suqinhuang: session is updated and need to run session.close() finally.

xutian:
Acked

virttest/utils_test/qemu.py
pingl2016: @CongLi I think session is needed, or it will login again before reboot vm.

pingl2016:
Acked-by: Ping Li pingl@redhat.com

PR #604, opened 2016-06-30
Update aexpect version requirement

olavphibm:
An extra argument "safe" was added to aexpect.ShellSession.cmd_output(),
which is used in virttest/virsh.py. It will break in earlier aexpect versions.


lmr:
OK, thanks.

PR #611, opened 2016-07-06
Create target for dir type pool

will-Do:
If the given pool target is not a valid path, create it in the tmp dir.


kylazhang:
ACK.

will-Do:
Thx, merge it now

PR #619, opened 2016-07-13
utils_v2v: retry 5 times to get driver info for windows guest

chloerh:
To avoid problem that occationally timeout when getting return data
from 'driverquery' command

### DIFF
--- a/virttest/utils_v2v.py
+++ b/virttest/utils_v2v.py
@@ -606,8 +606,23 @@ class WindowsVMCheck(VMCheck):
         cmd = "DRIVERQUERY"
         if signed:
             cmd += " /SI"
-        status, output = self.session.cmd_status_output(cmd)
-        return self.run_cmd(cmd)[1]
+        # Try 5 times to get driver info
+        output, count = '', 5
+        while count > 0:
+            logging.debug('%d times remaining for getting driver info' % count)
+            try:
+                output = self.session.cmd_output(cmd)
+            except Exception as detail:
+                logging.error(detail)
+                count -= 1
+                # Clean up output
+                self.session.cmd('cls')
+            else:
+                break
+        if not output:
+            logging.error('Fail to get driver info')
+        logging.debug("Command output:\n%s", output)
+        return output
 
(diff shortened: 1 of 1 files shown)


will-Do:
ACK

virttest/utils_v2v.py
```diff
@@ -606,8 +606,23 @@ def get_driver_info(self, signed=True):
...
+                logging.error(detail)
+                count -= 1
+                # Clean up output
+                self.session.cmd('cls')
+            else:
+                break
+        if not output:
+            logging.error('Fail to get driver info')
+        logging.debug("Command output:\n%s", output)
+        return output
```
PandaWei: Use utils_misc.wait_for(func, ) to deal with such situation maybe better than given a fixed operation times.

chloerh: But utils_misc.wait_for  only deals with functions which needs more time to get result. In this situation, command could fail once or twice. During my test, the correct result most comes from the second run, sometimes the 3rd, therefore retrying several times might be necessary.

PandaWei: It's not really that utils_misc.wait_for **only** deals with functions which need more time to get result.
We can make use of wait_for to check the cases that doesn't need consume time by decreasing the parameter ‘step‘.
And you need less than 3 times to get right result. maybe other guys needs more times, or just only once.
Anyhow, you try a fixed times can fix your problem.  That's just my advice.

chloerh: Thanks for the advice. : )  Sorry I haven't expressed clearly, this command will fail and raise error, after catching this error, output stream should be cleaned, otherwise it would mix with the next output of command, and utils_misc.wait_for may not actually meet the need.

will-Do: IIRC, utils_misc.wait_for always band with time, and seems here the problem is command return time is unpredictable.

PandaWei:
@chloerh  clean the output stream makes sense.   ACK

PandaWei:
Merge it with double ACKs from @will-Do and Me

PR #621, opened 2016-07-14
iscsi: Add discovery for iscsi target export function.

rbian:
Libvirt don't discovery iscsi target automatically currently, it
lead to fail to create iscsi pool. Add discovery step in target
export function.


will-Do:
Yes, I also met this failure before.
BTW, you need resolve the Travis CI failure.

rbian:
Thanks @will-Do , fixed.

will-Do:
ACK and merge this pr

PR #635, opened 2016-07-22
utils_test: Fix RemoteVMManager

dzhengfy:
1. Fix checking rsa private and public key on remote host instead of
local host because it is to execute the logon guest on remote host
instead of local host

2. Add some ssh option to avoid host key checking
3. Return the command execution result if required
4. Enable to avoid raising an exception when command execution occurs an
error
5. Fix syntax error

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -1947,13 +1947,20 @@ class RemoteVMManager(object):
         """
         pri_key = '~/.ssh/id_rsa'
         pub_key = '~/.ssh/id_rsa.pub'
-        if os.path.exists(pri_key) and os.path.exists(pub_key):
-            logging.info("SSH key pair already exist")
+        # Check the private key and public key file on remote host.
+        cmd = "ls %s %s" % (pri_key, pub_key)
+        result = self.runner.run(cmd, ignore_status=True)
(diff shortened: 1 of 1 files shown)


virttest/utils_test/__init__.py
```diff
@@ -2004,15 +2011,32 @@ def check_network(self, vm_ip, count=5, timeout=60):
...
+            will return CmdResult object. False, raise an exception.
+
+        :raise: exceptions.TestFail, if the command fails
+        :return: CmdResult object
         """
-        cmd = "ssh %s@%s '%s'" % (vm_user, vm_ip, command)
+        ssh_options = "%s %s" % ("-o UserKnownHostsFile=/dev/null",
+                                 "-o StrictHostKeyChecking=no")
+        cmd = 'ssh %s %s@%s "%s"' % (ssh_options, vm_user, vm_ip, command)
+        cmdres = None
```
waynesun09: s/cmdres/ret/g

dzhengfy: @waynesun09 , updated and please help checking.

waynesun09:
ACK

dzhengfy:
@will-Do , could you help review too?

will-Do:
ACK

PR #636, opened 2016-07-25
virttest: rework update_mac_ip_address function

xutian:
skip "IndexError" if a interface isn't belong the VM (eg,
virbr0, sit0 and so on). Update IP pool only status of the
interface is up. Add ipv6 support.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3062,6 +3062,88 @@ def get_correspond_ip(remote_ip):
     return local_ip
 
 
+def get_linux_mac(session, nic):
+    """
+    Get MAC address by nic name
+    """
+    sys_path = "%s/%s" % (SYSFS_NET_PATH, nic)
+    pattern = "(\w{2}:\w{2}:\w{2}:\w{2}\:\w{2}:\w{2})"
+    if session.cmd_status("test -d %s" % sys_path) == 0:
+        mac_index = 1
+        show_mac_cmd = "cat %s/address" % sys_path
+        out = session.cmd_output(show_mac_cmd)
+    else:
+        pattern = "(ether|HWaddr) %s" % pattern
+        mac_index = 2
+        show_mac_cmd = "ifconfig %s || ip link show %s" % (nic, nic)
+        out = session.cmd_output(show_mac_cmd)
+    try:
+        return re.search(pattern, out, re.M | re.I).group(mac_index)
+    except Exception:
+        logging.error("No HWaddr/ether found for nic %s: %s" % (nic, out))
+
+
+def get_linux_ipaddr(session, nic):
+    """
+    Get IP addresses by nic name
+    """
+    cmd = "ifconfig %s || ip address show %s" % (nic, nic)
+    out = session.cmd_output(cmd)
+    try:
+        inet = re.findall("inet (\S+)", out, re.M)[0]
+    except IndexError:
+        inet = None
+    inet6 = re.findall("inet6 (\S+)", out, re.M)
+    if not inet6:
+        inet6 = [None]
+    return (inet, inet6)
+
+
+def windows_mac_ip_maps(session):
+    """
+    Windows get MAC IP addresses maps
+    """
+    maps = {}
+    cmd = "wmic nicconfig where IPEnabled=True get ipaddress, macaddress"
+    out = session.cmd_output(cmd)
+    for line in out.splitlines()[1:]:
+        line = line.split(',')
+        inets = line[1].lstrip('{').rstrip('}').split(';')
+        try:
+            inet = inets[0]
+        except IndexError:
+            inet = None
+        inet6 = inets[1:]
+        if not inet6:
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -2620,6 +2620,86 @@ def get_free_mem(session, os_type):
...
+    """
+    Get guest MAC IP addresses maps
+    """
+    os = session.cmd_output("echo %OS%")
+    if "Win" in os:
+        maps = windows_mac_ip_maps(session)
+    else:
+        maps = linux_mac_ip_maps(session)
+    return maps
+
```
PandaWei: Why don't we put them in utils_net.py?
They're are related to network.

xutian:
@PandaWei  updated, would you like to review it, thanks

PandaWei:
Actually, I like this PR. it divides the lengthy function into small and short functions.
That's good. And I haven't find any problem.

xutian:
@CongLi   can you help review this PR, thanks.

virttest/utils_net.py
```diff
@@ -3062,6 +3062,88 @@ def get_correspond_ip(remote_ip):
...
+        maps[mac] = get_linux_ipaddr(session, nic)
+    return maps
+
+
+def get_mac_ip_maps(session):
+    """
+    Get guest MAC IP addresses maps
+    """
+    os = session.cmd_output("echo %OS%")
+    if "Win" in os:
```
CongLi: Why not use os_type directly?

virttest/utils_net.py
```diff
@@ -3062,6 +3062,88 @@ def get_correspond_ip(remote_ip):
...
+        out = session.cmd_output(show_mac_cmd)
+    else:
+        pattern = "(ether|HWaddr) %s" % pattern
+        mac_index = 2
+        show_mac_cmd = "ifconfig %s || ip link show %s" % (nic, nic)
+        out = session.cmd_output(show_mac_cmd)
+    try:
+        return re.search(pattern, out, re.M | re.I).group(mac_index)
+    except Exception:
+        logging.error("No HWaddr/ether found for nic %s: %s" % (nic, out))
```
CongLi: Why not raise error here?

virttest/utils_net.py
```diff
@@ -3062,6 +3062,88 @@ def get_correspond_ip(remote_ip):
...
+    """
+    cmd = "ifconfig %s || ip address show %s" % (nic, nic)
+    out = session.cmd_output(cmd)
+    try:
+        inet = re.findall("inet (\S+)", out, re.M)[0]
+    except IndexError:
+        inet = None
+    inet6 = re.findall("inet6 (\S+)", out, re.M)
+    if not inet6:
+        inet6 = [None]
```
CongLi: Why inet6 is a list, but inet not?

virttest/utils_net.py
```diff
@@ -3118,55 +3200,31 @@ def update_mac_ip_address(vm, params, timeout=None):
...
-        # _ip, mac are in different sequence in Fedora and RHEL guest.
-        if re.match(".\d+\.\d+\.\d+\.\d+", mac):
-            _ip, mac = mac, _ip
+    maps = get_mac_ip_maps(session)
+    if len(maps) < len(vm.virtnet):
+        logging.warn("Not all nic get IP address, restart networking")
+        default_cmd = "service network restart"
+        restart_network = params.get("restart_network", default_cmd)
+        session.cmd(restart_network, timeout=240)
+        maps = get_mac_ip_maps(session)
```
CongLi: Need judge  if len(maps) < len(vm.virtnet) again after restart network, right?
it may be a while() here instead of if ?

virttest/utils_net.py
xutian: it depends on params,  I think judge from guest session is independent.

virttest/utils_net.py
xutian: right,  better to try twice in if network some nic still can't get a IP address.

PandaWei: IMHO, Now we're checking code on paper. we don't need to worry about the things not happen.
repetition can be simply used to fix this kind of problem. but it's really bad solution

PandaWei: Keeping codes clean and simple makes sense.

virttest/utils_net.py
PandaWei: I think no need raise error here.
if no match mac, it will return none.
and Its' caller don't mind the none return value

virttest/utils_net.py
xutian: one nic can assign more ipv6 address but only one ipv4 address.

xutian:
@CongLi  any question and suggestion, if no let's merged this PR.

xutian:
@CongLi  and @PandaWei  thanks your comments. And if no new comments on this PR, I will merged it in tomorrow.

Thanks,
Xu

PR #642, opened 2016-08-01
Update devcon path

suqinhuang:
devcon.exe is copied to C:  in xcopy option, so C:devcon.exe is available,
but  C:\devcon.exe is used when run devcon.exe cmd, so always show not available cmd
update C: to C:\ in xcopy


xutian:
@suqinhuang  please update your commit log, it's hard to understand.  view from your code, it's a typo fix in devcon copy command.

suqinhuang:
@xutian updated, could you review again

xutian:
Acked. thanks @suqinhuang

PR #645, opened 2016-08-04
utils_test.libvirt: Do other function during migration

dzhengfy:
Sometimes there is a need to execute some commands when the migration is
in progress. So this enables the capability.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1151,7 +1151,8 @@ class MigrationTest(object):
                 self.RET_LOCK.release()
 
     def do_migration(self, vms, srcuri, desturi, migration_type, options=None,
-                     thread_timeout=60, ignore_status=False):
+                     thread_timeout=60, ignore_status=False, func=None,
+                     func_params=None):
         """
         Migrate vms.
 
@@ -1159,6 +1160,12 @@ class MigrationTest(object):
         :param srcuri: local uri, used when migrate vm from remote to local
         :param descuri: remote uri, used when migrate vm from local to remote
         :param migration_type: do orderly for simultaneous migration
+        :param options: migration options
+        :param thread_timeout: time out seconds for the migration thread running
+        :param ignore_status: determine if an exception is raised for errors
+        :param func: the function executed during migration thread is running
+        :param func_params: used by func
+
         """
         if migration_type == "orderly":
             for vm in vms:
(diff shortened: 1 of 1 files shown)


balamuruhans:
@dzhengfy, I tried to test your patch - placed it in "./local/lib/python2.7/dist-packages/avocado_plugins_vt-38.0-py2.7.egg/virttest/utils_test/libvirt.py" but still I am observing "<type 'exceptions.AttributeError'>: 'module' object has no attribute 'do_migration_asynch'"

balamuruhans:
@dzhengfy Thanks

balamuruhans:
ACK

virttest/utils_test/libvirt.py
```diff
@@ -1150,6 +1150,54 @@ def thread_func_migration(self, vm, desturi, options=None,
...
+        migration_thread = threading.Thread(target=self.thread_func_migration,
+                                            args=(vm, desturi, options,
+                                                  ignore_status))
+        stime = int(time.time())
+        migration_thread.start()
+        return (migration_thread, stime)
+
+    def check_migration_result(self, migration_thread, start_time,
+                               thread_timeout=60):
+        """
```
kylazhang: I don't think it is suitable for using this func name, the function is used for waiting migration finish, so what about migration_thread_wait?

kylazhang:
@dzhengfy  What about add a function parameter in do_migration? It may reduce the code duplication.

will-Do:
@dzhengfy 
How do you think about kyla's suggestion?

dzhengfy:
@kylazhang @will-Do , updated. Please check again. Thanks.

kylazhang:
ACK.

virttest/utils_test/libvirt.py
```diff
@@ -1151,22 +1151,36 @@ def thread_func_migration(self, vm, desturi, options=None,
                 self.RET_LOCK.release()
 
     def do_migration(self, vms, srcuri, desturi, migration_type, options=None,
-                     thread_timeout=60, ignore_status=False):
+                     thread_timeout=60, ignore_status=False, func=None,
+                     func_params=None):
```
will-Do: It's okay to add optional parameters here. But you should consider to use kwargs to avoid the parameter numbers become more and more.

will-Do:
ACK

PR #648, opened 2016-08-05
virttest.nfs: set ssh automatic recover

dzhengfy:
This is to enable the ability to set the ssh connection automatic
recovery.


kylazhang:
ACK. @will-Do

will-Do:
ACK

PR #656, opened 2016-08-11
vittest.test_setup: Change to use function process.system to install egd.pl

ngu-niny:
The original used build.make function could not run command on a subshell, so change to use function process.system with parameter 'shell=True' instead.

ID: 1359005


xutian:
Acked

CongLi:

xutian:
Merged, thanks @ngu-niny  and @CongLi

PR #657, opened 2016-08-11
virttest: add Q35 machine type with bridge devices support

xutian:

### DIFF
--- a/shared/cfg/guest-hw.cfg
+++ b/shared/cfg/guest-hw.cfg
@@ -5,9 +5,13 @@ variants:
     - rtl8139:
         no ppc64 ppc64le
         nic_model = rtl8139
+        q35:
+            nic_pci_bus = pci_bridge
     - e1000:
         no ppc64 ppc64le
         nic_model = e1000
+        q35:
+            nic_pci_bus = pci_bridge
     - e1000-82540em:
         no ppc64 ppc64le
         nic_model = e1000-82540em
@@ -28,6 +32,8 @@ variants:
         # following lines to enable the vhost support ( only available
         # for tap )
         #netdev_extra_params = ",vhost=on"
+        q35:
+            nic_pci_bus = pcie_switch
         enable_msix_vectors = yes
         whql.submission.device.net:
             test_device = VirtIO Ethernet Adapter$
@@ -60,6 +66,7 @@ variants:
 
 variants:
     - ide:
+        no q35
         drive_format=ide
         cd_format=ide
     - scsi:
@@ -71,6 +78,7 @@ variants:
         i440fx:
             cd_format=ide
         q35:
+            hba_pci_bus = pcie_switch
             cd_format=ahci
         # Add -drive ...boot=yes unless qemu-kvm is 0.12.1.2 or newer
         # then kvm_vm will ignore this option.
@@ -86,6 +94,8 @@ variants:
         cd_format=scsi-cd
         scsi_hba=virtio-scsi-pci
         libvirt_controller=virtio-scsi
+        q35:
+            hba_pci_bus = pcie_switch
     - spapr_vscsi:
         drive_format=scsi-hd
         cd_format=scsi-cd
@@ -98,6 +108,7 @@ variants:
         cd_format=scsi-cd
         scsi_hba=lsi53c895a
     - ahci:
+        only q35
         drive_format=ahci
         cd_format=ahci
     - usb2:
@@ -311,28 +322,27 @@ variants image_backend:
 
 variants:
     - @no_virtio_rng:
-    - rng_random:
+    - virtio_rng:
         no Host_RHEL.m5 Host_RHEL.m6.u0 Host_RHEL.m6.u1 Host_RHEL.m6.u2 Host_RHEL.m6.u3 Host_RHEL.m6.u4 Host_RHEL.m6.u5
-        virtio_rngs += " rng0"
+        virtio_rngs = "rng0"
         #max-bytes_virtio-rng-pci =
         #period_virtio-rng-pci =
-        backend_rng0 = rng-random
-        backend_type = passthrough
-        filename_passthrough = /dev/random
-    - rng_egd:
-        no Host_RHEL.m5 Host_RHEL.m6.u0 Host_RHEL.m6.u1 Host_RHEL.m6.u2 Host_RHEL.m6.u3 Host_RHEL.m6.u4 Host_RHEL.m6.u5
-        virtio_rngs += "rng0"
-        #max-bytes_virtio-rng-pci =
-        #period_virtio-rng-pci =
-        backend_rng0 = rng-egd
-        backend_type = chardev
-        chardev_backend = socket
-        socket_type = tcp
-        host_tcp = localhost
-        port_tcp = 8000
-        # if egd.pl started on remote host, please set setup_egd = no
-        setup_egd = yes
-
+        q35:
+            rng_pci_bus = pcie_switch
+        variants:
+             - rng_random:
+                 backend_rng0 = rng-random
+                 backend_type = passthrough
+                 filename_passthrough = /dev/random
+             - rng_egd:
+                 backend_rng0 = rng-egd
+                 backend_type = chardev
+                 chardev_backend = socket
+                 socket_type = tcp
+                 host_tcp = localhost
+                 port_tcp = 8000
+                 # if egd.pl started on remote host, please set setup_egd = no
+                 setup_egd = yes
 variants:
     - @default_bios:
     - ovmf:
--- a/shared/cfg/machines.cfg
+++ b/shared/cfg/machines.cfg
@@ -16,6 +16,14 @@ variants:
             cd_format = ahci
         enable_pvpanic = yes
         ioport_pvpanic = 0x505
+        pci_controllers = root_port pcie_switch dmi2pci_bridge pci_bridge
+        pci_bus = pci.0
+        type_root_port = ioh3420
+        type_pcie_switch = x3130
+        pci_bus_pcie_switch = root_port
+        type_dmi2pci_bridge = i82801b11-bridge
+        type_pci_bridge = pci-bridge
+        pci_bus_pci_bridge = dmi2pci_bridge
     - @pseries:
         only ppc64, ppc64le
         machine_type = pseries
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -384,6 +384,17 @@ class VM(virt_vm.BaseVM):
                 return fmt % (option, str(value))
             return ""
 
+        def _get_pci_bus(params, name=None):
+            """
+            Get device pattern pci bus.
+
+            :param params: device params
+            :param name: deivce type name
+            """
+            name = name and "%s_pci_bus" % name or "pci_bus"
+            pci_bus = {"aobject": params.get(name, "pci.0")}
+            return pci_bus
+
         # Wrappers for all supported qemu command line parameters.
         # This is meant to allow support for multiple qemu versions.
         # Each of these functions receives the output of 'qemu -help'
@@ -1244,7 +1255,7 @@ class VM(virt_vm.BaseVM):
         have_ahci = False
         have_virtio_scsi = False
         virtio_scsi_pcis = []
-        pci_bus = {'aobject': params.get('pci_bus', 'pci.0')}
+        pci_bus = _get_pci_bus(params)
 
         # init value by default.
         # PCI addr 0,1,2 are taken by PCI/ISA/IDE bridge and the GPU.
@@ -1459,7 +1470,8 @@ class VM(virt_vm.BaseVM):
                     if params.get('machine_type').startswith("arm64-mmio"):
                         dev = QDevice('virtio-serial-device')
                     else:
-                        dev = QDevice('virtio-serial-pci', parent_bus=pci_bus)
+                        parent_bus = _get_pci_bus(port_params, "virtio_port")
+                        dev = QDevice('virtio-serial-pci', parent_bus=parent_bus)
                     dev.set_param('id',
                                   'virtio_serial_pci%d' % no_virtio_serial_pcis)
                     devices.insert(dev)
@@ -1468,7 +1480,8 @@ class VM(virt_vm.BaseVM):
                     if params.get('machine_type').startswith("arm64-mmio"):
                         dev = QDevice('virtio-serial-device')
                     else:
-                        dev = QDevice('virtio-serial-pci', parent_bus=pci_bus)
+                        parent_bus = _get_pci_bus(port_params, "virtio_port")
+                        dev = QDevice('virtio-serial-pci', parent_bus=parent_bus)
                     dev.set_param('id', 'virtio_serial_pci%d' % i)
                     devices.insert(dev)
                     no_virtio_serial_pcis += 1
@@ -1487,13 +1500,15 @@ class VM(virt_vm.BaseVM):
 
         # Add virtio-rng devices
         for virtio_rng in params.objects("virtio_rngs"):
-            virtio_rng_params = params.object_params(virtio_rng)
-            add_virtio_rng(devices, virtio_rng_params, pci_bus)
+            rng_params = params.object_params(virtio_rng)
+            rng_bus = _get_pci_bus(rng_params, "rng")
+            add_virtio_rng(devices, rng_params, rng_bus)
 
         # Add logging
         devices.insert(StrDev('isa-log', cmdline=add_log_seabios(devices)))
         if params.get("anaconda_log", "no") == "yes":
-            add_log_anaconda(devices, pci_bus)
+            parent_bus = _get_pci_bus(params, "anacondalog")
+            add_log_anaconda(devices, parent_bus)
 
         # Add USB controllers
         usbs = params.objects("usbs")
@@ -1543,6 +1558,8 @@ class VM(virt_vm.BaseVM):
                     if self.last_boot_index > 0:
                         image_boot = False
                     self.last_boot_index += 1
+            image_pci_bus = _get_pci_bus(image_params, "hba")
+            image_params["pci_bus"] = image_pci_bus.get("aobject", "pci.0")
             devs = devices.images_define_by_params(image_name, image_params,
                                                    'disk', index, image_boot,
                                                    image_bootindex)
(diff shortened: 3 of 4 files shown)


xutian:
ID: 1339038

xutian:
my qemu command line:
```
MALLOC_PERTURB_=1  /usr/libexec/qemu-kvm \
    -S  \
    -name 'avocado-vt-vm1'  \
    -sandbox off  \
    -machine q35  \
    -nodefaults  \
    -vga cirrus \
    -device ioh3420,id=root_port,bus=pcie.0,addr=02 \
    -device x3130-upstream,id=pcie_switch,bus=root_port,addr=00 \
    -device i82801b11-bridge,id=dmi2pci_bridge,bus=pcie.0,addr=03 \
    -device pci-bridge,id=pci_bridge,bus=dmi2pci_bridge,addr=01,chassis_nr=1  \
    -chardev socket,id=qmp_id_qmpmonitor1,path=/var/tmp/avocado_JEOyFB/monitor-qmpmonitor1-20160811-180259-DaHAxqZT,server,nowait \
    -chardev socket,id=qmp_id_catch_monitor,path=/var/tmp/avocado_JEOyFB/monitor-catch_monitor-20160811-180259-DaHAxqZT,server,nowait \
    -device pvpanic,ioport=0x505,id=idzqfncb  \
    -chardev socket,id=serial_id_serial0,path=/var/tmp/avocado_JEOyFB/serial-serial0-20160811-180259-DaHAxqZT,server,nowait \
    -device isa-serial,chardev=serial_id_serial0 \
    -object rng-random,filename=/dev/random,id=passthrough-UbX1MLV6 \
    -device xio3130-downstream,bus=pcie_switch,id=pcie_switch.0,addr=00,chassis=1 \
    -device virtio-rng-pci,id=virtio-rng-pci-sVzluko7,rng=passthrough-UbX1MLV6,bus=pcie_switch.0,addr=00,disable-legacy=off,disable-modern=on  \
    -chardev socket,id=seabioslog_id_20160811-180259-DaHAxqZT,path=/var/tmp/avocado_JEOyFB/seabios-20160811-180259-DaHAxqZT,server,nowait \
    -device isa-debugcon,chardev=seabioslog_id_20160811-180259-DaHAxqZT,iobase=0x402 \
    -device ich9-usb-ehci1,id=usb1,addr=1d.7,multifunction=on,bus=pcie.0 \
    -device ich9-usb-uhci1,id=usb1.0,multifunction=on,masterbus=usb1.0,addr=1d.0,firstport=0,bus=pcie.0 \
    -device ich9-usb-uhci2,id=usb1.1,multifunction=on,masterbus=usb1.0,addr=1d.2,firstport=2,bus=pcie.0 \
    -device ich9-usb-uhci3,id=usb1.2,multifunction=on,masterbus=usb1.0,addr=1d.4,firstport=4,bus=pcie.0 \
    -drive id=drive_image1,if=none,snapshot=off,aio=native,cache=none,format=qcow2,file=/usr/share/avocado/data/avocado-vt/images/RHEL-Server-7.3-64-virtio.qcow2 \
    -device xio3130-downstream,bus=pcie_switch,id=pcie_switch.1,addr=01,chassis=2 \
    -device virtio-blk-pci,id=image1,drive=drive_image1,bootindex=0,bus=pcie_switch.1,addr=00,disable-legacy=off,disable-modern=on \
    -device xio3130-downstream,bus=pcie_switch,id=pcie_switch.2,addr=02,chassis=3 \
    -device virtio-net-pci,mac=9a:6c:6d:6e:6f:70,id=idIZmBTY,vectors=4,netdev=idxrfUAc,bus=pcie_switch.2,addr=00,disable-legacy=off,disable-modern=on  \
    -netdev tap,id=idxrfUAc,vhost=on,vhostfd=19,fd=18 \
    -m 4096  \
    -smp 2,maxcpus=2,cores=1,threads=1,sockets=2  \
    -cpu 'Penryn',+kvm_pv_unhalt \
    -vnc :0  \
    -rtc base=utc,clock=host,driftfix=slew  \
    -boot order=cdn,once=c,menu=off,strict=off \
    -enable-kvm

```

shared/cfg/machines.cfg
```diff
@@ -16,6 +16,14 @@ variants:
...
         enable_pvpanic = yes
         ioport_pvpanic = 0x505
+        pci_controllers = root_port pcie_switch dmi2pci_bridge pci_bridge
+        pci_bus = pci.0
+        type_root_port = ioh3420
+        type_pcie_switch = x3130
+        pci_bus_pcie_switch = root_port
+        type_dmi2pci_bridge = i82801b11-bridge
+        type_pci_bridge = pci-bridge
+        pci_bus_pci_bridge = dmi2pci_bridge
```
ldoktor: The q35 machine worked without this for many years so I don't think it should be part of the default definition. How about creating another type like `q35-with-bridge` and use it for this specific definition?

virttest/qemu_vm.py
```diff
@@ -2080,11 +2084,9 @@ def add_disable_modern(devices, dev, dev_type):
         for balloon_device in params.objects("balloon"):
             params_balloon = params.object_params(balloon_device)
             balloon_devid = params_balloon.get("balloon_dev_devid")
-            balloon_bus = None
             use_ofmt = params_balloon.get("balloon_use_old_format",
                                           "no") == "yes"
-            if params_balloon.get("balloon_dev_add_bus") == "yes":
-                balloon_bus = pci_bus
+            balloon_bus = {"aobject": params_balloon.get("balloon_bus", "pci.0")}
```
ldoktor: do you need this granularity? I think simple `pci_bus` should suffice. After that you can tweak some of the actual devices to a different bus by `pci_bus_$id`. Or do you expect the frequent need to put all the ballon devices to a different bus then the hba devices or the usb controllers?

xutian: Right, according feature owner requirement we need to support add device to different bus. eg. virtio-blk-pci can be connect to pcie.0 and  pcie-switch downstream port or pci-bridge and so on, so I have above change.

shared/cfg/machines.cfg
xutian: Of course, I can create a new type and move this part of configuration to it

xutian:
@ldoktor  updated PR according your comments, would you help review it. thanks very much.

shared/cfg/guest-hw.cfg
```diff
@@ -70,8 +77,10 @@ variants:
         drive_format=virtio
         i440fx:
             cd_format=ide
-        q35:
+        q35, q35-with-bridge:
```
ldoktor: Oh I didn't notice it'd sprawl so extensively, perhaps I was being to careful. I'm wondering what the q35 would look like in libvirt, do you have any news whether they're going to do similar setup? If (and only if) they are then we can probably just redefine the q35 and avoid the many new branches.

ldoktor: I looked at virtmanager in F23 and it creates a root complex + pci-pci bridge. So I think it's indeed a standard and the default q35 should create it. I'm sorry for confusion and I say nothing against that part anymore. Can you please revert it?

virttest/qemu_vm.py
ldoktor: Well you can do that by using `pci_bus_image1` or `pci_bus_nic1`. But I see that bounding those might make sense as you might want to put all disks to one pci_bus, the nic interfaces to other and so on. Still the idea of `nic_bus` and `ballon_bus` looks prone to clashes. How about `ballon_pci_bus`, `nic_pci_bus` and so on? I think that would be clearer.

virttest/qemu_vm.py
```diff
@@ -1487,8 +1489,9 @@ def add_disable_modern(devices, dev, dev_type):
 
         # Add virtio-rng devices
         for virtio_rng in params.objects("virtio_rngs"):
-            virtio_rng_params = params.object_params(virtio_rng)
-            add_virtio_rng(devices, virtio_rng_params, pci_bus)
+            rng_params = params.object_params(virtio_rng)
+            rng_bus = {"aobject": rng_params.get("rng_bus", "pci.0")}
```
ldoktor: I can see a pattern. How about using a function:

```python
def _get_pci_bus(name, params):
    return {"aobject": params.get(name + "_pci_bus", "pci.0")}

...
rng_bus = _get_pci_bus("rng", rng_params)
...
virtio_bus = _get_pci_bus("port", port_params)
```

virttest/qemu_vm.py
```diff
@@ -1459,7 +1459,8 @@ def add_disable_modern(devices, dev, dev_type):
                     if params.get('machine_type').startswith("arm64-mmio"):
                         dev = QDevice('virtio-serial-device')
                     else:
-                        dev = QDevice('virtio-serial-pci', parent_bus=pci_bus)
+                        parent_bus = {"aobject": port_params.get("port_bus", "pci.0")}
```
ldoktor: how about `vio_port_pci_bus` (there are many kind of ports...)

ldoktor: Also there is a virtio_serial_port added in `add_log_anaconda`. Shouldn't that get the same treatment?

ldoktor:
Hello @xutian, I really welcome this patch. Sorry for the initial confusion with the defaults, now I believe the default `q35` should be replaced with this new layout as it's indeed a new standard.

As for this patch, I'd explicitly use the format `${type}_pci_bus` to avoid clashes/confusions, so the `nic_bus` should IMO be `nic_pci_bus`. There are few remaining places with default handling, which should not prevent this from inclusion as those devices can for now end up in the main `pci.0` bus.

Would you please update this PR so I can re-test it's not breaking anything? I'm glad avocado's q35 is getting attention again.

xutian:
@ldoktor   thanks your comments and PR updated, would you like to help review it.

Thanks,
Xu

shared/cfg/guest-hw.cfg
```diff
@@ -28,6 +32,8 @@ variants:
         # following lines to enable the vhost support ( only available
         # for tap )
         #netdev_extra_params = ",vhost=on"
+        q35:
+            nic_pci_bus = pcie_switch
```
ldoktor: The only issue with this is, that if the test defined other nic:

```
nics += " test_nic"
nic_model_test_nic = e1000
```
it'd have been added into the `pcie_switch`. I'm not saying it's a problem, we should just check the existing (up/down-stream) configs for this.

An alternative solution would be not to depend on the pci name, but rather pci type. The framework is prepared for it, it can handle `{'type': 'PCI'}` or `PCIE`, if they are defined like that. Then one would only say I have this `nic`, it's `PCIE` and avocado should pick the first free `PCIE` bus. Obviously we'd have to still provide some hints on which one to use (to avoid using the root complex or some switches in between).

Anyway that was just for consideration, I believe right now the hardcoded values serve the best.

shared/cfg/guest-hw.cfg
```diff
@@ -60,6 +66,7 @@ variants:
 
 variants:
     - ide:
+        no q35
```
ldoktor: nice

shared/cfg/machines.cfg
ldoktor: This comment is outdated, we decide somewhere else that adjusting q35 is the way to go as it's the recommended solution for q35 machines. (just repeating it here so people are not confused)

virttest/qemu_vm.py
```diff
@@ -384,6 +384,17 @@ def _add_option(option, value, option_type=None, first=False):
...
+        def _get_pci_bus(params, name=None):
+            """
+            Get device pattern pci bus.
+
+            :param params: device params
+            :param name: deivce type name
+            """
+            name = name and "%s_pci_bus" % name or "pci_bus"
+            pci_bus = {"aobject": params.get(name, "pci.0")}
+            return pci_bus
```
ldoktor: Nice, I like it

virttest/qemu_vm.py
```diff
@@ -1543,6 +1558,8 @@ def add_disable_modern(devices, dev, dev_type):
                     if self.last_boot_index > 0:
                         image_boot = False
                     self.last_boot_index += 1
+            image_pci_bus = _get_pci_bus(image_params, "hba")
+            image_params["pci_bus"] = image_pci_bus.get("aobject", "pci.0")
```
ldoktor: I think this should be handled in `qcontainer` rather than overriding the params here. The `hba` should accept the `hba_pci_bus` instead of `pci_bus` (with fallback to `pci_bus`).

ldoktor: Btw do you consider the `hba_pci_bus` for `hba` only? Because this way it's also for the actual disks, when talking about `virtio_blk` disk type. If so then it should be probably called `disk_pci_bus` rather than `hba_pci_bus`, because the `hba` is misleading.

ldoktor: (what I mean is in case of virtio_blk the device is not host->bus adapter, but directly the pci disk device)

virttest/qemu_vm.py
```diff
@@ -1459,7 +1470,8 @@ def add_disable_modern(devices, dev, dev_type):
                     if params.get('machine_type').startswith("arm64-mmio"):
                         dev = QDevice('virtio-serial-device')
                     else:
-                        dev = QDevice('virtio-serial-pci', parent_bus=pci_bus)
+                        parent_bus = _get_pci_bus(port_params, "virtio_port")
```
ldoktor: This virtio_port is not set anywhere in the config, therefor it defaults to the main `pcie.0`. Is this expected?

virttest/qemu_vm.py
ldoktor: Oh actually there is a minor bug in this function. The default should be:

```
pci_bus = {"aobject": object_params.get(name, params.get("pci_type", "pci.0"))}
```
so it reflects the main chosen `pci_type` type.

shared/cfg/machines.cfg
```diff
@@ -16,6 +16,14 @@ variants:
             cd_format = ahci
         enable_pvpanic = yes
         ioport_pvpanic = 0x505
+        pci_controllers = root_port pcie_switch dmi2pci_bridge pci_bridge
+        pci_bus = pci.0
```
ldoktor: I'm wondering whether it would be better to set everything to the `pcie_switch` and only change it when explicitly stated (like for nics, ...).

And if we think that most devices should for compatibility issues go to `PCI`, then we should probably set this to `pci_bridge` instead, what do you think?

I tried the behavior on the `virtio_console.spread_linear.specifiable.virtserialport.virtio_console_smoke.open` test, which by default does not define the `virtio_port`, therefor the ports end up in `pcie.0`. When using this (after fixing the https://github.com/avocado-framework/avocado-vt/pull/657/commits/fde21e57f0db8b418821b3d881df41577a290538#r74894177 ) it puts them into the proper switch/bridge instead by default, which is what I'd expect.

ldoktor:
Hello @xutian, it seems to be working well. There is one issue regarding the defaults (to reflect the main_params' pci_type)and other with the `hba` vs. `disk` and the handling inside `qemu_devices` instead of overriding it in `qemu_vm`, but apart from that it looks good.

There might still be some broken tests, where the devices could end up in a wrong bus, but that's something which needs to be fixed when identified and mostly it should improve the situation.

I tried some functional tests and the only issue there is that non-specified devices end up in `pcie.0`, which is in my understanding incorrect and we should change the default to either `pcie_switch` or `pci_bridge`.

ngu-niny:
@xutian Apply the patch to avocado-vt failed:

# git am 657.patch
Applying: virttest: add Q35 machine type with bridge devices support
error: patch failed: virttest/qemu_vm.py:1487
error: virttest/qemu_vm.py: patch does not apply
Patch failed at 0001 virttest: add Q35 machine type with bridge devices support
The copy of the patch that failed is found in:
   /home/niny/Documents/avocado-vt/.git/rebase-apply/patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".

ldoktor:
@ngu-niny there were probably some changes. The simplest way to check this code out is to fetch it with the same basecode:

```
git fetch origin pull/657/head
git checkout FETCH_HEAD
```
(or you can directly fetch it into a new branch, but I tend to prefer using `FETCH_HEAD` instead)

Github does not complains, so perhaps `git am` has not enough information to merge it cleanly. When you fetch it directly you have all the details so the `git merge` should be able to apply it cleanly (for now). But I'd simply recommend reviewing the fetched version, @xutian will rebase after the review if needed.

ngu-niny:
@ldoktor Get it, thanks. It can be applied now.

virttest/qemu_vm.py
xutian: I think connect virtio pci device to downstream port or pcie.0 it acceptable, in my mind, if virtio pci device connect to pcie.0 hotplug/unplug is not allowed (I'm not sure about this).
Device tree model in Q35 is very very flexible,  framkly speaking it very hard to cover all in this time.

virttest/qemu_vm.py
xutian: Agree,  will rename it to disk_pci_bus.

shared/cfg/machines.cfg
xutian: Will make all virtio pci device connect to downstream port, USB and AHCI contorller connected to pcie.0 and other PCI device connect to pci-bridge.

virttest/qemu_vm.py
ldoktor: Yep, currently qemu does not restricts anything, one can mix-match pci and pcie freely so I was just wondering what defaults should we pick

shared/cfg/machines.cfg
ldoktor: Yep, that would be my recommendation as the mainstream is currently pci so we should use pci-bridge as base.

virttest/qemu_vm.py
ldoktor: OK, thanks

ngu-niny:
Seemed virtio-balloon and virtio-serial devices still use pcie.0:
-device virtio-balloon-pci,id=balloon0,bus=pcie.0,addr=04
-device virtio-serial-pci,id=virtio_serial_pci0,bus=pcie.0,addr=04

ldoktor:
Hello @xutian, what happened to this PR? I think it was a great foundation. Did you found any big issues preventing it from being merged?

xutian:
@ldoktor I got a baby, just comeback  paternity leave,  I will update the commit and reopen this PR later.

Thanks,
Xu

ldoktor:
Congratulation, Xu, what a wonderful reason to close pull request :smile:.

PandaWei:
Congratulation.

wei

PR #663, opened 2016-08-17
libvirt_xml.vm_xml: Added function to get cpu topology

sathnaga:
Added function which returns the cpu topology as dict
for example:
cputopology content in vm xml:
` <topology sockets='1' cores='4' threads='8'/>`

function returns:
{'cores': '4', 'threads': '8', 'sockets': '1'}

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -727,6 +727,17 @@ class VMXML(VMXMLBase):
             raise xcepts.LibvirtXMLError(
                 "The cpu mode '%s' is invalid!" % mode)
 
+    def get_cpu_topology(self):
+        """
+        Return cpu topology dict
+        """
+        topology = {}
+        try:
+            topology = self.cpu.topology
+        except:
+            logging.error("Can't find <cpu>/<topology> element")
+        return topology
+
     def get_disk_all(self):
         """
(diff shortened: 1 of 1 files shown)


sathnaga:
@waynesun09 @will-Do could you please merge if fine, Thanks in advance :-)

virttest/libvirt_xml/vm_xml.py
```diff
@@ -727,6 +727,22 @@ def check_cpu_mode(mode):
             raise xcepts.LibvirtXMLError(
                 "The cpu mode '%s' is invalid!" % mode)
 
+    def get_cpu_topology(self):
+        """
+        Return cpu topology dict
+        """
+        topology = {}
+        xmltreefile = self.__dict_get__('xml')
```
waynesun09: topology is element of cpu xml, so you could get it with
vmxml = vm_xml.VMXML.new_from_dumpxml(vm_name, virsh_instance=virsh_instance)
topology = vmxml.cpu.topology

as cpu could be without topology setting, so you need use try..except to return None if not found

sathnaga:
@waynesun09 Thanks Updated the commit with comments

balamuruhans:
ACK, Tested with PR https://github.com/autotest/tp-libvirt/pull/856

will-Do:
LGTM

PR #664, opened 2016-08-17
Fixup libvirt.py

balamuruhans:
Added sleep time to make sure migration is started to execute
command in parallel.

Eg: For postcopy migration, cmd "virsh migrate-postcopy domain name"
needs to be executed during migration. Without sleep time below error
is observed.

"Requested operation is not valid: post-copy can only be started while
outgoing migration is in progress"

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1175,6 +1175,8 @@ class MigrationTest(object):
                 migration_thread.start()
                 eclipse_time = 0
                 if func:
+                    # Need time for thread to start migration before executing func
+                    time.sleep(5)
                     stime = int(time.time())
                     func(func_params)
                     eclipse_time = int(time.time()) - stime


balamuruhans:
# avocado run --vt-type libvirt --vt-config /var/lib/libvirt/images/bala/avocado-fvt-wrapper/data/avocado-vt/backends/libvirt/cfg/migrate.cfg --vt-only-filter 'RHEL.7.2.ppc64le scsi qcow2' --vt-no-filter 'network user macvtap'
JOB ID     : 758b49df97d03c65572182a785ae822310da22e9
JOB LOG    : /var/lib/libvirt/images/bala/avocado-fvt-wrapper/results/job-2016-08-17T12.45-758b49d/job.log
TESTS      : 3
 (1/3) guest_migration.guest_import.qemu.qcow2.scsi.smp2.virtio_net.RHEL.7.2.ppc64le.powerkvm-qemu.unattended_install.import.import.default_install.aio_native: PASS (43.36 s)
 (2/3) guest_migration.postcopy_migration.qemu.qcow2.scsi.smp2.virtio_net.RHEL.7.2.ppc64le.powerkvm-libvirt.virsh.migrate.postcopy_migration.postcopy: PASS (176.90 s)
 (3/3) guest_migration.guest_remove.qemu.qcow2.scsi.smp2.virtio_net.RHEL.7.2.ppc64le.powerkvm-libvirt.remove_guest.without_disk: PASS (1.03 s)
RESULTS    : PASS 3 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0
JOB HTML   : /var/lib/libvirt/images/bala/avocado-fvt-wrapper/results/job-2016-08-17T12.45-758b49d/html/results.html
TESTS TIME : 221.29 s

balamuruhans:
https://github.com/autotest/tp-libvirt/pull/825

PandaWei:
Not all migration tests need to sleep(), right?
We'd better to prevent the abuse of sleep() in API function.

balamuruhans:
@PandaWei , but intention of tests which use func is to execute a command during migration is in progress. it is better to make sure migration is started properly and trigger the command.

PandaWei:
Do you have any idea to check if the migration is started ?
If you have, It's best to do check and then execute your command.
If you haven't, But only to sleep a while.
I suggest you to modify your command, just for example,  "sleep 5 && \<other command\>" .
If so, we can avoid to sleep for all of migration tests.

balamuruhans:
@PandaWei Okay done

PandaWei:
Please allow me to stress a bit ...
The best solution is to confirm the migration has already started, and then execute our planned command.  sleep is not good method.
If you have any new idea about it, welcome to open new PR.

balamuruhans:
@PandaWei I agree with you. I will work to find out a method to check whether migration is started or not. On other hand I tried using command as mentioned by you but I couldn't pass shell=True in func_params to run sleep 5 && \<command\>. without shell=True, process.run is failing, can you suggest a way to solve this

PandaWei:
Indeed, the runner fails when command line includes && or ||. and shell=True must be set.
In regard of function maybe have more than one parameter, I suggest to modify func_params to *args or ** args.
You can make two new PRs, one is used to modify the API do_migration(). another is used to modify all of the caller if it's necessary.

balamuruhans:
@PandaWei, sure

PR #669, opened 2016-08-18
virttest.qemu_vm: Change to set the disable-legacy and disable-modern option only for RHEL.7.2 host

ngu-niny:
The default virtio1.0 behaviour on latest RHEL.7.3 host is correct, so only need to set the disable-legacy and disable-modern option for RHEL.7.2 host

ID: 1366941

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1237,24 +1237,6 @@ class VM(virt_vm.BaseVM):
                 dev.set_param("id", devid)
             devices.insert(dev)
 
-        def add_disable_legacy(devices, dev, dev_type):
-            """
-            This function is used to add disable_legacy option for virtio-pci
-            """
-            options = devices.execute_qemu("-device %s,?" % dev_type)
-            if "disable-legacy" in options:
-                value = params.get("disable_legacy", "off")
-                dev.set_param("disable-legacy", value)
-
-        def add_disable_modern(devices, dev, dev_type):
-            """
-            This function is used to add disable_modern option for virtio-pci
-            """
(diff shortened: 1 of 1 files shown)


virttest/qemu_vm.py
```diff
@@ -2140,9 +2122,18 @@ def add_disable_modern(devices, dev, dev_type):
...
+            # Currently only need to add the options for RHEL.7.2 host, you can add them to other hosts if necessary
+            if host_kernel_ver == "Host_RHEL.7.2" and ("disable-legacy" and "disable-modern") in options:
+                set_disable_legacy_modern = True
+                disable_legacy = params.get("disable_legacy", "off")
+                disable_modern = params.get("disable_modern", "on")
+            else:
+                set_disable_legacy_modern = False
+            if set_disable_legacy_modern and dev_type in virtio_pci_devices:
+                device.set_param("disable-legacy", disable_legacy)
+                device.set_param("disable-modern", disable_modern)
```
PandaWei: Frankly speaking, I don't like this change.
2125 ~ 2136 looks messy.
And we can declare the variable disable_legacy and disable_modern in cfg to achieve your purpose , right？

ngu-niny: @PandaWei Thanks for your comment! I agree, keeping the original add_disable_* functions and declaring the variables disable_legacy and disable_modern in some cfg file is a better way, but I failed to find any suitalbe cfg file under avocado-vt. Besides, current virtio1.0 behaviour on RHEL.7.3 host is default, we don't need to specify the two options explictly, and we don't need to specify them on other host before RHEL.7.2, so only RHEL.7.2 host need them.

PR #670, opened 2016-08-19
virttest.qemu_vm: Change to not set the disable-legacy and disable-modern options

ngu-niny:
Currently virtio1.0 behaviour on latest RHEL.7.2/RHEL.7.3 hosts is default, we don't need to specify the disable-legacy and disable-modern options explicitly.

ID: 1366941

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -2083,9 +2083,16 @@ class VM(virt_vm.BaseVM):
                               "virtio-serial-pci", "virtio-rng-pci"]
         for device in devices:
             dev_type = device.get_param("driver")
+            # Currently virtio1.0 behaviour on latest RHEL.7.2/RHEL.7.3
+            # qemu versions is default, we don't need to specify the
+            # disable-legacy and disable-modern options explicitly.
+            set_disable_legacy = params.get("set_disable_legacy", "no")
+            set_disable_modern = params.get("set_disable_modern", "no")
             if dev_type in virtio_pci_devices:
-                add_disable_legacy(devices, device, dev_type)
-                add_disable_modern(devices, device, dev_type)
+                if set_disable_legacy == "yes":
+                    add_disable_legacy(devices, device, dev_type)
+                if set_disable_modern == "yes":
+                    add_disable_modern(devices, device, dev_type)
 
         return devices
 


virttest/qemu_vm.py
```diff
@@ -2083,7 +2083,16 @@ def add_disable_modern(devices, dev, dev_type):
...
             dev_type = device.get_param("driver")
-            if dev_type in virtio_pci_devices:
+            host_kernel_ver = process.system_output("uname -r",
+                                                    shell=True,
+                                                    verbose=False)
+            # Currently only need to add the options for RHEL.7.2 host,
+            # you can add them for other hosts if necessary
+            if "3.10.0-327." in host_kernel_ver:
+                set_disable_legacy_modern = True
+            else:
```
PandaWei: We had a discussion on this PR, right?
The record is gone...

virttest/qemu_vm.py
```diff
@@ -2083,7 +2083,16 @@ def add_disable_modern(devices, dev, dev_type):
                               "virtio-serial-pci", "virtio-rng-pci"]
         for device in devices:
             dev_type = device.get_param("driver")
-            if dev_type in virtio_pci_devices:
+            host_kernel_ver = process.system_output("uname -r",
+                                                    shell=True,
+                                                    verbose=False)
+            # Currently only need to add the options for RHEL.7.2 host,
+            # you can add them for other hosts if necessary
+            if "3.10.0-327." in host_kernel_ver:
```
PandaWei: The "-327" is only introduced and used by redhat release.
Can you distinguish it by other way? for example, qemu version?

ngu-niny:
@PandaWei Yes, I have closed the original PR 669, and this is a new opened one.

For the "-327", it really might be a problem for other linux product. However, I  have confirmed that current released RHEL.7.2 qemu version can also use virtio1.0 by default without setting the two options, so no need to keep the "-327" judgement. I planed to keep the two functions but  set two switch variables with default value FALSE following line 2085 to control the using of them, how do you think?

Thanks for the review above!

PandaWei:
That's good way. As you said that only RHEL.7.2 host need them, use FALSE as default is reasonable.

ngu-niny:
@PandaWei I have changed the PR, would you please help to review the PR again.
Thanks!

virttest/qemu_vm.py
```diff
@@ -2083,9 +2083,13 @@ def add_disable_modern(devices, dev, dev_type):
                               "virtio-serial-pci", "virtio-rng-pci"]
         for device in devices:
             dev_type = device.get_param("driver")
+            set_disable_legacy = params.get("set_disable_legacy", "False")
+            set_disable_modern = params.get("set_disable_modern", "False")
             if dev_type in virtio_pci_devices:
-                add_disable_legacy(devices, device, dev_type)
-                add_disable_modern(devices, device, dev_type)
+                if set_disable_legacy == "True":
```
PandaWei: "== True" looks strange. And I don't find any codes like that.
How about " set_disable_legacy = eval(params.get("set_disable_legacy", "False"))"?
and "if  set_disable_legacy: "

virttest/qemu_vm.py
```diff
@@ -2083,9 +2083,13 @@ def add_disable_modern(devices, dev, dev_type):
                               "virtio-serial-pci", "virtio-rng-pci"]
         for device in devices:
             dev_type = device.get_param("driver")
```
PandaWei: And I want you add explanation for why we use False as default.

ngu-niny:
@PandaWei Have added the explanation and changed to yes/no to replace the True/False.

PandaWei:
ACK

xutian:
looks good to me

xutian:
thanks @ngu-niny and @PandaWei

PR #672, opened 2016-08-19
Fix can't find setup_modules problem

waynesun09:
If AUTOTEST_PATH is set in env, then load the setup_modules will
fail due to wrong file name.


PandaWei:
ACK

will-Do:
ACK

PR #676, opened 2016-08-24
set the disks using gluster ,howerver the system disk don't use gluster

PyLearner:
it is used by the patch: https://github.com/autotest/tp-qemu/pull/736

### DIFF
--- a/virttest/storage.py
+++ b/virttest/storage.py
@@ -129,11 +129,13 @@ def get_image_filename(params, root_dir):
     enable_gluster = params.get("enable_gluster", "no") == "yes"
     enable_ceph = params.get("enable_ceph", "no") == "yes"
     image_name = params.get("image_name")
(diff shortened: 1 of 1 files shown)


virttest/storage.py
```diff
@@ -129,11 +129,15 @@ def get_image_filename(params, root_dir):
...
     image_name = params.get("image_name")
+    only_disk_enable_gluster = params.get("only_disk_enable_gluster", "no") == "yes"
     if image_name:
-        if enable_gluster:
-            image_name = params.get("image_name", "image")
-            image_format = params.get("image_format", "qcow2")
-            return gluster.get_image_filename(params, image_name, image_format)
+        if only_disk_enable_gluster:
+            pass
+        else:
```
balamuruhans: @PyLearner, instead you can have "if not only_disk_enable_gluster"

PyLearner:
hi @balamuruhans , thanks ,will change it .

PR #677, opened 2016-08-24
SLES-11-SP2.xml: SLES guest's non-eth0 interface support

aliang123:
Add script to create interface config file and up the interface.

ID:1361416


xutian:
looks good to me.

CongLi:

PR #679, opened 2016-08-24
remote: Correct scp info

aliang123:
Correct log info for scp operation between guest and host.

ID:1361403

### DIFF
--- a/virttest/remote.py
+++ b/virttest/remote.py
@@ -795,10 +795,10 @@ def throughput_transfer(func):
     print throughput if filesize is not none, else will print elapsed time
     """
     def transfer(*args, **kwargs):
-        if "to" in func.__name__:
+        if "from" in func.__name__:
             msg = (
                 "Copy file from %s:%s to %s, " %
-                (args[0], args[6], args[5]))
+                (args[0], args[5], args[6]))
         else:
             msg = (
                 "Copy file from %s to %s:%s, " %


virttest/remote.py
```diff
@@ -797,12 +797,12 @@ def throughput_transfer(func):
...
-                "Copy file from %s:%s to %s, " %
-                (args[0], args[6], args[5]))
-        else:
-            msg = (
                 "Copy file from %s to %s:%s, " %
                 (args[5], args[0], args[6]))
+        else:
+            msg = (
+                "Copy file from %s:%s to %s, " %
+                (args[0], args[5], args[6]))
```
PandaWei: Good catch!
How about fix it like below?
`--- a/virttest/remote.py
+++ b/virttest/remote.py
@@ -795,10 +795,10 @@ def throughput_transfer(func):
     print throughput if filesize is not none, else will print elapsed time
     """
     def transfer(*args, **kwargs):
-        if "to" in func.__name__:
+        if "from" in func.__name__:
             msg = (
                 "Copy file from %s:%s to %s, " %
-                (args[0], args[6], args[5]))
+                (args[0], args[5], args[6]))
`

PandaWei: If so, we need fewer changes

PandaWei:
And I want to describe clear what& why you want to do in commit message.

PandaWei:
ID:1361403 is used in your group.
Not all of us understand it.

aliang123:
ID:1361403

PandaWei:
I'm OK for this fix.

But the format of commit message have room for improvement.
In general, 
The title should be written like "Module: Summary",  and '.py' can be omitted.
No more than 79 characters in one line. and so on ...
You can refer to other accepted commits by `git log virttest/remote.py` next time

PandaWei:
Thanks @aliang123  and Merge now

PR #683, opened 2016-08-26
Run vm.get_address() method on certain interfaces.

Andrei-Stepanov:
Up to this moment vm.get_address() went through all host's NICs to get
VM's address. This was very time consuming for hosts with several NICs.
It was because each call vm.wait_for_login() always called
vm.get_address(). Wich, in its turn, could be paused for 1 minute for
'wrong' interface.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -2919,13 +2919,15 @@ def parse_arp():
     return ret
 
 
-def verify_ip_address_ownership(ip, macs, timeout=60.0):
+def verify_ip_address_ownership(ip, macs, timeout=60.0, devs=None):
     """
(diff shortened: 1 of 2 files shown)


Andrei-Stepanov:
Could someone try to run tests with this patch?
It works for me, but I do not know how it will be influence other tests.

will-Do:
This should affect lots of libvirt test cases, I will run our job with patch and give feedback later

will-Do:
libvirt tests work fine with this patch

Andrei-Stepanov:
Hi ) Could someone please merge this pull req?) @xutian ?
Thank you.

PR #684, opened 2016-08-29
utils_test/__init__: Fix autotest issues inside SLES guests

ngu-niny:
1. Fix failure to get 'autotest_dirname' issue: Add lines in function 'extract(vm, remote_path, dest_dir)' to decide whether the 'extracting autotest.tar.bz2 output' and 'autotest_dirname' is null, if so, try to continue to get it in the remained 'while/for' loops; otherwise, break and go on with following test
2. Fix the 'cat' results in host failure issue: Add a 'while' loop in function 'get_results(base_results_dir)' to make sure file 'status.json' already created before compressing the result

ID: 1360676, 1361468

Signed-off-by: Nini Gu ngu@redhat.com

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2652,13 +2652,13 @@ def get_used_mem(session, os_type):
     """
     if os_type == "windows":
         cmd = "systeminfo"
-        pattern = r'Virtual Memory: In Use: (.+) MB'
+        pattern = r'(Virtual Memory|Page File): In Use: (.+) MB'
     else:
         cmd = "free -m | grep 'Mem'"
-        pattern = r'Mem:\s+\d+\s+(\d+)\s+'
+        pattern = r'Mem:\s+(\d+)\s+(\d+)\s+'
     output = session.cmd_output(cmd)
     match = re.search(pattern, output, re.M | re.I)
-    used = "%sM" % ''.join(match.group(1).split(","))
+    used = "%sM" % ''.join(match.group(2).split(","))
     used = float(normalize_data_size(used, order_magnitude="M"))
     return int(used)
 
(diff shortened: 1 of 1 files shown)


virttest/utils_test/__init__.py
```diff
@@ -790,6 +791,7 @@ def get_results(base_results_dir):
         compress_cmd += "tar cjvf %s ./*" % results_tarball
         compress_cmd += " --exclude=*core*"
         compress_cmd += " --exclude=*crash*"
+        time.sleep(5)
```
PandaWei: Frankly, I hate the abuse of sleep().

ngu-niny:
@PandaWei Do you have any suggestion to resolve this type of issue without using sleep()? 
Thanks!

PandaWei:
what's mean that ''Fix the 'cat' autotest results failure issue"?

PandaWei:
If you worry about that compress operation happens before test is still runing and writing the log, 
You should ensure that Any tests in guest had already finished before calling get_result().

PandaWei:
I had a discussion with @balamuruhans for the similar problem. #664 
calling sleep() in API function is not a good choice.

ngu-niny:
@PandaWei Thanks a lot for above info. I have updated the patch, would you help to review it again?

virttest/utils_test/__init__.py
```diff
@@ -790,7 +795,12 @@ def get_results(base_results_dir):
         compress_cmd += "tar cjvf %s ./*" % results_tarball
         compress_cmd += " --exclude=*core*"
         compress_cmd += " --exclude=*crash*"
-        session.cmd(compress_cmd, timeout=600)
+        while True:
+            status_json = os.path.join(results_dir, "status.json")
+            if session.cmd_status("test -f %s" % status_json) != 0:
+                continue
```
CongLi: It has been done in get_results_summary().

virttest/utils_test/__init__.py
```diff
@@ -755,11 +755,16 @@ def extract(vm, remote_path, dest_dir):
         extract_cmd = "tar %s %s -C %s" % (options,
                                            basename,
                                            os.path.dirname(dest_dir))
-        output = session.cmd(extract_cmd, timeout=120)
+        output = ""
+        while True:
+            output = session.cmd(extract_cmd, timeout=120)
+            if output != "":
+                break
```
PandaWei: IMHO, There's no obvious improvement than sleep().
Even this method is worse, the loop maybe endless.
session.cmd() return means the extract command had been executed and finished. If the output is not your expectation, You should investigate why and find solution.  

If it's impossible to find the true reason and there's no better way to fix it, but by sleep().  I'm OK , but please add explanation for it.

I think you should investigate why

ngu-niny:
@PandaWei For the first 'extracting autotest.tar.bz2 output' failure issue, sleep() even 300 seconds could not help to resolve it. With my current resolution, I have added a sentence 'print "####output%d: %s" % (i, output)' to do some debug; from following test log output, it can be seen the 1st time running cmd 'output = session.cmd(extract_cmd, timeout=120)' gets null output, WHILE the session log showed the 1st time 'extract' cmd was already correctly executed. Not sure about the real reason now. Hence, shall we add a for loop instead of the possible endless while one?

02:07:10 DEBUG| Sending command: tar xvjmf autotest.tar.bz2 -C /usr/local
02:07:10 DEBUG| ####output1:
02:07:10 DEBUG| Sending command: tar xvjmf autotest.tar.bz2 -C /usr/local
02:07:11 INFO | Update MAC(34:40:b5:a0:ca:10)<->(10.73.196.127)IP pair into address_cache
02:07:11 DEBUG| ####output2: client/__init__.py
02:07:11 DEBUG| client/autotest_local.py
02:07:11 DEBUG| client/base_sysinfo.py
02:07:11 DEBUG| client/base_utils.py
02:07:11 DEBUG| client/base_utils_unittest.py

For the second 'cat' results in host failure issue, usually adding sleep(10) could help to resolve it.

PandaWei:
Yes, for loop is better than endless while.

virttest/utils_test/__init__.py
xutian: @ngu-niny  maybe '-v' or '--verbose' is helpful for you.

virttest/utils_test/__init__.py
```diff
@@ -790,7 +795,12 @@ def get_results(base_results_dir):
         compress_cmd += "tar cjvf %s ./*" % results_tarball
         compress_cmd += " --exclude=*core*"
         compress_cmd += " --exclude=*crash*"
-        session.cmd(compress_cmd, timeout=600)
+        while True:
```
xutian: Agree with @PandaWei 's comments, you should find out root cause of output of extra_cmd is not expect, not repeat, repeat and repeat...

xutian:
so long not update, let me close it. @ngu-niny please find out rootcause then send a new pull-request.

ngu-niny:
New PR is https://github.com/avocado-framework/avocado-vt/pull/949

Issue #685, opened 2016-08-29
mount issue in the nfs.py


svirt:
Looks like there is an issue at line 330

`        mount_cmd = ssh_cmd + "'mount -t nfs %s %s -o %s'" % (self.mount_src,
                                                              self.mount_dir,
                                                              self.mount_options)`

when  self.mount_options is None then the mount command fails as -o does not get any argument. I would suggest to add a check for self.mount_options if it's empty or not , then based on it to use -o option or not. If this approach is fine then I can send a patch

sathnaga:
@svirt yep, looks good

will-Do:
Yes, split 'mount_options' to a new line should be good

PR #687, opened 2016-08-30
fix for mount issue in the nfs.py #685

svirt:
The mount "-o" option is not needed when there is no argument for it.
"-o" option returns an error if there is no argument used after it.

### DIFF
--- a/virttest/nfs.py
+++ b/virttest/nfs.py
@@ -288,7 +288,7 @@ class NFSClient(object):
         """
         ssh_cmd = "ssh %s@%s " % (self.ssh_user, self.nfs_client_ip)
         logging.debug("Umount %s from %s" %
-                      (self.mount_dir, self.nfs_server_ip))
+                      (self.mount_dir, self.nfs_client_ip))
         umount_cmd = ssh_cmd + "'umount -l %s'" % self.mount_dir
         try:
             process.system(umount_cmd, verbose=True)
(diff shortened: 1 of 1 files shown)


virttest/nfs.py
```diff
@@ -327,9 +327,13 @@ def setup_remote(self):
...
-        mount_cmd = ssh_cmd + "'mount -t nfs %s %s -o %s'" % (self.mount_src,
-                                                              self.mount_dir,
-                                                              self.mount_options)
+        if self.mount_options:
+            mount_cmd = ssh_cmd + "'mount -t nfs %s %s -o %s'" % (self.mount_src,
+                                                                  self.mount_dir,
+                                                                  self.mount_options)
+        else:
+            mount_cmd = ssh_cmd + "'mount -t nfs %s %s'" % (self.mount_src,
+                                                            self.mount_dir)
```
apahim: Yes, good catch. I'd suggest to take the opportunity to improve the code as well, removing the duplications and improving the readability. Something like:
```python
        mount_cmd = "mount -t nfs %s %s" % (self.mount_src, self.mount_dir)
        if self.mount_options:
            mount_cmd += " -o %s" % self.mount_options
        try:
            cmd = "%s '%s'" % (ssh_cmd, mount_cmd)
            process.system(mount_cmd, verbose=True)
        except process.CmdError:
            raise exceptions.TestFail("Failed to run: %s" % cmd)
```

balamuruhans: @svirt, Please fix a logging issue in Line 291, where instead of self.nfs_client_ip,
self.nfs_server_ip is logged,

ssh_cmd = "ssh %s@%s " % (self.ssh_user, self.nfs_client_ip)
        logging.debug("Umount %s from %s" %
(self.mount_dir, self.nfs_server_ip))

svirt:
I will have a look there

virttest/nfs.py
```diff
@@ -327,11 +327,12 @@ def setup_remote(self):
...
-                                                              self.mount_options)
+        mount_cmd = "mount -t nfs %s %s" % (self.mount_src, self.mount_dir)
+        if self.mount_options:
+            mount_cmd += " -o %s" % self.mount_options
         try:
-            process.system(mount_cmd, verbose=True)
+            cmd = "%s '%s'" % (ssh_cmd, mount_cmd)
+            process.system(cmd, verbose=True)
         except process.CmdError:
             raise exceptions.TestFail("Failed to run: %s" % mount_cmd)
```
apahim: s/mount_cmd/cmd/

svirt:
the new patch is submitted for PR

svirt:
forgot to fix the last mount_cmd, a new patch will arrive shortly

apahim:
lgtm. thanks.

PR #690, opened 2016-09-01
Update shell_client of virtio_rng to telnet

suqinhuang:
random.exe cmd could not be  analysis correct with nc


xutian:
looks good to me,  thanks @suqinhuang

PR #692, opened 2016-09-02
utils_v2v:add support for input mode 'libvirtxml' on command building

chloerh:
Same with input mode 'disk' and 'ova'


will-Do:
ACK

PR #693, opened 2016-09-09
utils_test.libvirt: Use xfs as default mkfs type for fs pool

waynesun09:
For fs pool in pre_pool, by default using mkfs.ext4 to make file
system. A iscsi bug 1359702 with ext4 is blocking current test for
fs pool, update the fs default type to xfs which supported on both
6 and 7 to avoid the bug.

Signed-off-by: Wayne Sun gsun@redhat.com

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -787,7 +787,11 @@ def mkfs(partition, fs_type, options="", session=None):
     """
     Make a file system on the partition
     """
-    mkfs_cmd = "mkfs.%s -F %s %s" % (fs_type, partition, options)
+    if fs_type == "xfs":
+        force = "-f"
+    else:
+        force = "-F"
+    mkfs_cmd = "mkfs.%s %s %s %s" % (fs_type, force, partition, options)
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -787,7 +787,11 @@ def mkfs(partition, fs_type, options="", session=None):
     """
     Make a file system on the partition
     """
-    mkfs_cmd = "mkfs.%s -F %s %s" % (fs_type, partition, options)
+    if fs_type == "xfs":
+        force = "-f"
+    else:
+        force = "-F"
+    mkfs_cmd = "mkfs.%s %s %s %s" % (fs_type, force, partition, options)
```
will-Do: Actually, I wrote a patch which also touch this part long time ago. And maybe pass both fs_type as mkfs options is better, like:
```
+    mkfs_cmd = "mkfs -t %s %s %s" % (fs_type, options, partition)
```

sathnaga:
@will-Do @waynesun09 Consider rebasing?

PR #710, opened 2016-09-21
add set/get_migrate_parameter function for HumanMonitor and QMPMonitor

XueqiangWei:
Signed-off-by: Xueqiang Wei xuwei@redhat.com

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -1258,6 +1258,29 @@ class HumanMonitor(Monitor):
         value = cache_size_info.split(":")[1].split()[0].strip()
         return value
 
+    def set_migrate_parameter(self, parameter, value):
+        """
+        Set parameters of migrate.
+
+        :param parameter: the parameter which need to set
+        :param value: the value of parameter
+        """
+        cmd = "migrate_set_parameter"
+        self.verify_supported_cmd(cmd)
+        cmd += " %s %s" % (parameter, value)
+        return self.cmd(cmd)
+
+    def get_migrate_parameter(self, parameter):
+        """
+        Get the parameter value. e.g. cpu-throttle-initial: 30
+
+        :param parameter: the parameter which need to get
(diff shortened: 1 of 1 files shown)


XueqiangWei:
ID 1372198

virttest/qemu_monitor.py
```diff
@@ -2299,6 +2322,30 @@ def get_migrate_cache_size(self):
...
+    def get_migrate_parameter(self, parameter):
+        """
+        Get the value of parameter.
+
+        :param parameter: parameter which need to get.
+        """
+        parameter_info = self.query("migrate-parameters")
+        if parameter in parameter_info:
+            return parameter_info[parameter]
+        else:
```
xutian: return null looks better than False.

virttest/qemu_monitor.py
```diff
@@ -1258,6 +1258,29 @@ def get_migrate_cache_size(self):
...
+        """
+        Set parameters of migrate.
+
+        :param parameter: the parameter which need to set
+        :param value: the value of parameter
+        """
+        cmd = "migrate_set_parameter"
+        self.verify_supported_cmd(cmd)
+        cmd += " %s %s" % (parameter, value)
+        return self.cmd(cmd)
```
xutian: how about self.cmd("%s %s %s" % (cmd, parameter, values))

XueqiangWei: Yes, I agree. But, I think they are all ok, just keep the coding style in the framework, like other functions

XueqiangWei: No, it is not good. self.cmd is a function, and its first parameter is cmd, so call it by self.cmd(cmd)

vivianQizhu:
LGTM, Acked-by: Qianqian Zhu qizhu@redhat.com

xutian:
tested and it works for me. thanks @XueqiangWei

PR #714, opened 2016-09-30
Typo fix

will-Do:


xutian:
Merged, thanks @will-Do

PR #715, opened 2016-09-30
libvirt_storage: Catch failure when get volume list

will-Do:


will-Do:
Using ```try...except``` to catch CmdError in list_volumes(), so no need in other places where calling this function.

PR #716, opened 2016-09-30
Added maxmemory option support for virt-install

sathnaga:
Added mammemory option support for virt-install
Added a method to check if a sub-option present in the
virt-install command line for a given parameter

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -370,6 +370,12 @@ class VM(virt_vm.BaseVM):
         def has_os_variant(os_text, os_variant):
             return bool(re.search(r"%s" % os_variant, os_text, re.MULTILINE))
 
+        def has_sub_option(option, sub_option):
+            option_help_text = process.system_output("%s --%s help" %
+                                                     (virt_install_binary, option),
+                                                     verbose=False)
+            return bool(re.search(r"%s" % sub_option, option_help_text, re.MULTILINE))
+
         # Wrappers for all supported libvirt command line parameters.
         # This is meant to allow support for multiple libvirt versions.
         # Each of these functions receives the output of 'libvirt --help' as a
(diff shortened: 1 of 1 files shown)


sathnaga:
@will-Do could you please check and merge if fine, Thanks in advance :-)

virttest/libvirt_vm.py
```diff
@@ -370,6 +370,12 @@ def has_option(help_text, option):
         def has_os_variant(os_text, os_variant):
             return bool(re.search(r"%s" % os_variant, os_text, re.MULTILINE))
 
+        def has_sub_option(option, sub_option):
+            option_help_text = process.system_output("%s --%s help" %
+                                                     (virt_install_binary, option),
+                                                     verbose=False)
+            return bool(re.search(r"%s" % sub_option, option_help_text, re.MULTILINE))
```
will-Do: It's not the perfect place to add this function here, but if you can't find a more appropriate place, lets keep it here

virttest/libvirt_vm.py
```diff
@@ -394,8 +400,15 @@ def add_hvm_or_pv(help_text, hvm_or_pv):
                 logging.warning("Unknown virt type hvm_or_pv, using default.")
                 return ""
 
-        def add_mem(help_text, mem):
-            return " --ram=%s" % mem
+        def add_mem(help_text, mem, maxmem=None):
+            logging.debug("satheesh")
```
will-Do: This should be removed

will-Do:
And there's a typo in your commit message
s/mammemory/maxmemory/

virttest/libvirt_vm.py
sathnaga: @will-Do Other helper functions for command line sits there, that's the reason have placed it,
not finding any other appropriate place right now, Hope it should be fine?

virttest/libvirt_vm.py
sathnaga: @will-Do Sorry, typo, removed

sathnaga:
@will-Do Tnx, have fixed the comments and pushed :-)

sathnaga:
@will-Do Could you please merge this PR, thanks in advance :-)

sathnaga:
@will-Do Thanks :-)

PR #719, opened 2016-10-07
Huge pages value is not reflecting in 'nr_hugepages'

nasastry:
Huge pages value is not reflecting in
'/proc/sys/vm/nr_hugepages'

1. While reading the value from nr_hugepages file, new line need
   to be handled
2. File open and close need to be handled properly.

This patch addresses the above issues.

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -486,19 +486,28 @@ class HugePageConfig(object):
         """
         error_context.context(
             "setting hugepages limit to %s" % self.target_hugepages)
-        hugepage_cfg = open(self.kernel_hp_file, "r+")
-        hp = hugepage_cfg.readline()
+        try:
+            hugepage_cfg = open(self.kernel_hp_file, "r+")
+            hp = hugepage_cfg.readline().strip()
+            hugepage_cfg.close()
+        except IOError:
+            raise exceptions.TestSetupFail("Can't read kernel hugepage file")
         while int(hp) < self.target_hugepages:
             loop_hp = hp
-            hugepage_cfg.write(str(self.target_hugepages))
-            hugepage_cfg.flush()
-            hugepage_cfg.seek(0)
-            hp = int(hugepage_cfg.readline())
+            try:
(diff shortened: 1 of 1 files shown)


virttest/test_setup.py
```diff
@@ -487,18 +487,20 @@ def set_hugepages(self):
         error_context.context(
             "setting hugepages limit to %s" % self.target_hugepages)
         hugepage_cfg = open(self.kernel_hp_file, "r+")
-        hp = hugepage_cfg.readline()
+        hp = hugepage_cfg.readline().strip()
+        hugepage_cfg.close()
         while int(hp) < self.target_hugepages:
             loop_hp = hp
+            hugepage_cfg = open(self.kernel_hp_file, "r+")
```
balamuruhans: it is better to use try and except with IOError and raise exceptions.TestError()

virttest/test_setup.py
```diff
@@ -487,18 +487,20 @@ def set_hugepages(self):
         error_context.context(
             "setting hugepages limit to %s" % self.target_hugepages)
         hugepage_cfg = open(self.kernel_hp_file, "r+")
```

virttest/test_setup.py
```diff
@@ -486,19 +486,28 @@ def set_hugepages(self):
...
+            try:
+                hugepage_cfg = open(self.kernel_hp_file, "r+")
+                hugepage_cfg.write(str(self.target_hugepages))
+                hugepage_cfg.flush()
+                hugepage_cfg.seek(0)
+                hp = int(hugepage_cfg.readline().strip())
+                hugepage_cfg.close()
+            except IOError:
+                msg = "Can't read/write from kernel hugepage file"
+                raise exceptions.TestSetupFail(msg)
```
PandaWei: Do we need open/close the nr_hugepages twice?

nasastry: I have tried with out closing, file is not reflecting the value user try to put in.

nasastry:
@will-Do Can you please help me in merging this patch. Thanks in advance.

xutian:
Looks OK to me. Merged, thanks @nasastry

PR #723, opened 2016-10-11
nfs.py: Ensure rpcbind running before starting nfs

dzhengfy:
The nfs service will fail to start with "rpc.nfsd: unable to set any
sockets for nfsd" when the rpcbind service is not running. So the fix is
to ensure rpcbind is running before starting the nfs service.


will-Do:
So you need update the unittest  as well

dzhengfy:
@will-Do @waynesun09 , please help review.

PR #726, opened 2016-10-12
shared.cfg: add e1000e configuration in guest-hw.cfg

xutian:


weliao:
Acked-by: Wei Liao

xutian:
merged, thanks @weliao

PR #728, opened 2016-10-14
Minor fix: Fail testcase with proper message

bssrikanth:
Currently if ovs packages are not present on host, tests error out with improper message. This minor fix will Fail test with proper message

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -1128,8 +1128,7 @@ def __init_openvswitch(func):
                 if (not __ovs.check()):
                     raise Exception("Check of OpenVSwitch failed.")
             except Exception, e:
-                logging.debug("Host does not support OpenVSwitch: %s", e)
-
+                raise exceptions.TestFail("Host does not support OpenVSwitch")
         return func(*args, **kargs)
     return wrap_init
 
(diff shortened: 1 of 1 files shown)


bssrikanth:
Incase host does not have ovs packages, currently all tests under virtual_network->iface_ovs ERRORs out with below message:

04:43:08 ERROR| AttributeError: 'NoneType' object has no attribute 'list_br'
04:43:08 ERROR|

After fix now we get proper message:

05:08:46 ERROR| TestFail: Host does not support OpenVSwitch

bssrikanth:
@sathnaga @will-Do my local branch was in sync with the master branch. Not sure why getting branch conflict.

virttest/utils_net.py
```diff
@@ -1129,7 +1129,7 @@ def wrap_init(*args, **kargs):
                     raise Exception("Check of OpenVSwitch failed.")
             except Exception, e:
                 logging.debug("Host does not support OpenVSwitch: %s", e)
-
+                raise exceptions.TestFail("Host does not support OpenVSwitch")
```
will-Do: It's some kinds of duplication with the above debug message

bssrikanth: I can remove logging.debug message. If that is ok.
But not failing test at this point will lead test to ERROR out with below message which is misleading:

04:43:08 ERROR| AttributeError: 'NoneType' object has no attribute 'list_br'
04:43:08 ERROR|

PR #729, opened 2016-10-18
Adding maxcpus for given arch in guest capability

sathnaga:
Added maxcpus field in a guest capability method
to get the default maxcpus for a given arch.

Right now it takes the maxcpus of a default(first)
machine for a given arch which is needed for
the https://github.com/autotest/tp-libvirt/pull/878
the same can be extended further for each machine later
when there is a requirement as identified in TODO.


sathnaga:
@will-Do This PR needs to be merged for autotest/tp-libvirt#878 to work, please merge if fine, Thanks in advance.

PR #743, opened 2016-11-02
utils_test: Directly load stress tool on host

dzhengfy:
Loading stress tool on host can be much easiler. There is no need to
depend on autotest.

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -1660,12 +1660,14 @@ class HostStress(object):
         load stress tool on host.
         """
         # Run stress tool on host.
-        from autotest.client import common
-        autotest_client_dir = os.path.dirname(common.__file__)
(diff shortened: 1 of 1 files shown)


virttest/utils_test/__init__.py
```diff
@@ -1660,12 +1660,8 @@ def load_stress_tool(self):
...
         """
         # Run stress tool on host.
-        from autotest.client import common
-        autotest_client_dir = os.path.dirname(common.__file__)
-        autotest_local_path = os.path.join(autotest_client_dir,
-                                           "autotest-local")
-        args = [autotest_local_path, '--args=%s' % self.stress_args,
-                self.control_path, '--verbose']
+        args = [self.stress_type, '--args=%s' % self.stress_args,
+                '--verbose']
```
will-Do: So the stress tool already present on the host?

PR #744, opened 2016-11-02
shared.guest-hw.cfg: set "cd_format=ide" for win_virtio_drive…

CongLi:
…r_update_test.with_vioscsi.

cdrom won't be recognized if virtio_scsi driver is uninstalled,
which will cause install driver failed after uninstall.

id: 1231072

### DIFF
--- a/shared/cfg/guest-hw.cfg
+++ b/shared/cfg/guest-hw.cfg
@@ -99,6 +99,10 @@ variants:
         cd_format=scsi-cd
         scsi_hba=virtio-scsi-pci
         libvirt_controller=virtio-scsi
+        win_virtio_driver_update_test.with_vioscsi:
+            cd_format = ide
+            q35:
+                cd_format = ahci
         q35:
             disk_pci_bus = pcie_switch
             cdrom_pci_bus = ${disk_pci_bus}


shared/cfg/guest-hw.cfg
```diff
@@ -102,6 +102,8 @@ variants:
         q35:
             disk_pci_bus = pcie_switch
             cdrom_pci_bus = ${disk_pci_bus}
+        win_virtio_driver_update_test.with_vioscsi:
+            cd_format = ide
```
suqinhuang: could you put these two lines above the line of q35

CongLi: @suqinhuang 

Updated, could you help review it again?

Thanks.

shared/cfg/guest-hw.cfg
```diff
@@ -99,6 +99,8 @@ variants:
         cd_format=scsi-cd
         scsi_hba=virtio-scsi-pci
         libvirt_controller=virtio-scsi
+        win_virtio_driver_update_test.with_vioscsi:
+            cd_format = ide
```
xutian: if machine type is Q35, cd format should be ahci

CongLi: @xutian 

Yes, thanks.
could you help review it again?

shared/cfg/guest-hw.cfg
```diff
@@ -99,6 +99,10 @@ variants:
         cd_format=scsi-cd
         scsi_hba=virtio-scsi-pci
         libvirt_controller=virtio-scsi
+        win_virtio_driver_update_test.with_vioscsi:
```
xutian: And I think this kind of configuration should be in qemu/cfg/xxx.cfg

CongLi: this will be covered by guest-hw.cfg.

PR #747, opened 2016-11-03
Fixup libvirt_vm.py

balamuruhans:
Install package and remove package in guest was not working for
ubuntu, this patch fixes it.

Added get_distro() API to get Guest distro name.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -1340,7 +1340,7 @@ class VM(virt_vm.BaseVM):
             output = session.cmd_output(cmd)
             # Sample output of 'df /swapfile':
             # Filesystem 1K-blocks     Used Available Use% Mounted on
-            #/dev/vdb    52403200 15513848  36889352  30% /
+            # /dev/vdb    52403200 15513848  36889352  30% /
             device = output.splitlines()[1].split()[0]
 
             # Set kernel parameters.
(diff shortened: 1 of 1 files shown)


virttest/libvirt_vm.py
```diff
@@ -2485,9 +2513,13 @@ def remove_package(self, name):
         :param name: Name of package to be removed
         """
         session = self.wait_for_login()
+        vm_distro = self.get_distro()
         try:
             # Remove the package if it exists
-            cmd = "! rpm -q %s || rpm -e %s" % (name, name)
+            if vm_distro.lower() == 'ubuntu':
+                cmd = "! (dpkg -l | grep %s) || apt remove -y %s" % (name, name)
```
balamuruhans: @sathnaga done

virttest/libvirt_vm.py
```diff
@@ -2449,6 +2449,24 @@ def get_interface_mac(self, interface):
...
+            status, output = session.cmd_status_output(cmd, timeout=300)
+            if status:
+                raise virt_vm.VMError("Unable to get the distro name: %s"
+                                      % output)
+            else:
+                distro = output.split('=')[1].strip()
+        finally:
+            session.close()
+            return distro
+
```
will-Do: You can just use 'platform'  module, chech: https://docs.python.org/2/library/gg.html#platform.linux_distribution

balamuruhans: @will-Do platform.dist() gives distro name of host but we need to check the distro of guest by using session.cmd_status_output. so I thought this will be fine.

balamuruhans: Also vm.get_distro() can be used by other tests

balamuruhans:
@will-Do please help in merging this PR

PR #755, opened 2016-11-09
Replace letter in cmd to real winutils cdrom letter

suqinhuang:
Some cmd is put in winutils.iso, we can set a letter of winutils.iso in cmd (WIN_UTILS as default),
and replace it with real winutils.iso letter with this function

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2696,6 +2696,19 @@ def get_winutils_vol(session, label="WIN_UTILS"):
     return get_win_disk_vol(session, condition="VolumeName='%s'" % label)
 
 
+def set_winutils_letter(session, cmd, label="WIN_UTILS"):
+    """
+    Replace label in command to real winutils CDROM drive letter.
+
+    :param session: session Objest
+    :param cmd: cmd path in winutils.iso
+    :param label: volume label of WIN_UTILS.iso
+    """
+    if label in cmd:
+        return cmd.replace(label, get_winutils_vol(session))
+    return cmd
+
+
 def format_windows_disk(session, did, mountpoint=None, size=None, fstype="ntfs"):
     """
     Create a partition on disk in windows guest and format it.


pingl2016:
Acked-by: Ping Li <pingl@redhat.com>

virttest/utils_misc.py
```diff
@@ -2675,6 +2675,17 @@ def get_winutils_vol(session, label="WIN_UTILS"):
     return get_win_disk_vol(session, condition="VolumeName='%s'" % label)
 
 
+def set_winutils_letter(session, cmd, params):
```
xutian: where used param argument? and winutils volume letter is not always 'X', so I think you need to rework this function

virttest/utils_misc.py
```diff
@@ -2675,6 +2675,17 @@ def get_winutils_vol(session, label="WIN_UTILS"):
...
+        """
+        Replace 'X:' in command to real winutils CDROM drive letter.
+        """
+        vol = "X:"
+        if params["os_type"] != "linux":
+            vol = utils_misc.get_winutils_vol(session)
```
pingl2016: Need to remove "utils_misc."

virttest/utils_misc.py
suqinhuang: user need to use X: in cmd if you want to call this function

virttest/utils_misc.py
```diff
@@ -2675,6 +2675,16 @@ def get_winutils_vol(session, label="WIN_UTILS"):
     return get_win_disk_vol(session, condition="VolumeName='%s'" % label)
 
 
+def set_winutils_letter(session, cmd, label="WIN_UTILS"):
+        """
+        Replace 'WIN_UTILS:' in command to real winutils CDROM drive letter.
+        """
+        if label in cmd:
```
xutian: if label in cmd:
    return cmd.replace(get_winutils_vol(session), label)
return cmd

xutian:
please update your commit message like: introduce API to format winutils placeholder

pingl2016:
It's ok for me.

PR #756, opened 2016-11-10
virttest: append utils_test.qemu into packages list

xutian:
utils_test.qemu module introduced in commit 4448e1b, this commit
add the module in packages list.


xutian:
just a hot fix, merged.

PR #766, opened 2016-11-14
Add module virtadmin

iccaszhulili:
In order to automate the test case of the new feature virt-admin, I have
added module virtadmin.


iccaszhulili:
@will-Do  I have modified the file according to your comments. Please help me review it. Thanks very much.

will-Do:
As no more concern about this new module, and it only used for virt-admin related cases for now, I'm going to merge it.

PR #769, opened 2016-11-15
virttest.env_process: optimize the check for kvm module

PandaWei:
we should throw an exception earlier if /dev/kvm doesn't exist.
No need to add "enable-kvm" option in make_create_command(),
which will lead the following error,

"Could not access KVM kernel module: No such file or directory\n
 failed to initialize KVM: No such file or directory'"


xutian:
looks good to me, merged thanks @PandaWei

PR #770, opened 2016-11-15
shared.guest-hw.cfg: set "cd_format=ide" for win_virtio_driver_update…

CongLi:
…_test.with_vioscsi.

cdrom won't be recognized if virtio_scsi driver is uninstalled,
which will cause install driver failed after uninstall.

id: 1231072


CongLi:
@suqinhuang @xutian 

Could you help review it?

Thanks.

suqinhuang:
the config is a little muddledness now, we'd better to have a discuss and make a rule

CongLi:
Close this PR, will add the cfg to tp-qemu repo with "_fixed".

Thanks.

PR #771, opened 2016-11-17
virttest.remote: Enable to match the console prompt

dzhengfy:
When the aexpect matches the expected contents, sometimes the console
session will prompt '...Escape character is...' string, then the user
need to type an Enter key to show the 'login:' prompt. So this special
string should be considered to match.


waynesun09:
ACK

PR #778, opened 2016-11-18
remote: Fix typo in regex to avoid failure

Hao-Liu:
In commit f5ccf3da, the regex introduced is not valid and has an asterisk
at the beginning of it. It will cause failure following failure of many
cases using this function:
```
    error: nothing to repeat
```


Hao-Liu:
I'll quick merge this fix to avoid more potential failure since this is a simple fix.

lmr:
Sure

PR #779, opened 2016-11-21
Option to set ssh_key config

balamuruhans:
Some tests execute commands to test machine remotely from other server
using,

virsh -c qemu+ssh://\<ipaddress\>\/system \<command\>

where it performs StrictHostKeyChecking and test fails even after
successful ssh key setup. For permitting virsh to use ssh with
StrictHostKeyChecking=no, the option can be set from ~/.ssh/config

Also if any other tests needs to config options can be used by this
same API.


balamuruhans:
@will-Do done

PR #784, opened 2016-11-24
shared.cfg.guest-os.Windows:update iso for Win10-32&Win10-64

aliang123:
Win10.i386:
1.update Win10-32 iso and its md5.
2.update its maxmem to 3G

Win10.x86_64:
1.update Win10-64 iso and its md5.

Signed-off-by: Aihua Liang<aliang@redhat.com>

ID:1386122


aliang123:
As shuang's advice, no need to update upstream, so close the PR.

PR #790, opened 2016-11-26
virttest.qemu_monitor: balloon(): Add status check before getting QMP event AS there is no memory change until the paused guest is resumed

ngu-niny:
For guest that in paused status, there is no need to get the event after balloon guest memory SINCE the memory will change until the guest is resumed

ID: 1398838 

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -2276,9 +2276,12 @@ class QMPMonitor(Monitor):
         # Send a balloon monitor command
         self.send_args_cmd("%s value=%s" % (cmd, size))
         # Look for BALLOON QMP events
-        if not utils_misc.wait_for(lambda: self.get_event(qmp_event), 120):
-            raise QMPEventError(cmd, qmp_event, self.name)
-        logging.info("%s QMP event received" % qmp_event)
+        if self.verify_status('running'):
+            # For guest in paused status, no need to get the event
+            # since the memory will change util guest is resumed
+            if not utils_misc.wait_for(lambda: self.get_event(qmp_event), 120):
+                raise QMPEventError(cmd, qmp_event, self.name)
+            logging.info("%s QMP event received" % qmp_event)
 
     def set_migrate_capability(self, state, capability):
         """


virttest/qemu_monitor.py
```diff
@@ -2276,9 +2276,12 @@ def balloon(self, size):
...
         self.send_args_cmd("%s value=%s" % (cmd, size))
         # Look for BALLOON QMP events
-        if not utils_misc.wait_for(lambda: self.get_event(qmp_event), 120):
-            raise QMPEventError(cmd, qmp_event, self.name)
-        logging.info("%s QMP event received" % qmp_event)
+        if self.vm.params.get("boot_in_pause", "no") == "no":
+            # For guest that boots with '-S', no need to get the event
+            # since the memory will change util guest is resumed
+            if not utils_misc.wait_for(lambda: self.get_event(qmp_event), 120):
+                raise QMPEventError(cmd, qmp_event, self.name)
```
CongLi: I think need over write / write a new func for such (negative) tests.

ngu-niny:
@CongLi Code changed to use verify_status() function, could you help to review the patch again? Thanks!

CongLi:
Code is fine for me.
But could you please update the commit log?
"For guest that boots with '-S'" is not clear, I think "status is paused/ running..." is better.

Thanks.

ngu-niny:
@CongLi Done.

CongLi:
"virttest.qemu_monitor: Change to not get event after balloon memory for guest that in paused status "

Hi @ngu-niny 

could you specify the balloon() func for "Change to not get event" ?
Maybe "virttest.qemu_monitor: balloon(): Add status check before getting QMP event.
As*****************"

ngu-niny:
@CongLi Changed. Thanks!

CongLi:
LGTM,
acked-by: Cong Li <coli@redhat.com>

virttest/qemu_monitor.py
```diff
@@ -2276,9 +2276,12 @@ def balloon(self, size):
         # Send a balloon monitor command
         self.send_args_cmd("%s value=%s" % (cmd, size))
         # Look for BALLOON QMP events
-        if not utils_misc.wait_for(lambda: self.get_event(qmp_event), 120):
-            raise QMPEventError(cmd, qmp_event, self.name)
-        logging.info("%s QMP event received" % qmp_event)
+        if self.verify_status('running'):
```
xutian: verify_status()  always return None,  so code in if branch will not be execute. and is VM status is pause (in your case), this line will raise exception  virt_vm.VMStatusError.

xutian: sorry,  misunderstand it. you are right.

ngu-niny:
@xutian Could you help to recheck the PR? Thanks.

virttest/qemu_monitor.py
aliang123: With my comment in https://github.com/autotest/tp-qemu/pull/840, i think it's better to keep original. We need to confirm "No balloon change event received" before we change vm's status to "running" in case "balloon guest when it starts with paused status".

ngu-niny: @aliang123 Thanks a lot! Close the PR.

PR #791, opened 2016-11-28
virttest.qemu_storage: Support strict mode  for qemu_img compare

pingl2016:
Support strict mode  for qemu_img compare

Signed-off-by: Ping Li <pingl@redhat.com>

ID: 1395550

### DIFF
--- a/virttest/qemu_storage.py
+++ b/virttest/qemu_storage.py
@@ -436,32 +436,41 @@ class QemuImg(storage.QemuImg):
 
         return supports_cmd
 
-    def compare_images(self, image1, image2, verbose=True):
+    def compare_images(self, image1, image2, strict_mode=False, verbose=True):
         """
         Compare 2 images using the appropriate tools for each virt backend.
 
         :param image1: image path of first image
         :param image2: image path of second image
+        :param strict_mode: Boolean value, True for strict mode,
+                            False for default mode.
         :param verbose: Record output in debug file or not
+
+        :return: process.CmdResult object containing the result of the command
         """
         compare_images = self.support_cmd("compare")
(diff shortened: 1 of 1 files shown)


virttest/qemu_storage.py
```diff
@@ -462,6 +469,7 @@ def compare_images(self, image1, image2, verbose=True):
                 raise exceptions.TestFail("Compared images differ")
             else:
                 raise exceptions.TestError("Error in image comparison")
+            return rv
```
PyLearner: hi, why return rv ?

pingl2016: For all function of qemu-img, I think it is better to return process.CmdResult object. It is very useful when we need to check stdout in case.

suqinhuang:
LGTM

virttest/qemu_storage.py
suqinhuang: suggest to use a more clear name, others LGTM

virttest/qemu_storage.py
```diff
@@ -436,32 +436,41 @@ def support_cmd(self, cmd):
...
+            compare_cmd += " %s %s" % (image1, image2)
+            CmdResult = process.run(compare_cmd, ignore_status=True,
+                                    shell=True)
 
             if verbose:
-                logging.debug("Output from command: %s" % rv.stdout)
+                logging.debug("Output from command: %s" % CmdResult.stdout)
 
-            if rv.exit_status == 0:
+            if CmdResult.exit_status == 0:
```
xutian: take care your var name.

xutian:
merged, thanks @pingl2016

PR #765, opened 2016-11-11
Support to ensure firewall to permit nfs service

balamuruhans:
Add nfs service to firewall to permit mounting nfs directory
from remote machine.

Signed-off-by: Balamuruhan S <bala24@linux.vnet.ibm.com>

PR #793, opened 2016-11-28
utils_test: Revert mistakenly removed importing

Hao-Liu:
These three imports are mistakenly removed in commit 6e1d2bc4.
They looks like unused imports, but is acctually useful to make these
submodules accessible directly from utils_test name space.

Removing these imports will make many test cases those using this way
fail. Added comments and a pylint disablement to avoid future removal.

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -52,6 +52,17 @@ from .. import utils_net
 from .. import virt_vm
 from ..staging import utils_memory
 
+# Get back to importing submodules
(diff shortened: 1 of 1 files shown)


balamuruhans:
https://github.com/avocado-framework/avocado-vt/pull/765
@Hao-Liu, I think issue I am facing in Travis error due to this.

Hao-Liu:
@balamuruhans, I believe its caused by this. Thanks for notifying me this. Since this issue is not only causing case failures but also make Travis CI not working. I'll merge it now to avoid more impact.

PR #794, opened 2016-11-29
Controller: Add two attributes

dzhengfy:
The two attributes are needed in some test scenarios.


waynesun09:
ACK

PR #795, opened 2016-11-30
libvirt_vm.py: support to form uri with protocol and port

balamuruhans:
Forming uri with given protocol and port where migration tests can use it to form uri dynamically and also avoids redundancy

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -49,20 +49,25 @@ def normalize_connect_uri(connect_uri):
     return result
 
 
-def complete_uri(ip_address):
+def complete_uri(ip_address, protocol=None, port=None):
     """
     Return a complete URI with the combination of ip_address and local uri.
     It is useful when you need to connect remote hypervisor.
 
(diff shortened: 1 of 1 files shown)


virttest/libvirt_vm.py
```diff
@@ -49,20 +49,25 @@ def normalize_connect_uri(connect_uri):
...
 
 
-def complete_uri(ip_address):
+def complete_uri(ip_address, protocol=None, port=None):
     """
     Return a complete URI with the combination of ip_address and local uri.
     It is useful when you need to connect remote hypervisor.
 
     :param ip_address: an ip address or a hostname
+    :param protocol: protocol for uri eg: tcp, spice etc.,
```
will-Do:
```
+    :param protocol: protocol for uri eg: tcp, spice etc.,

+    :param protocol: protocol for uri, eg: tcp, spice, etc.
```

balamuruhans: @will-Do sure will change it

balamuruhans: done

PR #774, opened 2016-11-17
Support to add or remove iptable rules to local or remote machine

balamuruhans:
Some of the tests requires to add or remove iptable rules locally or remotely,
Example: For migration tests, 49152 to 49216 ports have to be opened
for libvirt to permit migration

This patch add the support and can be used by other tests.

Signed-off-by: Balamuruhan S <bala24@linux.vnet.ibm.com>

PR #804, opened 2016-12-05
Pre-setup for migration to succeed

balamuruhans:
1. iptables have to permit libvirt to perform migration,
so 49152 to 49216 ports have to be enabled in iptables.

2. In remote Power8 machine, SMT have to be turned off.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -43,8 +43,8 @@ from .. import libvirt_storage
 from .. import utils_net
 from .. import gluster
 from .. import remote
+from .. import test_setup
 from ..staging import lv_utils
-from ..test_setup import LibvirtPolkitConfig
 from ..utils_libvirtd import service_libvirtd_control
 from ..libvirt_xml import vm_xml
 from ..libvirt_xml import network_xml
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1159,6 +1159,54 @@ def thread_func_migration(self, vm, desturi, options=None,
                 self.RET_MIGRATION = False
                 self.RET_LOCK.release()
 
+    def migrate_pre_setup(self, desturi, params=None, cleanup=False):
+        """
+        # Setup before migration,
+        # 1. To enable migration ports using iptables
+        # 2. Turn off SMT for power8 machine in remote machine to migrate
```
will-Do: 1. After https://github.com/avocado-framework/avocado-vt/pull/774 merge, you can use is directly without a new function here to call it.
2. There's already a step to do it on local machine( [here](https://github.com/avocado-framework/avocado-vt/blob/master/virttest/env_process.py#L571 ) ), and it have bug for x86, so 
I think you can fix it, please check 'setup_test.disable_smt()'

balamuruhans: @will-Do , sure
1. But I added a new method migrate_pre_setup()  here is all the files virsh_migrate.py, migrate_vm.py, migrate_multi_vms.py, virsh_migrate_stress.py, virsh_migrate_virtio_scsi.py, virsh_migrate_compcache.py, virsh_migrate_set_get_speed.py requires this setup and this can be called before and after migration for setting up and cleaning up and helps us to avoid redundancy.

2. As this SMT have to be switched off only for ppc where as x86 doesn't have this limitation, For local machine, it takes cares for PPC in test_setup before our test even starts, but for remote destination PPC machine in migration test can be taken care here.

Please correct me if I am wrong.

PR #805, opened 2016-12-05
Turn off SMT in remote machine for Power8

balamuruhans:
Migration test may fail if SMT is not turned off on Power8
machine, so this patch adds support to disable SMT on remote
Power8 machines and to clean up for reverting default setting

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -24,6 +24,7 @@ from . import error_context
 from . import utils_misc
 from . import versionable_class
 from . import openvswitch
+from . import remote
 from .staging import service
 from .staging import utils_memory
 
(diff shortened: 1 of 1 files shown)


balamuruhans:
PR #804 depends on this PR

balamuruhans:
@xutian please help me in reviewing this PR

virttest/test_setup.py
```diff
@@ -2029,16 +2030,55 @@ def stop(self):
...
+        return
+    try:
+        if params:
+            cmd_output = server_session.cmd_status_output(cmd)
+            server_session.close()
+            if (cmd_output[0] != 0):
+                raise exceptions.TestSetupFail("VM cant be started :%s",
+                                               cmd_output[1])
+            else:
+                logging.debug("SMT turned off successfully in remote server")
```
balamuruhans: @sathnaga sure, I will add enable_smt()

virttest/test_setup.py
```diff
@@ -2029,16 +2030,55 @@ def stop(self):
...
+    if ((smt_output != "SMT is off") and not (cleanup)):
+        cmd = "ppc64_cpu --smt=off"
+    elif ((smt_output == "SMT is off") and cleanup):
+        cmd = "ppc64_cpu --smt=on"
+    else:
+        logging.debug("SMT is in expected state: %s", smt_output)
+        return
+    try:
+        if params:
+            cmd_output = server_session.cmd_status_output(cmd)
```
balamuruhans: okay, I will do the check.

virttest/test_setup.py
```diff
@@ -2029,16 +2030,55 @@ def stop(self):
...
+            server_session.close()
+            if (cmd_output[0] != 0):
+                raise exceptions.TestSetupFail("VM cant be started :%s"
+                                               % cmd_output[1])
+            else:
+                logging.debug("SMT turned off successfully in remote server")
```
sathnaga: Cleanup is not required here, add a separate method to enabled_smt.

virttest/test_setup.py
```diff
@@ -2029,16 +2030,55 @@ def stop(self):
...
+    & disable smt in Power8 for local and remote machine respectively.
+
+    :param params: Test params dict for remote machine details to turn off smt
+    :param cleanup: If True, revert back to default setting for cleanup
     """
-    smt_output = process.system_output("ppc64_cpu --smt", shell=True).strip()
-    if smt_output != "SMT is off":
+    cmd = "ppc64_cpu --smt"
+    smt_output = ""
+    if params:
```
sathnaga: better not to use params as the parameter to check for remote, have a separate flag for remote host

balamuruhans:
@sathnaga, have made changes as per your suggestion.

virttest/test_setup.py
```diff
@@ -2029,16 +2030,54 @@ def stop(self):
...
 
-def disable_smt():
+def disable_smt(rmt=False, params=None):
     """
     Checks whether smt is on, if so disables it in PowerPC system
-    This function is used in env_process to check & disable smt in Power8
+    This function is used in env_process and in libvirt.py to check
+    & disable smt in Power8 for local and remote machine respectively.
+
+    :param rmt: If True, turns off smt on remote machine
```
will-Do: No need 'rmt', just deal with it by checking required parameters(server/remote_ip ...)

sathnaga: @will-Do rmt can be additional check that we do not enter the block in case of disabling in the
local machine in case of accidental pass of params, what say?

virttest/test_setup.py
```diff
@@ -2029,16 +2030,50 @@ def stop(self):
...
-        try:
-            utils_misc.verify_running_as_root()
-            process.run("ppc64_cpu --smt=off", verbose=True, shell=True)
-            logging.debug("SMT turned off successfully")
-        except process.CmdError, info:
-            raise exceptions.TestSetupFail("VM can not be started :%s", info)
+    smt_output = ""
+    cmd = "ppc64_cpu --smt"
+    if params:
+        server_ip = params.get("server_ip")
```
xutian: please use param["server_ip"] to avoid server_ip missed in test parms, 'server_pwd' also need to update.

balamuruhans: @xutian sure, will handle this

balamuruhans:
@xutian made the changes as per your suggestion.

virttest/test_setup.py
```diff
@@ -2029,16 +2030,54 @@ def stop(self):
...
+                                                   server_user, server_pwd,
+                                                   r"[\#\$]\s*$")
+            cmd_output = server_session.cmd_status_output(cmd)
+            if (cmd_output[0] == 0):
+                smt_output = cmd_output[1].strip()
+            else:
+                server_session.close()
+                raise exceptions.TestSetupFail("Couldn't get SMT of server: %s"
+                                               % cmd_output[1])
+        except KeyError:
```
xutian: No need catch this exception, framework will raise a better exception.

xutian:
Looks OK to me.
@luckyh  can you second review this pull request?

Thanks,
Xu

virttest/test_setup.py
balamuruhans: @xutian sure, then I will remove it. Thanks for your review

PR #806, opened 2016-12-05
utils_test/libvirt: Load stress tool support

dzhengfy:
This is to prepare stress tool on host or VM for the testing with stress.
Stress package is not included in some repos. So the user need prepare
the installation package of stress and provide the URL of the package
location.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -24,13 +24,16 @@ import shutil
 import threading
 import time
 import sys
+import platform
 
 import aexpect
 from avocado.core import exceptions
 from avocado.utils import path as utils_path
 from avocado.utils import process
 from avocado.utils import stacktrace
+from avocado.utils import download
 from avocado.utils import linux_modules
+from avocado.utils.software_manager import RpmBackend
 
 from .. import virsh
 from .. import xml_utils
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -2924,6 +2927,61 @@ def get_iothreadsinfo(vm_name, options=None):
     return info_dict
 
 
+def load_stress_tool(params):
+    """
+    Download and install stress tool with specified params.
+
+    :param params: Parameters for running stress tool
+       stress_download_url: URL to download the package
+       vm_migration: The VM instance for migration
```
will-Do: It's weird to have this 'vm_migration'  in this  stress tool load function

virttest/utils_test/libvirt.py
```diff
@@ -2924,6 +2927,61 @@ def get_iothreadsinfo(vm_name, options=None):
...
+       vm_session: The session to the VM
+       in_vm: True if run stress within the guest. False if run stress
+       on the host.
+    """
+    stress_download_url = params.get("stress_download_url")
+    vm = params.get("vm_migration")
+    vm_session = params.get("vm_session")
+    in_vm = params.get("in_vm", True)
+    file = "/etc/redhat-release"
```
will-Do: you should avoid using these kind of names

will-Do:
Let's talk this later

PR #807, opened 2016-12-05
Added cpu topology support for vcpu change method

sathnaga:
This patch enables the support for considering
topology xml change during vcpu changes in xml.

Before this patch xml fails to define incase
of cpu topology present in the xml.

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -701,34 +701,57 @@ class VMXML(VMXMLBase):
         vmxml.sync()
 
     @staticmethod
-    def set_vm_vcpus(vm_name, value, current=None, virsh_instance=base.virsh):
+    def set_vm_vcpus(vm_name, vcpus, current=None, sockets=None, cores=None,
+                     threads=None, add_topology=False,
+                     virsh_instance=base.virsh):
         """
-        Convenience method for updating 'vcpu' and 'current' attribute property
-        of a defined VM
+        Convenience method for updating 'vcpu', 'current' and
+        'cpu topology' attribute property with of a defined VM
 
         :param vm_name: Name of defined vm to change vcpu elemnet data
-        :param value: New data value, None to delete.
+        :param vcpus: New vcpus count, None to delete.
         :param current: New current value, None will not change current value
+        :param sockets: number of socket, default None
+        :param cores: number of cores, default None
+        :param threads: number of threads, default None
+        :param add_topology: True to add new topology definition if not present
+        :parma virsh_instance: virsh instance
         """
         vmxml = VMXML.new_from_dumpxml(vm_name, virsh_instance=virsh_instance)
-        if value is not None:
+        if vcpus is not None:
             if current is not None:
                 try:
-                    current_int = int(current)
+                    if(int(current) > vcpus):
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -705,11 +705,27 @@ def set_vm_vcpus(vm_name, value, current=None, virsh_instance=base.virsh):
...
+                # Do change here with respect to vcpu
+                sockets = topology['sockets']
+                cores = topology['cores']
+                threads = topology['threads']
+                if (int(value) % int(threads)) != 0:
+                    # Do not raise test fail here,
+                    # lets leave it to libvirt to detect while define and start
+                    logging.error("vcpu: %s not the multiples of"
+                                  " threads: %s", value, threads)
+                new_cores = int(value)/(int(threads) * int(sockets))
```
will-Do: This tell us you just update the cores, as you want to make the topology configurable, this is not the good way.

sathnaga: @will-Do ping

balamuruhans:
LGTM

virttest/libvirt_xml/vm_xml.py
will-Do: I just think we should deal with the calculation/checking in particular cases, and there can set 'sockets', 'cores' and 'threads' directly, so

sathnaga: @will-Do problem is changing of just vcpu will fail the testcase as it happens in e:g:- vcpucount.py, 
having handled in API will be good as many testcases use this function already.

sathnaga: @will-Do ping

will-Do: if the case fail for this reason, then what we need do it fix the case, and keep the API as simple as possible. And there may have some negative cases that we want to set the invalid core/thread or sth else.

sathnaga: @will-Do Agreed, though the vcpus hotplug going to be core granularity, negative cases will have a probelmand API doing a calculation can also cause a issue later, changed now to take inputs from user.

virttest/libvirt_xml/vm_xml.py
```diff
@@ -693,34 +693,57 @@ def set_pm_suspend(vm_name, mem="yes", disk="yes", virsh_instance=base.virsh):
...
+        :param threads: number of threads, default None
+        :param add_topology: True to add new topology definition if not present
+        :parma virsh_instance: virsh instance
         """
         vmxml = VMXML.new_from_dumpxml(vm_name, virsh_instance=virsh_instance)
-        if value is not None:
+        if vcpus is not None:
             if current is not None:
                 try:
                     current_int = int(current)
```
balamuruhans: `current_int` is used only once,
```
try:
    if not current:
        try:
            if (int(current) > vcpus):
                raise ...
        except ValueError:
            raise ...
```

sathnaga: @balamuruhans Tnx, fixed :-)

sathnaga:
@will-Do pls help merge if no further comments, Tnx

sathnaga:
@will-Do Thanks :-)

PR #808, opened 2016-12-05
Fixup: Changed to log xml content during failure

sathnaga:
It will be helpful to get xml file content in
case of failure rather than the temporary xml filename.

This patches changes the error log to print the
failed xml file content.

Signed-off-by: Satheesh Rajendran sathnaga@linux.vnet.ibm.com

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -608,8 +608,8 @@ class VMXML(VMXMLBase):
         if not self.define(virsh_instance=virsh_instance):
             if backup:
                 backup.define(virsh_instance=virsh_instance)
-            raise xcepts.LibvirtXMLError("Failed to define %s, from %s."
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -608,8 +608,8 @@ def sync(self, options=None, virsh_instance=base.virsh):
         if not self.define(virsh_instance=virsh_instance):
             if backup:
                 backup.define(virsh_instance=virsh_instance)
-            raise xcepts.LibvirtXMLError("Failed to define %s, from %s."
-                                         % (self.vm_name, self.xml))
+            raise xcepts.LibvirtXMLError("Failed to define %s, from xml:\n%s."
+                                         % (self.vm_name, self.xmltree))
```
will-Do: this make sense

sathnaga:
@will-Do If no other comments can we merge this one, Thanks in advance

PR #815, opened 2016-12-09
virttest.virsh: Add one parameter for virsh option

dzhengfy:
Add a parameter providing virsh option when executing virsh command.
Migration function also adds this parameter for its caller to use.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -1726,7 +1726,7 @@ class VM(virt_vm.BaseVM):
             lockfile.close()
 
     def migrate(self, dest_uri="", option="--live --timeout 60", extra="",
-                ignore_status=False, debug=False):
+                ignore_status=False, debug=False, virsh_opt=""):
         """
         Migrate a VM to a remote host.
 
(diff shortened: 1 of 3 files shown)


virttest/libvirt_vm.py
```diff
@@ -1687,7 +1687,7 @@ def create(self, name=None, params=None, root_dir=None, timeout=5.0,
             lockfile.close()
 
     def migrate(self, dest_uri="", option="--live --timeout 60", extra="",
-                ignore_status=False, debug=False):
+                virsh_opt=None, ignore_status=False, debug=False):
```
will-Do: The above 'options' is not what you want?

will-Do:
@dzhengfy Just want to confirm with you that all callers from both avocado-vt and tp-libvirt should be not effect by this change.

dzhengfy:
@will-Do yes, Latest changes add the new option in the last place. I have checked the other usages and should not have side effects.

dzhengfy:
@waynesun09 , could you help review ? Thanks.

dzhengfy:
I did codes update today. I changed virttest/utils_test/libvirt.py again.
Because I need another parameter for extra options for migration which is already supported by vm.migrate() function.  Due to the new changes are closed to the original changes, I think they will get merging issue if they are in different PRs. So I changed them together. Thanks for understanding.

will-Do:
@dzhengfy 
please address this conflict problem before we move on

dzhengfy:
@will-Do done and please go ahead.

PR #718, opened 2016-10-05
Fixup: VM with multiple interfaces boot failure

sathnaga:
When VM have multiple interfaces and
when IP gets assigned to random interface,
get_address hang on single interface and timeout for boot(import)test results to failure.

This patch will address it by introducing
a parameter""flexible_nic_index"",
when set it will try to poke all the interface and
check which has valid IP and connect to it.

Fixes: https://github.com/avocado-framework/avocado-vt/issues/617

PR #816, opened 2016-12-09
Fixup: VM with multiple interfaces boot failure[V2]

sathnaga:
When VM have multiple interfaces and
IP address can get assigned to any(index) interface

Right now, get_address hang on single interface and
timeout for boot(import) test.

This patch will address it by introducing
a parameter""flexible_nic_index"",
when set it will try to poke all the interface and
check which has valid IP and tries connect to it.

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -624,7 +624,7 @@ class BaseVM(object):
         except KeyError:
             raise VMMACAddressMissingError(nic_index)
 
-    def get_address(self, index=0):
+    def get_address(self, index=0, flexible_index=False):
         """
         Return the IP address of a NIC or guest (in host space).
 
@@ -638,6 +638,7 @@ class BaseVM(object):
         :raise VMAddressVerificationError: If the MAC-IP address mapping cannot
                 be verified (using arping)
         """
+
         nic = self.virtnet[index]
         self.ip_version = self.params.get("ip_version", "ipv4").lower()
         # TODO: Determine port redirection in use w/o checking nettype
@@ -656,8 +657,6 @@ class BaseVM(object):
             arp_ip = self.address_cache.get(nic.mac.lower())
             # Make sure IP is assigned to one or more macs for this guest
             macs = self.virtnet.mac_list()
-            if not arp_ip:
-                arp_ip = self.address_cache.get(nic.mac.upper())
             devs = set()
             if nic.has_key('netdst'):
                 """At this point we know that VM's NIC is attached to a certain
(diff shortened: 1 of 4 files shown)


sathnaga:
@will-Do @xutian have come up with version 2 addressing all the comments and shortcomings in https://github.com/avocado-framework/avocado-vt/pull/718

sathnaga:
```
TESTS      : 4
 (1/4) guest_sanity.boundary.network_max.sparvlan.qemu.qcow2.scsi.smp2.spapr-vlan.Ubuntu.16.10.ppc64le.powerkvm-qemu.unattended_install.import.import.default_install.aio_native: PASS (220.36 s)
 (2/4) guest_sanity.boundary.network_max.sparvlan.qemu.qcow2.scsi.smp2.spapr-vlan.Ubuntu.16.10.ppc64le.powerkvm-libvirt.remove_guest.without_disk: PASS (1.74 s)
 (3/4) guest_sanity.boundary.network_max.virtionet.qemu.qcow2.scsi.smp2.virtio_net.Ubuntu.16.10.ppc64le.powerkvm-qemu.unattended_install.import.import.default_install.aio_native: PASS (164.80 s)
 (4/4) guest_sanity.boundary.network_max.virtionet.qemu.qcow2.scsi.smp2.virtio_net.Ubuntu.16.10.ppc64le.powerkvm-libvirt.remove_guest.without_disk: PASS (1.76 s)
RESULTS    : PASS 4 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0

```

virttest/virt_vm.py
```diff
@@ -638,6 +638,7 @@ def get_address(self, index=0):
         :raise VMAddressVerificationError: If the MAC-IP address mapping cannot
                 be verified (using arping)
         """
+
         nic = self.virtnet[index]
         self.ip_version = self.params.get("ip_version", "ipv4").lower()
```
will-Do: The original idea is not add a new param in function define, but get it from test params dict, just like the 'ip_version':
```
self.flexible_index = "yes" == self.params.get("flexible_index", "no")
```
then no need update these callers.
And maybe an info message about using  flexible index instead of index=0 is necessary

sathnaga: @will-Do I was doing so in my initial version but based on your suggestion as below have updated this PR, now am confused :-(

will-Do 24 days ago Collaborator
@sathnaga
According to this solution, my suggestion is make flexible_index as the 3rd param of this function:

     def get_address(self, index=0, flexible_index=False):
And use index=0 as high priority

will-Do: @sathnaga 
My fault, guess I mess it up. I didn't notice that when add a new param here, it will cause a lot of changes, and for new caller, we always need pass it explicit. Maybe we can reopen the original one, what do you think? Sorry for make you confused

sathnaga:
@will-Do Sure, np, have reopened the original one and pls do merge, Thanks in advance :-)

PR #819, opened 2016-12-13
virsh: Add function for guestvcpus

Grace-chhu:
Work together with tp-libvirt: PR #959
PR #959: virsh.guestvcpus: Add test cases for guestvcpus

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -759,6 +759,23 @@ def setvcpus(name, count, extra="", **dargs):
     return command(cmd, **dargs)
 
 
+def guestvcpus(name, cpu_list=None, options=None, **dargs):
+    """
+    Query or modify state of vcpu in the guest (via agent)
+    :param name: name of domain
+    :param cpu_list: list of cpus to enable or disable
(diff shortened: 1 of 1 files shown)


virttest/virsh.py
```diff
@@ -759,6 +759,23 @@ def setvcpus(name, count, extra="", **dargs):
...
+    """
+    Query or modify state of vcpu in the guest (via agent)
+    :param name: name of domain
+    :param cpu_list: list of cpus to enable or disable
+    :param options: --enable, --disable
+    :param dargs: standardized virsh function API keywords
+    :return: CmdResult object
+    """
+    cmd = "guestvcpus --domain %s" % name
+    if cpu_list is not None:
```
balamuruhans: `if cpu_list:` will be sufficient

virttest/virsh.py
```diff
@@ -759,6 +759,23 @@ def setvcpus(name, count, extra="", **dargs):
...
+        cmd += " --cpulist %s" % cpu_list
+    if options is not None:
```
balamuruhans: same here

PR #820, opened 2016-12-14
Update the order of cpu model

suqinhuang:
host_cpu_model equal to the first of cpu_support_model
need to put the more advance cpu model at the beginning.

id: 1403716


xutian:
LGTM

suqinhuang:
updated, could you review it again @xutian 
Thanks

xutian:
looks OK to me.

PR #823, opened 2016-12-16
utils_netperf: Extract "prompt,linesep and status_test_command" as params

aliang123:
utils_netperf:
1.extract shell,linesep and status_test_command as params for both "ssh" and "nc"

Signed-off-by: Aihua Liang<aliang@redhat.com>

ID:1261786,1263919

### DIFF
--- a/virttest/utils_netperf.py
+++ b/virttest/utils_netperf.py
@@ -50,7 +50,8 @@ class NetperfTestError(NetperfError):
 class NetperfPackage(remote.Remote_Package):
 
     def __init__(self, address, netperf_path, md5sum="", netperf_source="",
-                 client="ssh", port="22", username="root", password="123456"):
+                 client="ssh", port="22", username="root", password="123456",
+                 prompt="^root@.*[\#\$]\s*$|", linesep="\n", status_test_command="echo $?"):
         """
         Class NetperfPackage just represent the netperf package
         Init NetperfPackage class.
@@ -75,14 +76,7 @@ class NetperfPackage(remote.Remote_Package):
         self.netperf_base_dir = self.remote_path
         self.netperf_file = os.path.basename(self.netperf_source)
 
-        if client == "nc":
-            self.prompt = r"^\w:\\.*>\s*$"
-            self.linesep = "\r\n"
-            self.status_test_command = "echo %errorlevel%"
-        else:
-            self.prompt = "^\[.*\][\#\$]\s*$"
-            self.linesep = "\n"
-            self.status_test_command = "echo $?"
+        if client == "ssh":
             if self.netperf_source.endswith("tar.bz2"):
                 self.pack_suffix = ".tar.bz2"
                 self.decomp_cmd = "tar jxvf"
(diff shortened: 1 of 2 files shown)


virttest/utils_netperf.py
```diff
@@ -75,14 +76,7 @@ def __init__(self, address, netperf_path, md5sum="", netperf_source="",
...
 
-        if client == "nc":
-            self.prompt = r"^\w:\\.*>\s*$"
-            self.linesep = "\r\n"
-            self.status_test_command = "echo %errorlevel%"
-        else:
-            self.prompt = "^\[.*\][\#\$]\s*$"
-            self.linesep = "\n"
-            self.status_test_command = "echo $?"
+        if client == "ssh":
```
CongLi: if so, no need to judge ssh here.

CongLi:

virttest/utils_netperf.py
xutian: why is ssh?

virttest/utils_netperf.py
```diff
@@ -106,9 +100,9 @@ def __init__(self, address, netperf_path, md5sum="", netperf_source="",
         logging.debug("Create remote session")
         self.session = remote.remote_login(self.client, self.address,
                                            self.port, self.username,
-                                           self.password, self.prompt,
-                                           self.linesep, timeout=360,
-                                           status_test_command=self.status_test_command)
+                                           self.password, prompt,
```
xutian: looks should be self.prompt ?

virttest/utils_netperf.py
```diff
@@ -106,9 +100,9 @@ def __init__(self, address, netperf_path, md5sum="", netperf_source="",
...
+                                           linesep, timeout=360,
+                                           status_test_command=status_test_command)
```
xutian: self.linsep ?

virttest/utils_netperf.py
```diff
@@ -245,6 +232,7 @@ class NetperfServer(Netperf):
 
     def __init__(self, address, netperf_path, md5sum="", netperf_source="",
                  client="ssh", port="22", username="root", password="redhat",
+                 prompt="", linesep="", status_test_command="",
```
xutian: same issue

virttest/utils_netperf.py
```diff
@@ -197,25 +192,17 @@ def __init__(self, address, netperf_path, md5sum="", netperf_source="",
...
-                                      password)
+                                      password, prompt, linesep, status_test_command)
         self.netserver_path, self.netperf_path = self.package.install(install,
                                                                       compile_option)
         logging.debug("Create remote session")
         self.session = remote.remote_login(client, address, port, username,
-                                           password, self.prompt,
-                                           self.linesep, timeout=360,
-                                           status_test_command=self.status_test_command)
+                                           password, prompt,
```
xutian: same question here

aliang123: In original code, self.prompt/self.linesep/self.status_test_command will be redefined in class NetperfPackage and Netperf, and these definations are the same. Finally, all these params will be only used by remote.remote_login()

Although we have introduce self.prompt/self.linesep/self.status_test_command as params, i think there is no need to use the self.* for these three params when passing them to Netperf, NetperfPackage, remote.remote_login().

virttest/utils_netperf.py
aliang123: ssh has different decomp_cmd for different pack_suffix(".tar.gz" or ".tar.bz2").

aliang123:
Test Result, see https://github.com/avocado-framework/avocado-vt/pull/823

xutian:
looks Ok to me

xutian:
looks Ok to me

PR #827, opened 2016-12-20
utils_net: support to check network interfaces of remote machine

balamuruhans:
get_sorted_net_if() and get_net_if() can be used to get network
interfaces of remote machines

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -1304,6 +1304,43 @@ def get_sorted_net_if():
     return (phy_interfaces, vir_interfaces)
 
 
+def get_remote_host_net_ifs(session, state=None):
+    """
+    Get all network interfaces of remote host and sort them as physical
+    and virtual interfaces.
+
+    :param session: remote session object
+    :param state: regex used for state
+    :return: Tuple of physical and virtual interfaces lists
+    """
+    phy_interfaces = []
+    vir_interfaces = []
+    cmd = "ip link"
+    if not state:
+        state = ".*"
+    cmd_output = session.cmd_status_output(cmd)
+    if cmd_output[0]:
+        exceptions.TestError("Failed to fetch %s from remote"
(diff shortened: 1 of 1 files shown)


balamuruhans:
@sathnaga Thank you

virttest/utils_net.py
```diff
@@ -1265,41 +1265,60 @@ def local_runner_status(cmd, timeout=None, shell=False):
...
     :return: List of network interfaces.
     """
-    if runner is None:
-        runner = local_runner
+    cmd = "ip link"
     if state is None:
         state = ".*"
-    cmd = "ip link"
-    result = runner(cmd)
+    if session:
```
xutian: it doesn't work, if OS installed in remote machine is windows

balamuruhans: @xutian, this API was already with command "ip link", even without this support for remote machine, I doubt it works for windows.

xutian: @balamuruhans  right.  original code only works on host, so it linux OS only.  after this change, session maybe a remote windows guest shell session. so if you can add windows session support is better. And wmic is helpful to implement it.

balamuruhans: @xutian, okay but I dont have env to work with windows

xutian: @balamuruhans  it better write another function like get_remote_host_net_ifs(). I have a plan to rewrite get_net_ifs and get_sorted_net_ifs with netifaces module to make utils_net looks more clear.

balamuruhans: @xutian sure. I will write another function `get_remote_host_net_ifs()`

balamuruhans:
@xutian, have written another function `get_remote_host_net_ifs()` as suggested.

xutian:
LGTM

xutian:
works for me, merged thanks @balamuruhans

PR #829, opened 2016-12-26
shared.deps: fix typo in finish.bat

xutian:


yanan-fu:
As we have tested, this patch can fix the problem
Acked-by: Yanan Fu yfu@redhat.com

PR #831, opened 2016-12-27
cartesian: Parsing suffix before yielding dict

Hao-Liu:
The current Cartesian parsing performance is very low. It'll take
several minutes to list all the cases in a test provider.

After some CPU performance profiling, we can find the bottleneck of
current Cartesian parsing process is a function named postfix_parse.
This function is used to clip the range of a value if its key ends with
'_min', '_max' or '_fixed' and is called after the whole configuration
is combined and generated. Since the combination may generates very huge
number possible results, this function then will be called too many
times.

This patch reduce the calculation by calling this function right before
parser reach a leaf node. This will greatly decrease the number of
calling this function thus increase the performance of list large number
of test cases.

### DIFF
--- a/virttest/cartesian_config.py
+++ b/virttest/cartesian_config.py
@@ -1939,7 +1939,6 @@ class Parser(object):
             for d in self.get_dicts_plain(node, ctx, content, shortname, dep):
                 if parent:
                     self.drop_suffixes(d)
-                postfix_parse(d)
                 yield d
         else:
             # Rewrite all separate joins in one node as many `only'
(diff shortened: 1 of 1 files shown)


Hao-Liu:
@xutian, Please help check if this patch can fix the performance problem you met. Thanks.

xutian:
@Hao-Liu thanks your commit, I will test it later.

xutian:
tested, this commit works fine on my host. thanks @Hao-Liu

xutian:
@CongLi and @suqinhuang  can you second review this commit. thanks

CongLi:

PR #837, opened 2017-01-04
Sysconfig file location is set for RHEL by default

sudeeshjohn:
The sysconfig file location is set for RHEL by default, but it different for Ubuntu. This patch setting the config path based on the distro.

Signed-off-by: Sudeesh John sudeesh@linux.vnet.ibm.com


suqinhuang:
ack

xutian:
merged, thanks @sudeeshjohn

PR #841, opened 2017-01-06
virttest.qemu_vm: add vm status check

xutian:
need to check VM status after send monitor command pause
or cont to ensure the command take effect really.


xutian:
thanks @sathnaga, small fix so merged with one ack

PR #845, opened 2017-01-09
shared.cfg: fix typo in base.cfg

xutian:


xutian:
sample hotfix to avoid it block test on ppc platform.

PR #847, opened 2017-01-09
nfs.py: Fix nfs umount error

dzhengfy:
Sometimes umount operation fails unexpectedly. This is a workaround
to always restart the service before the testing which can avoid of
the umount error.

### DIFF
--- a/virttest/nfs.py
+++ b/virttest/nfs.py
@@ -203,10 +203,9 @@ class Nfs(object):
         service and exportfs too.
         """
         if self.nfs_setup:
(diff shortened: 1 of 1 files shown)


virttest/nfs.py
```diff
@@ -203,10 +203,9 @@ def setup(self):
         service and exportfs too.
         """
         if self.nfs_setup:
-            if not self.nfs_service.status():
```
PandaWei: remove the judgment is not good choice.
you should find out the cause leads to the failure of umount operation.

PR #851, opened 2017-01-11
virttest.bootstrap: fix KeyError exception

xutian:
fix KeyError, if 'type_' is  prefix for a key of test params.

ID: 1405926


suqinhuang:
Ack

xutian:
merged, thanks @suqinhuang

PR #855, opened 2017-01-13
Support for ipv6 connection to Spice client/server

rdkdd:
Instead of using utils_net.get_host_ip_address(self.params) to translate ipv4 address to ipv6, real ipv6 address of utilized interface is used.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -994,17 +994,10 @@ class VM(virt_vm.BaseVM):
                 set_value("port=%s", "spice_port")
 
             set_value("password=%s", "spice_password", "disable-ticketing")
-            if optget("listening_addr") == "ipv4":
-                host_ip = utils_net.get_host_ip_address(self.params)
-                self.spice_options['listening_addr'] = "ipv4"
+            ip_ver = optget("listening_addr")
+            if ip_ver:
+                host_ip = utils_net.get_host_ip_address(self.params, ip_ver)
                 spice_opts.append("addr=%s" % host_ip)
-                # set_value("addr=%s", "listening_addr", )
-            elif optget("listening_addr") == "ipv6":
-                host_ip = utils_net.get_host_ip_address(self.params)
-                host_ip_ipv6 = utils_misc.convert_ipv4_to_ipv6(host_ip)
-                self.spice_options['listening_addr'] = "ipv6"
-                spice_opts.append("addr=%s" % host_ip_ipv6)
-
             set_yes_no_value(
                 "disable_copy_paste", yes_value="disable-copy-paste")
             set_value("addr=%s", "spice_addr")
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3007,26 +3007,28 @@ def get_ip_address_by_interface(ifname):
             "Error while retrieving IP address from interface %s." % ifname)
 
 
-def get_host_ip_address(params):
+def get_host_ip_address(params, ip_version="ipv4"):
     """
     Returns ip address of host specified in host_ip_addr parameter if provided.
     Otherwise ip address on interface specified in netdst parameter is returned.
     In case of "nettype == user" "netdst" parameter is left empty, then the
     default interface of the system is used.
     :param params
+    :param ip_version: Host IP version, ipv4 or ipv6.
     """
     host_ip = params.get('host_ip_addr', None)
     if host_ip:
         logging.debug("Use IP address at config %s=%s", 'host_ip_addr', host_ip)
         return host_ip
     net_dev = params.get('netdst')
-    if not net_dev:
+    if net_dev not in get_host_iface():
         net_dev = get_host_default_gateway(iface_name=True)
     if not net_dev:
         raise NetError("Error while retrieving IP address for host.")
     logging.warning("No IP address of host was provided, using IP address"
                     " on %s interface", str(net_dev))
-    host_ip = get_ip_address_by_interface(net_dev)
+    nic_address = get_net_if_addrs(net_dev)
+    host_ip = nic_address[ip_version][0]
     return host_ip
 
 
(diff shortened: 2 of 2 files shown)


rdkdd:
please take a look @will-Do.
thanks.

virttest/utils_net.py
```diff
@@ -3026,7 +3026,12 @@ def get_host_ip_address(params):
         raise NetError("Error while retrieving IP address for host.")
     logging.warning("No IP address of host was provided, using IP address"
                     " on %s interface", str(net_dev))
-    host_ip = get_ip_address_by_interface(net_dev)
+    # host_ip = get_ip_address_by_interface(net_dev)
```
Andrei-Stepanov: You can safely remove line with commented out function.
Commenting out code is a bad approach for public code.
You always can get previous version using git.

virttest/utils_net.py
```diff
@@ -3026,7 +3026,12 @@ def get_host_ip_address(params):
...
+    ips = get_net_if_and_addrs()
+    ip_v = params.get('listening_addr')
+    if not ip_v:
+        ip_v = 'ipv4'
```
Andrei-Stepanov: redundant code

Example:
a = {}   # Empty dic
a.get('listening_addr', 'ipv4')
Out[3]: 'ipv4'

virttest/utils_net.py
```diff
@@ -3026,7 +3026,12 @@ def get_host_ip_address(params):
...
+    host_ip = ips[net_dev][ip_v][0]
```
Andrei-Stepanov: Very-very dangerous code.
There is no any guaranty that ips holds 'net_dev' and it holds 'ip_v' and it has at least one address.

virttest/qemu_vm.py
```diff
@@ -994,16 +994,10 @@ def set_value(opt_string, key, fallback=None):
                 set_value("port=%s", "spice_port")
 
             set_value("password=%s", "spice_password", "disable-ticketing")
-            if optget("listening_addr") == "ipv4":
-                host_ip = utils_net.get_host_ip_address(self.params)
-                self.spice_options['listening_addr'] = "ipv4"
+            if optget("listening_addr"):
+                host_ip = utils_net.get_host_ip_address(
+                    self.params, ip_version=optget("listening_addr"))
```
Andrei-Stepanov: Make it more obvious:

ip_ver = optget("listening_addr")
if ip_ver:
    ost_ip = utils_net.get_host_ip_address(self.params, ip_version=ip_ver)

virttest/utils_net.py
```diff
@@ -3007,26 +3007,28 @@ def get_ip_address_by_interface(ifname):
...
     host_ip = params.get('host_ip_addr', None)
     if host_ip:
         logging.debug("Use IP address at config %s=%s", 'host_ip_addr', host_ip)
         return host_ip
     net_dev = params.get('netdst')
-    if not net_dev:
+    if net_dev not in get_host_iface():
         net_dev = get_host_default_gateway(iface_name=True)
-    if not net_dev:
+    if net_dev not in get_host_iface():
```
Andrei-Stepanov: if net_dev not in get_host_iface():
is redundant.

It is responsibility get_host_iface() to return right interface name.

if not net_dev: -- is enough

rdkdd:
Please take a look @will-Do.
The ipv6 addresss of host interface cannot be obtained using get_ip_address_by_interface(ifname).
I changed it, so now is obtained by get_net_if_addrs(net_dev). I also enhanced  get_host_ip_address function by ip_version parameter.

Spice-qe team needs it to test ipv6 SPICE proxy (https://github.com/spiceqa/tp-spice/pull/51/commits/36df6fca9002b03edd71ce1d89f3bbaf71fc01d3).

thanks.

rdkdd:
Could you please take a look  @xutian?
Thanks!

virttest/utils_net.py
```diff
@@ -3007,26 +3007,28 @@ def get_ip_address_by_interface(ifname):
...
     :param params
+    :param ip_version: Host IP version, ipv4 or ipv6.
     """
     host_ip = params.get('host_ip_addr', None)
     if host_ip:
         logging.debug("Use IP address at config %s=%s", 'host_ip_addr', host_ip)
         return host_ip
     net_dev = params.get('netdst')
-    if not net_dev:
+    if net_dev not in get_host_iface():
```
xutian: Why not make get_ip_address_by_interface() support get IPv6 address?

xutian: I perfer
I prefer simple code like:

```
default_dev = get_host_default_gateway(iface_name=True)
net_dev = params.get("netdst", default_dev)
return get_ip_address_by_interface(netdev, ip_ver)
```

rdkdd: >I prefer simple code like:
default_dev = get_host_default_gateway(iface_name=True)
net_dev = params.get("netdst", default_dev)
return get_ip_address_by_interface(netdev, ip_ver)

ok, I can change it this way

> Why not make get_ip_address_by_interface() support get IPv6 address?

@xutian I do not know how to do this using similar approach as in get_ip_address_by_interface() (e.g. fcntl.ioctl and socket). I would have to divide this function by condition into two branches ipv4 & ipv6 - it does not seem to me practical enough. It seems easier to use get_net_if_addrs. What is your opinion?

xutian: Hi @rdkdd ，

 please use 'netifaces' module, it will make your problem becomes very simple.

thanks,
Xu

xutian: @rdkdd 
here has some example http://www.programcreek.com/python/example/81895/netifaces.interfaces

rdkdd: @xutian yes solution with netifaces is pretty easy and neat. But the module included in python standard library. Should I add the package to requirements.txt?

xutian: yes,  you are right.

rdkdd:
```
default_dev = get_host_default_gateway(iface_name=True)
net_dev = params.get("netdst", default_dev)
return get_ip_address_by_interface(netdev, ip_ver)
```
I cannot make it like this, because netdst is defined in params as empty string ('netdst='). get() function does not return default_dev in this case. So I must use if condition.

Made a new PR https://github.com/avocado-framework/avocado-vt/pull/880. Please take a look @xutian

PR #861, opened 2017-01-18
Empty soundcards params for ppc64le platform

Zhengtong:
ID: 1414244

For ppc platform , no soundcards support for qemu, remove it

### DIFF
--- a/shared/cfg/machines.cfg
+++ b/shared/cfg/machines.cfg
@@ -34,6 +34,7 @@ variants:
         # Currently only supports standard VGA
         vga = std
         del rtc_drift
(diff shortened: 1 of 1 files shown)


shared/cfg/machines.cfg
```diff
@@ -34,6 +34,7 @@ variants:
         # Currently only supports standard VGA
         vga = std
         del rtc_drift
+        soundcards = ''
```
xutian: better to
del soundcards

Zhengtong: It doesn't matter so much, updated

PR #877, opened 2017-02-07
utils_misc: Get meminfo for all nodes

will-Do:

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -1748,6 +1748,25 @@ class NumaInfo(object):
 
         return node_distance.strip().split()
 
+    def get_all_node_meminfo(self):
+        """
+        Get the complete meminfo of all nodes.
+
+        :return: All nodes' meminfo
+        :rtype: Dict
+        """
+        meminfo = {}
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -1748,6 +1748,25 @@ def get_node_distance(self, node_id):
...
+        meminfo_file = os.path.join(self.numa_sys_path, "node%s/meminfo")
+        for node in get_all_nodes():
+            node_meminfo = {}
+            meminfo_f = open(meminfo_file % node, 'r')
+            for info in meminfo_f.readlines():
+                key, value = re.match(r'Node \d+ (\S+):\s+(\d+)', info).groups()
+                node_meminfo[key] = value
+            meminfo_f.close()
+            meminfo.append(node_meminfo)
+        return meminfo
```
balamuruhans: I will be good to have `meminfo` as a dict with key as `node` instead of list, as we can't rely on the order in which list gets appended

will-Do: OKay, will update it, thx

PR #882, opened 2017-02-10
utils.misc: add new funtion for "balloon_illegal" test

aliang123:
virttest.utils_misc:
1.add function "get_win_mon_free_mem()"

Signed-off-by:Aihua Liang <aliang@redhat.com>

ID:1372566

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2620,8 +2620,8 @@ def get_free_disk(session, mount):
     """
     Get FreeSpace for given mount point.
 
-    :parm session: shell Object.
-    :parm mount: mount point(eg. C:, /mnt)
+    :param session: shell Object.
+    :param mount: mount point(eg. C:, /mnt)
 
     :return string: freespace M-bytes
     """
@@ -2639,12 +2639,27 @@ def get_free_disk(session, mount):
     return int(free)
 
 
+def get_win_mon_free_mem(session):
+    """
+    Get Performace Monitored Free memory for Windows.
+
+    :param session: shell Object.
+
+    :return string: freespace M-bytes
+    """
+    cmd = 'typeperf "\Memory\Free & Zero Page List Bytes" -sc 1'
+    output = session.cmd_output(cmd)
+    free = "%s" % re.findall("\d+\.\d+", output)[2]
+    free = float(normalize_data_size(free, order_magnitude="M"))
+    return int(free)
+
+
 def get_free_mem(session, os_type):
     """
     Get Free memory for given OS.
 
-    :parm session: shell Object.
-    :parm os_type: os type (eg. linux or windows)
+    :param session: shell Object.
+    :param os_type: os type (eg. linux or windows)
 
     :return string: freespace M-bytes
     """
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -2626,6 +2626,22 @@ def get_free_disk(session, mount):
     return int(free)
 
 
+def get_win_mon_free_mem(session, os_type):
```
suqinhuang: does it used for windows only? if so ,no need to transfer os_type

aliang123: yes, it's only used for "windows", will update it later.

xutian: no need os_type is it's windows only

aliang123: ok, i'll update it later.

virttest/utils_misc.py
```diff
@@ -2626,6 +2626,21 @@ def get_free_disk(session, mount):
     return int(free)
 
 
+def get_win_mon_free_mem(session):
+    """
+    Get Performace Monitored Free memory for Windows.
+
+    :parm session: shell Object.
```
balamuruhans: %s/parm/param

aliang123: ok, will correct it, thanks.

xutian:
split into two commits looks better,  one for fix typo

virttest/utils_misc.py
```diff
@@ -2639,12 +2639,27 @@ def get_free_disk(session, mount):
...
+def get_win_mon_free_mem(session):
+    """
+    Get Performace Monitored Free memory for Windows.
+
+    :param session: shell Object.
+
+    :return string: freespace M-bytes
+    """
+    cmd = 'typeperf "\Memory\Free & Zero Page List Bytes" -sc 1'
+    output = session.cmd_output(cmd)
```
balamuruhans: as status is not checked here, if typeperf is not installed then it might break the API

aliang123: will update, thanks!

virttest/utils_misc.py
```diff
@@ -2639,12 +2639,30 @@ def get_free_disk(session, mount):
...
+    if output.find("not recognized") == -1:
```
pingl2016: "str.find("substr") == -1" means "substr not in str"

virttest/utils_misc.py
aliang123: @balamuruhans 
   "typeperf" is an internal cmd of windows, so we don't need to install it manually.

CongLi: @aliang123 

Agree with @balamuruhans .

Although "typeperf" is an internal cmd of windows, no need to check 
whether it's installed, but it's better to check cmd execution status.

aliang123: Like "free -m" in linux,  "typeperf" is old for windows, it exists since WinXP, mainly used for writing performance counter data to the command window, or to a supported log file format.
   As a windows built-in cmd, "typeperf" always has a returned value.
   So, from my point of view, no need to check its execution status, thanks.

xutian:
Please split into two commit, one for fix typo. and another one should be in tp-qemu/qemu/tests/balloon_check.py: BallooningTestWin

Thanks,
Xu

aliang123:
According to Xu's suggestion, has moved the function in tp-qemu/tests/balloon_check.py, so close the pullreq.

PR #890, opened 2017-02-16
Fix auto_ip lookup for unattended_install.py.

Andrei-Stepanov:


apahim:
lgtm

PR #891, opened 2017-02-18
utils_misc: Fix function name scope typos

Hao-Liu:
There are two function references in a latest patch that mistaken
referred class functions as module functions, thus leading several
test cases failed with:

```
NameError: global name 'get_all_node_meminfo' is not defined
```


waynesun09:
ACK

Hao-Liu:
Thanks @waynesun09, Merging this.

PR #897, opened 2017-02-24
qemu_vm: add params for add memory devices

xutian:
Reset vm memory size according host free memory size if 'automem' params is 'yes'(default is 'yes'). For
migration related test case 'automem' should be 'no' to avoid migration failed.
And add device check before add 'slots' and 'maxmem' option for memory devices.

ID: 1426892

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -1810,8 +1810,8 @@ class DevContainer(object):
         pc-dimm devices.
         """
         devices = []
-        if not self.has_option("object"):
-            logging.warn("QOM does not support by your qemu")
+        if not self.has_device("pc-dimm"):
+            logging.warn("'PC-DIMM' does not support by your qemu")
             return devices
         mem = self.memory_object_define_by_params(params, name)
         if mem:
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -896,25 +896,32 @@ class VM(virt_vm.BaseVM):
 
             :param devices: VM devices container
             """
-            options = []
+            options, devs = [], []
+            params.setdefault("mem", "512")
             mem_params = params.object_params("mem")
-            if mem_params.get("maxmem"):
+            mem_params.setdefault("automem", "yes")
+            automem = mem_params["automem"] == "yes"
+            if automem:
                 normalize_data_size = utils_misc.normalize_data_size
-                values = map(lambda x: "%sM" % mem_params.get(x, 512), ["maxmem", "mem"])
-                values = map(lambda x: int(float(normalize_data_size(x))), values)
-                params["mem"] = min(values)
-            options.append("%s" % params.get("mem", 512))
-            if mem_params.get("slots") and mem_params.get("maxmem"):
-                options.append("slots=%s" % mem_params["slots"])
-                options.append("maxmem=%s" % mem_params["maxmem"])
-
-            cmdline = "-m %s" % ",".join(options)
-            dev = StrDev("mem", cmdline=cmdline)
-            devices.insert(dev)
-            for name in mem_params.objects("mem_devs"):
-                memdev_params = mem_params.object_params(name)
-                mem_devs = devices.memory_define_by_params(memdev_params, name)
-                devices.insert(mem_devs)
+                mem_size_m = "%sM" % mem_params["mem"]
+                mem_size_m = int(float(normalize_data_size(mem_size_m)))
+                max_usable_mem_m = int(mem_params["max_usable_mem"])
+                if mem_size_m >= max_usable_mem_m:
+                    logging.debug("Host no enough free memory, reset guest"
+                                  " memory size to %s MB" % max_usable_mem)
+                    params["mem"] = max_usable_mem_m
+            options.append(params["mem"])
+            if devices.has_device("pc-dimm"):
+                if mem_params.get("slots") and mem_params.get("maxmem"):
+                    options.append("slots=%s" % mem_params["slots"])
+                    options.append("maxmem=%s" % mem_params["maxmem"])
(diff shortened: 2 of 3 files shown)


virttest/qemu_vm.py
```diff
@@ -898,11 +898,17 @@ def add_memorys(devices, params):
             """
             options = []
             mem_params = params.object_params("mem")
-            if mem_params.get("maxmem"):
+            mem_params.setdefault("mem", "512")
+            automem = mem_params.get("automem", "yes") == "yes"
```
suqinhuang: the memory may over commit for migration with the method, and I rem migration will failed when mem over commit

xutian: right,  if there no enough memory to do migration,  test fail is expect.

xutian: if source VM memory size not equal to destination VM  memory size, migration maybe failed too, so  that's why https://github.com/autotest/tp-qemu/pull/933 disable automem  for migration test.

xutian: origin code will reset destination VM memory size if no enough free memory space to  launch destination VM.  if migration failed because different memory size, it's very hard for tester to found it. I think better to make test failed with real reason.

xutian:
@suqinhuang  any other comments ?

virttest/qemu_vm.py
```diff
@@ -896,25 +896,35 @@ def add_memorys(devices, params):
...
             :param devices: VM devices container
             """
-            options = []
+            options, devs = [], []
+            params.setdefault("mem", "512")
             mem_params = params.object_params("mem")
-            if mem_params.get("maxmem"):
+            mem_params.setdefault("automem", "yes")
+            automem = mem_params["automem"] == "yes"
+            if automem:
```
luckyh: What kind of scenario do we need to automatically get memory size for guest?

virttest/qemu_vm.py
```diff
@@ -896,25 +896,35 @@ def add_memorys(devices, params):
...
+                mem_keys = ["maxmem", "mem", "max_usable_mem", "maxmem_mem"]
+                mem_values = [mem_params[x] for x in mem_keys if mem_params.get(x)]
+                mem_values = map(
+                    lambda x: int(
+                        float(
+                            normalize_data_size(
+                                '%sM' %
+                                x))),
+                    mem_values)
+                params["mem"] = min(mem_values)
```
luckyh: And why does the memory size has to be set as the minimum value of the above params?

virttest/qemu_vm.py
xutian: request memory size bigger than free memory size + 256M or request memory bigger than maxmen size, will reset it to avoid create VM failed.

virttest/qemu_vm.py
xutian: because ‘maxmem’, 'max_usable_mem', 'maxmem_mem' are roof of memory guest can use.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1812,8 +1812,8 @@ def memory_define_by_params(self, params, name):
         pc-dimm devices.
         """
         devices = []
-        if not self.has_option("object"):
-            logging.warn("QOM does not support by your qemu")
+        if not self.has_device("pc-dimm"):
+            logging.warn("'PC-DIMM' does not support by your qemu")
```
nasastry: `self.has_device` returns `True` when `pc-dimm` device is available. Should remove `not` to match with the `logging.warn` message.
`if self.has_device("pc-dimm"):`
          ` logging.warn("'PC-DIMM' does not support by your qemu")`

virttest/qemu_vm.py
luckyh: Requesting guest memory size bigger than the value of `maxmem` in the qemu `-m` option is a negative test case, and in this situation, a VM create error is what we expected.

We can let user decide to shrink the memory size to be close to the free value, but if we got a very small value (such as 256 MB), the guest os is more likely to refuse to boot up. I don't know if setting this as default is a good choice.

Moreover, if the guest is using a memory file backend (hugetlbfs), I'm not sure if these code can calculate the value correctly.

virttest/qemu_devices/qcontainer.py
xutian: if "pc-dimm" devices not support, skip params for pc-dimm and memory object,  it looks current logical is right.

xutian: not support raise warning, then skip add devices.

virttest/qemu_vm.py
```diff
@@ -896,25 +896,35 @@ def add_memorys(devices, params):
 
             :param devices: VM devices container
             """
-            options = []
+            options, devs = [], []
+            params.setdefault("mem", "512")
             mem_params = params.object_params("mem")
-            if mem_params.get("maxmem"):
+            mem_params.setdefault("automem", "yes")
```
balamuruhans: @xutian you have mentioned migration will not work if `automem` is `yes`, so should we have a check here if a migration test uses to add memory controller, then we have to  `mem_params.setdefault("automem", "no")`

xutian: @balamuruhans  have sent pull-request for migration related case to set  automem = "no",  details see https://github.com/autotest/tp-qemu/pull/933.
Thanks your comments.
Xu

balamuruhans: @xutian cool then.

virttest/qemu_vm.py
xutian: it's really a issue for memory in 1G hugepage,  I think we 1G to align memory size can avoid issues you mention in above.

xutian:
@luckyh  any other comments for this PR ?

PR #906, opened 2017-03-01
virttest.virt_vm:ensure to remove stale ip from address_cache

maxujun:
Signed-off-by: Xujun Ma <xuma@redhat.com>
id:1421906

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -677,8 +677,11 @@ class BaseVM(object):
             # get valid IPADDR
             if arp_ip:
                 if not utils_net.verify_ip_address_ownership(arp_ip, macs, devs=devs):
-                    self.address_cache[nic.mac.lower()] = None
-                    arp_ip = None
+                    if self.address_cache.get(nic.mac.lower()) == arp_ip:
+                        self.address_cache[nic.mac.lower()] = None
+                        arp_ip = None
+                    else:
+                        arp_ip =None
 
             if (not arp_ip) and (len(macs) > 1) and flexible_index:
                 # Try to poke for each index
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -681,8 +681,8 @@ class BaseVM(object):
                         self.address_cache[nic.mac.lower()] = None
                         arp_ip = None
                     else:
-                        arp_ip =None
-
+                        arp_ip = self.address_cache.get(nic.mac.lower())
+                        
             if (not arp_ip) and (len(macs) > 1) and flexible_index:
                 # Try to poke for each index
                 for idx in range(len(macs)):


virttest/qemu_vm.py
```diff
@@ -3022,6 +3023,10 @@ def _cleanup(self, free_mac_addresses):
             except OSError:
                 pass
 
+        nic = self.virtnet[0]
```
xutian: I think this lines should be add after line#3034.

virttest/qemu_vm.py
```diff
@@ -3022,6 +3023,10 @@ def _cleanup(self, free_mac_addresses):
...
+        if nic.mac:
+            self.address_cache[nic.mac.lower()] = None
```
xutian: it should be
del self.address_cache[nic.mac.lower()]

xutian:
After a short think, I think it should be add in wait_for_get_address function before raise verify_ip_mac relationship failed.

virttest/qemu_vm.py
xutian: we can't add it here,  if VM use a static IP address, it will remove IP/MAC mapping,  so add it in wait_for_get_address  looks more reasonable.

maxujun:
@xutian but the  wait_for_get_address function isn't used by login function,do you mean I need to modify it in get_address function?

xutian:
it's looks need in get_address.

virttest/virt_vm.py
```diff
@@ -659,6 +659,7 @@ def get_address(self, index=0):
             raise VMMACAddressMissingError(index)
         if self.ip_version == "ipv4":
             # Get the IP address from arp cache
+            del self.address_cache[nic.mac.lower()]
```
XueqiangWei: Delete key "nic.mac.lower()" is not ok, in a test loop other test cases also use it. The error likes:
Unhandled KeyError: '9a:72:26:02:68:f0'    [context: (run) --> logging into 'virt-tests-vm1' --> (wait_for_get_address)]

sathnaga: It is not good idea to determine the stale ip from this code path, we would endup in above mentioned issues.

maxujun:
Have updated.

xutian:
@maxujun  it looks we just need to sed IP to None  before raise VMAddressVerificationError in line#737

maxujun:
@sathnaga I have updated this patch ,could you help review it?

virttest/virt_vm.py
maxujun: @sathnaga Setting ip to None maybe cover the ip has updated in address_cache, so I think shoud modify it in this code path.

xutian:
@maxujun  I think https://github.com/avocado-framework/avocado-vt/pull/906 fixed your issue, can you have a try. thanks.

virttest/virt_vm.py
```diff
@@ -677,9 +677,12 @@ def get_address(self, index=0):
             # get valid IPADDR
             if arp_ip:
                 if not utils_net.verify_ip_address_ownership(arp_ip, macs, devs=devs):
-                    self.address_cache[nic.mac.lower()] = None
-                    arp_ip = None
-
+                    if self.address_cache.get(nic.mac.lower()) == arp_ip:
```
nasastry: Instead of calling `nic.mac.lower()` each every time can this be put in to a variable and then use?
`mac_lower = nic.mac.lower()`

xutian:
@nasastry  PR https://github.com/avocado-framework/avocado-vt/pull/936  include fix for this issue,  can you review it, if you are interesting.

Thanks,
Xu

nasastry:
@xutian Sure. Thanks!!

xutian:
get_address function has re-factored, please fetch latest code to have a try when free. now let's close this pull-request.  thanks your contribution.

PR #916, opened 2017-03-06
Deal with guest undefine in aarch

kylazhang:
adding --nvram

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -4,6 +4,7 @@ http://libvirt.org/formatdomain.html
 """
 
 import logging
+import platform
 
 from .. import xml_utils
 from .. import utils_misc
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -593,6 +594,13 @@ def get_device_class(type_name):
 
     def undefine(self, options=None, virsh_instance=base.virsh):
         """Undefine this VM with libvirt retaining XML in instance"""
+        # add --nvram in undefine for aarch64
```
balamuruhans: just curious to know whether undefining guest with nvram in aarch64 always requires this flag, else we can have check for libvirt version which requires this flag to work.

kylazhang: Because the arm guest boot up must have 
    <loader readonly='yes' type='pflash'>/usr/share/AAVMF/AAVMF_CODE.fd</loader>
    <nvram>/var/lib/libvirt/qemu/nvram/virt-tests-vm3_VARS.fd</nvram>
so undefine will need --nvram, it do not related to libvirt version, but the configure

balamuruhans: Thanks for the clarification

kylazhang:
@balamuruhans Thanks and added :)

balamuruhans:
@kylazhang, Thanks for the Patch

PR #921, opened 2017-03-09
get_ip_address_by_interface gives exceptions if a network interface is up without ip address configured

sudeeshjohn:
The net_utils function get_ip_address_by_interface gives exceptions if a network interface is up without ip address configured
Signed-off-by: Sudeesh John <sudeesh@linux.vnet.ibm.com>

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3054,7 +3054,8 @@ def get_ip_address_by_interface(ifname, ip_ver="ipv4"):
         ver = netifaces.AF_INET
     try:
         addr = netifaces.ifaddresses(ifname).get(ver)
-        return [a['addr'] for a in addr if not a['addr'].startswith('fe80')][0]
+        if addr is not None:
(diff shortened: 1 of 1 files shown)


virttest/utils_net.py
```diff
@@ -3054,7 +3054,8 @@ def get_ip_address_by_interface(ifname, ip_ver="ipv4"):
         ver = netifaces.AF_INET
     try:
         addr = netifaces.ifaddresses(ifname).get(ver)
-        return [a['addr'] for a in addr if not a['addr'].startswith('fe80')][0]
+        if addr is not None:
+            return [a['addr'] for a in addr if not a['addr'].startswith('fe80')][0]
```
sathnaga: This method is to get the ip address and error explain the same.
with this fix, this function won't return anything if it does not find IP

sudeeshjohn:
https://github.com/avocado-framework/avocado-vt/pull/922

PR #922, opened 2017-03-09
net_utils function get_ip_address_by_interface gives exceptions if a network interface is up without ip address configured.

sudeeshjohn:
net_utils function get_ip_address_by_interface gives exceptions a network interface is up without ip address configured.
Signed-off-by: Sudeesh John <sudeesh@linux.vnet.ibm.com>

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3054,7 +3054,10 @@ def get_ip_address_by_interface(ifname, ip_ver="ipv4"):
         ver = netifaces.AF_INET
     try:
         addr = netifaces.ifaddresses(ifname).get(ver)
-        return [a['addr'] for a in addr if not a['addr'].startswith('fe80')][0]
+        if addr is not None:
+            return [a['addr'] for a in addr if not a['addr'].startswith('fe80')][0]
+        else:
+            return None
     except Exception, e:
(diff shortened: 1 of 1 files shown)


virttest/utils_net.py
```diff
@@ -3054,7 +3054,10 @@ def get_ip_address_by_interface(ifname, ip_ver="ipv4"):
         ver = netifaces.AF_INET
     try:
         addr = netifaces.ifaddresses(ifname).get(ver)
-        return [a['addr'] for a in addr if not a['addr'].startswith('fe80')][0]
+        if addr is not None:
+            return [a['addr'] for a in addr if not a['addr'].startswith('fe80')][0]
+        else:
+            return None
```
sathnaga: It is fine to return a value from library method by default in all cases(positive and no ip), in that case the below test error has to revisited, as we would get to the exception only when user gives wrong ifname and anyways that becomes obvious in normal scenario.

How about "return None" instead of TestError so that we cover both the cases?

sudeeshjohn:
Opened a new PR https://github.com/avocado-framework/avocado-vt/pull/1075

virttest/utils_net.py
sudeeshjohn: Yeah, we can do that

PR #928, opened 2017-03-14
Store host_ip as vm parameter

rdkdd:
Make host_ip variable accessible after vm creation if listening_addr specified.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -999,7 +999,7 @@ class VM(virt_vm.BaseVM):
             ip_ver = optget("listening_addr")
             if ip_ver:
                 host_ip = utils_net.get_host_ip_address(self.params, ip_ver)
-                spice_opts.append("addr=%s" % host_ip)
+                self.spice_options['spice_addr'] = host_ip
             set_yes_no_value(
                 "disable_copy_paste", yes_value="disable-copy-paste")
             set_value("addr=%s", "spice_addr")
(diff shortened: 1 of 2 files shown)


virttest/utils_net.py
```diff
@@ -3448,7 +3448,7 @@ def check_listening_port_by_service(service, port, listen_addr='0.0.0.0',
     """
     Check TCP/IP listening by service
     """
-    cmd = "netstat -tunlp | grep -E '^tcp.*LISTEN.*%s.*'" % service
+    cmd = "netstat -tunlpW | grep -E '^tcp.*LISTEN.*%s.*'" % service
```
rdkdd: Added W (--wide) to netstat in case of long ipv6 address, which was trimmed before.

rdkdd:
Can you take a look @xutian @will-Do and possibly merge?
Thanks.

virttest/qemu_vm.py
```diff
@@ -999,7 +999,7 @@ def set_value(opt_string, key, fallback=None):
             ip_ver = optget("listening_addr")
             if ip_ver:
                 host_ip = utils_net.get_host_ip_address(self.params, ip_ver)
-                spice_opts.append("addr=%s" % host_ip)
```
xutian: after drop this line, spice will listen on '0.0.0.0' , right?

rdkdd: @xutian no this line is redundant now because `spice_addr` in `spice_options` is defined. So spice address will be added by  #L1005 (set_value("addr=%s", "spice_addr")). Otherwise `addr=`  would be placed two times as qemu-kvm -spice parameter

rdkdd: @xutian  please could you review?
thanks

xutian:
looks OK to me.

will-Do:
Fine for me

PR #931, opened 2017-03-15
fix set_block_job_speed function

hjydxy:
In qemu2.6,the para "device" is "drive_image1 (#block856)".But we
expect it as "drive_image1".So seperate the str and take the
first part.This modify is compatible with qemu1.5 or earlier version.

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -968,7 +968,7 @@ class HumanMonitor(Monitor):
         if correct:
             cmd = self.correct(cmd)
         self.verify_supported_cmd(cmd)
-        cmd += " %s %sB" % (device, speed)
+        cmd += " %s %sB" % (device.split(" ")[0], speed)
         return self.cmd(cmd)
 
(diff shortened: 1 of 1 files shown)


virttest/qemu_monitor.py
```diff
@@ -968,7 +968,7 @@ def set_block_job_speed(self, device, speed=0,
         if correct:
             cmd = self.correct(cmd)
         self.verify_supported_cmd(cmd)
-        cmd += " %s %sB" % (device, speed)
+        cmd += " %s %sB" % (device.split(" ")[0], speed)
```
xutian: why need this, it looks works fine in our test.
cc @vivianQizhu to look at it.

vivianQizhu: @xutian Agree, 'device' here does not depends on qemu version, and we can get accurate device name by qemu_vm.get_block(), it works well on both qemu 2.6 and qemu 1.5. Why this modification?

hjydxy: @xutian @vivianQizhu It's seem my misunderstand.Thanks for the guidance...

PR #942, opened 2017-03-21
virttest.utils_package: Add package manager module

waynesun09:
Use LocalPackageMgr which is subclass from software_manager in
avocado to decide the package manager and install or remove
package on local host.

If remote session is given, use the RemotePackageMgr class
to find out the package manager on remote for install or remove
packages.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -385,7 +385,7 @@ def is_port_free(port, address):
     """
     try:
         s = socket.socket()
-        #s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        # s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         if address == "localhost":
             s.bind(("localhost", port))
             free = True
@@ -912,7 +912,7 @@ def get_guest_service_status(session, service, service_former=None):
             output.count("Active: active") > 0):
         status = "active"
     elif (re.search(r"running", output.lower(), re.M) and
-            not re.search(r"not running", output.lower(), re.M)):
+          not re.search(r"not running", output.lower(), re.M)):
         status = "active"
 
     return status
@@ -1611,34 +1611,6 @@ def get_cpu_info(session=None):
     return cpu_info
 
 
-def yum_install(pkg_list, session=None, timeout=300):
-    """
-    Try to install packages on system
-
-    :param pkg_list: list of packages
-    :session: session Object
-    :return: True if all packages installed, False if any error
-    """
-    if not isinstance(pkg_list, list):
-        logging.error("Parameter exceptions.")
-        return False
-    yum_cmd = "rpm -q {0} || yum -y install {0}"
-    for pkg in pkg_list:
-        if session:
-            status = session.cmd_status(yum_cmd.format(pkg),
-                                        timeout=timeout)
-        else:
-            status = process.run(yum_cmd.format(pkg),
-                                 timeout=timeout,
-                                 ignore_status=False,
-                                 shell=True).exit_status
-        if status:
-            logging.error("Failed to install package: %s"
-                          % pkg)
-            return False
-    return True
-
-
 def add_identities_into_ssh_agent():
     """
     Adds RSA or DSA identities to the authentication agent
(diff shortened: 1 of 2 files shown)


virttest/utils_misc.py
```diff
@@ -1609,6 +1610,69 @@ def get_cpu_info(session=None):
     return cpu_info
 
 
+def remote_pkg_install(pkg_list, session, timeout=300):
+    """
+    Try to install packages on remote system with session.
+
+    :param pkg_list: list of packages
+    :session: session Object
+    :timeout: timeout for install with session
```
balamuruhans:
```
:param session:  session Object
:param timeout: timeout for install with session
```

virttest/utils_misc.py
```diff
@@ -1609,6 +1610,69 @@ def get_cpu_info(session=None):
...
+    return True
+
+
+def package_install(pkg_list, session=None, timeout=300):
+    """
+    Try to install packages on system with package manager.
+
+    :param pkg_list: list of packages
+    :session: session Object
+    :timeout: timeout for install with session
```
balamuruhans: same here

virttest/utils_misc.py
```diff
@@ -1609,6 +1610,69 @@ def get_cpu_info(session=None):
...
+    :return: True if all packages installed, False if any error
+    """
+    if not isinstance(pkg_list, list):
+        raise exceptions.TestError("Parameter exception on pkg_list")
+    if session:
+        return remote_pkg_install(pkg_list, session, timeout)
```
balamuruhans: 👍

virttest/utils_misc.py
```diff
@@ -1609,6 +1610,69 @@ def get_cpu_info(session=None):
...
+    """
+    Try to install packages on remote system with session.
+
+    :param pkg_list: list of packages
+    :param session: session Object
+    :param timeout: timeout for install with session
+    :return: True if all packages installed, False if any error
+    """
+    if not session or not isinstance(pkg_list, list):
+        raise exceptions.TestError("Parameters exception on session or "
```
sathnaga: if this function is always going to be wrapped by` package_install()`, then check for  `not isinstance(pkg_list, list):` becomes redundant?

waynesun09: it could be used separately, so the check is enforced. I hope remote install could use the new remote function, while update exist yum_install to package_install in the test cases.

virttest/utils_misc.py
```diff
@@ -1611,30 +1617,144 @@ def get_cpu_info(session=None):
...
 
-def yum_install(pkg_list, session=None, timeout=300):
+class RemotePackage(object):
     """
-    Try to install packages on system
+    The remote package manage class
+    """
+    def __init__(self, session, pkg):
+        """
+        :param session: sessioon object
```
balamuruhans: %s/sessioon/session

virttest/utils_misc.py
```diff
@@ -1611,30 +1617,144 @@ def get_cpu_info(session=None):
...
+        self.session = session
+
+        # Inspect and set package commands
+        for pkg_mgr in PACKAGE_MANAGERS:
+            cmd = 'which ' + pkg_mgr
+            if not self.session.cmd_status(cmd):
+                self.package_manager = pkg_mgr
+                break
+
+            if not self.package_manager:
```
balamuruhans: should this come out of for loop, so that it checks for all the `which pkg_mgr` and then errors out else if 1st iteration failing will raise the exception

virttest/utils_misc.py
```diff
@@ -1611,30 +1617,144 @@ def get_cpu_info(session=None):
...
+        for pkg in pkg_list:
+            cmd = self.remove_cmd + pkg
+            if (("*" in pkg or self.is_installed(pkg)) and
+                    self.session.cmd_status(cmd, timeout)):
+                logging.error("Remove package %s failed", pkg)
+                return False
+        return True
+
+
+def package_install(pkg, session=None, timeout=300):
```
balamuruhans: just a suggestion, both `package_install` and `package_remove` are almost same instead we can have a single function, 
```
def package_manager(pkg, session=None, func=sm_inst.install, timeout=300):
    for pkg in pkg_list:
        if (('*' in pkg or not sm_inst.check_installed(pkg)) and
            not func(pkg)):
```

virttest/utils_misc.py
```diff
@@ -1611,30 +1617,144 @@ def get_cpu_info(session=None):
...
+        """
+        Check the package install status
+
+        :param pkg_name: package name
+        :return: True or False
+        """
+        cmd = self.query_cmd + pkg_name
+        return not self.session.cmd_status(cmd)
+
+    def install(self, pkg_list, timeout=300):
```
balamuruhans: same for `install()` and `remove()`

virttest/utils_package.py
```diff
@@ -0,0 +1,213 @@
...
+class RemotePackageMgr(object):
+
+    """
+    The remote package manage class
+    """
+
+    def __init__(self, session, pkg):
+        """
+        :param session: session object
+        :param pkg: pacakge name or list
```
balamuruhans: %s/pacakge/package

virttest/utils_package.py
```diff
@@ -0,0 +1,213 @@
...
+        :return: if remove succeed return True, else False
+        """
+        self.cmd = self.remove_cmd
+        return self.operate(timeout, True)
+
+
+class LocalPackageMgr(software_manager.SoftwareManager):
+
+    """
+    Local pacakge manage class
```
balamuruhans: %s/pacakge/package

virttest/utils_package.py
```diff
@@ -0,0 +1,213 @@
...
+            else:
+                self.pkg_list = [pkg, ]
+        else:
+            self.pkg_list = pkg
+        super(LocalPackageMgr, self).__init__()
+        self.func = None
+
+    def operate(self, default_status):
+        """
+        Use pacakge manager to operate pacakges
```
balamuruhans: %s/pacakge/package

balamuruhans:
@waynesun09 Thanks for the patch

PR #946, opened 2017-03-22
Add public config file for v2v

chloerh:
These configurations including infomations of source hypervisors
and destinations are needed by almost every v2v cases. Therefore
organize and store them in public config files will be make future
case implementing job easier and cfg files lighter.

### DIFF
--- a/backends/v2v/cfg/convert_destination.cfg
+++ b/backends/v2v/cfg/convert_destination.cfg
@@ -1,0 +1,53 @@
+# Configurations of different conversion targets.
+# Please replace the example params with actual values.
+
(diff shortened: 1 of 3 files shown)


backends/v2v/cfg/convert_destination.cfg
```diff
@@ -0,0 +1,53 @@
+# Configurations of different conversion targets.
+# Please replace the example params with actual values.
+
+variants:
+    - dest_none:
+    - v2v_libvirt:
```
will-Do: Just thinking a more readable variant name here

PR #952, opened 2017-03-24
qemu_vm: fix bug of the reboot function

luckyh:
Commit 9ca0c3ea incorrectly removed some codes that is used for waiting
for guest accomplishing the shutdown operation in the case of using a
net session. This patch will bring them back.

ID: 1434307


sathnaga:
LGTM.

xutian:
looks OK to me.

PR #958, opened 2017-03-28
virttest.utils_misc: ignore unicode encode errors

xutian:
ID: 1436501


xutian:
small fix, merge it to avoid it block our test.

PR #960, opened 2017-03-28
Added option to not specify spice_port

rdkdd:

Some of our testcases does not require to specify `--spice port=foo` option.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -992,7 +992,8 @@ class VM(virt_vm.BaseVM):
                 else:
                     self.spice_options['spice_port'] = self.spice_port
                     spice_opts.append("port=%s" % self.spice_port)
-            else:
+            # spice_port = no: spice_port value is not present on qemu cmdline
+            elif optget("spice_port") != "no":
                 set_value("port=%s", "spice_port")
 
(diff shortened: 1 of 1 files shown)


virttest/qemu_vm.py
```diff
@@ -992,7 +992,7 @@ def set_value(opt_string, key, fallback=None):
                 else:
                     self.spice_options['spice_port'] = self.spice_port
                     spice_opts.append("port=%s" % self.spice_port)
-            else:
+            elif optget("spice_port") != "no":
```
lmr: The `no` here is being used on an attempt to be consistent with the values of other parameters (`no` means disabled), right? If so, sure, looks good.

rdkdd: yes `no` is used because of consistency.

Andrei-Stepanov: For clarity's sake I would ask to add a comment:
"""
spice_port = no, means that 'spice_port' parameter is absent in qemu cmdline.
"""
It takes a time to understand what "no" means.
Having a such comment greatly helps to understand what is "no".

xutian:
Looks OK to me, thanks @rdkdd

PR #961, opened 2017-03-29
utils_misc: Avoid configuring Selinux for Ubuntu

balamuruhans:
Configuring selinux for ubuntu fails the test case, so it is handled to
avoid from API, selinux will remain disabled in ubuntu as it is not supported
or selinux commands will not be available. so selinux can be configured only if is 
available and enabled.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -3824,16 +3824,23 @@ class SELinuxBoolean(object):
 
     def __init__(self, params):
         self.server_ip = params.get("server_ip", None)
+        self.ssh_user = params.get("server_user", "root")
+        self.ssh_cmd = "ssh %s@%s " % (self.ssh_user, self.server_ip)
         if self.server_ip:
             # Setup SSH connection
             from utils_conn import SSHConnection
             self.ssh_obj = SSHConnection(params)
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -3820,16 +3820,18 @@ class SELinuxBoolean(object):
...
         self.server_ip = params.get("server_ip", None)
+        self.ssh_user = params.get("server_user", "root")
+        self.params = params
         if self.server_ip:
             # Setup SSH connection
             from utils_conn import SSHConnection
             self.ssh_obj = SSHConnection(params)
             ssh_timeout = int(params.get("ssh_timeout", 10))
             self.ssh_obj.conn_setup(timeout=ssh_timeout)
+            self.check_if_ubuntu_remote = self.check_if_ubuntu(remote=True)
```
will-Do: Instead hard code a new function for ubuntu in this Class, I think you may need a more general way to check if 'selinux' is supported in the place where calling the config function.

balamuruhans: Okay, I will change it accordingly

balamuruhans:
@will-Do have made changes as per your suggestion

balamuruhans:
@will-Do Thanks for review

PR #962, opened 2017-03-29
vm_xml: Added uri to support new_from_dumpxml from remote machine

balamuruhans:
Migration test cases requires to check the xml attributes after migrating to
remote host which requires this support

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -557,18 +557,21 @@ class VMXML(VMXMLBase):
         self.xml = u"<domain type='%s'></domain>" % hypervisor_type
 
     @staticmethod  # static method (no self) needed b/c calls VMXML.__new__
-    def new_from_dumpxml(vm_name, options="", virsh_instance=base.virsh):
+    def new_from_dumpxml(vm_name, options="", virsh_instance=base.virsh,
+                         uri=None):
         """
         Return new VMXML instance from virsh dumpxml command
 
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -557,18 +557,21 @@ def __init__(self, hypervisor_type='kvm', virsh_instance=base.virsh):
         self.xml = u"<domain type='%s'></domain>" % hypervisor_type
 
     @staticmethod  # static method (no self) needed b/c calls VMXML.__new__
-    def new_from_dumpxml(vm_name, options="", virsh_instance=base.virsh):
+    def new_from_dumpxml(vm_name, options="", virsh_instance=base.virsh,
```
will-Do: Here we have virsh_instance, uri is just part of it, so there's no need to add this param

balamuruhans: Yes, I am using the uri from virsh_instance which is just part of it. But using virsh.dumpxml() directly returns string, but for performing some operations with xml object of remote machine we can use new_from_dumpxml() with uri which in turn uses virsh_instance uri.

will-Do: @balamuruhans 
Sorry, I'm not that clear about your meaning. If you want to operate xml object of remote machine, just use create a remote virsh instance and pass it to this method is not enough?

balamuruhans:
Thanks @will-Do, I will use the way you suggested. Closing this PR

PR #966, opened 2017-03-30
docs: Add test status description section for writing simple tests

will-Do:
And, also fix the list index number re-count problem.

### DIFF
--- a/docs/source/WritingTests/WritingSimpleTests.rst
+++ b/docs/source/WritingTests/WritingSimpleTests.rst
@@ -33,7 +33,7 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
    function ``run``. Let's start with it and just put the keyword
    pass, which is a no op. Our test will be like:
 
-.. code-block:: python
+   .. code-block:: python
 
        def run(test, params, env):
            """
@@ -48,7 +48,7 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
    is, most of the time, ``vm1``. ``env.get_vm`` returns a vm object, which
    we'll store on the variable vm. It'll be like this:
 
-.. code-block:: python
+   .. code-block:: python
 
        def run(test, params, env):
            """
@@ -67,7 +67,7 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
    because sometimes due to a bug the vm process might be dead on the
    water, or the monitors are not responding:
 
-.. code-block:: python
+   .. code-block:: python
 
        def run(test, params, env):
            """
@@ -88,7 +88,7 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
    this method does have a default timeout value. Back to business,
    picking up login timeout from our dict of parameters:
 
-.. code-block:: python
+   .. code-block:: python
 
        def run(test, params, env):
            """
@@ -102,7 +102,7 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
 #. Now we'll call ``wait_for_login()`` and pass the timeout to it,
    storing the resulting session object on a variable named session:
 
-.. code-block:: python
+   .. code-block:: python
 
        def run(test, params, env):
            """
@@ -126,7 +126,7 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
    is as simple as calling ``cmd()`` with 'uptime' as a parameter and
    storing the result in a variable called uptime:
 
-.. code-block:: python
+   .. code-block:: python
 
        def run(test, params, env):
            """
@@ -138,22 +138,17 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
            session = vm.wait_for_login(timeout=timeout)
            uptime = session.cmd('uptime')
 
-.. warning:: Some guests OS's do not respect terminal ``echo`` setting,
-   corrupting the output. There are some workaround described in
-   https://github.com/avocado-framework/avocado-vt/issues/231
+   .. warning:: Some guests OS's do not respect terminal ``echo`` setting,
+      corrupting the output. There are some workaround described in
+      github `issue#231 <https://github.com/avocado-framework/avocado-vt/issues/231>`_.
 
 #. If you want to just print this value so it can be seen on the test
    logs, just log the value of uptime using the logging library. Since
    that is all we want to do, we may close the remote connection, to
    avoid ssh/rss sessions lying around your test machine, with the
-   method ``close()``. Now, note that all failures that might happen
-   here are implicitly handled by the methods called. If a test
-   went from its beginning to its end without unhandled exceptions,
-   autotest assumes the test automatically as PASSed, *no need to mark a
-   test as explicitly passed*. If you have explicit points of failure,
-   for more complex tests, you might want to add some exception raising:
+   method ``close()``.
 
-.. code-block:: python
+   .. code-block:: python
 
        def run(test, params, env):
            """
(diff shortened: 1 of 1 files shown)


will-Do:
@ldoktor 
According to the discussion in https://github.com/avocado-framework/avocado-vt/pull/941, I think we should   add this part and let all the test contributors notice it.

will-Do:
[Write document for tp-libvirt on exception raising standard](https://trello.com/c/Pug8x0Qy/123-write-document-for-tp-libvirt-on-exception-raising-standard)

docs/source/WritingTests/WritingSimpleTests.rst
```diff
@@ -167,6 +162,42 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
...
+           session.close()
+           expected_skip_msg = '0 min'
+           expected_err_msg = '1 day'
+           expected_fail_msg = '10 days'
+           if expected_skip_msg in uptime:
+               test.skip('Skip message')
+           if expected_err_msg in uptime:
+               test.error('Error message')
+           if expected_fail_msg in uptime:
+               test.fail('Fail message')
```
ldoktor: This is actually really good example I never thought about and it probably requires more attention. I never thought people will use `test` to raise exceptions mainly because it would not work when running avocado_vt from autotest (unless I'm wrong).

Therefor I'd recommend keep using `from autotest.client.shared import error`. Anyway you are right we want to move away from Autotest and therefor we'd like to actually start using `test` instead, the question is whether we are ready.

Based on your example I guess you are already using it so we can say Autotest+Avocado-vt is not supported at all and we can encourage people to follow your description. The only issue I see here is that the `test.skip` is about to be discontinued (at the end of 2017) and people should use `self.cancel` instead. Anyway that one is not available in 36lts.

@clebergnu, @lmr, did you consider this usage?

clebergnu: The use of `test` is really something we did not anticipate, but as a public API it should be supported IMO.  And @ldoktor is indeed right about the `skip` status, it will affect this use case in the future.

As another notice, some exceptions have been moved to the test API recently (public and supported): https://github.com/avocado-framework/avocado/pull/1875

clebergnu: @lmr do you want to comment here?

docs/source/WritingTests/WritingSimpleTests.rst
```diff
@@ -167,6 +162,42 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
...
+           timeout = float(params.get("login_timeout", 240))
+           session = vm.wait_for_login(timeout=timeout)
+           uptime = session.cmd('uptime')
+           logging.info("Guest uptime result is: %s", uptime)
+           session.close()
+           expected_skip_msg = '0 min'
+           expected_err_msg = '1 day'
+           expected_fail_msg = '10 days'
+           if expected_skip_msg in uptime:
+               test.skip('Skip message')
```
balamuruhans: yes as @ldoktor mentioned, test.skip() gives error msg suggesting to use test.cancel() from inside test,
```
2017-04-05 02:26:41,069 stacktrace       L0044 ERROR| Traceback (most recent call last):
2017-04-05 02:26:41,070 stacktrace       L0044 ERROR|   File "/usr/local/lib/python2.7/dist-packages/avocado_plugins_vt-48.0-py2.7.egg/avocado_vt/test.py", line 260, in runTest
2017-04-05 02:26:41,070 stacktrace       L0044 ERROR|     raise self.__status
2017-04-05 02:26:41,070 stacktrace       L0044 ERROR| TestSetupSkip: [WARNING: self.skip() will be deprecated. Use 'self.cancel()' or the skip decorators] Test Skip
2017-04-05 02:26:41,070 stacktrace       L0045 ERROR|
2017-04-05 02:26:41,070 test             L0748 ERROR| ERROR 1-migration.qemu.qcow2.virtio_scsi.smp2.virtio_net.Ubuntu.17.04.ppc64le.powerkvm-libvirt.virsh.exception.skip -> TestError: Calling skip() in places other than setUp() is not allowed in avocado, you must fix your test. Original skip exception: [WARNING: self.skip() will be deprecated. Use 'self.cancel()' or the skip decorators] Test Skip
```

will-Do: Thanks, I will update the 'skip' part

docs/source/WritingTests/WritingSimpleTests.rst
```diff
@@ -167,6 +162,45 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
            logging.info("Guest uptime result is: %s", uptime)
            session.close()
 
+#. Note that all failures that might happen here are implicitly handled
+   by the methods called. If a test went from its beginning to its end
+   without unhandled exceptions, avocado assumes the test automatically
+   as **PASS**, *no need to mark a test as explicitly passed*. If you
+   have explicit points of failure, for more complex tests, you might
+   want to mark it explicitly. ``test.cancel`` makes the test **CANCEL**
+   (for avocado 36lts version, as this method not backport yet, you may
```
ldoktor: It is a new feature, so it will never be backported, so to not give false hopes I'd say `(it is a new feature not supported in avocado 36lts ...`.

docs/source/WritingTests/WritingSimpleTests.rst
```diff
@@ -167,6 +162,45 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
...
+   want to use ``test.skip`` to make the test **SKIP** to achieve the
+   similar purpose), ``test.error`` makes the test **Error**, and
+   ``test.fail`` makes the test **Fail**. Note, people should never
+   import ``exceptions`` from ``avocado.core`` to raise them in test
```
ldoktor: But in recent Avocado versions some of the exceptions are part of public API available directly in `avocado` and these are fine to be used. (and soon they will be part of the new LTS version)

docs/source/WritingTests/WritingSimpleTests.rst
will-Do: okay, will update it. Thanks

docs/source/WritingTests/WritingSimpleTests.rst
will-Do: Yes, it is, will update it

docs/source/WritingTests/WritingSimpleTests.rst
```diff
@@ -167,6 +162,48 @@ to pick up a living guest, connect to it via ssh, and return its uptime.
...
+   without unhandled exceptions, avocado assumes the test automatically
+   as **PASS**, *no need to mark a test as explicitly passed*. If you
+   have explicit points of failure, for more complex tests, you might
+   want to mark it explicitly. ``test.cancel`` makes the test **CANCEL**
+   (it is a new feature which not supported in avocado 36lts, you may
+   want to use ``test.skip`` to make the test **SKIP** to achieve the
+   similar purpose), ``test.error`` makes the test **Error**, and
+   ``test.fail`` makes the test **Fail**. And, in recent Avocado version
+   (since commit 7ecf09fa), there're three exceptions: ``TestFail``,
+   ``TestError``, and ``TestCancel`` are added to avocado namespace,
```
ldoktor: there'`re` ... `are` added. => you have two `are`s in one sentence. How about

    ...since commit 7ecf09fa) three exceptions: ... were added to ...

ldoktor:
@will-Do there is one minor English issue but apart from that it looks good.

balamuruhans:
Thanks @will-Do @ldoktor, let us merge this.

PR #967, opened 2017-03-30
qemu_vm: add spice plaintext channel option

rdkdd:

We need to test whether SPICE plaintext channels are properly created.

Can you check @will-Do @xutian @Andrei-Stepanov ?

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1060,6 +1060,11 @@ class VM(virt_vm.BaseVM):
                     for item in tmp.split(","):
                         spice_opts.append("tls-channel=%s" % (item.strip()))
 
+                tmp = optget("spice_plaintext_channels")
+                if tmp:
+                    for item in tmp.split(","):
+                        spice_opts.append("plaintext-channel=%s" % (item.strip()))
+
             # Less common options
             set_value("seamless-migration=%s", "spice_seamless_migration")
             set_value("image-compression=%s", "spice_image_compression")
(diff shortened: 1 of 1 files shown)


virttest/qemu_vm.py
```diff
@@ -2043,8 +2048,8 @@ def sort_key(dev):
                     "spice_x509_key_file", "spice_x509_cacert_file",
                     "spice_x509_key_password", "spice_x509_secure",
                     "spice_x509_cacert_subj", "spice_x509_server_subj",
-                    "spice_secure_channels", "spice_image_compression",
-                    "spice_jpeg_wan_compression",
+                    "spice_secure_channels", "spice_plaintext_channels",
```
Andrei-Stepanov: spice_plaintext_channels is not suitable here.

"spice_keys" list holds all known params for qemu-kvm

There is no such option as "spice_plaintext_channels"

rdkdd: from man page of `qemu-kvm`:

> 
           tls-channel=[main|display|cursor|inputs|record|playback]
           plaintext-channel=[main|display|cursor|inputs|record|playback]
               Force specific channel to be used with or without TLS
               encryption.  The options can be specified multiple times to
               configure multiple channels.  The special name "default" can be
               used to set the default mode.  For channels which are not
               explicitly forced into one mode the spice client is allowed to
               pick tls/plaintext as he pleases.

Andrei-Stepanov: Okay, seems fine for me.
@rdkdd please update tp-spice/spice/lib/stest.py 

```
KVM_SPICE_KNOWN_PARAMS = [
    "spice_port",
    "spice_password",
...
```

When this pull req is merged.

ACK.

rdkdd: ok @Andrei-Stepanov done. Could you approve this please?

Andrei-Stepanov: @rdkdd I would be happy to merge. But it is necessary one more ACK from maintainers. The rule of two ack.

PR #986, opened 2017-04-17
virt_vm: Using a more accurate prompt

will-Do:
The original prompt is a bit of inaccurate, when running a command with
'#' or '$' in the output(such as install a package), it will lead to
get_cmd_status fail. So use a more accurate prompt to fix the problem.


xutian:
Also need to update it in 'serial_login' and  'remote_login' function.

will-Do:
@xutian 
Thanks, just update ```serial_login```, and keep the 'prompt' value as original in ```commander``` as I didn't touch it.

xutian:
Looks OK to me.

PR #987, opened 2017-04-17
virttest.utils_test: update regex for getting ping loss ratio

xutian:
output of ping command in windows is different with it output
in linux(output in win10 like:'.. Lost = 8 (80% loss)',
output in linux like:' 100% packet loss, time 8999ms'),so update
regex to made it works on both windows and linux.

ID: 1397690

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -1222,14 +1222,9 @@ def get_loss_ratio(output):
     :param output: Ping output.
     """
     try:
-        return int(re.findall('(\d+)% packet loss', output)[0])
+        return int(re.findall('(\d+)%.*loss', output)[0])
     except IndexError:
-        pass
-    try:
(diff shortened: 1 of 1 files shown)


CongLi:

virttest/utils_test/__init__.py
```diff
@@ -1222,7 +1222,7 @@ def get_loss_ratio(output):
     :param output: Ping output.
     """
     try:
-        return int(re.findall('(\d+)% packet loss', output)[0])
+        return int(re.findall('(\d+)%.*loss', output)[0])
     except IndexError:
         pass
     try:
```
pingl2016: I think The second try can be dropped.

pingl2016:
Acked-by: Ping Li \<pingl@redhat.com\>

xutian:
thanks @CongLi and @pingl2016

PR #989, opened 2017-04-18
qemu_vm: Remove thread check from function migrate

luckyh:
It is unreasonable to check whether a new thread is created for
doing migration in function `migrate` because the behavior is
implementation-dependent. Suggest to make this as an unittest for the
corresponding version of QEMU if do have the need.

ID: 1321505


CongLi:

CongLi:
It's better to remove the variant from tp-qemu too.

XueqiangWei:
LGTM,
Acked-by Xueqaing Wei <xuwei@redhat.com>

luckyh:
Thank @CongLi and @XueqiangWei

PR #992, opened 2017-04-19
Clearing Existing Targetcli Configuration before iscsci setup

prudmirya:

### DIFF
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -830,6 +830,9 @@ class Iscsi(object):
             logging.error(err_msg, "iscsi-initiator-utils")
         try:
             path.find_command("targetcli")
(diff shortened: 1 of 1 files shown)


virttest/iscsi.py
```diff
@@ -830,6 +830,9 @@ def create_iSCSI(params, root_dir=data_dir.get_tmp_dir()):
             logging.error(err_msg, "iscsi-initiator-utils")
         try:
             path.find_command("targetcli")
+            cmd = "targetcli clearconfig confirm=true"
+            if process.system(cmd, shell=True) != 0:
+                logging.error(err_msg, "targetcli configuration unable to clear")
```
will-Do: This is not appropriate place to do cleanup

PR #993, opened 2017-04-20
virt_vm:fix issue of cannot connect to guests after v2v conversion

chloerh:
Raise exception that next procedure could catch and deal with
other than throw it out if get_mac_address returns None for mac
address.

set_mac_address method should be called with an object of
virttest.utils_net.VirtNet

update param list of function get_address() of ovirt.VMManager

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -638,9 +638,11 @@ class BaseVM(object):
         """
         try:
             mac = self.virtnet[nic_index].mac
-            return mac
         except KeyError:
             raise VMMACAddressMissingError(nic_index)
+        if mac is None:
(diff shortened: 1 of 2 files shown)


virttest/virt_vm.py
```diff
@@ -666,6 +666,8 @@ def _get_mac_addr(nic_index):
             mac = None
             try:
                 mac = self.get_mac_address(nic_index)
+                if not mac:
+                    raise VMMACAddressMissingError(nic_index)
```
will-Do: 'get_mac_address' will not raise this exception and return None? then the logic should be fixed, but not using try.except to control the workflow

PR #994, opened 2017-04-20
fix add SGA device failed issue

xutian:
SGA (Serial Graphics Adapter) support in qemu as device, so
replace qcontainor.has_option with qcontainor.has_device to
avoid add 'sga' device failed.

ID: 1443432


xutian:
thanks @luckyh , and merged the small fix.

PR #996, opened 2017-04-24
virttest.virt_vm: add wrapper function for get_address

xutian:
Param 'flexible_nic_index' should also works to get IPv6 address of nic

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -644,6 +644,28 @@ class BaseVM(object):
 
     def get_address(self, index=0, ip_version="ipv4"):
         """
+        Wrapper for self._get_address. if 'flexible_nic_index' is 'yes',
+        will traverses from first nic to first element toward the end
+        util get a reachable IP address;
+        """
+        flexible_nic_index = ("yes" == self.params.get("flexible_nic_index", "no"))
+        macs, nics_index = self.virtnet.mac_list(), [index]
+        flexible_nic_index &= bool(len(macs) > 1)
+        if flexible_nic_index:
+            nics_index += [i for i in xrange(len(macs)) if i != index]
+        for index in nics_index:
+            try:
+                return self._get_address(index, ip_version)
+            except VMAddressError:
+                if not flexible_nic_index:
+                    raise
+                else:
+                    continue
+        else:
+            raise VMAddressError("VM '%s' no reachable %s address found" % (self.name, ip_version))
+
+    def _get_address(self, index=0, ip_version="ipv4"):
+        """
         Return the IP address of a NIC or guest (in host space).
 
         :param index: Name or index of the NIC whose address is requested.
@@ -719,24 +741,6 @@ class BaseVM(object):
                         os.system("arp -d %s" % ip_addr)
                     logging.debug("Clean up outdated address info, "
                                   "MAC: %s, IP: %s" % (mac, ip_addr))
-                    raise
-            except VMIPAddressMissingError:
-                flexible_index = ("yes" == self.params.get("flexible_nic_index", "no"))
-                macs = self.virtnet.mac_list()
-                if not flexible_index or len(macs) < 2:
-                    raise
-                indexs = range(len(macs))
-                indexs.pop(index)
-                for index in indexs:
-                    try:
-                        mac = _get_mac_addr(index)
-                        ip_addr = self.get_address(index)
-                        break
-                    except (VMAddressVerificationError,
-                            VMIPAddressMissingError,
-                            VMMACAddressMissingError):
-                        continue
-                else:
                     raise
             if not ip_addr:
                 raise VMIPAddressMissingError(mac, ip_version)
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -815,17 +815,15 @@ class BaseVM(object):
         self.virtnet.free_mac_address(nic_index_or_name)
 
     @error_context.context_aware
-    def wait_for_get_address(self, nic_index_or_name, timeout=30,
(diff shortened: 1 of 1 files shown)


virttest/virt_vm.py
```diff
@@ -644,6 +644,28 @@ def get_mac_address(self, nic_index=0):
...
     def get_address(self, index=0, ip_version="ipv4"):
         """
+        Wrapper for self._get_address. if 'flexible_nic_index' is 'yes',
+        will traverses from first nic to first element toward the end
+        util get a reachable IP address;
+        """
+        flexible_nic_index = ("yes" == self.params.get("flexible_nic_index", "no"))
+        macs, nics_index = self.virtnet.mac_list(), [index]
+        if flexible_nic_index:
+            nics_index += [i for i in xrange(len(macs)) if i != index]
```
luckyh: Looks like you need a `set()`.

virttest/virt_vm.py
```diff
@@ -644,6 +644,28 @@ def get_mac_address(self, nic_index=0):
 
     def get_address(self, index=0, ip_version="ipv4"):
         """
+        Wrapper for self._get_address. if 'flexible_nic_index' is 'yes',
+        will traverses from first nic to first element toward the end
+        util get a reachable IP address;
+        """
```
yanan-fu: params and  the type ?

virttest/virt_vm.py
```diff
@@ -644,6 +644,28 @@ def get_mac_address(self, nic_index=0):
...
+        flexible_nic_index = ("yes" == self.params.get("flexible_nic_index", "no"))
+        macs, nics_index = self.virtnet.mac_list(), [index]
+        if flexible_nic_index:
+            nics_index += [i for i in xrange(len(macs)) if i != index]
+        flexible_nic_index &= bool(len(macs) > 1)
```
luckyh: Better to combine this with line 651 together to reduce loop overhead.

virttest/virt_vm.py
```diff
@@ -644,6 +644,28 @@ def get_mac_address(self, nic_index=0):
...
+        for index in nics_index:
+            try:
+                return self._get_address(index, ip_version)
+            except VMIPAddressMissingError:
```
luckyh: What about `VMAddressVerificationError` and `VMMACAddressMissingError` two type of errors .

virttest/virt_vm.py
```diff
@@ -644,6 +644,28 @@ def get_mac_address(self, nic_index=0):
...
+                if not flexible_nic_index:
+                    raise
+                else:
+                    continue
+        else:
+            raise
```
luckyh: Can we raise an `IndexError` or some else?

virttest/virt_vm.py
xutian: No,  index should be try first, so use list here.

virttest/virt_vm.py
xutian: agree, with fix in next version

virttest/virt_vm.py
xutian: It's better to raise real exception here.

virttest/virt_vm.py
xutian: agree, will fix in next version.

xutian:
@luckyh  updated, would you like to help review it, thanks.

virttest/virt_vm.py
luckyh: Alright.

virttest/virt_vm.py
luckyh: But the "real" exception is a little misleading. If we get here, it means no reachable IP addrs have been found for all the NICs, not just for the last one.

virttest/virt_vm.py
```diff
@@ -644,6 +644,28 @@ def get_mac_address(self, nic_index=0):
...
+        """
+        flexible_nic_index = ("yes" == self.params.get("flexible_nic_index", "no"))
+        macs, nics_index = self.virtnet.mac_list(), [index]
+        flexible_nic_index &= bool(len(macs) > 1)
+        if flexible_nic_index:
+            nics_index += [i for i in xrange(len(macs)) if i != index]
+        for index in nics_index:
+            try:
+                return self._get_address(index, ip_version)
+            except (VMIPAddressMissingError, VMMACAddressMissingError):
```
luckyh: I think, when `flexible_nic_index` is `True`, `VMAddressVerificationError` could be ignored like it used to be, until it found a valid address.

virttest/virt_vm.py
xutian: Sounds reasonable, let's raise VMAddressError("VM '%s' no reachable %s address found" % (self.name, ip_version) ).
what's your opinion?

luckyh: Looks good.

virttest/virt_vm.py
xutian: right, it should catch VMAddressError. fixed in new version. thanks

virttest/virt_vm.py
xutian: right, here should ignore VMAddressError, has updated in new version.

xutian:
@luckyh  any other comments for this version?

CongLi:
ID: 1442685

CongLi:

@xutian 
Could you please check @yanan-fu comment to add doc string for param in get_address() ?

Thanks.

xutian:
@CongLi and @yanan-fu  as doc string said it's wrapper for self._get_address, I think it's needless to add duplicate docs in get_address function.

xutian:
thanks @CongLi and @luckyh

PR #999, opened 2017-04-27
virttest.virt_vm: fix exception not catch issue

xutian:
ID: 1445756


CongLi:

PR #1001, opened 2017-04-28
utils_net:it should be 'IPAddrGetError' in get_ip_address_by_interface

chloerh:
'IPAddrGetError' could be caught afterwards by other functions while
'TestError' would be thrown out.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3095,8 +3095,8 @@ def get_ip_address_by_interface(ifname, ip_ver="ipv4"):
         addr = netifaces.ifaddresses(ifname).get(ver)
         return [a['addr'] for a in addr if not a['addr'].startswith('fe80')][0]
     except Exception, e:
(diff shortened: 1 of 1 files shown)


virttest/utils_net.py
```diff
@@ -3095,8 +3095,8 @@ def get_ip_address_by_interface(ifname, ip_ver="ipv4"):
         addr = netifaces.ifaddresses(ifname).get(ver)
         return [a['addr'] for a in addr if not a['addr'].startswith('fe80')][0]
     except Exception, e:
-        raise exceptions.TestError("Cannot get %s address form %s interface: %s"
-                                   % (ip_ver, ifname, e))
+        raise NetError("Cannot get %s address form %s interface: %s"
```
xutian: it looks raise IPAddrGetError looks more suitable at here.

will-Do:
Thanks @xutian

PR #1010, opened 2017-05-12
libvirt_vm: Fix create console bug

will-Do:
When close console, the object is still exist, but not function anymore.


xutian:
looks OK to me.

will-Do:
@xutian Thanks, merging this fix now

PR #1013, opened 2017-05-15
libvirt.py: empty string param might break the API

balamuruhans:
handled setting default value if empty string is used


will-Do:
Thanks, @balamuruhans

PR #1023, opened 2017-06-01
Enable readonly attribute when attaching rbd disk with snapshot

chunfuwen:
since readonly attribute need be set explicitly


chunfuwen:
@will-Do 
@balamuruhans 

Please help take a look.

PR #1025, opened 2017-06-05
utils_net: enable get_network_cfg_file() and create_network_script() for host

balamuruhans:
Enable APIs to provide distro specific network config file and create network
script in host and guest

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -12,6 +12,7 @@ import commands
 import signal
 import netifaces
 import netaddr
+import platform
 
 import aexpect
 from avocado.core import exceptions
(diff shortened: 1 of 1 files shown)


virttest/utils_net.py
```diff
@@ -1609,21 +1613,24 @@ def get_vm_network_cfg_file(vm, iface_name):
     return iface_cfg_file
 
 
-def create_vm_network_script(vm, iface_name, mac_addr, boot_proto, net_mask,
-                             ip_addr=None, **dargs):
+def create_network_script(iface_name, mac_addr, boot_proto, net_mask,
+                          vm=None, ip_addr=None, **dargs):
     """
-    Form network script with its respective network param for vm instance
+    Form network script with its respective network param for vm or Host.
```
xutian: Better to doc what should include in  dargs dict.

xutian: View from function name,  this function use to generate configuration file to network service, but  checked it implementation, it looks not and it just return  param list. And the function is not complete.

balamuruhans: @xutian, sure I will work it out.

balamuruhans:
@xutian, I have made the changes suggested by you.

balamuruhans:
Thanks @xutian , had made API to generate configuration file instead of returning the params list.
Merging now.

PR #1026, opened 2017-06-06
Fixup: pm-utils install failure for guest agent tests

sathnaga:
pm-utils is not common in all distributions in
default repo and fails to install and breaks the flow
of guest agent test cases, even if we do not require
functionalities given by pm-utils.

So, added ignore_status param for package installation
method and and updated the pm-utils install in guest agent
setup accordingly.


balamuruhans:
Thanks @sathnaga

PR #1034, opened 2017-06-14
Add secret operation in libvirt.py

lento-sun:
With this change, we can define a secret with create_secret()
and set base64 encoded password to it with set_secret_pwd()

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -24,8 +24,8 @@ import shutil
 import threading
 import time
 import sys
-
 import aexpect
+
 from avocado.core import exceptions
 from avocado.utils import path as utils_path
 from avocado.utils import process
@@ -55,6 +55,7 @@ from ..libvirt_xml import IPXML
 from ..libvirt_xml import pool_xml
 from ..libvirt_xml import nwfilter_xml
 from ..libvirt_xml import vol_xml
+from ..libvirt_xml import secret_xml
 from ..libvirt_xml.devices import disk
 from ..libvirt_xml.devices import hostdev
 from ..libvirt_xml.devices import controller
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -3095,3 +3097,70 @@ def virsh_cmd_has_option(cmd, option, raise_skip=True):
...
+    try:
+        sec_uuid = re.findall(r".+\S+(\ +\S+)\ +.+\S+",
+                              ret.stdout)[0].lstrip()
+    except IndexError:
+        raise exceptions.TestError("Fail to get newly created secret uuid")
+
+    return sec_uuid
+
+
+def set_secret_pwd(sec_uuid, password):
```
balamuruhans: we can have the encoding part in `virsh.secret_set_value`
```
def secret_set_value(uuid, password, options=None, **dargs):
    if dargs.get("encode", False):
        password = base64.b64encode(password)
```

lento-sun: Thx for reviewing, but IMHO, the virsh command methods should be kept clean to do exactly what they can do. And "darags" should only contain valid virsh API keywords, such as debug/uri/session_id, etc, which could be found in virsh.command(). But definitely put the encoding in virsh.secret_set_value is much more convenient.
What do you think about it? @will-Do

balamuruhans: Agree, fine then let us have the param out of **dargs,
`secret_set_value(uuid, password, encode=False, options=None, **dargs)`

lento-sun: another PR submitted for this, thx for the advice
https://github.com/avocado-framework/avocado-vt/pull/1036

lento-sun:
@balamuruhans pls have a review, I've removed the  "set_secret_pwd" method and did a modification in  https://github.com/avocado-framework/avocado-vt/pull/1036

balamuruhans:
Thanks @lento-sun

PR #1036, opened 2017-06-15
Add 'encode' parameter for secret_set_value

lento-sun:
It's a easier way to just provide a plaintext password and
do the encoding inside this method. So when encode=True,
a base64 encoding will take effect.

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -28,8 +28,9 @@ import re
 import weakref
 import time
 import select
-
+import base64
 import aexpect
+
 from avocado.utils import path
 from avocado.utils import process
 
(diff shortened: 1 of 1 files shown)


virttest/virsh.py
```diff
@@ -3606,19 +3607,25 @@ def secret_get_value(uuid, options=None, **dargs):
...
-def secret_set_value(uuid, base64, options=None, **dargs):
+def secret_set_value(uuid, password, encode=False, options=None, **dargs):
     """
     Set a secret value
 
     :param uuid: secret UUID
-    :param base64: base64-encoded secret value
+    :param password: secret value
+    :param encode: if FALSE, that means you've already provided a base64-encoded
+                   password. if TRUE, will base64-encode password before use.
```
balamuruhans: Boolean value is case sensitive, so let's have doc string accordingly, False/True instead of FALSE/TRUE

balamuruhans:
Thanks @lento-sun

PR #1041, opened 2017-06-22
Added support to enable libvirtd log for tests

sathnaga:
Added support for enabling libvirtd debug log
per test based on the user params.

Usage:
Set "enable_libvirtd_debug_log = True" param in test config inorder
to capture libvirtd log into a file, bydefault it sets the log level to '1'
which is debug, same can be controlled by param, "libvirtd_debug_level" and
by default it stores as libvirtd.log into specific test debug directory
and the same can be controlled by param, "libvirtd_debug_file"

E.g:-
libvirtd_debug_level = 3
libvirtd_debug_file = "/tmp/libvirtd.log"

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -25,6 +25,8 @@ from . import utils_misc
 from . import versionable_class
 from . import openvswitch
 from . import remote
+from . import utils_libvirtd
+from . import utils_config
 from .staging import service
 from .staging import utils_memory
 
@@ -2079,3 +2081,45 @@ def disable_smt(params=None):
                 logging.debug("SMT turned off successfully")
             except process.CmdError, info:
                 raise exceptions.TestSetupFail("VM can not be started :%s" % info)
+
+
+class LibvirtdDebugLog(object):
+    """
+    Enable libvirtd log for testcase incase
+    with the use of param "enable_libvirtd_debug_log",
+    with additional params log level("libvirtd_debug_level")
+    and log file path("libvirtd_debug_file") can be controlled.
+    """
+    def __init__(self, test, log_level=1, log_file=""):
+        """
+        initialize variables
+
+        :param test: Test object
+        :param log_level: debug level for libvirtd log
+        :param log_file: debug file path
+        """
+        self.log_level = log_level
+        self.log_file = log_file
+        self.test = test
+        self.libvirtd = utils_libvirtd.Libvirtd()
+        self.libvirtd_conf = utils_config.LibvirtdConfig()
+        self.backupfile = "%s.bkup.debuglog" % self.libvirtd_conf.conf_path
+
+    def enable(self):
+        """ Enable libvirtd debug log """
+        if not self.log_file or not os.path.isdir(os.path.dirname(self.log_file)):
+            self.log_file = utils_misc.get_path(self.test.debugdir,
+                                                "libvirtd.log")
+        logging.debug("libvirtd debug log stored in: %s", self.log_file)
+        self.libvirtd_conf["log_level"] = self.log_level
+        self.libvirtd_conf["log_outputs"] = '"%s:file:%s"' % (self.log_level,
+                                                              self.log_file)
+        self.libvirtd.restart()
+        fd = open(self.backupfile, "w")
+        fd.write(self.libvirtd_conf.backup_content)
+        fd.close()
+
+    def disable(self):
+        """ Disable libvirtd debug log """
+        os.rename(self.backupfile, self.libvirtd_conf.conf_path)
+        self.libvirtd.restart()
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -600,6 +600,14 @@ def preprocess(test, params, env):
 
     vm_type = params.get('vm_type')
 
+    if vm_type == 'libvirt' and params.get("enable_libvirtd_debug_log", False):
+        log_level = params.get("libvirtd_debug_level", 1)
+        log_file = params.get("libvirtd_debug_file", "")
+        libvirtd_debug_log = test_setup.LibvirtdDebugLog(test,
+                                                         log_level,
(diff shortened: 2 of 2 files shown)


virttest/test_setup.py
```diff
@@ -2079,3 +2081,46 @@ def disable_smt(params=None):
...
+        self.backupfile = "%s.bkup.debuglog" % self.libvirtd_conf.conf_path
+
+    def enable(self):
+        """ Enable libvirtd debug log """
+        if not self.log_file or not os.path.isfile(self.log_file):
+            self.log_file = utils_misc.get_path(self.test.debugdir,
+                                                "libvirtd.log")
+        logging.debug("libvirtd debug log stored in: %s", self.log_file)
+        self.libvirtd_conf["log_level"] = self.log_level
+        self.libvirtd_conf["log_outputs"] = '"1:file:%s"' % self.log_file
```
balamuruhans:
```
self.libvirtd_conf["log_outputs"] = '"%s:file:%s"' % (self.log_level,
                                                      self.log_file)
```

virttest/test_setup.py
```diff
@@ -2079,3 +2081,46 @@ def disable_smt(params=None):
...
+    and log file path("libvirtd_debug_file") can be controlled.
+    """
+    def __init__(self, env, test, log_level=1, log_file=""):
+        """
+        initialize variables
+
+        :param env: ENV object
+        :param test: Test object
+        :param level: debug level for libvirtd log
+        :param debug_file: debug file path
```
balamuruhans:
```
:param log_level: debug level for libvirtd log
:param log_file: debug file path
```

virttest/test_setup.py
```diff
@@ -2079,3 +2081,46 @@ def disable_smt(params=None):
...
+        """
+        self.env = env
```
balamuruhans: `env` is not used in this here

virttest/test_setup.py
```diff
@@ -2079,3 +2081,46 @@ def disable_smt(params=None):
...
+        self.log_level = log_level
+        self.log_file = log_file
+        self.test = test
+        self.libvirtd = utils_libvirtd.Libvirtd()
+        self.libvirtd_conf = utils_config.LibvirtdConfig()
+        self.backupfile = "%s.bkup.debuglog" % self.libvirtd_conf.conf_path
+
+    def enable(self):
+        """ Enable libvirtd debug log """
+        if not self.log_file or not os.path.isfile(self.log_file):
```
balamuruhans: if user have given the log_file but file may or may not exist, if file doesn't exist then we can open a new file with user given log_file path?

virttest/test_setup.py
```diff
@@ -2079,3 +2081,46 @@ def disable_smt(params=None):
...
+            self.log_file = utils_misc.get_path(self.test.debugdir,
+                                                "libvirtd.log")
+        logging.debug("libvirtd debug log stored in: %s", self.log_file)
+        self.libvirtd_conf["log_level"] = self.log_level
+        self.libvirtd_conf["log_outputs"] = '"1:file:%s"' % self.log_file
+        self.libvirtd.restart()
+        fd = open(self.backupfile, "w")
+        fd.write(self.libvirtd_conf.backup_content)
```
balamuruhans: it is better to backup first and then to change the `log_level` and `log_outputs`

virttest/test_setup.py
```diff
@@ -2079,3 +2081,46 @@ def disable_smt(params=None):
...
                 raise exceptions.TestSetupFail("VM can not be started :%s" % info)
+
+
+class LibvirtdDebugLog(object):
+    """
+    Enable libvirtd log for testcase incase
+    with the use of param "enable_libvirtd_debug_log",
+    with additional params log level("libvirtd_debug_level")
+    and log file path("libvirtd_debug_file") can be controlled.
+    """
```
balamuruhans: This is really a good one to have it for each testcase if required, Nice one 👍

virttest/test_setup.py
sathnaga: @balamuruhans backup is already taking in the init, we are just using the backedup content here

virttest/test_setup.py
sathnaga: @balamuruhans sure :+1:

virttest/test_setup.py
sathnaga: @balamuruhans yes, this needs a improvement, file as such is not required as libvirtd will create, we need to make sure the parent folder exists and if not create one, will do it :+1:

virttest/test_setup.py
sathnaga: sure, will check and remove

virttest/test_setup.py
sathnaga: :+1:

virttest/test_setup.py
```diff
@@ -2079,3 +2081,46 @@ def disable_smt(params=None):
...
+        :param level: debug level for libvirtd log
+        :param debug_file: debug file path
+        """
+        self.env = env
+        self.log_level = log_level
+        self.log_file = log_file
+        self.test = test
+        self.libvirtd = utils_libvirtd.Libvirtd()
+        self.libvirtd_conf = utils_config.LibvirtdConfig()
+        self.backupfile = "%s.bkup.debuglog" % self.libvirtd_conf.conf_path
```
will-Do: utils_config already provide the ``restore`` method, so no need do it manually

sathnaga: @will-Do yep initially tried using that, problem we call the restore is in different place of the code flow and content gets re-backedup with the changed content, so no use of that method here.

sathnaga:
@balamuruhans @will-Do fixed the review comments, pls merge if fine, Thanks in advance :-)

balamuruhans:
Thanks @sathnaga, let us merge this.

PR #1042, opened 2017-06-27
Clearing the targetcli configuration

prudmirya:
while running regression running test case clear the iscsi devices, but still i found the configuration issues for next test case, so i clear the targetcli configuration. so i added the code.

### DIFF
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -347,6 +347,9 @@ class _IscsiComm(object):
             restart_iscsid()
         if self.export_flag:
             self.delete_target()
+        cmd = "targetcli clearconfig confirm=true"
+        if process.system(cmd, shell=True) != 0:
+            logging.error("targetcli configuration unable to clear")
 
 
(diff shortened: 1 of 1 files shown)


virttest/iscsi.py
```diff
@@ -347,6 +347,9 @@ def cleanup(self):
             restart_iscsid()
         if self.export_flag:
             self.delete_target()
+        cmd = "targetcli clearconfig confirm=true"
+        if process.system(cmd, shell=True) != 0:
+            logging.error("targetcli configuration unable to clear")
```
luckyh: This cleanup operation should not be done by the base class `_IscsiComm`, the right place is somewhere in `IscsiLIO`.

prudmirya:
working on it

virttest/iscsi.py
prudmirya: Hi luckyh,

i want to clear the configuration before device created, IscsiLIO will effect the after device created.

PR #1048, opened 2017-07-03
Fix params sequence of secret_set_value

lento-sun:
Some existing cases will use virsh.secret_set_value() as follow:
 result = virsh.secret_set_value(uuid, secret_string, options, debug=True)
So avocado-vt PR#1036 will cause case failure because an new param 'encode'
inserted before 'options'.

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -3607,7 +3607,7 @@ def secret_get_value(uuid, options=None, **dargs):
     return command(cmd, **dargs)
 
 
-def secret_set_value(uuid, password, encode=False, options=None, **dargs):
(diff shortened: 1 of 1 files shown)


lento-sun:
this is to fix https://github.com/avocado-framework/avocado-vt/pull/1036/files
`-def secret_set_value(uuid, base64, options=None, **dargs):
+def secret_set_value(uuid, password, encode=False, options=None, **dargs):`
which cause a params sequence issue for some exiting cases.
pls have a check
@balamuruhans @will-Do

balamuruhans:
@lento-sun Thanks!

PR #1052, opened 2017-07-05
test_setup: request_devs() function doesn't have error_context decorator

balamuruhans:
request_devs() function doesn't have error_context decorator which causes
traceback in error_context as "IndexError: list assignment index out of range"


balamuruhans:
Thanks! @luckyh

PR #1066, opened 2017-07-17
utils_net: simple ipv4 address generator

balamuruhans:
API to generate ipv4 address and can be used inside tests for
assigning static IP address to guests/hosts

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3080,6 +3080,31 @@ def generate_mac_address_simple():
     return mac
 
 
+def gen_ipv4_addr(network_num="10.0.0.0", network_prefix="24", exclude_ips=[]):
+    """
+    generate ipv4 address
+    :param network_num: network number to be used
+    :param network_prefix: prefix used to get subnet mask to calculate ip range
+    :param exclude_ips: the list of ipaddress should be excluded
+
+    :return: ipaddress of type str
+    """
+    ip_regex = "^\d+.\d+.\d+.\d+$"
+    exclude_ips = set(exclude_ips)
+    if not re.match(ip_regex, network_num):
+        network_num = "10.0.0.0"
+    if not exclude_ips and network_prefix == "24":
+        exclude_ips.add(network_num)
+        exclude_ips.add('.'.join(network_num.split('.')[0:3]) + ".%s" %
+                        str(1))
+        exclude_ips.add(('.'.join(network_num.split('.')[0:3]) + ".%s" %
+                        str(255)))
+    network = netaddr.IPNetwork("%s/%s" % (network_num, network_prefix))
+    for ip_address in network:
+        if str(ip_address) not in exclude_ips:
+            yield str(ip_address)
+
+
 def get_ip_address_by_interface(ifname, ip_ver="ipv4"):
     """
     returns ip address by interface


balamuruhans:
@luckyh done.

luckyh:
@balamuruhans I found that package `netaddr` have provided an implementation for this, and the package has already been imported by `utils_net.py`. So sorry, I think you'd better change the code, and make it as a wrapper.

References:
https://netaddr.readthedocs.io/en/latest/api.html#ip-functions-and-generators
https://netaddr.readthedocs.io/en/latest/api.html#ip-networks-and-subnets

balamuruhans:
@luckyh, Thanks for the suggestion, made the changes.

CongLi:

luckyh:
@balamuruhans why not make it for common use, like:
```python
def gen_ipv4_addr(addr, exclude=[]):
    net = netaddr.IPNetwork(addr)
    for ip in net:
        if str(ip) in exclude:
            continue
        yield (str(ip),
               str(net.netmask),
               str(net.broadcast))

>>> NETADDR = "10.0.0.0/24"
>>> EXCLUDE = ["10.0.0.1", "10.0.0.255"]
>>> g = gen_ipv4_addr(NETADDR, EXCLUDE)
>>> for i, m, b in g:
...     print i, m, b
...
10.0.0.0 255.255.255.0 10.0.0.255
10.0.0.2 255.255.255.0 10.0.0.255
...
```

balamuruhans:
@luckyh, Thanks for the idea, but the API `gen_ipv4_addr()` should generate ip address similar to a API generates mac address in `utils_net` and I feel returning netmask and broadcast is not the right way.

luckyh:
@balamuruhans ok, we can only return `ip`.

The main point is I think user must specify ip range or give a network definition to calculate the range, otherwise, the code have to decide a range by hardcoding, which is not a good idea.

virttest/utils_net.py
```diff
@@ -3080,6 +3080,30 @@ def generate_mac_address_simple():
...
+    """
+    ip_regex = "^\d+.\d+.\d+.\d+$"
+    if not re.match(ip_regex, network_num):
+        network_num = "10.0.0.0"
+    if not exclude_ips:
+        exclude_ips.append(network_num)
+        exclude_ips.append('.'.join(network_num.split('.')[0:3]) + ".%s" %
+                           str(1))
+        exclude_ips.append(('.'.join(network_num.split('.')[0:3]) + ".%s" %
+                           str(255)))
```
luckyh: The above two statements are not always correct for every subnet.

balamuruhans: this is for default value when `network_prefix = "24"`, if user gives different network prefix for his required subnet, then user can also give `exclude_ips` too.

luckyh: If so, I suggest appending the judgement to L3095. And it is better to make `exclude_ips` as a `set()`.

luckyh: If so, I suggest appending the judgement to L3095. And it is better to make `exclude_ips` as a `set()`.

balamuruhans: @luckyh Agreed, I have made the changes as suggested.

balamuruhans:
Thank you @luckyh

PR #1072, opened 2017-07-20
Added method to get near real guest memory size

sathnaga:
MemTotal in /proc/meminfo would display total
usable memory(i.e. physical ram minus a few reserved
bits and the kernel binary code)

So, this method helps to get physical memory as discovered
by system in Kb

Note: Have tried with simpler solution to get
no. of blocks, failed due to the below issue,
so had to choose this little time consuming approach.
E:g:-
```
# time grep 1 */online|wc -l
-bash: /bin/grep: Argument list too long
0

# time for i in `ls`;do [ -f $i/online ] && a=$(<$i/online) && count=$(( $count + $a ));a=0;done;echo $count

real	3m14.631s
user	2m30.930s
sys	0m53.794s
131072

```

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -1387,6 +1387,27 @@ class BaseVM(object):
         cmd = self.params.get("mem_chk_cur_cmd")
         return self.get_memory_size(cmd)
 
+    def get_totalmem_sys(self):
+        """
+        To get the total guest memory(ram) as detected by system
+        MemTotal in /proc/meminfo would display
+        total usable memory(i.e. physical ram minus
+        a few reserved bits and the kernel binary code)
+
+        :return: system memory in Kb as float
+        """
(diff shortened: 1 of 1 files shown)


virttest/virt_vm.py
```diff
@@ -1387,6 +1387,27 @@ def get_current_memory_size(self):
         cmd = self.params.get("mem_chk_cur_cmd")
         return self.get_memory_size(cmd)
 
+    def get_totalmem_sys(self):
+        """
+        To get the total guest memory(ram) as detected by system
+        MemTotal in /proc/meminfo would display
```
luckyh: How about windows system?

sathnaga: @luckyh not much idea

sathnaga:
can this be merged if no further comments, Thanks in advance.

PR #1075, opened 2017-07-24
net_utils function get_ip_address_by_interface raising exception

sudeeshjohn:
net_utils function get_ip_address_by_interface raising exception if a network interface is up without ip address configured.

Signed-off-by: Sudeesh John sudeesh@linux.vnet.ibm.com


balamuruhans:
Thanks @sudeeshjohn, let us merge this.

PR #1077, opened 2017-07-24
Added support for guest dmesg verify

sathnaga:
Modified verify_host_dmesg method to accomodate
guest dmesg verification aswell.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -3391,12 +3391,13 @@ def monotonic_time():
         return monotonic_time_os()
 
 
-def verify_host_dmesg(dmesg_log_file=None, ignore_result=False, level_check=3):
+def verify_dmesg(dmesg_log_file=None, ignore_result=False, level_check=3,
+                 session=None):
     """
-    Find host call trace in dmesg log.
+    Find host/guest call trace in dmesg log.
 
     :param dmesg_log_file: The file used to save host dmesg. If None, will save
-                           host dmesg to logging.debug.
+                           guest/host dmesg to logging.debug.
     :param ignore_result: True or False, whether to fail test case on issues
     :param level_check: level of severity of issues to be checked
                         1 - emerg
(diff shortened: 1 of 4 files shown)


sathnaga:
Working to find a right place to call/use guest dmesg verify function
Approach 1:
Call it in postprocess(), but issue is we can not guarantee vm alive there, but will work if vm shutdown not handled in testcase.

Approach 2:
provide a method to VM class, so that tester can use it flexibly inside the test.

Approach 3:
Provide both Approach 1(with check for vm alive and proceed) and Approach 2 for ease of test,(similar to verify_kernel_crash, can be replacement)

virttest/utils_misc.py
```diff
@@ -3389,33 +3389,45 @@ class struct_timespec(ctypes.Structure):
         return monotonic_time_os()
 
 
-def verify_host_dmesg(dmesg_log_file=None, ignore_result=False, level_check=3):
+def verify_dmesg(dmesg_log_file=None, ignore_result=False, level_check=3,
+                 session=None):
```
lmr: Seems relatively straight forward, so OK. I was going to suggest that instead of a session and handle things differently, that a runner callable was passed, defaulting to `process.run`, plus **runner_kwargs, but that's not that big a deal.

sathnaga: @lmr Sure, any usage example for my reference to check in case of implementation.

lmr: I'm inclined to prefer option number 2.

sathnaga: @lmr  agree, that would require changes in all our tests in case of usage, any suggestion how to better handle that?

sathnaga:
Addressed review comments, let me know further comments to help merge it.

sathnaga:
can we merge if no further comments

sathnaga:
Merging as it got approved long back and pending.

PR #1078, opened 2017-07-24
Added guest dmesg verification during postprocess

sathnaga:
Added guest dmesg verification during postprocess

Config usage:
    verify_guest_dmesg = yes(by default)
    guest_dmesg_level = 3(by default)
    guest_dmesg_ignore = False(by default)

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -901,6 +901,10 @@ def postprocess(test, params, env):
     """
     error_context.context("postprocessing")
     err = ""
+    if params.get("verify_guest_dmesg", "yes") == "yes" and params.get("start_vm", "no") == "yes":
+        guest_dmesg_log_file = params.get("guest_dmesg_logfile", "guest_dmesg.log")
+        guest_dmesg_log_file = utils_misc.get_path(test.debugdir, guest_dmesg_log_file)
+        virt_vm.verify_dmesg(dmesg_log_file=guest_dmesg_log_file)
 
     # Postprocess all VMs and images
     try:


virttest/env_process.py
```diff
@@ -901,6 +901,21 @@ def postprocess(test, params, env):
...
+        try:
+            session = vm.wait_for_login()
+        except:
+            logging.warning("Could not get VM session to verify dmesg")
+        else:
+            guest_dmesg_log_file = params.get("guest_dmesg_logfile", "guest_dmesg.log")
+            level = params.get("guest_dmesg_level", 3)
+            ignore_result = params.get("guest_dmesg_ignore", "no") == "yes"
+            dmesg_log_file = utils_misc.get_path(test.debugdir, dmesg_log_file)
+            utils_misc.verify_dmesg(dmesg_log_file=dmesg_log_file,
```
balamuruhans: can we reuse your API here, `vm.verify_dmesg()`

sathnaga: @balamuruhans both are RFC patchs, not sure if we need both of them yet, moreover it is better to use `utils_misc.verify_dmesg()` directly for arguments flexibility, etc.

balamuruhans: I feel, it will be good to have default params with argument flexibility even for vm.

virttest/env_process.py
```diff
@@ -901,6 +901,21 @@ def postprocess(test, params, env):
     """
     error_context.context("postprocessing")
     err = ""
+    vm = env.get_vm(params.get('main_vm'))
+    if params.get("verify_guest_dmesg", "no") == "yes" and vm.is_alive():
```
CongLi: I think it's better to enable 'verify_guest_dmesg' as default.

sathnaga:
Addressed review comments, let me know further comments to help merge it.

sathnaga:
@CongLi @balamuruhans Can we merge this if no further comments

sathnaga:
Merging as no further comments and dependency PR merge

PR #1085, opened 2017-07-27
Increase memory assigned to guest when installing using URL v2

KarMez-Redhat:
Setting memory parameter to 4096 for RHEL 7.0, 7.1, 7.2, 7.3 and 7.devel
guests when running unattended_install.url test as installing RHEL over
http fails with memory set to 1024.


KarMez-Redhat:
Changes from v1:
Originally had updated multiple cfg files under the 7.x subdirectories  (7.2/ppc64.cfg, 7.2/x86_64.cfg etc). Instead, moved up one directory to avocado-vt/shared/cfg/guest-os/Linux/RHEL and updated the 7.x.cfg files there.

xutian:
Acked-by:  Xu Tian

PR #1087, opened 2017-07-31
Add support to boot guest with pf/vf on x86_64 host.

huangyum:
Add support to boot guest with pf/vf on x86_64 host.

ID: 1468525, 1468528

### DIFF
--- a/shared/cfg/guest-hw.cfg
+++ b/shared/cfg/guest-hw.cfg
@@ -158,6 +158,7 @@ variants:
         pci_assignable = no
     - pf_assignable:
         pci_assignable = pf
+        host_setup_flag = 3
         # device_name is optional. If device_name not set, it will try to get
         # first available pf. If device_name set, but the device is not
         # available, it will also try to get first available pf.
@@ -173,6 +174,7 @@ variants:
     - vf_assignable:
         pci_assignable = vf
         nettype = bridge
+        host_setup_flag = 3
         # Driver (kernel module) that supports SR-IOV hardware.
         # As of today (30-11-2009), we have 2 drivers for this type of hardware:
         # Intel® 82576 Gigabit Ethernet Controller - igb
@@ -193,7 +195,7 @@ variants:
         # the default below is for the igb driver
         # For PowerPC with mlx5_core: driver_option = "10" where 10 is numer of
         # virtual functions (VF). Maximum supported is 63 VF per port
-        driver_option = "max_vfs=7"
+        driver_option = "7"
         mac_ip_filter = "inet (.\d+.\d+.\d+.\d+).*?ether (.\w+:\w+:\w+:\w+:\w+:\w+)"
         RHEL.6, RHEL.5:
             mac_ip_filter = "HWaddr (.\w+:\w+:\w+:\w+:\w+:\w+)\s+?inet addr:(.\d+\.\d+\.\d+\.\d+)"
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -1139,9 +1139,7 @@ class PciAssignable(object):
         :rtype: builtin.list
         """
 
-        cmd = "lspci | awk '/%s/ {print $1}'" % self.pf_filter_re
-        pf_ids = [i for i in process.system_output(
-            cmd, shell=True).splitlines()]
+        pf_ids = self.get_pf_ids()
         pf_vf_dict = []
         for pf_id in pf_ids:
             pf_info = {}
@@ -1162,14 +1160,6 @@ class PciAssignable(object):
                 vf_ids.append(vf_info)
             pf_info["vf_ids"] = vf_ids
             pf_vf_dict.append(pf_info)
-        # When the VFs loaded on a PF are > 10 [I have tested till 63 which is
-        # max VF supported by Mellanox CX4 cards],VFs probe on host takes bit
-        # more time than usual.
-        # Without this sleep ifconfig would fail With below error, since not all
-        # probed interfaces of VF would have got properly renamed:
-        # CmdError: Command 'ifconfig -a' failed (rc=1)
-        if int(self.driver_option) > 10:
-            time.sleep(60)
         if_out = process.system_output("ifconfig -a")
         ethnames = re.findall(self.nic_name_re, if_out)
         for eth in ethnames:
@@ -1313,6 +1303,21 @@ class PciAssignable(object):
             cmd, shell=True, timeout=60).splitlines()
         return (PF_devices)
 
+    def get_pf_ids(self):
+        """
+        Get the PF devices on x86 host
+        """
+        cmd = "lspci | grep -v 'Virtual Function' |awk '/%s/ {print $1}'" % self.pf_filter_re
+        PF_devices = [i for i in process.system_output(
+                cmd, shell=True).splitlines()]
+        if not PF_devices:
+            raise exceptions.TestSkipError("No specified pf found in the host!")
+        pf_ids = []
+        for pf in PF_devices:
+            pf_id = utils_misc.get_full_pci_id(pf)
+            pf_ids.append(pf_id)
+        return pf_ids
+
     def check_vfs_count(self):
         """
         Check VFs count number according to the parameter driver_options.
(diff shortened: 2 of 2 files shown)


virttest/test_setup.py
```diff
@@ -1342,10 +1349,22 @@ def set_vf(self, pci_pf, vf_no="0"):
         :params vf_no: Number of Vfs to be virtualized
         :return: True on success, Fail on failure
         """
-        cmd = "echo %s > /sys/class/net/%s/device/sriov_numvfs" % (vf_no, pci_pf)
+        if ARCH == 'ppc64le':
+            path = "/sys/class/net/%s/device/sriov_numvfs" % pci_pf
+        elif ARCH == 'x86_64':
+            path = "/sys/bus/pci/devices/%s/sriov_numvfs" % pci_pf
```
balamuruhans: both paths are same, we can use 1 of them by removing the arch check.

shared/cfg/guest-hw.cfg
```diff
@@ -181,13 +180,9 @@ variants:
         # Mellanox CX4 MT27700 Family
         # For PowerPC driver (kernel module) that supports SR-IOV hardware is
         # mlx5_core
-        driver = igb
+        #driver = igb
```
balamuruhans: we don't have to comment the driver's default value, configure the required one and run it.

shared/cfg/guest-hw.cfg
```diff
@@ -181,13 +180,9 @@ variants:
...
         # vf_filter_re: Regex used to filter vf from lspci.
         vf_filter_re = "Virtual Function"
-        # pf_filter_re: Regex used to filter pf from lspci. It will be used
-        # when device_name could not filter the pf.
-        # For PowerPC: pf_filter_re = "MT27700 Family \[ConnectX-4\]"
-        pf_filter_re = "82576 Gigabit Network"
```
balamuruhans: same here

shared/cfg/guest-hw.cfg
```diff
@@ -163,7 +163,6 @@ variants:
         # available, it will also try to get first available pf.
         device_name = eth1
         nettype = bridge
-        pf_filter_re = "82576 Gigabit Network"
```
balamuruhans: why do we remove this param ?

virttest/test_setup.py
```diff
@@ -1313,6 +1303,23 @@ def get_pf_mlx(self):
             cmd, shell=True, timeout=60).splitlines()
         return (PF_devices)
 
+    def get_pf_ids(self):
+        """
+        Get the PF devices on x86 host
+        """
+        cmd = "lspci | grep -v 'Virtual Function' |awk '/%s/ {print $1}'" % self.pf_filter_re
```
balamuruhans: This will return VF devices right?

huangyum: No, it will return pf devices.

virttest/test_setup.py
huangyum: As I just checked, the second path exists on both arches. Update to use the second one. Thanks!

shared/cfg/guest-hw.cfg
huangyum: It is because we define it in other cfg for different pfs.  You can refer to https://github.com/autotest/tp-qemu/pull/1026.

shared/cfg/guest-hw.cfg
huangyum: I comment it because the driver is different for different pfs.  Since several kinds of pfs can be on one host at same time, we want to test them by running avocado only once, just like https://github.com/autotest/tp-qemu/pull/1026#issuecomment-318313569.

shared/cfg/guest-hw.cfg
huangyum: Same as above. Thanks for your review :)

virttest/test_setup.py
yanan-fu: why not use  cmd = "lspci | awk '/%s/ {print $1}'" % self.pf_filter_re  as before ? 
IMHO, this is mush easier to understand, and the one you added use "Virtual Function" is easy to misunderstanding.

huangyum: Because some vfs may contain the key words which is used to filter the pf.  
e.g. 
PF: Broadcom Limited NetXtreme II BCM57810 10 Gigabit Ethernet (rev 10)
VF: Broadcom Limited NetXtreme II BCM57810 10 Gigabit Ethernet Virtual Function

yanan-fu: oh~  got it,  you are right,  non-matching "Virtual Function",   thx!

shared/cfg/guest-hw.cfg
balamuruhans: okay, `# driver = igb` have a space after #

huangyum: Thanks, updated!

virttest/test_setup.py
balamuruhans: got it. Thanks.

shared/cfg/guest-hw.cfg
balamuruhans: other sriov tests also uses this param, we can override it in the --vt-config, what do you say ?

huangyum: How about this, I comment this line instead of removing it.  If someone wants to run other sriov tests, he could uncomment it and reconfigure it as he wants. After all, this param needs to be reconfigured according to the kinds of pf.  Thanks!

virttest/test_setup.py
```diff
@@ -1543,8 +1565,8 @@ def sr_iov_cleanup(self):
 
         re_probe = False
         # if lsmod lists the driver then remove it to clean up
-        if not process.system('lsmod | grep %s' % self.driver,
-                              ignore_status=True, shell=True):
+        if not self.driver or not process.system('lsmod | grep %s' % self.driver,
```
balamuruhans: if `self.driver` is None, why do we remove it?

huangyum: You are right.  Updated, please help review again, thanks!

virttest/test_setup.py
```diff
@@ -1362,13 +1375,15 @@ def remove_driver(self, driver=None):
...
-        if process.system(cmd, ignore_status=True, shell=True):
-            logging.debug("Failed to remove driver: %s", driver)
-            return False
+            if process.system(cmd, ignore_status=True, shell=True):
+                logging.debug("Failed to remove driver: %s", driver)
+                return False
+        elif ARCH == 'x86_64':
+            pf_devices = self.get_pf_ids()
+        for PF in pf_devices:
+            if not self.set_vf(PF):
```
balamuruhans: At this point, for `ARCH == 'ppc64le'` driver would have been removed and we cannot echo 0 to sriov_numvfs, please move it ahead.

virttest/test_setup.py
```diff
@@ -1387,7 +1402,7 @@ def modprobe_driver(self, driver=None, option=None):
         cmd = "modprobe %s" % driver
         if not option:
             option = self.driver_option
-        if driver and driver != "mlx5_core":
+        if driver and driver != "mlx5_core" and ARCH != 'x86_64':
```
balamuruhans: earlier this was used by x86_64, as ppc doesn't have any other than mlx, so is it sure that x86_64 never uses it?

huangyum: The option is used to generate vf. Now function `set_vf` does that.  So it can be removed for x86_64 platform.

virttest/test_setup.py
```diff
@@ -1544,7 +1566,7 @@ def sr_iov_cleanup(self):
         re_probe = False
         # if lsmod lists the driver then remove it to clean up
         if not process.system('lsmod | grep %s' % self.driver,
-                              ignore_status=True, shell=True):
+                                                 ignore_status=True, shell=True):
```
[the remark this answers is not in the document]
huangyum: Updated, please help review again, thanks!

virttest/test_setup.py
```diff
@@ -1362,10 +1375,12 @@ def remove_driver(self, driver=None):
...
             logging.info("Mellanox PF devices '%s'", pf_devices)
-            for PF in pf_devices:
-                if not self.set_vf(PF):
-                    return False
             cmd = "rmmod mlx5_ib;modprobe -r mlx5_core;modprobe mlx5_ib"
+        elif ARCH == 'x86_64':
+            pf_devices = self.get_pf_ids()
+        for PF in pf_devices:
+            if not self.set_vf(PF):
+                return False
```
balamuruhans: @huangyum, In Older code you can see `set_vf()` is performed before removing mlx drivers for ppc64le, with this change the mlx5 drivers will be removed in L1378 after which `set_vf()` will fail.

huangyum: Thanks for your review. L1378 only sets 'cmd', which will be executed later in L1384.  set_vf() is executed before that so it would not fail.

balamuruhans: Got it, Thanks!

balamuruhans:
Thanks @huangyum, lets merge it.

PR #1089, opened 2017-08-01
nfs: Cleanup new created export dir

waynesun09:
Add option to support clean up new created dir in nfs module.
Enable cleanup both new created export and mount dir in libvirt
nfs setup and clean function.

### DIFF
--- a/virttest/nfs.py
+++ b/virttest/nfs.py
@@ -152,7 +152,8 @@ class Nfs(object):
         self.mount_src = params.get("nfs_mount_src")
         self.nfs_setup = False
         path.find_command("mount")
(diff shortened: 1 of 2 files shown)


virttest/nfs.py
```diff
@@ -153,6 +153,7 @@ def __init__(self, params):
         self.nfs_setup = False
         path.find_command("mount")
         self.mk_mount_dir = False
+        self.mk_export_dir = False
```
balamuruhans: can we have variable name like `self.rm_export_dir` as mk misinterprets to make

waynesun09: Yeah, updating now.

PR #1092, opened 2017-08-02
libvirt_vm: Fixup virt-install takes conflicting memory unit with --memory and --cpu

balamuruhans:
virt-install takes MiB as default memory unit for --memory option and
in KiB as default memory unit for --cpu option while adding numa nodes.
Due to this behavior virt-install boots guest with memory value in --memory
but with KiB due to libvirt issue.


xutian:
looks OK to me.

balamuruhans:
Thanks @sathnaga @xutian

PR #1103, opened 2017-08-09
test_setup:Modify the adjudgement for set_hugepages 

hahakiki2010:
After the set operation, it shall compare the hugepages with the target
hugepages number, but not the one before set. Otherwise "Can't allocate
memory" might be raised during vm start.

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -513,8 +513,8 @@ class HugePageConfig(object):
             except IOError:
                 msg = "Can't read/write from kernel hugepage file"
                 raise exceptions.TestSetupFail(msg)
-            if loop_hp == hp:
-                raise ValueError("Cannot set the kernel hugepage setting "
+            if hp != self.target_hugepages:
+                raise exceptions.TestCancel("Cannot set the kernel hugepage setting "
                                  "to the target value of %d hugepages." %
                                  self.target_hugepages)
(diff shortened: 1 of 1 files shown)


virttest/test_setup.py
```diff
@@ -513,8 +513,8 @@ def set_hugepages(self):
             except IOError:
                 msg = "Can't read/write from kernel hugepage file"
                 raise exceptions.TestSetupFail(msg)
-            if loop_hp == hp:
-                raise ValueError("Cannot set the kernel hugepage setting "
+            if hp != self.target_hugepages:
+                raise exceptions.TestCancel("Cannot set the kernel hugepage setting "
```
balamuruhans: This would be a TestSetupFail(), also please fix indentation in below 2 lines for travis.

virttest/test_setup.py
```diff
@@ -513,8 +513,8 @@ def set_hugepages(self):
             except IOError:
                 msg = "Can't read/write from kernel hugepage file"
                 raise exceptions.TestSetupFail(msg)
-            if loop_hp == hp:
-                raise ValueError("Cannot set the kernel hugepage setting "
+            if hp != self.target_hugepages:
```
CongLi: It's a loop to set hugepage.

hahakiki2010: yes, I think you are right. I will revert the commit instead.

PR #1106, opened 2017-08-10
virttest.qbuses: support to specify the port of usb device on the bus

hereischen:
The param port_prefix is initialized with value None.
It will not work, if set usb devices up by specifying usb device port
on the bus. Because it will always return False or try to get len of
a None object.

id: 1480109


CongLi:
Ack.

luckyh:
Thanks all.

PR #1112, opened 2017-08-14
utils_misc.py: add support for guest disk mount

vivianQizhu:
1. Add guest support for mount function in utils_disk
2. Import utils_disk directly in utils_misc for same function, it should be removed, but just keep it for existed import.
id: 1490235

### DIFF
--- a/virttest/utils_disk.py
+++ b/virttest/utils_disk.py
@@ -49,35 +49,63 @@ def copytree(src, dst, overwrite=True, ignore=''):
             shutil.copy(src_file, dst_dir)
 
 
-def is_mount(src, dst):
+def is_mount(src, dst=None, fstype=None, options=None, verbose=False,
+             session=None):
     """
     Check is src or dst mounted.
 
-    :param src: source device or directory, if None will skip to check
+    :param src: source device or directory
     :param dst: mountpoint, if None will skip to check
+    :param fstype: file system type, if None will skip to check
+    :param options: mount options should be seperated by ","
+    :param session: check within the session if given
 
-    :return: if mounted mountpoint or device, else return False
+    :return: True if mounted, else return False
     """
-    if dst and os.path.ismount(dst):
-        return dst
-    if src and (src in str(open('/proc/mounts', 'r')) or
-                src in process.system_output('losetup -a')):
-        return src
+    mount_str = "%s %s %s" % (src, dst, fstype)
+    mount_str = mount_str.replace('None', '').strip()
+    mount_list_cmd = 'cat /proc/mounts'
+
+    if session:
+        mount_result = session.cmd_output_safe(mount_list_cmd)
+    else:
+        mount_result = process.system_output(mount_list_cmd, shell=True)
+    if verbose:
+        logging.debug("/proc/mounts contents:\n%s", mount_result)
+
+    for result in mount_result.splitlines():
+        if mount_str in result:
+            if options:
+                options = options.split(",")
+                options_result = result.split()[3].split(",")
+                for op in options:
+                    if op not in options_result:
+                        if verbose:
+                            logging.info("%s is not mounted with given"
+                                         " option %s", src, op)
+                        return False
+            if verbose:
+                logging.info("%s is mounted", src)
+            return True
+    if verbose:
+        logging.info("%s is not mounted", src)
     return False
 
 
-def mount(src, dst, fstype=None, options=None, verbose=False):
+def mount(src, dst, fstype=None, options=None, verbose=False, session=None):
     """
     Mount src under dst if it's really mounted, then remout with options.
 
-    :param src: source device or directory, if None will skip to check
-    :param dst: mountpoint, if None will skip to check
+    :param src: source device or directory
+    :param dst: mountpoint
     :param fstype: filesystem type need to mount
+    :param options: mount options
+    :param session: mount within the session if given
 
     :return: if mounted return True else return False
     """
     options = (options and [options] or [''])[0]
-    if is_mount(src, dst):
+    if is_mount(src, dst, fstype, options, verbose, session):
         if 'remount' not in options:
             options = 'remount,%s' % options
     cmd = ['mount']
@@ -87,24 +115,32 @@ def mount(src, dst, fstype=None, options=None, verbose=False):
         cmd.extend(['-o', options])
     cmd.extend([src, dst])
     cmd = ' '.join(cmd)
+    if session:
+        return session.cmd_status(cmd, safe=True) == 0
     return process.system(cmd, verbose=verbose) == 0
 
 
-def umount(src, dst, verbose=False):
+def umount(src, dst, fstype=None, verbose=False, session=None):
     """
     Umount src from dst, if src really mounted under dst.
 
-    :param src: source device or directory, if None will skip to check
-    :param dst: mountpoint, if None will skip to check
+    :param src: source device or directory
+    :param dst: mountpoint
+    :param fstype: fstype used to check if mounted as expected
+    :param session: umount within the session if given
 
     :return: if unmounted return True else return False
     """
-    mounted = is_mount(src, dst)
+    mounted = is_mount(src, dst, fstype, verbose=verbose, session=session)
     if mounted:
-        fuser_cmd = "fuser -km %s" % mounted
-        process.system(fuser_cmd, ignore_status=True, verbose=True)
-        umount_cmd = "umount %s" % mounted
-        return process.system(umount_cmd, ignore_status=True, verbose=True) == 0
+        fuser_cmd = "fuser -km %s" % dst
+        umount_cmd = "umount %s" % dst
+        if session:
+            session.cmd_output_safe(fuser_cmd)
+            return session.cmd_status(umount_cmd, safe=True) == 0
+        else:
+            process.system(fuser_cmd, ignore_status=True, verbose=True)
+            return process.system(umount_cmd, ignore_status=True, verbose=True) == 0
     return True
 
 
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -45,6 +45,7 @@ from . import data_dir
 from . import error_context
 from . import cartesian_config
 from . import utils_selinux
+from . import utils_disk
 from .staging import utils_koji
 from .xml_utils import XMLTreeFile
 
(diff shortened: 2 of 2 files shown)


virttest/utils_misc.py
```diff
@@ -1088,7 +1088,8 @@ def umount(src, mount_point, fstype, verbose=False, fstype_mtab=None):
         return True
 
 
-def mount(src, mount_point, fstype, perm=None, verbose=False, fstype_mtab=None):
+def mount(src, mount_point, fstype, perm=None, verbose=False, fstype_mtab=None,
+          session=None):
```
balamuruhans: Also can we consider support for umount as it will be used in cleanup code, otherwise code looks good to me.

vivianQizhu: @balamuruhans Thanks, I will merge this with utils_disk.mount(), where include umount() too.

balamuruhans: Sure, Thanks!

vivianQizhu:
11:14:07 INFO | PASS 1-smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.myprovider.live_backup.multi_disks

11:13:48 DEBUG| Sending command (safe): cat /proc/mounts |cut -d " " -f 2
11:13:48 DEBUG| Sending command (safe): losetup -a |cut -d " " -f 1
11:13:48 DEBUG| Sending command (safe): cat /proc/mounts |cut -d " " -f 1
11:13:48 DEBUG| Sending command (safe): losetup -a |cut -d " " -f 3
11:13:48 DEBUG| Sending command (safe): mount -t ext4 /dev/vdc /mnt/vdc
11:13:49 DEBUG| Sending command (safe): echo $?
11:13:49 DEBUG| Sending command (safe): cat /proc/mounts |cut -d " " -f 2
11:13:49 DEBUG| Sending command (safe): losetup -a |cut -d " " -f 1
11:13:49 DEBUG| Sending command (safe): cat /proc/mounts |cut -d " " -f 1
11:13:49 DEBUG| Sending command (safe): losetup -a |cut -d " " -f 3
11:13:49 DEBUG| Sending command (safe): mount -t ext4 /dev/vdb /mnt/vdb
11:13:49 DEBUG| Sending command (safe): echo $?

vivianQizhu:
@CongLi @luckyh Would you please help review? Thanks.

virttest/utils_disk.py
```diff
@@ -87,24 +113,31 @@ def mount(src, dst, fstype=None, options=None, verbose=False):
...
     """
     Umount src from dst, if src really mounted under dst.
 
     :param src: source device or directory, if None will skip to check
     :param dst: mountpoint, if None will skip to check
+    :param session: umount within the session if given
 
     :return: if unmounted return True else return False
     """
     mounted = is_mount(src, dst)
```
luckyh: Need to pass `session` to this function.

vivianQizhu:
@luckyh Updated, please help review again, thanks.

CongLi:
@vivianQizhu 

Could you rebase your code at first?

vivianQizhu:
@CongLi Updated, please help review, thanks.

CongLi:
1. please fix the typo in commit log.
'utils: Merge mount function in utils_misk and utils_disk, and add sup…'
'Import utils_disk directly in util_misk for same function, it should be removed,
but just keep it for existed import.'
s/utils_misk/utils_misc

virttest/utils_misc.py
```diff
@@ -1073,22 +1074,11 @@ def umount(src, mount_point, fstype, verbose=False, fstype_mtab=None):
     :param fstype_mtab: file system type in mtab could be different
     :type fstype_mtab: str
     """
-    if fstype_mtab is None:
-        fstype_mtab = fstype
+    return utils_disk.umount(src, mount_point, verbose, session)
```
CongLi: how about the param fstype_mtab?

virttest/utils_misc.py
```diff
@@ -1137,32 +1112,7 @@ def is_mounted(src, mount_point, fstype, perm=None, verbose=False,
...
-    logging.debug("Searching '%s' in mtab...", mount_string)
-    if verbose:
-        logging.debug("/etc/mtab contents:\n%s", file("/etc/mtab").read())
-    if re.findall(mount_string.strip(), file("/etc/mtab").read()):
-        logging.debug("%s is mounted.", src)
-        return True
-    else:
-        logging.debug("%s is not mounted.", src)
-        return False
+    return utils_disk.is_mount(src, mount_point, verbose, session)
```
CongLi: same as above, fstype_mtab?

vivianQizhu: @CongLi Thanks. The parameter fstype and fstype_mtab is only meaningful when justify is_mounted by file /etc/mtab, while this will not give correct result if only simply mount by command 'mount -type src des', which is what mount() in both utils_disk and utils_misk do. So I inherited method in utils_disk to check is mounted or not by /proc/mounts, in this case, fstype is not really necessary for the check.  In umount(), fstype is also useless besides is_mount(). Only keep these parameters for safe since it is already imported by some other module.

vivianQizhu:
@CongLi Updated commit log. Thanks.

virttest/utils_disk.py
```diff
@@ -49,35 +49,61 @@ def copytree(src, dst, overwrite=True, ignore=''):
...
-def is_mount(src, dst):
+def is_mount(src, dst, verbose=False, session=None):
     """
     Check is src or dst mounted.
 
     :param src: source device or directory, if None will skip to check
     :param dst: mountpoint, if None will skip to check
+    :param session: check within the session if given
 
     :return: if mounted mountpoint or device, else return False
```
luckyh: We can't know if the `src` is mounted to the `dst` by the current implementation, that's a problem.

vivianQizhu:
@luckyh Updated is_mount() check method to check if src mounted to dst. Please help review again, many thanks.

virttest/utils_disk.py
```diff
@@ -49,35 +49,56 @@ def copytree(src, dst, overwrite=True, ignore=''):
             shutil.copy(src_file, dst_dir)
 
 
-def is_mount(src, dst):
+def is_mount(src, dst, verbose=False, session=None):
```
luckyh: IMHO, we should add `fstype` and `option` here.

vivianQizhu: @luckyh Updated, please help review, thanks.

vivianQizhu:
Test passed.
15:46:49 DEBUG| [stdout] 127.0.0.1:/home/nfs /mnt/nfs nfs4 rw,relatime,vers=4.0,rsize=1048576,wsize=1048576,namlen=255,hard,proto=tcp,port=0,timeo=600,retrans=2,sec=sys,clientaddr=127.0.0.1,local_lock=none,addr=127.0.0.1 0 0
15:46:49 INFO | 127.0.0.1:/home/nfs is mounted

@luckyh Could you help review again? Thanks.

virttest/utils_disk.py
```diff
@@ -87,24 +111,31 @@ def mount(src, dst, fstype=None, options=None, verbose=False):
         cmd.extend(['-o', options])
     cmd.extend([src, dst])
```
luckyh: What if no `src`, or `src` is a relative path?

vivianQizhu: @luckyh I supposed this requires src must exist and must be a real path. Otherwise the case should fail with the command execution failure.

luckyh: ok, make sense, seems "no src" would not happen, and user can avoid to use relative path.

virttest/utils_disk.py
```diff
@@ -49,35 +49,59 @@ def copytree(src, dst, overwrite=True, ignore=''):
...
+    if verbose:
+        logging.debug("/proc/mounts contents:\n%s", mount_result)
+
+    for result in mount_result.splitlines():
+        if mount_str in result:
+            if options:
+                options = options.split(",")
+                options_result = result.split()[3].split(",")
+                for op in options:
+                    if op not in options_result:
```
luckyh: How about the `nfs -> nfs4` issue?

virttest/utils_disk.py
```diff
@@ -49,35 +49,59 @@ def copytree(src, dst, overwrite=True, ignore=''):
...
+                        logging.info("%s is not mounted with given option %s", src, op)
```
luckyh: `if verbose`?

virttest/utils_disk.py
vivianQizhu: @luckyh Since this version I only get first 3 parameters for mount string. It would be `127.0.0.1:/home/nfs /mnt/nfs nfs` in `127.0.0.1:/home/nfs /mnt/nfs nfs4`, it will return `True` anyway. So I supposed no need to handle it as exception. Not sure for other fstype, but we can add exceptions when we meet it.

virttest/utils_disk.py
vivianQizhu: @luckyh Updated, thanks.

vivianQizhu:
@luckyh Could you help re-review it? Thanks.

vivianQizhu:
@hereischen Would you please help review it? Many thanks.

virttest/utils_disk.py
```diff
@@ -87,24 +115,32 @@ def mount(src, dst, fstype=None, options=None, verbose=False):
...
+    :param dst: mountpoint
+    :param fstype: fstype used to check if mounted as expected
+    :param session: umount within the session if given
 
     :return: if unmounted return True else return False
     """
-    mounted = is_mount(src, dst)
+    mounted = is_mount(src, dst, fstype, verbose=verbose, session=session)
     if mounted:
         fuser_cmd = "fuser -km %s" % mounted
```
hereischen: `mounted is True`, which makes `fuser_cmd = "fuser -km True"`.

vivianQizhu:
@hereischen Thanks for your comment, updated and retest:
>>> from virttest import utils_disk
>>> fstype="nfs"
>>> src="10.66.8.133:/home/nfs"
>>> dst="/mnt/nfs"
>>> options="rw"
>>> utils_disk.mount(src, dst, fstype=fstype, options=options)
True
>>> utils_disk.is_mount(src, dst, fstype=fstype, options=options)
True
>>> utils_disk.umount(src, dst, fstype=fstype)
True
>>> utils_disk.is_mount(src, dst, fstype=fstype, options=options)
False

vivianQizhu:
utils_misc test:
>>> from virttest import utils_misc
>>> utils_misc.mount(src, dst, fstype=fstype, perm=options)
True
>>> utils_misc.is_mounted(src, dst, fstype=fstype, perm=options)
True
>>> utils_misc.umount(src, dst, fstype=fstype)
True
>>> utils_misc.is_mounted(src, dst, fstype=fstype, perm=options)
False
>>>

hereischen:

vivianQizhu:
@luckyh Would you please help merge it if no more comments? Many thanks.

luckyh:
Let's merge this, thanks all.

PR #1113, opened 2017-08-14
Add wait_for for mounted status check

hahakiki2010:
Add wait_for(is_mounted) to make sure mount info has updated in /etc/mtab already

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -1113,7 +1113,8 @@ def mount(src, mount_point, fstype, perm=None, verbose=False, fstype_mtab=None):
         process.system(mount_cmd, verbose=verbose)
     except process.CmdError:
         return False
-    return is_mounted(src, mount_point, fstype, perm, verbose, fstype_mtab)
+    return wait_for(lambda: is_mounted(src, mount_point, fstype, perm, verbose, fstype_mtab),
+                    timeout=5)
 
 
 def is_mounted(src, mount_point, fstype, perm=None, verbose=False,


virttest/nfs.py
```diff
@@ -225,7 +225,7 @@ def setup(self):
         if not os.path.isdir(self.mount_dir):
             os.makedirs(self.mount_dir)
             self.rm_mount_dir = True
-        self.mount()
+        utils_misc.wait_for(self.mount(), 5, text="Waiting for nfs mount")
```
CongLi: for retry, you could use 'soft' mode of 'mount' option like : 'mount -o soft,timeo=20,retrans=3'.

hahakiki2010: actually, I'm not sure if I shall retry the nfs mount or not. The issue is sometimes the mount info just not timely updated in /etc/mtab after nfs mount. So it may cause nfs pool fail to create. In this case, I think wait_for(is_mounted) is more suitable than nfs retry. I changed the code, pls help review again.

CongLi: I see. but step=0.1 is too short I think which will cause frequent call.

hahakiki2010: so how about step=1 ?

CongLi: step=1 is fine for me.

hahakiki2010: done, set step=1

hahakiki2010:
@lmr , rebase done

CongLi:

waynesun09:
ACK

PR #1117, opened 2017-08-16
Add perf function in virttest/perf.py

jinqi7:

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -4069,6 +4069,22 @@ def blkdeviotune(name, device=None, options=None,
     return command(cmd, **dargs)
 
 
+def perf(domain, options="", events="", other_opt="", **dargs):
+    """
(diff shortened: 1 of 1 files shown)


lmr:
@jinqi7, please rebase against the latest master, I've fixed Travis CI

virttest/virsh.py
```diff
@@ -4069,6 +4069,22 @@ def blkdeviotune(name, device=None, options=None,
     return command(cmd, **dargs)
 
 
+def perf(domain="", options="", events="", other_opt="", **dargs):
```
balamuruhans: as domain is mandate, let us not have domain as default argument.

jinqi7: Changed 
def perf(domain="", options="", events="", other_opt="", **dargs):
to 
def perf(domain, options="", events="", other_opt="", **dargs):

PR #1129, opened 2017-08-22
Revert "libvirt_xml: Fix pylint E1003"

waynesun09:
This reverts commit f1e87c1178975fa04d214374c216647626672f2f.

The super will fail with imp.load_module as load_module will
reload the class and got TypeError with: obj must be an
instance or subtype of type.

This problem exist for a long time and in commit 0bb4f7a4b33
fixed it with using self.__class__, E1003 also been disabled
in other classes, so revert to disable E1003 check and avoid
the problem.


kylazhang:
Merged quickly as it affect our daily test with so many failure, any further problem please noticed, thanks.

PR #1130, opened 2017-08-24
Add support of pool_autostart

hahakiki2010:
libvirt_storage: Add support of pool_autostart, to get
pool autostart status

### DIFF
--- a/virttest/libvirt_storage.py
+++ b/virttest/libvirt_storage.py
@@ -172,17 +172,20 @@ class StoragePool(object):
         :return: active/inactive, and None when something wrong.
         """
         try:
-            pools = self.list_pools()
-        except process.CmdError:
+            return self.list_pools()[name]['State']
+        except process.CmdError, KeyError:
             return None
 
-        if self.pool_exists(name):
-            details_dict = pools[name]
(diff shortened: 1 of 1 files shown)


virttest/libvirt_storage.py
```diff
@@ -184,6 +184,24 @@ def pool_state(self, name):
...
+            pools = self.list_pools()
+        except process.CmdError:
+            return None
+        if self.pool_exists(name):
+            details_dict = pools[name]
+            try:
+                return details_dict['Autostart']
+            except KeyError:
+                pass
+        return None
```
balamuruhans:
```
try:
    return self.list_pools()[name]["Autostart"]
except process.CmdError, KeyError:
    return None
```

hahakiki2010: Is it for more simple or something else?

balamuruhans: `pool_exists` also calls `list_pools()`, this redundant calls can be avoided.

hahakiki2010: Agreed and updated code. beside, pool_state is updated in this way also.

PR #1133, opened 2017-08-28
utils_test.libvirt: Support to keep nfs export dir

dzhengfy:
When clean up, nfs export dir is forcely removed. But sometimes we do
need keep the export dir. So this is to provide a way to determine if we
want to remove it by the user function.


waynesun09:
ACK

PR #1136, opened 2017-08-31
pool_xml: Add support of del_uuid

hahakiki2010:
Add new function 'del_uuid' for storage cases developing purpose


hahakiki2010:
find another more simple way to del uuid. this function no more needed.

PR #1137, opened 2017-09-01
Utils_misc: add support to get machine types for ppc and s390

aliang123:
Utils_misc: add support to get machine types for ppc and s390

Signed-off-by:Aihua Liang<aliang@redhat.com>

Bug id:1487520

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2091,9 +2091,14 @@ def get_support_machine_type(qemu_binary="/usr/libexec/qemu-kvm"):
     """
     Get the machine type the host support,return a list of machine type
     """
-    o = process.system_output("%s -M ?" % qemu_binary)
-    s = re.findall("(\S*)\s*RHEL[-\s]", o)
-    c = re.findall("(RHEL.*)", o)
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -2091,9 +2091,14 @@ def get_support_machine_type(qemu_binary="/usr/libexec/qemu-kvm"):
...
     Get the machine type the host support,return a list of machine type
     """
-    o = process.system_output("%s -M ?" % qemu_binary)
-    s = re.findall("(\S*)\s*RHEL[-\s]", o)
-    c = re.findall("(RHEL.*)", o)
+    o = process.system_output("%s -M ?" % qemu_binary).split("\n")
+    s = []
+    c = []
+    for item in o[1:]:
+        machine_list = re.split(r"\s", item, 1)
```
luckyh: Suggest changing the pattern to `\s+`, so that `machine_list[1]` doesn't to be stripped later.

aliang123: @luckyh, have updated the code, thanks for pointing it out.

PR #1139, opened 2017-09-01
Introduced function for setting static IP for the PF devices of mlx5_core driver.

santwana:
The static IP assignment for PF devices for mlx5_core driver is happening after re-probe.
Introduced a function for setting the static IP for the PF devices , which can be referenced before probe and after as well.

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -9,7 +9,7 @@ import random
 import math
 import shutil
 import platform
-from netaddr import *
+import netaddr
 
 from avocado.utils import process
 from avocado.utils import archive
@@ -27,7 +27,6 @@ from . import openvswitch
 from . import remote
 from . import utils_libvirtd
 from . import utils_config
-from . import utils_net
 from .staging import service
 from .staging import utils_memory
 
@@ -1301,6 +1300,29 @@ class PciAssignable(object):
             pf_ids.append(pf_id)
         return pf_ids
 
+    def assign_static_ip(self):
+        """
+        Set the static IP for the PF devices for mlx5_core driver
+        """
+        # This function assigns static IP for the PF devices
+        pf_devices = self.get_pf_ids()
+        if (not self.start_addr_PF) or (not self.net_mask):
+            raise exceptions.TestSetupFail(
+                "No IP / netmask found, please populate starting IP address for PF devices in configuration file")
+        ip_addr = netaddr.IPAddress(self.start_addr_PF)
+        for PF in pf_devices:
+            ifname = utils_misc.get_interface_from_pci_id(PF)
+            ip_assign = "ifconfig %s %s netmask %s up" % (
+                ifname, ip_addr, self.net_mask)
+            logging.info("assign IP to PF device %s : %s", PF,
+                         ip_assign)
+            cmd = process.system(ip_assign, shell=True, ignore_status=True)
+            if cmd:
+                raise exceptions.TestSetupFail("Failed to assign IP : %s"
+                                               % cmd)
+            ip_addr += 1
+        return True
+
     def check_vfs_count(self):
         """
         Check VFs count number according to the parameter driver_options.
@@ -1450,6 +1472,8 @@ class PciAssignable(object):
         # If driver is available and pa_type is vf then set VFs
         elif self.pa_type == 'vf':
             pf_devices = self.get_pf_ids()
+            if (self.static_ip):
+                self.assign_static_ip()
             for PF in pf_devices:
                 if not self.set_vf(PF, self.driver_option):
                     re_probe = True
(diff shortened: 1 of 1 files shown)


virttest/test_setup.py
```diff
@@ -1482,26 +1508,9 @@ def sr_iov_setup(self):
...
-                        logging.info("assign IP to PF device %s : %s", PF,
-                                     ip_assign)
-                        if process.system(ip_assign, shell=True,
-                                          ignore_status=True):
-                            return False
-                        ip_addr = ip_addr + 1
-                    logging.info("PF device '%s'.", PF)
-                    if not self.set_vf(PF, self.driver_option):
-                        return False
+                ip_assign = self.assign_ip_pf_mlx()
```
balamuruhans: Static ip assignment is not specific to ppc64le, we can keep it out.
we should check for the flag `self.static_ip` before calling the function instead.

virttest/test_setup.py
```diff
@@ -1466,7 +1490,9 @@ def sr_iov_setup(self):
         # If driver is available then set VFs for ppc64le
         else:
             if ARCH == 'ppc64le' and self.driver == 'mlx5_core':
+                ip_assign = self.assign_ip_pf_mlx()
```

virttest/test_setup.py
```diff
@@ -1313,6 +1312,31 @@ def get_pf_mlx(self):
...
+            set_ip = 1
+        for PF in pf_devices:
+            if set_ip:
+                ip_assign = "ifconfig %s %s netmask %s up" % (
+                    PF, ip_addr, self.net_mask)
+                logging.info("assign IP to PF device %s : %s", PF,
+                             ip_assign)
+                process.system(ip_assign, shell=True, ignore_status=True)
+                ip_addr += 1
+            logging.info("PF device '%s'.", PF)
```
balamuruhans: Not required as L1333 logs it.

virttest/test_setup.py
```diff
@@ -1313,6 +1312,31 @@ def get_pf_mlx(self):
...
         return (PF_devices)
 
+    def assign_ip_pf_mlx(self):
+        """
+        Set the static IP for the PF devices for mlx5_core driver
+        """
+        # This function assigns static IP for the PF device of
+        # mellanox cx4 card
+        set_ip = 0
+        pf_devices = self.get_pf_mlx()
```
balamuruhans: This looks redundant, we call this function again in L1494, lets pass it as parameter and use it instead calling twice.

virttest/test_setup.py
```diff
@@ -1313,6 +1312,31 @@ def get_pf_mlx(self):
             cmd, shell=True, timeout=60).splitlines()
         return (PF_devices)
 
+    def assign_ip_pf_mlx(self):
```
balamuruhans: `def assign_static_ip()` can be used.

virttest/test_setup.py
```diff
@@ -1482,26 +1508,9 @@ def sr_iov_setup(self):
...
-                                     ip_assign)
-                        if process.system(ip_assign, shell=True,
-                                          ignore_status=True):
-                            return False
-                        ip_addr = ip_addr + 1
-                    logging.info("PF device '%s'.", PF)
-                    if not self.set_vf(PF, self.driver_option):
-                        return False
+                ip_assign = self.assign_ip_pf_mlx()
+                if not self.set_vf(PF, self.driver_option):
```
huangyum: Undefined PF ?

santwana:
@balamuruhans Addressed the changes, please review.

virttest/test_setup.py
```diff
@@ -1301,6 +1300,31 @@ def get_pf_ids(self):
...
+            if (not self.start_addr_PF) or (not self.net_mask):
+                raise exceptions.TestSetupFail(
+                    "No IP / netmask found, please populate starting IP address for PF devices in configuration file")
+            ip_addr = netaddr.IPAddress(self.start_addr_PF)
+            set_ip = 1
+        for PF in pf_devices:
+            if set_ip:
+                ifname = utils_misc.get_interface_from_pci_id(PF)
+                ip_assign = "ifconfig %s %s netmask %s up" % (
+                    ifname, ip_addr, self.net_mask)
```
balamuruhans: so this will try to assign IP even if `self.static_ip` is not configured, also it would probably break here as `ip_addr` is referenced before assignment in case of `self.static_ip` is not configured.

I would suggest to have check of `self.static_ip` and then call `assign_static_ip()` from `sr_iov_setup()`

virttest/test_setup.py
```diff
@@ -1450,6 +1474,7 @@ def sr_iov_setup(self):
         # If driver is available and pa_type is vf then set VFs
         elif self.pa_type == 'vf':
             pf_devices = self.get_pf_ids()
+            ip_assign = self.assign_static_ip()
```
balamuruhans: this variable is not used.

virttest/test_setup.py
```diff
@@ -1465,26 +1490,7 @@ def sr_iov_setup(self):
...
-                        ip_assign = "ifconfig %s %s netmask %s up" % (
-                            ifname, ip_addr, self.net_mask)
-                        logging.info("assign IP to PF device %s : %s", PF,
-                                     ip_assign)
-                        if process.system(ip_assign, shell=True,
-                                          ignore_status=True):
-                            return False
-                        ip_addr = ip_addr + 1
-                    logging.info("PF device '%s'.", PF)
+            ip_assign = self.assign_static_ip()
```

virttest/test_setup.py
santwana: @balamuruhans Addressed the changes suggested, please review. Thanks!

virttest/test_setup.py
```diff
@@ -1301,6 +1300,30 @@ def get_pf_ids(self):
...
+        # This function assigns static IP for the PF devices
+        set_ip = 0
+        pf_devices = self.get_pf_ids()
+        if (not self.start_addr_PF) or (not self.net_mask):
+            raise exceptions.TestSetupFail(
+                "No IP / netmask found, please populate starting IP address for PF devices in configuration file")
+        ip_addr = netaddr.IPAddress(self.start_addr_PF)
+        set_ip = 1
+        for PF in pf_devices:
+            if set_ip:
```
balamuruhans: In your new re-factor I don't think `set_ip` is required.

santwana: @balamuruhans Addressed the changes. Please review.

virttest/test_setup.py
```diff
@@ -1301,6 +1300,26 @@ def get_pf_ids(self):
...
+            raise exceptions.TestSetupFail(
+                "No IP / netmask found, please populate starting IP address for PF devices in configuration file")
+        ip_addr = netaddr.IPAddress(self.start_addr_PF)
+        for PF in pf_devices:
+            ifname = utils_misc.get_interface_from_pci_id(PF)
+            ip_assign = "ifconfig %s %s netmask %s up" % (
+                ifname, ip_addr, self.net_mask)
+            logging.info("assign IP to PF device %s : %s", PF,
+                         ip_assign)
+            process.system(ip_assign, shell=True, ignore_status=True)
```
balamuruhans: please have a check on the return value and raise `TestSetupFail()` if command fails

santwana: @balamuruhans Incorporated the changes suggested, please review. Thanks.

balamuruhans:
Thanks @santwana let's merge it.

PR #1143, opened 2017-09-04
libvirt: configure selinux for nfs on distro other than ubuntu

balamuruhans:
selinux configuration on ubuntu fails and interrupts the testcases
so consider it to avoid it for ubuntu. For negative scenarios to be
enabled 'enable_selinux_ubuntu' param can be set


balamuruhans:
https://github.com/autotest/tp-libvirt/pull/1203 needs this fix for ubuntu

PR #1144, opened 2017-09-05
Empty vm.spice_options before its another population

rdkdd:
In `vm.spice_options` are preserved outdated `params` from previous testcase. This is 
because `vm` instance is not recreated when another test run from job suite is performed.

example:
TC 1 `params`:
> spice_port = no
> spice_port_closed = no
> spice_seamless_migration = on
> spice_secure_channels = 
> spice_ssl = yes
> spice_tls_ciphers = DEFAULT
> spice_tls_port = blabla

qemu-kvm spice cmd:

    -spice disable-ticketing,tls-port=blabla,x509-dir=/tmp/spice_x509d,seamless-migration=on

`params` for following TC_2 in the same job:

> spice_addr = 127.0.0.1
> spice_port = generate
> spice_port_closed = no
> spice_seamless_migration = on

(notice `spice_tls_port = blabla` is not in `params`)

qemu-kvm spice cmd:

    -spice port=3000,disable-ticketing,addr=127.0.0.1,tls-port=blabla,x509-dir=/tmp/spice_x509d,seamless-migration=on

`tls-port=blabla` makes this 2. TC to fail.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -939,7 +939,7 @@ class VM(virt_vm.BaseVM):
             """
             processes spice parameters on rhel5 host.
 
-            :param spice_options - dict with spice keys/values
+            :param spice_params - dict with spice keys/values
             :param port_range - tuple with port range, default: (3000, 3199)
             """
 
@@ -950,9 +950,12 @@ class VM(virt_vm.BaseVM):
             spice_help = ""
             if devices.has_option("spice-help"):
                 spice_help = commands.getoutput("%s -device \\?" % qemu_binary)
-            s_port = str(utils_misc.find_free_port(*port_range))
-            self.spice_options['spice_port'] = s_port
-            cmd += " port=%s" % s_port
+            if not self.is_alive():
+                s_port = str(utils_misc.find_free_port(*port_range))
+                self.spice_port = s_port
+                cmd += " port=%s" % s_port
+            else:
+                cmd += "port=%s" % self.spice_port
             for param in spice_params.split():
                 value = params.get(param)
                 if value:
@@ -970,7 +973,7 @@ class VM(virt_vm.BaseVM):
                 cmd += " -qxl %s" % qxl_dev_nr
             return cmd
 
-        def add_spice(port_range=(3000, 3199),
+        def add_spice(spice_options, port_range=(3000, 3199),
                       tls_port_range=(3200, 3399)):
             """
             processes spice parameters
@@ -983,7 +986,7 @@ class VM(virt_vm.BaseVM):
 
             def optget(opt):
                 """a helper function"""
-                return self.spice_options.get(opt)
+                return spice_options.get(opt)
 
             def set_yes_no_value(key, yes_value=None, no_value=None):
                 """just a helper function"""
@@ -1001,14 +1004,14 @@ class VM(virt_vm.BaseVM):
                     spice_opts.append(opt_string % tmp)
                 elif fallback:
                     spice_opts.append(fallback)
-            s_port = str(utils_misc.find_free_port(*port_range))
             if optget("spice_port") == "generate":
                 if not self.is_alive():
-                    self.spice_options['spice_port'] = s_port
+                    s_port = str(utils_misc.find_free_port(*port_range))
+                    spice_options['spice_port'] = s_port
                     spice_opts.append("port=%s" % s_port)
                     self.spice_port = s_port
                 else:
-                    self.spice_options['spice_port'] = self.spice_port
+                    spice_options['spice_port'] = self.spice_port
                     spice_opts.append("port=%s" % self.spice_port)
             # spice_port = no: spice_port value is not present on qemu cmdline
             elif optget("spice_port") != "no":
@@ -1018,21 +1021,22 @@ class VM(virt_vm.BaseVM):
             ip_ver = optget("listening_addr")
             if ip_ver:
                 host_ip = utils_net.get_host_ip_address(self.params, ip_ver)
-                self.spice_options['spice_addr'] = host_ip
+                spice_options['spice_addr'] = host_ip
             set_yes_no_value(
                 "disable_copy_paste", yes_value="disable-copy-paste")
             set_value("addr=%s", "spice_addr")
 
             if optget("spice_ssl") == "yes":
                 # SSL only part
-                t_port = str(utils_misc.find_free_port(*tls_port_range))
                 if optget("spice_tls_port") == "generate":
                     if not self.is_alive():
-                        self.spice_options['spice_tls_port'] = t_port
+                        t_port = str(utils_misc.
+                                     find_free_port(*tls_port_range))
+                        spice_options['spice_tls_port'] = t_port
                         spice_opts.append("tls-port=%s" % t_port)
                         self.spice_tls_port = t_port
                     else:
-                        self.spice_options[
+                        spice_options[
                             'spice_tls_port'] = self.spice_tls_port
                         spice_opts.append("tls-port=%s" % self.spice_tls_port)
                 else:
@@ -1357,7 +1361,7 @@ class VM(virt_vm.BaseVM):
         # If nothing changed and devices exists, return immediately
         if (name is None and params is None and root_dir is None and
                 self.devices is not None):
-            return self.devices
+            return self.devices, self.spice_options
 
         if name is None:
             name = self.name
@@ -2075,15 +2079,15 @@ class VM(virt_vm.BaseVM):
                     "disable_copy_paste", "spice_seamless_migration",
                     "listening_addr"
                 )
-
+                spice_options = {}
                 for skey in spice_keys:
                     value = params.get(skey, None)
                     if value is not None:
                         # parameter can be defined as empty string in Cartesian
                         # config.  Example: spice_password =
-                        self.spice_options[skey] = value
+                        spice_options[skey] = value
 
-                cmd += add_spice()
+                cmd += add_spice(spice_options)
         if cmd:
             devices.insert(StrDev('display', cmdline=cmd))
 
@@ -2288,7 +2292,7 @@ class VM(virt_vm.BaseVM):
                 if set_disable_modern == "yes":
                     add_disable_modern(devices, device, dev_type)
 
-        return devices
+        return devices, spice_options
 
     def _nic_tap_add_helper(self, nic):
         if nic.nettype == 'macvtap':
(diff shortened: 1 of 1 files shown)


rdkdd:
Could you please check this PR @ldoktor, @will-Do, @Andrei-Stepanov and merge?

Thanks!

ldoktor:
[Fix the `VM.spice_options` usage to allow `VM.create` with different `params`.](https://trello.com/c/EkWc1DC0/910-fix-the-vmspiceoptions-usage-to-allow-vmcreate-with-different-params)

rdkdd:
Thank you @ldoktor for in-depth review but I am afraid I do not get it fully.. I have someremarks:
The third step:

> 3. Second case is started, env_process.preprocess_vm checks whether it can keep the previous VM by using vm.needs_restart()
> a. vm.needs_restart issues vm.make_create_command() which cleans the self.spice_options and uses first free slot, which is 3001.

As I can see it you mean line 580 of `virt_vm.py`:
`need_restart = (self.make_create_command() !=
                            self.make_create_command(name, params, basedir))`
`self.make_create_command()` returns `vm.devices` in `qemu_vm.py:1358`:
```
if (name is None and params is None and root_dir is None and
                self.devices is not None):
            return self.devices
```
so no cleaning there of `vm.spice_options` as I can see. or change of port number.
> b. vm.needs_restart issues vm.make_create_command(name, params, basedir) which again cleans the self.spice_options and uses first free slot, which is (as a coincidence) still 3001

ok, but I cannot reproduce this even with `keep_guest_running = True` in `vt.conf`. I also tried to set `kill_vm=no` but with no avail. I cannot get over `self.is_alive()` on line 576 in `virt_vm.py` and `True` is returned.
> c. vm is not restarted, because they match,

yes because only `vm.devices` is compared and not `vm.spice_options` <- **that is the problem as I understand**
So the problem are dynamically created parameters to `spice_options`, when vm guest  keeps running.

> Anyway I understand the concept os spice_options is widely spread and it'd be hard and error-prone to split and fix it properly, so similarly to previous discussion I think you could simply use local (and always new) variable for spice_options in make_create_command, return it together with devices (around line 2295 return devices, spice_options)

ok this way we can compare even `spice_options` in `vm.needs_restart()`

>  and assign it on create (around line 2719 self.devices, self.spice_options = self.make_create_command). That way you'd get functional needs_restart as well as always up2date spice_options. It'd still check whether port 3001 is port 3001, but after this operation the vm.spice_options would report the correct port 3000.
> 
> Hopefully this is understandable, if you want to I can take care of it (as I already debugged this twice...) or you can send v2 which would cover this scenario (feel free to disagree about the form, but the result must be that spice_options are only updated on create, which is when the cmd is executed.

I will send v2 tomorrow...

rdkdd:
Hi @ldoktor . After another deeper examination of `qemu_vm.py` file, I found that your scenario will probably not happen. `spice_port`, as dynamically generated variable, is also stored in `vm.spice_port` and is reloaded and used to qemu cmd-line and to feed `vm.spice_options` in case the vm is running. See `qemu_vm.py:1011`:
```
            if optget("spice_port") == "generate":
                if not self.is_alive():
                    self.spice_options['spice_port'] = s_port
                    spice_opts.append("port=%s" % s_port)
                    self.spice_port = s_port
                else:
                    self.spice_options['spice_port'] = self.spice_port
                    spice_opts.append("port=%s" % self.spice_port
```
the same is performed with `tls_port`. Is my assumption right? 

Even in that case `need_restart` in `needs_restart()` is not properly evaluated as old and new `vm.spice_options` are not compared to each other. So I think it is no problem to have `vm.spice_options` in `make_create_command`. But it is good idea to return `self.spice_options` together with `devices` as you proposed, so that new and old `vm.spice_options` will be compared. What is your opinion?

ldoktor:
Dne 7.9.2017 v 15:58 Radek Duda napsal(a):
> Thank you @ldoktor for in-depth review but I am afraid I do not get it fully.. I have someremarks:
> The third step:
> 
>> 3. Second case is started, env_process.preprocess_vm checks whether it can keep the previous VM by using vm.needs_restart()
>> a. vm.needs_restart issues vm.make_create_command() which cleans the self.spice_options and uses first free slot, which is 3001.
> 
> As I can see it you mean line 580 of `virt_vm.py`:
> `need_restart = (self.make_create_command() !=
>                             self.make_create_command(name, params, basedir))`
> `self.make_create_command()` returns `vm.devices` in `qemu_vm.py:1358`:
> ```
> if (name is None and params is None and root_dir is None and
>                 self.devices is not None):
>             return self.devices
> ```
> so no cleaning there of `vm.spice_options` as I can see. or change of port number.

Yes, sorry I was wrong about the first `self.make_create_command()`. The question is whether the second call `self.make_create_command(name, params, basedir)` (which unquestionably re-creates the `vm.spice_options`) can break things. If all spice_options produce visible changes to `devices`, then it should work properly (wort case it should force-recreate machine when it's not necessary), but if some spice_options does not affects `devices`, it might contain different value and still use the previously created machine.

Even worse is that `vm.make_create_command` can be used outside the `qemu_vm` (eg. on failure it's used in `avocado_vt/test.py`) or even from tests. Then the `vm.spice_options` would get refreshed and every since then it might contain different values.

Therefor yes, it's better than the current situation, but it can still break things silently. Do you want me to send the version which I think would work, or are you going to look into that a bit deeper?

Lukáš



>> b. vm.needs_restart issues vm.make_create_command(name, params, basedir) which again cleans the self.spice_options and uses first free slot, which is (as a coincidence) still 3001
> 
> ok, but I cannot reproduce this even with `keep_guest_running = True` in `vt.conf`. I also tried to set `kill_vm=no` but with no avail. I cannot get over `self.is_alive()` on line 576 in `virt_vm.py` and `True` is returned.
>> c. vm is not restarted, because they match,
> 
> yes because only `vm.devices` is compared and not `vm.spice_options` <- **that is the problem as I understand**
> So the problem are dynamically created parameters to `spice_options`, when vm guest  keeps running.
> 
>> Anyway I understand the concept os spice_options is widely spread and it'd be hard and error-prone to split and fix it properly, so similarly to previous discussion I think you could simply use local (and always new) variable for spice_options in make_create_command, return it together with devices (around line 2295 return devices, spice_options)
> 
> ok this way we can compare even `spice_options` in `vm.needs_restart()`
> 
>>  and assign it on create (around line 2719 self.devices, self.spice_options = self.make_create_command). That way you'd get functional needs_restart as well as always up2date spice_options. It'd still check whether port 3001 is port 3001, but after this operation the vm.spice_options would report the correct port 3000.
>>
>> Hopefully this is understandable, if you want to I can take care of it (as I already debugged this twice...) or you can send v2 which would cover this scenario (feel free to disagree about the form, but the result must be that spice_options are only updated on create, which is when the cmd is executed.
> 
> I will send v2 tomorrow...
>

rdkdd:
Hello @ldoktor, 
I made `spice_options` local and returned this together with `devices` as you proposed. Please take a look and if you do not find this version appropriate, please make your own as you already suggested.

Thank you!

ldoktor:
Yes, this is exactly what I had in mind. The `add_spice_rhel5` still overrides the `self.spice_options` rather than `spice_options` so please change that. Additionally I noticed the other spice values `self.spice_port` and `self.spice_tls_port` which suffer the same issue. Why do they exist in the first place? Can't they be just obtained from `spice_options`?

Anyway this seems as a correct fix to this issue, please fix the remaining spot and either here, or in following patch also take care of the `self.spice_*port` variables (simplest solution to me would be to turn it into a property which gets the value from `self.spice_options`, cleanest would be to adjust tests to not use it and get rid of them completely, or you can simply set them directly in `vm.create` based on `spice_options`. Either solution is fine by me).

rdkdd:
Hello @ldoktor and thanks for reply. Few comments:
> Yes, this is exactly what I had in mind. The add_spice_rhel5 still overrides the self.spice_options rather than spice_options so please change that. 

ok I will do that

> Additionally I noticed the other spice values self.spice_port and self.spice_tls_port which suffer the same issue. Why do they exist in the first place? Can't they be just obtained from spice_options?

You probably did not read https://github.com/avocado-framework/avocado-vt/pull/1144#issuecomment-328045503? They are there to prevent using incorrect spice port in case vm is not restarted. So .. now we do not need this, right?

ldoktor:
Ideally dynamic params should be treated by `dynamic` params (see https://github.com/avocado-framework/avocado-vt/blob/master/virttest/qemu_vm.py#L523 for example usage) so it might still be necessary as you don't use these (unless I missed them).

rdkdd:
@ldoktor since https://github.com/avocado-framework/avocado-vt/pull/1144#issuecomment-328045503 I find `vm.spice*_port` essential to store dynamically created spice port in case that vm is not restarted (`keep_guest_running = True`). Furthermore this possibility is badly treated in `add_spice_rhel5()`, thats why I added `vm.is_alive()` test on line 950 similarly as in `vm.add_spice()`. I will upload the code. I think `spice_options` are not needed anymore in `add_spice_rhel5()`. If you find it useful, I can try to incorporate `vi.spice-*port` options in `vm.devices` as you suggested in previous comment.
Radek

rdkdd:
Hi @ldoktor I have uploaded new version of PR. Could you look on it again please?
Thanks!

virttest/qemu_vm.py
```diff
@@ -950,9 +950,12 @@ def add_spice_rhel5(devices, spice_params, port_range=(3100, 3199)):
             spice_help = ""
             if devices.has_option("spice-help"):
                 spice_help = commands.getoutput("%s -device \\?" % qemu_binary)
-            s_port = str(utils_misc.find_free_port(*port_range))
-            self.spice_options['spice_port'] = s_port
-            cmd += " port=%s" % s_port
+            if not self.is_alive():
```
ldoktor: The thing here is that this works now, but in the future we might want to change the condition to skip evaluation from `is_alive` to something else and then this would break. Can I ask you one more change and then I promise I won't complain (about such issues) and that is to store those ports in `spice_options` and then update them in `create()` after retrieving the `spice_options`? Something like:

```diff
diff --git a/virttest/qemu_vm.py b/virttest/qemu_vm.py
index 97cd697..2349bde 100644
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -2717,6 +2717,8 @@ class VM(virt_vm.BaseVM):
             try:
                 self.devices, self.spice_options = self.make_create_command()
                 self.update_vga_global_default(params, migration_mode)
+                self.spice_port = self.spice_options.get("spice_port")
+                self.spice_tls_port = self.spice_options.get("spice_tls_port")
                 logging.debug(self.devices.str_short())
                 logging.debug(self.devices.str_bus_short())
                 qemu_command = self.devices.cmdline()
```

PS: Don't forget to store them, in this patch you don't keep it in some branches.

ldoktor:
Hello @rdkdd, it's pretty close, this approach works for now but is a "skeleton in the closet" as on change of behavior it might become hard-to-debug issue. Would you please move the `self.spice_port` and such to `vm.create` section after you receive the new params? It should behave the same but the location clearly states "update on recreate" and not "update when not running".

rdkdd:
Replaced by https://github.com/avocado-framework/avocado-vt/pull/1171.

PR #1145, opened 2017-09-05
Improve the process of sr_iov_setup

huangyum:
Improve sr_iov_setup process to make sure the driver is installed.
ID: 1488321

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -1475,7 +1475,7 @@ class PciAssignable(object):
                     time.sleep(3)
         re_probe = False
         # If driver not available after modprobe try to remove it and reprobe
-        if process.system("lsmod | grep %s" % self.driver, ignore_status=True,
+        if process.system("lsmod | grep -w %s" % self.driver, ignore_status=True,
                           shell=True):
             re_probe = True
         # If driver is available then set VFs for ppc64le
@@ -1496,7 +1496,7 @@ class PciAssignable(object):
         # Re-probe driver with proper number of VFs once more and raise
         # exception
         if re_probe:
-            if not self.remove_driver() and not self.modprobe_driver():
+            if not self.remove_driver() or not self.modprobe_driver():
                 return False
             if ARCH == 'ppc64le' and self.driver == 'mlx5_core':
                 set_ip = 0


virttest/test_setup.py
```diff
@@ -1496,7 +1496,7 @@ def sr_iov_setup(self):
         # Re-probe driver with proper number of VFs once more and raise
         # exception
         if re_probe:
-            if not self.remove_driver() and not self.modprobe_driver():
+            if not self.modprobe_driver():
```
yanan-fu: I am not faimlar ppc part and mlx5_core driver well.
"remove_driver"  here is used to disable all of the vfs (vf_no=0) before modprobe pf driver.
I think it is ok here. no need to be removed.

yanan-fu: Before commit  af59aa5797c8c9fcb569914dfeffd17c872242bd,  the function "remove_driver" only disable vf (set vf num to 0) for ppc and mlx5_core, i guess this is need for specific card. In that commit, x86 was added to disable vf. this is not must for x86, but it work too. So, consider ppc+mlx5_core, this function can not be removed here. And, before modprobe_driver,  remove_driver again even if it is not exist is ok.  it will return True when remove a no-exist driver.

huangyum: Yes, it cannot be removed. But there is something wrong with the logic. I will update.  Thanks!

virttest/test_setup.py
```diff
@@ -1496,7 +1496,7 @@ def sr_iov_setup(self):
         # Re-probe driver with proper number of VFs once more and raise
         # exception
         if re_probe:
-            if not self.remove_driver() and not self.modprobe_driver():
+            if not self.remove_driver() or not self.modprobe_driver():
```
balamuruhans: `or` should be removed. when reprobe have to be done, clean up the VF and remove driver if any, and perform a probe once before raising.

virttest/test_setup.py
balamuruhans: @yanan-fu right!

virttest/test_setup.py
huangyum: remove_driver() and modprobe_driver() return True on success.  If we keep `and` here,  when remove_driver() success,  modprobe_driver() will not be executed. That's why I use `or` instead. Thanks!

yanan-fu: Yes,  agree with @huangyum ,  `or` is the correct logic i think. @balamuruhans

yanan-fu:
LGTM

balamuruhans:
Thanks @huangyum @yanan-fu.

PR #1149, opened 2017-09-07
libvirt_storage: minor fix of multiple exceptions in one line

hahakiki2010:
when with multiple exceptions in one line, "as" shall be used.
the old way that separating the exceptions from the variable
with a comma is now deprecated.


sathnaga:
Closing this PR as it already got addressed and merged by another one.

PR #1151, opened 2017-09-11
Implementation of getting the iommu pci devices address

iccaszhulili:
Add the function of getting the pci devices addresses
sharing the same iommuGroup.

### DIFF
--- a/virttest/libvirt_xml/nodedev_xml.py
+++ b/virttest/libvirt_xml/nodedev_xml.py
@@ -126,22 +126,42 @@ class PCIXML(CAPXML):
     """
 
     # Example:
-    #<capability type='pci'>
-    #  <domain>0</domain>
-    #  <bus>7</bus>
-    #  <slot>0</slot>
-    #  <function>0</function>
-    #  <product id='0x1521'>I350 Gigabit Network Connection</product>
-    #  <vendor id='0x8086'>Intel Corporation</vendor>
-    #  <capability type='virt_functions'>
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/nodedev_xml.py
```diff
@@ -140,8 +140,27 @@ class for capability whose type is pci.
     #  <numa node='0'/>
     #</capability>
 
+    # or the Example of iommu:
+    #<capability type="pci">
```
chunfuwen: follow the comment format(leave one white space between # and following words):# xx ,and the following have similar issues.

iccaszhulili: Updated

virttest/libvirt_xml/nodedev_xml.py
```diff
@@ -162,6 +181,10 @@ def __init__(self, virsh_instance=base.virsh):
                                  parent_xpath='/capability',
                                  marshal_from=self.marshal_from_address,
                                  marshal_to=self.marshal_to_address)
+        accessors.XMLElementList('iommuGroup_number', self,
+                                 parent_xpath='/iommuGroup',
+                                 marshal_from=self.marshal_from_address,
+                                 marshal_to=self.marshal_to_address)
```
kylazhang: The iommuGroup_number should be an attribute? And here should be the 'iommuGroup_address' ?

iccaszhulili: @kylazhang Updated, please have a check

PR #1152, opened 2017-09-11
virttest.test_setup: Set vf only when test vf cases

huangyum:
Update test_setup.py to set vf only when test vf cases.
ID: 1489264

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -875,7 +875,8 @@ class PciAssignable(object):
 
     def __init__(self, driver=None, driver_option=None, host_set_flag=None,
                  kvm_params=None, vf_filter_re=None, pf_filter_re=None,
-                 device_driver=None, nic_name_re=None, static_ip=None, net_mask=None, start_addr_PF=None):
+                 device_driver=None, nic_name_re=None, static_ip=None,
+                 net_mask=None, start_addr_PF=None, pa_type=None):
         """
         Initialize parameter 'type' which could be:
         vf: Virtual Functions
@@ -909,7 +910,7 @@ class PciAssignable(object):
         :type pf_filter_re: string
         :param static_ip: Flag to be set if the test should assign static IP
         :param start_addr_PF: Starting private IPv4 address for the PF interface
-
+        :param pa_type: pci_assignable type, pf or vf
         """
         self.devices = []
         self.driver = driver
@@ -924,6 +925,7 @@ class PciAssignable(object):
         self.static_ip = static_ip
         self.net_mask = net_mask
         self.start_addr_PF = start_addr_PF
+        self.pa_type = pa_type
 
         if nic_name_re:
             self.nic_name_re = nic_name_re
@@ -1284,20 +1286,9 @@ class PciAssignable(object):
             pci_ids = out.split()
         return pci_ids
 
-    def get_pf_mlx(self):
-        """
-        Get the PF devices for mlx5_core driver
-        """
-        # This function returns the network device associated with
-        # mellanox cx4 card
-        cmd = "lshw -c network -businfo | grep '%s' | awk '{print $2}'" % self.pf_filter_re
-        PF_devices = process.system_output(
-            cmd, shell=True, timeout=60).splitlines()
-        return (PF_devices)
-
     def get_pf_ids(self):
         """
-        Get the PF devices on x86 host
+        Get the id of PF devices
         """
         cmd = "lspci | grep -v 'Virtual Function' |awk '/%s/ {print $1}'" % self.pf_filter_re
         PF_devices = [i for i in process.system_output(
@@ -1349,7 +1340,7 @@ class PciAssignable(object):
         # Without this sleep ifconfig would fail With below error, since not all
         # probed interfaces of VF would have got properly renamed:
         # CmdError: Command 'ifconfig -a' failed (rc=1)
-        if int(self.driver_option) > 10:
+        if int(vf_no) > 10:
             time.sleep(60)
         return True
 
@@ -1365,14 +1356,12 @@ class PciAssignable(object):
         if driver and driver != "mlx5_core":
             cmd = "modprobe -r %s" % driver
         if ARCH == 'ppc64le' and driver == 'mlx5_core':
-            pf_devices = self.get_pf_mlx()
-            logging.info("Mellanox PF devices '%s'", pf_devices)
-            cmd = "rmmod mlx5_ib;modprobe -r mlx5_core;modprobe mlx5_ib"
-        elif ARCH == 'x86_64':
             pf_devices = self.get_pf_ids()
-        for PF in pf_devices:
-            if not self.set_vf(PF):
-                return False
+            logging.info("Mellanox PF devices '%s'", pf_devices)
+            for PF in pf_devices:
+                if not self.set_vf(PF):
+                    return False
+            cmd = "rmmod mlx5_ib;modprobe -r mlx5_core;modprobe mlx5_ib"
         if process.system(cmd, ignore_status=True, shell=True):
             logging.debug("Failed to remove driver: %s", driver)
             return False
@@ -1458,26 +1447,24 @@ class PciAssignable(object):
         if process.system("lsmod | grep -w %s" % self.driver, ignore_status=True,
                           shell=True):
             re_probe = True
-        # If driver is available then set VFs for ppc64le
-        else:
-            if ARCH == 'ppc64le' and self.driver == 'mlx5_core':
-                pf_devices = self.get_pf_mlx()
-            elif ARCH == 'x86_64':
-                pf_devices = self.get_pf_ids()
+        # If driver is available and pa_type is vf then set VFs
+        elif self.pa_type == 'vf':
+            pf_devices = self.get_pf_ids()
             for PF in pf_devices:
                 if not self.set_vf(PF, self.driver_option):
                     re_probe = True
             if not self.check_vfs_count():
                 re_probe = True
-            if not re_probe:
-                self.setup = None
-                return True
+        if not re_probe:
+            self.setup = None
+            return True
 
         # Re-probe driver with proper number of VFs once more and raise
         # exception
         if re_probe:
             if not self.remove_driver() or not self.modprobe_driver():
                 return False
+            pf_devices = self.get_pf_ids()
             if ARCH == 'ppc64le' and self.driver == 'mlx5_core':
                 set_ip = 0
                 if (self.static_ip):
@@ -1488,8 +1475,9 @@ class PciAssignable(object):
                     set_ip = 1
                 for PF in pf_devices:
                     if set_ip:
+                        ifname = utils_misc.get_interface_from_pci_id(PF)
                         ip_assign = "ifconfig %s %s netmask %s up" % (
-                            PF, ip_addr, self.net_mask)
+                            ifname, ip_addr, self.net_mask)
                         logging.info("assign IP to PF device %s : %s", PF,
                                      ip_assign)
                         if process.system(ip_assign, shell=True,
@@ -1497,16 +1485,13 @@ class PciAssignable(object):
                             return False
                         ip_addr = ip_addr + 1
                     logging.info("PF device '%s'.", PF)
-                    if not self.set_vf(PF, self.driver_option):
-                        return False
-            elif ARCH == 'x86_64':
-                pf_devices = self.get_pf_ids()
+            if self.pa_type == 'vf':
                 for PF in pf_devices:
                     if not self.set_vf(PF, self.driver_option):
                         return False
-            if not self.check_vfs_count():
-                # Even after re-probe there are no VFs created
-                return False
+                if not self.check_vfs_count():
+                    # Even after re-probe there are no VFs created
+                    return False
             dmesg = process.system_output("dmesg", timeout=60,
                                           ignore_status=True,
                                           verbose=False)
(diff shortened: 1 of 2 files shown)


virttest/test_setup.py
```diff
@@ -1365,14 +1367,7 @@ def remove_driver(self, driver=None):
...
             cmd = "modprobe -r %s" % driver
         if ARCH == 'ppc64le' and driver == 'mlx5_core':
-            pf_devices = self.get_pf_mlx()
-            logging.info("Mellanox PF devices '%s'", pf_devices)
             cmd = "rmmod mlx5_ib;modprobe -r mlx5_core;modprobe mlx5_ib"
-        elif ARCH == 'x86_64':
-            pf_devices = self.get_pf_ids()
-        for PF in pf_devices:
-            if not self.set_vf(PF):
-                return False
```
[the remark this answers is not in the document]
huangyum: Why do we need to disable vf first ? Is it a specific need for ppc+mlx5_core ?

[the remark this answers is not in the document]
yanan-fu: This is the old design,  keep it is recommended, as we don't have this card.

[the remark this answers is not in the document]
yanan-fu: @huangyum  i have test ppc + mlx5_core,  disable vf before remove pf is must for this condition.  Otherwise, vf can not be disabled when unload and load nic driver again, and it will caused generate vf failed after this.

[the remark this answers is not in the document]
huangyum: Thanks, updated. Please help review again.

virttest/test_setup.py
```diff
@@ -1367,12 +1369,10 @@ def remove_driver(self, driver=None):
         if ARCH == 'ppc64le' and driver == 'mlx5_core':
             pf_devices = self.get_pf_mlx()
             logging.info("Mellanox PF devices '%s'", pf_devices)
+            for PF in pf_devices:
+                if not self.set_vf(PF):
```
yanan-fu: with  `pf_devices = self.get_pf_mlx()` , `pf_devices` is the interface name of the nic, such as "em1", it can not be used in `set_vf(PF)` .
In short, `get_pf_mlx()` return nic interface, `get_pf_ids`  return full device id. they are different .

virttest/test_setup.py
```diff
@@ -1490,33 +1485,27 @@ def sr_iov_setup(self):
...
+            if (self.static_ip):
+                if (not self.start_addr_PF) or (not self.net_mask):
+                    raise exceptions.TestSetupFail(
+                        "No IP / netmask found, please populate starting IP address for PF devices in configuration file")
+                ip_addr = utils_net.IPAddress(self.start_addr_PF)
+                set_ip = 1
+            for PF in pf_devices:
+                if set_ip:
+                    ip_assign = "ifconfig %s %s netmask %s up" % (
+                        PF, ip_addr, self.net_mask)
```
yanan-fu:
```
     pf_devices = self.get_pf_mlx()
elif ARCH == 'x86_64':
    pf_devices = self.get_pf_ids()
```
same with the comment above, pf_device is different for  get_pf_mlx() and get_pf_ids().
for example,  we can not do "ifconfig 0000:04:00.0".

virttest/test_setup.py
huangyum: I didn't test it on ppc host and I just kept it as same as before. If so, the former code couldn't work at all. Anyway, I will update. Thanks for your review.

yanan-fu: "set_vf" was changed and get_pf_ids was added  in commit af59aa5797c8c9fcb569914dfeffd17c872242bd. I think you can refer the code before this commit.

huangyum: Thanks, I updated set_vf() as it was before. It should work now.

virttest/test_setup.py
```diff
@@ -1349,7 +1351,7 @@ def set_vf(self, pci_pf, vf_no="0"):
         # Without this sleep ifconfig would fail With below error, since not all
         # probed interfaces of VF would have got properly renamed:
         # CmdError: Command 'ifconfig -a' failed (rc=1)
-        if int(self.driver_option) > 10:
+        if int(vf_no) > 10:
```
yanan-fu: Good !

virttest/test_setup.py
```diff
@@ -1528,6 +1520,7 @@ def sr_iov_setup(self):
             utils_misc.log_line(file_name, dmesg)
             self.setup = None
             return True
+        return True
```
yanan-fu: No need, should be removed.

virttest/test_setup.py
```diff
@@ -1490,33 +1487,28 @@ def sr_iov_setup(self):
...
-                for PF in pf_devices:
-                    if not self.set_vf(PF, self.driver_option):
+            set_ip = 0
+            if (self.static_ip):
+                if (not self.start_addr_PF) or (not self.net_mask):
+                    raise exceptions.TestSetupFail(
+                        "No IP / netmask found, please populate starting IP address for PF devices in configuration file")
+                ip_addr = utils_net.IPAddress(self.start_addr_PF)
+                set_ip = 1
+            pf_devices = self.get_pf_mlx()
```
yanan-fu: `pf_devices`  be used not only once in this function, such as L1475,  suggest to move it out of if-else.

virttest/test_setup.py
```diff
@@ -1286,7 +1288,7 @@ def get_same_group_devs(self, pci_id):
 
     def get_pf_mlx(self):
         """
-        Get the PF devices for mlx5_core driver
+        Get the ifname of PF devices
```
yanan-fu: In this PR, this function was not used for mlx5_core driver only, so how about change the function name too ?  and in fact, i prefer to delete this function,  its function is repeated with get_pf_ids partly.

virttest/test_setup.py
```diff
@@ -1339,7 +1341,7 @@ def set_vf(self, pci_pf, vf_no="0"):
         :params vf_no: Number of Vfs to be virtualized
         :return: True on success, False on failure
         """
-        cmd = "echo %s > /sys/bus/pci/devices/%s/sriov_numvfs" % (vf_no, pci_pf)
+        cmd = "echo %s > /sys/class/net/%s/device/sriov_numvfs" % (vf_no, pci_pf)
```
yanan-fu: yes,  this updated set_vf() can work with get_pf_mlx().  i have checked the whole file , get_pf_mlx() only used for get pf_devices when execute set_vf().  So I  prefer to use get_pf_ids()  instead of  get_pf_mlx(),  no need to change set_vf().

virttest/test_setup.py
huangyum: Good idea to update the function name. Actually, this function is useful when setting ip for the interface(L1500).

virttest/test_setup.py
huangyum: We use get_pf_mlx() to get pf_devices not only when set_vf  but also when set ip for pf. You can refer to L1500, which sets ip for the interface. With updated set_vf, get_pf_mlx can work well for both.

virttest/test_setup.py
huangyum: No, it can not be moved out. If the driver is not installed, it will return wrong value. It must be placed after  L1488.

virttest/test_setup.py
huangyum: Actually, it is needed.  It will be executed when re_probe is False.

virttest/test_setup.py
yanan-fu: Recommended to get interface name with "ls /sys/bus/pci/devices/0000\:04\:00.0/net/". refer the comment  for set_vf()

virttest/test_setup.py
yanan-fu: Yes, you are right !

yanan-fu:
please update commit and  PR title, add "virttest.test_setup" as prefix.

virttest/test_setup.py
huangyum: Yes, you are right. I will update.

virttest/test_setup.py
huangyum: Thanks :)

virttest/test_setup.py
huangyum: Yes, it can avoid the error. I will update, thanks!

huangyum: Yes, it can avoid the error if lshw is not installed. But it will call get_pf_ids() to get the id first, which would be less efficient . I'm not sure which way is better.

yanan-fu: I think , "lshw" is not a package avocado depend on, it is only used for sr-iov test.  From my understanding, avocado dependency is something to support the framework.  So, i prefer to change another method to get the nic interface.
@vivianQizhu   what do you think about this problem ?

vivianQizhu: @yanan-fu Agree with you, I don't prefer to add a not very necessary dependency if have other soft way work around.

huangyum: Thanks @yanan-fu @vivianQizhu , updated, please help review again.

virttest/test_setup.py
```diff
@@ -1470,53 +1470,46 @@ def sr_iov_setup(self):
...
-            elif ARCH == 'x86_64':
-                pf_devices = self.get_pf_ids()
-                for PF in pf_devices:
-                    if not self.set_vf(PF, self.driver_option):
+            set_ip = 0
+            if (self.static_ip):
+                if (not self.start_addr_PF) or (not self.net_mask):
+                    raise exceptions.TestSetupFail(
+                        "No IP / netmask found, please populate starting IP address for PF devices in configuration file")
+                ip_addr = utils_net.IPAddress(self.start_addr_PF)
```
yanan-fu: @huangyum  could you explain this line ?

huangyum: I just kept it as it was since I saw other pr (https://github.com/avocado-framework/avocado-vt/pull/1139) handling this part. Thanks!

yanan-fu: @huangyum  i can not understand it well, evenif you follow the design from other PR, i think it is needed to make it clearly.  thx!

huangyum: OK. I keep it the original way for now. Please help review again, thanks!

yanan-fu:
LGTM
@huangyum  could you add the test result about the new change?

huangyum:
@yanan-fu Thanks for your review.  I ran pf_assignable cases but failed with "VMIPAddressMissingError: No ipv4 DHCP lease". I checked the log and lspci info on host, no vf generated and guest boot with pf successfully. So I think the patch is ok. What do you think?

yanan-fu:
@huangyum  thx, check the test log too, it is ok for me,  two small issue that are not related with this fix can be improved. i think you can update it at the same time.
1. It is not easy to read debug.log when print all the host dmesg info before exeute sr_iov_setup:
```
1428    o = process.system_output("dmesg")
```
You can refer L1521~1524, save the dmesg log to a separate file i think.

2. In modprobe_driver():
```
1390   msg = "Loading the driver '%s' without option" % driver
```
I think it should be put before 1394, to avoid "Loading the driver 'None' without option"  in debug log.

huangyum:
@yanan-fu  This patch aims to fix the issue to not set vf when run pf cases.  As for the log issue you mentioned, I will send another patch to fix it. Thanks :)

yanan-fu:
@huangyum  it is ok for me to send another patch for the log issue.  And, please remember to update the code that call `PciAssignable()` , given right `pa_type`.

LGTM

huangyum:
@yanan-fu Sure, I'm handling this right now. Thanks!

vivianQizhu:
Code looks good to me, but pls make sure your commit title under 70 characters, you may add description if title can not conclude all things. And suggest describe like "Update class PciAssignable to..."

huangyum:
@vivianQizhu  Thanks for your review.  Updated, thanks!

huangyum:
@luckyh Could you please help review and merge this pr? Thanks!

virttest/test_setup.py
```diff
@@ -1284,20 +1286,20 @@ def get_same_group_devs(self, pci_id):
             pci_ids = out.split()
         return pci_ids
 
-    def get_pf_mlx(self):
+    def get_pf_ifname(self, pci_pf):
```
balamuruhans: we can remove this function, we can use `get_interface_from_pci_id()` from utils_misc.

virttest/test_setup.py
```diff
@@ -1284,20 +1286,20 @@ def get_same_group_devs(self, pci_id):
...
-        return (PF_devices)
+        # This function returns the interface name associated with the pf
+        cmd = "ls /sys/bus/pci/devices/%s/net/" % pci_pf
+        ifname = process.system_output(cmd, shell=True, timeout=60)
+        return ifname
 
     def get_pf_ids(self):
         """
-        Get the PF devices on x86 host
+        Get the id of PF devices
```
balamuruhans: even this function can be removed with `get_pci_id_using_filter()` from utils_misc, these will make your changes more good.

huangyum: I'm afraid it can't be replaced with `get_pci_id_using_filter()`.  `get_pci_id_using_filter()` can only filter the device by a filter regex. We need to exclude 'Virtual Function' here by `grep -v` as L1304. Thanks!

virttest/test_setup.py
huangyum: Thanks, updated, please help review again.

balamuruhans:
Thanks @yanan-fu @vivianQizhu @santwana for reviews.
Looks good to me, let us merge this.

PR #1161, opened 2017-09-15
iscsi: Handle create_iSCSI() to install iscsi target and initiator packages

balamuruhans:
Tests gets interrupted if iscsi target and initiator packages are not available,
handle it to install the package and proceed.

### DIFF
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -15,10 +15,12 @@ from avocado.core import exceptions
 from avocado.utils import data_factory
 from avocado.utils import process
 from avocado.utils import path
+from avocado.utils import distro
 
 from . import utils_selinux
 from . import utils_net
 from . import data_dir
+from . import utils_package
 from .staging import service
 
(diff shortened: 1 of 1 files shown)


virttest/iscsi.py
```diff
@@ -824,6 +826,17 @@ class Iscsi(object):
     def create_iSCSI(params, root_dir=data_dir.get_tmp_dir()):
         iscsi_instance = None
         err_msg = "Please install package(s): %s"
+        ubuntu = distro.detect().name == 'Ubuntu'
```
prudmirya: try to clear the configuration before using iscsi if need(not sure)

balamuruhans: @prudmirya Thanks! agreed it is good to consider clearing the configuration before using iscsi, but please fix it from PR #1042, This patch will handle the pre-requisite of having the packages installed to use iscsi.

balamuruhans:
Thanks @prudmirya, let us merge this PR.

PR #1171, opened 2017-09-25
Empty vm.spice_options before its another population

rdkdd:
This is an update of previous PR https://github.com/avocado-framework/avocado-vt/pull/1144.

In `vm.spice_options` are preserved outdated `params` from previous testcase. This is 
because `vm` instance is not recreated when another test run from job suite is performed.

example:
TC 1 `params`:
> spice_port = no
> spice_port_closed = no
> spice_seamless_migration = on
> spice_secure_channels = 
> spice_ssl = yes
> spice_tls_ciphers = DEFAULT
> spice_tls_port = blabla

qemu-kvm spice cmd:

    -spice disable-ticketing,tls-port=blabla,x509-dir=/tmp/spice_x509d,seamless-migration=on

`params` for following TC_2 in the same job:

> spice_addr = 127.0.0.1
> spice_port = generate
> spice_port_closed = no
> spice_seamless_migration = on

(notice `spice_tls_port = blabla` is not in `params`)

qemu-kvm spice cmd:

    -spice port=3000,disable-ticketing,addr=127.0.0.1,tls-port=blabla,x509-dir=/tmp/spice_x509d,seamless-migration=on

`tls-port=blabla` makes this 2. TC to fail.

- To correct this behaviour, instead of `self.spice_options` was used local variable `spice_options`, which is recreated on `vm.make_create_command()` and assigned to `vm` instance only by `vm.create()` command.

- Another problem presents dynamically generated `spice_port` value (`spice_*port = generate` parameter must be set) while `keep_guest_running == True` option. After first test case finishes, `vm` is not restarted, but newly created `vm.spice*_port` reports different spice port, then virtual machine is actually listening on. For following test case - this might cause fake test FAIL. To workaround this problem, `vm.is_alive()` test was removed from `if optget("spice*_port") == "generate"` condition, so that `vm.needs_restart(..)` always returns `True`. This recreates vm instance with proper `spice*_port`. As a trade-off we loose functionality of `keep_guest_running` option for SPICE tests while `spice_port == "generate"`. Nevertheless we, as SPICE QE team, do not use this option. This way we prevent to create unrestarted VM with changed parameters. 

- Final thing is to drop `vm.spice*_port` option in favour of `vm.spice_options["spice*_port"]`. There is no reason to store this value twice now. Thus `vm.spice-*port` temporary properties were created.

Deeper analysis can be found here: https://www.redhat.com/archives/avocado-devel/2017-February/msg00010.html

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -951,7 +951,6 @@ class VM(virt_vm.BaseVM):
             if devices.has_option("spice-help"):
                 spice_help = commands.getoutput("%s -device \\?" % qemu_binary)
             s_port = str(utils_misc.find_free_port(*port_range))
-            self.spice_options['spice_port'] = s_port
             cmd += " port=%s" % s_port
             for param in spice_params.split():
                 value = params.get(param)
@@ -968,12 +967,13 @@ class VM(virt_vm.BaseVM):
             if devices.has_option("qxl"):
                 qxl_dev_nr = params.get("qxl_dev_nr", 1)
                 cmd += " -qxl %s" % qxl_dev_nr
-            return cmd
+            return cmd, s_port
 
-        def add_spice(port_range=(3000, 3199),
+        def add_spice(spice_options, port_range=(3000, 3199),
                       tls_port_range=(3200, 3399)):
             """
             processes spice parameters
+            :param spice_options - dict with spice keys/values
             :param port_range - tuple with port range, default: (3000, 3199)
             :param tls_port_range - tuple with tls port range,
                                     default: (3200, 3399)
@@ -983,7 +983,7 @@ class VM(virt_vm.BaseVM):
 
             def optget(opt):
                 """a helper function"""
-                return self.spice_options.get(opt)
+                return spice_options.get(opt)
 
             def set_yes_no_value(key, yes_value=None, no_value=None):
                 """just a helper function"""
@@ -1001,15 +1001,12 @@ class VM(virt_vm.BaseVM):
                     spice_opts.append(opt_string % tmp)
                 elif fallback:
                     spice_opts.append(fallback)
-            s_port = str(utils_misc.find_free_port(*port_range))
             if optget("spice_port") == "generate":
-                if not self.is_alive():
-                    self.spice_options['spice_port'] = s_port
-                    spice_opts.append("port=%s" % s_port)
-                    self.spice_port = s_port
-                else:
-                    self.spice_options['spice_port'] = self.spice_port
-                    spice_opts.append("port=%s" % self.spice_port)
+                # FIXME: This makes the "needs_restart" to always re-create the
+                # machine.
+                s_port = str(utils_misc.find_free_port(*port_range))
+                spice_options['spice_port'] = s_port
+                spice_opts.append("port=%s" % s_port)
             # spice_port = no: spice_port value is not present on qemu cmdline
             elif optget("spice_port") != "no":
                 set_value("port=%s", "spice_port")
@@ -1018,24 +1015,20 @@ class VM(virt_vm.BaseVM):
             ip_ver = optget("listening_addr")
             if ip_ver:
                 host_ip = utils_net.get_host_ip_address(self.params, ip_ver)
-                self.spice_options['spice_addr'] = host_ip
+                spice_options['spice_addr'] = host_ip
             set_yes_no_value(
                 "disable_copy_paste", yes_value="disable-copy-paste")
             set_value("addr=%s", "spice_addr")
 
             if optget("spice_ssl") == "yes":
                 # SSL only part
-                t_port = str(utils_misc.find_free_port(*tls_port_range))
                 if optget("spice_tls_port") == "generate":
-                    if not self.is_alive():
-                        self.spice_options['spice_tls_port'] = t_port
-                        spice_opts.append("tls-port=%s" % t_port)
-                        self.spice_tls_port = t_port
-                    else:
-                        self.spice_options[
-                            'spice_tls_port'] = self.spice_tls_port
-                        spice_opts.append("tls-port=%s" % self.spice_tls_port)
-                else:
+                    t_port = str(utils_misc.find_free_port(*tls_port_range))
+                    spice_options['spice_tls_port'] = t_port
+                    spice_opts.append("tls-port=%s" % t_port)
+                # spice_tls_port = no: spice_port value is not present on qemu
+                # cmdline
+                elif optget("spice_tls_port") != "no":
                     set_value("tls-port=%s", "spice_tls_port")
 
                 prefix = optget("spice_x509_prefix")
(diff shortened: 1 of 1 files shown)


rdkdd:
@ldoktor could you please check this PR?
Thank you!

virttest/virt_vm.py
```diff
@@ -577,6 +577,9 @@ def needs_restart(self, name, params, basedir):
             return True
 
         try:
+            # If `params.get("spice*_port") == generate` and
+            # `keep_guest_running == True`, `need_restart` will
+            # always be True to prevent assigning of undesirable SPICE port.
```
ldoktor: The `virt_vm` is generic implementation and this is a specific information about `qemu_vm` and only when using `spice` vga. Would you please move this note to the place where the `spice_*port` is being set? Also I'd suggest the `FIXME: ` prefix.

virttest/qemu_vm.py
```diff
@@ -939,7 +939,7 @@ def add_spice_rhel5(devices, spice_params, port_range=(3100, 3199)):
             """
             processes spice parameters on rhel5 host.
 
-            :param spice_options - dict with spice keys/values
+            :param spice_params - dict with spice keys/values
```
ldoktor: nice touch, ideally should be in separate commit

virttest/virt_vm.py
ldoktor: Also please include that in commit message (and please be more verbose in the commit message)

virttest/qemu_vm.py
```diff
@@ -2711,8 +2702,10 @@ def create(self, name=None, params=None, root_dir=None,
                 self.update_system_dependent_devs()
             # Make qemu command
             try:
-                self.devices = self.make_create_command()
+                self.devices, self.spice_options = self.make_create_command()
                 self.update_vga_global_default(params, migration_mode)
+                self.spice_port = self.spice_options.get("spice_port")
+                self.spice_tls_port = self.spice_options.get("spice_tls_port")
```
ldoktor: Nice, how about this as the next commit (`git am` should get it):

```patch
From 0783e3acc741db8fe4e91fc071a85a6d43de5b4c Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Luk=C3=A1=C5=A1=20Doktor?= <ldoktor@redhat.com>
Date: Mon, 25 Sep 2017 16:15:31 +0200
Subject: [PATCH] virttest.qemu_vm: Deprecate `VM.spice_*port` variables
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

The `VM.spice_port` and `VM.spice_tls_port` are a convenience variables
which unnecessarily pollute the VM namespace as the values are already
available in `VM.spice_options`.

Let's remove them completely at the end of the year 2017 and until then
show a deprecation message.

---
 virttest/qemu_vm.py | 16 ++++++++++++++--
 1 file changed, 14 insertions(+), 2 deletions(-)

index 009d0dc..89066dc 100644
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -2461,6 +2461,20 @@ class VM(virt_vm.BaseVM):
                 qdev = qdevices.QGlobal(vga_type, pro, val)
                 self.devices.insert(qdev)
 
+    @property
+    def spice_port(self):
+        logging.warning("'VM.spice_port' will be removed by the end of "
+                        "the year 2017, please use 'self.spice_options."
+                        "get(\"spice_port\")' instead")
+        return self.spice_options.get("spice_port")
+
+    @property
+    def spice_tls_port(self):
+        logging.warning("'VM.spice_tls_port' will be removed by the end of "
+                        "the year 2017, please use 'self.spice_options."
+                        "get(\"spice_tls_port\")' instead")
+        return self.spice_options.get("spice_tls_port")
+
     @error_context.context_aware
     def create(self, name=None, params=None, root_dir=None,
                timeout=120, migration_mode=None,
@@ -2704,8 +2718,6 @@ class VM(virt_vm.BaseVM):
             try:
-                self.spice_port = self.spice_options.get("spice_port")
-                self.spice_tls_port = self.spice_options.get("spice_tls_port")
                 logging.debug(self.devices.str_short())
                 logging.debug(self.devices.str_bus_short())
                 qemu_command = self.devices.cmdline()
-- 
2.9.5

```

virttest/qemu_vm.py
```diff
@@ -1001,15 +1000,10 @@ def set_value(opt_string, key, fallback=None):
...
-            s_port = str(utils_misc.find_free_port(*port_range))
             if optget("spice_port") == "generate":
-                if not self.is_alive():
-                    self.spice_options['spice_port'] = s_port
-                    spice_opts.append("port=%s" % s_port)
-                    self.spice_port = s_port
-                else:
-                    self.spice_options['spice_port'] = self.spice_port
-                    spice_opts.append("port=%s" % self.spice_port)
+                s_port = str(utils_misc.find_free_port(*port_range))
```
ldoktor: This is where I'd expect the `FIXME: This makes the "needs_restart" to always re-create the machine...` message.

virttest/qemu_vm.py
```diff
@@ -2056,9 +2046,12 @@ def sort_key(dev):
         elif params.get("display") == "nographic":
             cmd += add_nographic()
         elif params.get("display") == "spice":
+            spice_options = {}
             if params.get("rhel5_spice"):
                 spice_params = params.get("spice_params")
-                cmd += add_spice_rhel5(devices, spice_params)
+                tmp_cmd, spice_options["spice_port"] =\
```
ldoktor: In style guide we encourage the use of brackets to escaped `\n`s. How about `ret = add_spice_rhel5(...); tmp_cmd, spice_options... = ret`?

virttest/qemu_vm.py
```diff
@@ -2075,15 +2068,13 @@ def sort_key(dev):
                     "disable_copy_paste", "spice_seamless_migration",
                     "listening_addr"
                 )
-
```
ldoktor: also style fix, please move it to a separate commit along with the docstring fix (or remove it)

rdkdd:
Hi @ldoktor, please check this PR and possibly merge.

Thanks!

virttest/qemu_vm.py
```diff
@@ -2288,7 +2281,7 @@ def sort_key(dev):
                 if set_disable_modern == "yes":
                     add_disable_modern(devices, device, dev_type)
 
-        return devices
+        return devices, spice_options
```
ldoktor: Actually the `spice_options` might not be defined so please initialize it on top and not in if-branch.

rdkdd: yes you are right. I overlooked it. I will correct it.

rdkdd:
Hello @ldoktor and @clebergnu .
I have update those commits according to requests. I have also improved a doc string for `spice-params` (https://github.com/avocado-framework/avocado-vt/pull/1171/files#diff-3f2317089f10636f453dca6c4f880f6cR942). Hope it is fine this time. And I am sorry for overlooking this `spice_options` initialization. So please can you review once more?

ldoktor:
[Fix the `VM.spice_options` usage to allow `VM.create` with different `params`.](https://trello.com/c/EkWc1DC0/910-fix-the-vmspiceoptions-usage-to-allow-vmcreate-with-different-params)

PR #1175, opened 2017-09-27
virttest.test_setup: Improve log info for sr_iov test

huangyum:
1. Remove dmesg info in log
2. Remove unused option when modprobe driver

ID: 1496365

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -1379,33 +1379,21 @@ class PciAssignable(object):
         return True
 
     @error_context.context_aware
-    def modprobe_driver(self, driver=None, option=None):
+    def modprobe_driver(self, driver=None):
         """
         Method to modprobe driver
 
         :param driver: driver name
-        :param option: modprobe options to driver
         :return: True on success, False on failure
         """
-        msg = "Loading the driver '%s' without option" % driver
-        error_context.context(msg, logging.info)
         if not driver:
             driver = self.driver
+        msg = "Loading the driver '%s'" % driver
+        error_context.context(msg, logging.info)
         cmd = "modprobe %s" % driver
-        if not option:
-            option = self.driver_option
-        if driver and driver != "mlx5_core" and ARCH != 'x86_64':
-            if option:
-                cmd += " %s" % option
         if process.system(cmd, ignore_status=True, shell=True):
             logging.debug("Failed to modprobe driver: %s", driver)
             return False
-        if ARCH == 'ppc64le' and self.driver == 'mlx5_core':
-            pf_devices = self.get_pf_mlx()
-            if (self.get_vfs_count() > 0):
-                for PF in pf_devices:
-                    if not self.set_vf(PF, option):
-                        return False
         return True
 
     @error_context.context_aware
@@ -1425,8 +1413,8 @@ class PciAssignable(object):
                               logging.info)
         if ARCH != 'ppc64le':
             kvm_re_probe = True
-            o = process.system_output("dmesg")
-            ecap = re.findall("ecap\s+(.\w+)", o)
+            dmesg = process.system_output("dmesg", verbose=False)
+            ecap = re.findall("ecap\s+(.\w+)", dmesg)
             if not ecap:
                 logging.error("Fail to check host interrupt remapping support")
             else:


jingzhao84:
@huangyum double checked the code style since checked failed.

huangyum:
@jingzhao84 Thanks, updated!

virttest/test_setup.py
```diff
@@ -1379,32 +1379,26 @@ def remove_driver(self, driver=None):
...
-                cmd += " %s" % option
         if process.system(cmd, ignore_status=True, shell=True):
             logging.debug("Failed to modprobe driver: %s", driver)
             return False
         if ARCH == 'ppc64le' and self.driver == 'mlx5_core':
             pf_devices = self.get_pf_mlx()
             if (self.get_vfs_count() > 0):
                 for PF in pf_devices:
-                    if not self.set_vf(PF, option):
+                    if not self.set_vf(PF, self.driver_option):
```
yanan-fu: We can not  `set_vf ` with non-zero value (such as `self.driver_option` here) if  `get_vfs_count `  is not zero.

yanan-fu:
About commit message,  i prefer to add more explanations for deleting the parameter "option".   
1. Remove dmesg info in debug.log
2. Remove unused parameter "option" in function modprobe_driver. As we don't generate VFs when modprobe driver, we generate VFs with function `set_vf`.
Just a suggestion !

virttest/test_setup.py
huangyum: I agree. How about we remove L1403~1408 directly? IMO, it should not be here in the modprobe_driver().  set_vf will be executed later in L1477 anyway.

yanan-fu: Yes, it will be executed later, agree to move it here.

huangyum:
@yanan-fu Updated, please help review again, thanks!

virttest/test_setup.py
```diff
@@ -1379,33 +1379,21 @@ def remove_driver(self, driver=None):
...
         if process.system(cmd, ignore_status=True, shell=True):
             logging.debug("Failed to modprobe driver: %s", driver)
             return False
-        if ARCH == 'ppc64le' and self.driver == 'mlx5_core':
-            pf_devices = self.get_pf_mlx()
-            if (self.get_vfs_count() > 0):
-                for PF in pf_devices:
-                    if not self.set_vf(PF, option):
-                        return False
         return True
```
yanan-fu: As L1403-1408  be removed,  how about change `return True`  to  
```
else:
    return True
```

huangyum: I'd prefer to keep 'return True' only and 'else' is not necessary here. It will return when modprobe driver successfully in L1394.

yanan-fu: Same,  it is ok for me too.

virttest/test_setup.py
```diff
@@ -1425,7 +1413,7 @@ def sr_iov_setup(self):
                               logging.info)
         if ARCH != 'ppc64le':
             kvm_re_probe = True
-            o = process.system_output("dmesg")
+            o = process.system_output("dmesg", verbose=False)
```
balamuruhans: commit msg says, "Remove dmesg info in log" but only verbose is turned off, lets have more clear msg.

huangyum: Thanks for your review. Updated. How about now ?

virttest/test_setup.py
```diff
@@ -1379,33 +1379,21 @@ def remove_driver(self, driver=None):
...
 
         :param driver: driver name
-        :param option: modprobe options to driver
         :return: True on success, False on failure
         """
-        msg = "Loading the driver '%s' without option" % driver
-        error_context.context(msg, logging.info)
         if not driver:
             driver = self.driver
+        msg = "Loading the driver '%s' without option" % driver
```
yanan-fu: `Loading driver '%s' ` is ok, no need to add `without option`.

virttest/test_setup.py
huangyum: @balamuruhans  Could you please help review again, thanks!

balamuruhans:
let us address @yanan-fu's review comments and merge this.

virttest/test_setup.py
balamuruhans: please use proper variable name instead of `o`

huangyum:
@yanan-fu @balamuruhans Thanks, updated!

yanan-fu:
LGTM

balamuruhans:
@huangyum @yanan-fu Thanks.

PR #1178, opened 2017-09-30
Fix postprocess bug

hsmj1412:
When vm starts and then pauses quickly, it may cause some setup problems

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -909,7 +909,7 @@ def postprocess(test, params, env):
     if params.get("verify_guest_dmesg", "yes") == "yes" and params.get("start_vm", "no") == "yes":
         guest_dmesg_log_file = params.get("guest_dmesg_logfile", "guest_dmesg.log")
         guest_dmesg_log_file = utils_misc.get_path(test.debugdir, guest_dmesg_log_file)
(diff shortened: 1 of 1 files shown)


virttest/env_process.py
```diff
@@ -911,6 +911,8 @@ def postprocess(test, params, env):
         guest_dmesg_log_file = utils_misc.get_path(test.debugdir, guest_dmesg_log_file)
         living_vms = [vm for vm in env.get_all_vms() if vm.is_alive()]
         for vm in living_vms:
+            if vm.is_paused():
+                vm.resume()
```
sathnaga: Ideally test case should restore the `vm` states after the test is done, we expect the VM in running state, even if not running, `waitforlogin` method will error out.
My opinion not to handle here, otherwise we might have to handle the corner cases of other states like `in shutdown` etc, 
Can you point me the failures log?

sathnaga: @hsmj1412 Just gone through the logs, we need further refining of `only running` vms, to be used,
it is good  not to change the vm states after test is over.

sathnaga: @hsmj1412 tried addressing this way https://github.com/avocado-framework/avocado-vt/pull/1181, can you try and let me know your results and comments, Thanks in advance.

PR #1181, opened 2017-10-06
Fixup: verify_dmesg: Handle 'not running' guests

sathnaga:
This patch handles dmesg check for live, not running
vms by simply skipping it, as during postprocess it
is not good to manipulate or change the vm state.

Reported-by: Junxiang Li <junli@redhat.com>

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -889,13 +889,17 @@ class BaseVM(object):
         """
         Verify guest dmesg
         """
-        session = self.wait_for_login()
-        level = self.params.get("guest_dmesg_level", 3)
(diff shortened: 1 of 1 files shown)


hsmj1412:
Sorry, I do not think this patch is the best one to fix up. wait_for_login will hang in wait_for_get_address, and this function's timeout is 600 seconds. So, the so long time could not be accepted.
There are 2 different ways to fix it.
1. Use if vm.is_alive() and not vm.is_paused()
2. Use wait_for_login before the vm pause and after it start.(I'm not sure it will work well, and I'm not sure paused vm if to do this step or not)

hsmj1412:
Updated #1178

sathnaga:
@hsmj1412 Sure,` 1. `makes sense

PR #1185, opened 2017-10-09
libvirt_vm:rule out rescue kernel when set debug kernel to default

chloerh:
Add condition to rule out rescue kernel in case to be set as
default kernel when it shoud be debug kernel to be set as default.


balamuruhans:
Thanks @chloerh, let us merge this.

PR #1187, opened 2017-10-11
libvirt: initialize param selinux_status_bak to avoid key error

balamuruhans:
param selinux_status_bak is not initialized for ubuntu and breaks with keyerror
in set_vm_disk() where this param is retrieved as res['selinux_status_bak']


balamuruhans:
This is a trivial fix, so let's merge it.

PR #1158, opened 2017-09-14
Add method to get numa nodes with memory

sathnaga:
Add method to get numa nodes with memory

Signed-off-by: Satheesh Rajendran <sathnaga@linux.vnet.ibm.com>

PR #1194, opened 2017-10-16
Add method to get cells with memory

hsmj1412:
Add method to get cells with memory

### DIFF
--- a/virttest/libvirt_xml/capability_xml.py
+++ b/virttest/libvirt_xml/capability_xml.py
@@ -255,12 +255,14 @@ class TopologyXML(base.LibvirtXMLBase):
         self.xmltreefile.reroot("/host/topology")
         self.xmltreefile.write()
 
-    def get_cell(self):
+    def get_cell(self, withmem=False):
         """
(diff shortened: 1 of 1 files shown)


hsmj1412:
Adapt to PR #1158 
@sathnaga

virttest/libvirt_xml/capability_xml.py
```diff
@@ -268,6 +268,21 @@ def get_cell(self):
             cell_list.append(new_cell)
         return cell_list
 
+    def get_cell_with_memory(self):
```
chunfuwen: suggest use old get_cell_with_memory(self) and add one parameter with default value to diff them rather than create new method

hsmj1412: Where is the old one ? Do you mean get_cell()?

chunfuwen: yes, reuse get_cell()

hsmj1412: Done

PR #1197, opened 2017-10-20
Add hostdev network xml implementation

yafu-1:
Add hostdev network xml implementation.

Signed-off-by: Yan Fu <yafu@redhat.com>

### DIFF
--- a/virttest/libvirt_xml/network_xml.py
+++ b/virttest/libvirt_xml/network_xml.py
@@ -7,6 +7,7 @@ import logging
 
 from virttest import xml_utils
 from virttest.libvirt_xml import base, xcepts, accessors
+from virttest.libvirt_xml.devices import librarian
 
 
 class RangeList(list):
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/network_xml.py
```diff
@@ -414,6 +422,50 @@ def __init__(self, virsh_instance=base.virsh):
                                tag_name='virtualport', attribute='type')
         super(NetworkXMLBase, self).__init__(virsh_instance=virsh_instance)
 
+    class Address(base.LibvirtXMLBase):
```
kylazhang: Can we use the existing address class in device?

yafu-1: The existing address class does not define any attributes, so it will be more complicated to use it to get and set values, or wants to get the value of some attribute.

yafu-1: @kylazhang  have updated to use the existing address class in device. Could you help to review it again please? Thanks.

PR #1201, opened 2017-10-26
virttest.utils_test.qemu: check virtio-win drivers' state for windows guest

xiagao:
Before using virtio-win driver, it's better to check if it's installed and running. So create a function to check it basically.

ID: 1455810

### DIFF
--- a/virttest/utils_test/qemu/__init__.py
+++ b/virttest/utils_test/qemu/__init__.py
@@ -186,6 +186,35 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
         session.close()
 
 
+@error_context.context_aware
+def windrv_verify_running(session, test, driver, timeout=300):
+    """
+    Check if driver is running for windows guest.
+    :param session: VM session
+    :param test: Kvm test object
+    :param driver: The driver which needs to check.
+    :param timeout: Timeout in seconds.
+    :return: A tuple (status, output) where status is the driver check result
+    and output is the detail information.
+    """
+    error_context.context("Check %s driver state." % driver,
+                          logging.info)
+    # It's a temp method to check driver,if there's better one,will modify.
+    msinfo_cmd = "start /wait msinfo32 /report c:\msinfo.txt"
+    driver_check_cmd = "type c:\msinfo.txt |findstr /i \"%s.sys\" " \
+                       "|findstr /i \"running\"" % driver
+    end_time = time.time() + timeout
+    while time.time() < end_time:
+        session.cmd(msinfo_cmd, timeout)
+        status = session.cmd_status(driver_check_cmd)
+        if not status:
+            break
+    if status:
+        test.error("%s driver is not running" % driver)
+    else:
+        logging.info("%s driver is running" % driver)
+
+
 def setup_runlevel(params, session):
     """
     Setup the runlevel in guest.


virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,38 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
+    :param timeout: Timeout in seconds.
+    """
+    session = vm.wait_for_login(timeout=timeout)
+    logging.info("Sleep a while before check guest msinfo.")
+    time.sleep(10)
+    try:
+        error_context.context("Check %s driver state." % driver,
+                              logging.info)
+        msinfo_cmd = "start /wait msinfo32 /report c:\msinfo.txt"
+        driver_check_cmd = "type c:\msinfo.txt |findstr /i \"%s.sys\"" % driver
```
hereischen: I think both `msinfo_cmd`  and `driver_check_cmd` should be be passed in as params.

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,38 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
+        if session.cmd_status(msinfo_cmd, timeout=360):
+            test.error("Get msinfo32 failed.")
```
hereischen: This part maybe better to `return False` as well.
In fact, all these if clauses may better `return (bool , output)` , so you can check the bool, and test.error(output) accordingly in tp-qemu.
You can also remove these logging.info() there.

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,38 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
+        s_driver, o_driver = session.cmd_status_output(driver_check_cmd)
+        if s_driver:
+            logging.info("%s driver is not installed." % driver)
+            return False
+        if o_driver.lower().find("running", 0) == -1:
```
hereischen: `o_driver.lower().find("running")` would be fine

virttest/utils_test/qemu/__init__.py
xiagao: I have modified these code, pls have another check.
Thanks.

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,38 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
+                              logging.info)
+        msinfo_cmd = "start /wait msinfo32 /report c:\msinfo.txt"
+        driver_check_cmd = "type c:\msinfo.txt |findstr /i \"%s.sys\"" % driver
+
+        if session.cmd_status(msinfo_cmd, timeout=360):
+            test.error("Get msinfo32 failed.")
+
+        s_driver, o_driver = session.cmd_status_output(driver_check_cmd)
+        if s_driver or "running" in o_driver.lower():
+            msg = "%s driver is not installed or state is not running,the " \
```
kimi1978wy: I think should be "if s_driver or not "running" "

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,38 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
+                  "detailed info is %s." % (driver, o_driver)
+            test.fail(msg)
+
+        logging.info("%s driver is running." % driver)
+    finally:
+        session.close
+
```
kimi1978wy: should be session.close ()

virttest/utils_test/qemu/__init__.py
xiagao: I changed this module. Will return the result,not do judegement in this funciton.
Pls have a check.

virttest/utils_test/qemu/__init__.py
xiagao: It's a typo when I update to github. Thanks for reminder.

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,32 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
 
 
+@error_context.context_aware
+def check_win_driver(test, driver, vm, timeout=300):
+    """
+    Check if driver is installed and state is running for windows guest.
+    :param test: QEMU test object.
+    :param driver: The driver which needs enable the verifier.
+    :param vm: VM object.
+    :param timeout: Timeout in seconds.
```
hereischen: Better have a return doc string as well, as it is not well clear at first look,

xiagao: Agree, updated.

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,34 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
+    :return: A tuple (status, output) where status is the driver check result
+    and output is the detail information.
+    """
+    session = vm.wait_for_login(timeout=timeout)
```
luckyh: Suggest using an existed `session` instead of creating a new one.

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,34 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
+    logging.info("Sleep a while before check guest msinfo.")
+    time.sleep(10)
```
luckyh: Is there some reason to do so?

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,34 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
+    try:
+        error_context.context("Check %s driver state." % driver,
+                              logging.info)
+        msinfo_cmd = "start /wait msinfo32 /report c:\msinfo.txt"
+        driver_check_cmd = "type c:\msinfo.txt |findstr /i \"%s.sys\" " \
+                           "|findstr /i \"running\"" % driver
+        if session.cmd_status(msinfo_cmd, timeout=360):
+            test.error("Get msinfo32 failed.")
```
luckyh: You can call `session.cmd()`, thus you do not have to handle the error by yourself.

virttest/utils_test/qemu/__init__.py
xiagao: As when booting up guest,it need  some time to start some service.
If there is no time.sleep(), the driver service is in stopped status. So the check result is wrong.

virttest/utils_test/qemu/__init__.py
xiagao: ok, pass session parameter.

virttest/utils_test/qemu/__init__.py
xiagao: @luckyh  
I have updated ,could you help to review it?

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,32 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
         session.close()
 
 
+@error_context.context_aware
+def check_win_driver(session, driver, timeout=300):
+    """
+    Check if driver is installed and state is running for windows guest.
+    :param session: VM session
+    :param driver: The driver which needs enable the verifier.
```
lijinlijin: I guess this line is copied from setup_win_driver_verifier function,please modify it.

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,32 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
+                          logging.info)
+    msinfo_cmd = "start /wait msinfo32 /report c:\msinfo.txt"
+    driver_check_cmd = "type c:\msinfo.txt |findstr /i \"%s.sys\" " \
+                       "|findstr /i \"running\"" % driver
+    end_time = time.time() + timeout
+    while time.time() < end_time:
+        session.cmd(msinfo_cmd, timeout)
+        status, output = session.cmd_status_output(driver_check_cmd)
+        if status:
+            time.sleep(4)
```
lijinlijin: why need to sleep?

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,32 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
+        else:
+            break
+    return (status, output)
```
lijinlijin: When driver is not installed,the output is always empty.
I saw win_irq_check and pvpanic cases both call this function and print the output msg.
 - https://github.com/autotest/tp-qemu/pull/1134
 - https://github.com/autotest/tp-qemu/pull/1119
If this is  not what you expect,please adjust it.
Thanks

virttest/utils_test/qemu/__init__.py
xiagao: Modified.

virttest/utils_test/qemu/__init__.py
xiagao: Wait a moment to let driver become running status. Otherwise the cmd will be run too frequently.

virttest/utils_test/qemu/__init__.py
xiagao: It's what I expect, from the output we can see if the driver is installed and if the driver is in stopped status.
Can't adjust it in this function, as it is only check driver's status.  How to use it should be adjusted by the call function.

Thanks.

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,35 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
...
+    while time.time() < end_time:
+        session.cmd(msinfo_cmd, timeout)
+        status, output = session.cmd_status_output(driver_check_cmd)
+        if status:
+            time.sleep(4)
+        else:
+            break
+    # If driver is not installed,the output from cmd is empty.
+    if not output:
+        output = "%s driver is not installed." % driver
```
lijinlijin: when driver is in disabled status,the output is also empty,but actually the driver is installed.

xiagao: Updated, pls have a review. thanks.

virttest/utils_test/qemu/__init__.py
```diff
@@ -186,6 +186,35 @@ def clear_win_driver_verifier(driver, vm, timeout=300):
         session.close()
 
 
+@error_context.context_aware
+def check_win_driver(session, test, driver, timeout=300):
```
luckyh: Suggest `windrv_verify_running()`.

xiagao: @luckyh  updated.

luckyh:
Thanks all.

PR #1206, opened 2017-10-27
Clear the targetcli configuration before use it

prudmirya:
it is better to clear targetcli configuration before use it, otherwise i seen few errors like "AttributeError: 'NoneType' object has no attribute 'export_flag'"


prudmirya:
Because of this activity the running time is reduced to 3/4th of original time. when there is any issue with configuration, script take max avocado fvt wrapper timeout to exit from running. with this clear configuration it went smoothly.

balamuruhans:
@prudmirya please have proper commit msg explaining the fix.

PR #1211, opened 2017-11-02
qemu_virtio_port: add pattern to match exit status in GuestWorker

sitoliu:
GuestWorker is a process in guest,  receiving commands and printing results on screen, these results will be got by session. Since there are several commands operated in sequence, session might get several results together and host will be confused with common pattern `("^PASS:", "^FAIL:")`.

I saw a relevant issue in my work, so suggest to import this patch to framework. Thanks.


yanan-fu:
Add id please.

PR #1216, opened 2017-11-08
Add libvirt san lock config class

chunfuwen:
associated elements in san lock config are added

### DIFF
--- a/virttest/utils_config.py
+++ b/virttest/utils_config.py
@@ -463,3 +463,19 @@ class LibvirtGuestsConfig(LibvirtConfigCommon):
         'BYPASS_CACHE': 'boolean',
         'SYNC_TIME': 'boolean',
     }
+
+
+class LibvirtSanLockConfig(LibvirtConfigCommon):
+
+    """
+    Class for libvirt san lock config file.
+    """
+    conf_path = '/etc/libvirt/qemu-sanlock.conf'
+    __option_types__ = {
+        'auto_disk_leases': 'boolean',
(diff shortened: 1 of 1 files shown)


chunfuwen:
@kylazhang 
@balamuruhans 

please help take a look.

virttest/utils_config.py
```diff
@@ -463,3 +463,19 @@ class LibvirtGuestsConfig(LibvirtConfigCommon):
...
+
+    """
+    Class for libvirt san lock config file.
+    """
+    conf_path = '/etc/libvirt/qemu-sanlock.conf'
+    __option_types__ = {
+        'auto_disk_leases': 'boolean',
+        'disk_lease_dir': 'string',
+        'host_id': 'int',
+        'require_lease_for_disks': 'boolean',
```
balamuruhans: `auto_disk_leases` and `require_lease_for_disks` looks to be boolean from libvirt source, but as part of configuration, it asks to set as 1 to enable, should be consider it as int for handling from configuration file.

chunfuwen: Yes, somehow you are right. But since config file doesn't support True or False value, so it need set to "1" or "0". But from user point of view(libvirt), boolean is more meaningful ,and actually it will transfer to 1 if True was set . I think on other config such as qemu.conf ,libvirtd are also the same.

balamuruhans: cool then 👍

balamuruhans:
Thanks @chunfuwen @kylazhang, let's merge it.

PR #1217, opened 2017-11-08
Modify check_vfs_count to include the total no.of PF interfaces rather than fixed values.

santwana:
The function should fetch the total no.of PF devices in the host, and multiply the same with the VF's to get the total number of VF's.
Multiplying with a fixed number of PF's will error out if there is mismatch in number of
network interfaces.

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -1305,10 +1305,9 @@ class PciAssignable(object):
         """
         Check VFs count number according to the parameter driver_options.
         """
-        # Network card 82576 has two network interfaces and each can be
-        # virtualized up to 7 virtual functions, therefore we multiply
-        # two for the value of driver_option 'max_vfs'.
-        expected_count = int((re.findall("(\d+)", self.driver_option)[0])) * 2
+        # The VF count should be multiplied with the total no.of PF's
+        # present, rather than fixed number of network interfaces.
+        expected_count = int((re.findall("(\d+)", self.driver_option)[0])) * len(self.get_pf_ids())
         return (self.get_vfs_count() == expected_count)
 
     def is_binded_to_stub(self, full_id):


virttest/test_setup.py
```diff
@@ -1305,10 +1305,9 @@ def check_vfs_count(self):
         """
         Check VFs count number according to the parameter driver_options.
         """
-        # Network card 82576 has two network interfaces and each can be
-        # virtualized up to 7 virtual functions, therefore we multiply
-        # two for the value of driver_option 'max_vfs'.
-        expected_count = int((re.findall("(\d+)", self.driver_option)[0])) * 2
+        # The VF count should should be multiplied with the total no.of PF's
+        # present , rather than fixed number of network interfaces.
+        expected_count = int((re.findall("(\d+)", self.driver_option)[0])) * len(self.get_pf_ids())
```
balamuruhans: 👍

virttest/test_setup.py
```diff
@@ -1305,10 +1305,9 @@ def check_vfs_count(self):
         """
         Check VFs count number according to the parameter driver_options.
         """
-        # Network card 82576 has two network interfaces and each can be
-        # virtualized up to 7 virtual functions, therefore we multiply
-        # two for the value of driver_option 'max_vfs'.
-        expected_count = int((re.findall("(\d+)", self.driver_option)[0])) * 2
+        # The VF count should should be multiplied with the total no.of PF's
+        # present , rather than fixed number of network interfaces.
```
balamuruhans: \# present, rather than fixed number of network interfaces

virttest/test_setup.py
```diff
@@ -1305,10 +1305,9 @@ def check_vfs_count(self):
         """
         Check VFs count number according to the parameter driver_options.
         """
-        # Network card 82576 has two network interfaces and each can be
-        # virtualized up to 7 virtual functions, therefore we multiply
-        # two for the value of driver_option 'max_vfs'.
-        expected_count = int((re.findall("(\d+)", self.driver_option)[0])) * 2
+        # The VF count should should be multiplied with the total no.of PF's
```
balamuruhans: typo of should is typed twice.

santwana: @balamuruhans made changes.

balamuruhans:
@santwana Thanks for the patch, let's merge.

PR #1219, opened 2017-11-09
Add qemu version methods in virttest/utils_misc

lento-sun:
Different qemu versions may have different cmd params, we'll need
to get qemu versions' info. This commit add get_qemu_version and
compare_qemu_version, and use them in get_image_info and
get_image_snapshot to avoid qemu lock introduced by rfe bz1378241.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2370,6 +2370,63 @@ def get_qemu_best_cpu_model(params):
     return params.get("default_cpu_model", None)
 
 
+def get_qemu_version(params):
+    """
+    Get the qemu-kvm(-rhev) version info.
+
+    :param params: Passed to get_qemu_binary, can set to {} if not needed.
+    :return: A dict contain qemu versoins info as {'major': int, 'minor': int,
+    'update': int, 'is_rhev': bool}
+    """
+    version = {'major': None, 'minor': None, 'update': None, 'is_rhev': False}
+    regex = r'\s*[Ee]mulator [Vv]ersion\s*(\d+)\.(\d+)\.(\d+)'
+
+    qemu_binary = get_qemu_binary(params)
+    version_raw = commands.getoutput("%s -version" % qemu_binary).splitlines()
+    for line in version_raw:
+        search_result = re.search(regex, line)
+        if search_result:
+            version['major'] = int(search_result.group(1))
+            version['minor'] = int(search_result.group(2))
+            version['update'] = int(search_result.group(3))
+        if "rhev" in str(line).lower():
+            version['is_rhev'] = True
+    if None in version.values():
+        logging.error("Local install qemu version cannot be detected, "
+                      "the version info is: %s" % version_raw)
+        return None
+    return version
+
+
+def compare_qemu_version(major, minor, update, is_rhev=True, params={}):
+    """
+    Check if local install qemu versions is newer than provided version.
+
+    :param major: The major version to be compared.
+    :param minor: The minor version to be compared.
+    :param update: The update version to be compared.
+    :param is_rhev: If the qemu is a rhev version.
+    :param params: Other params.
+    :return: True means local installed version is equal or newer than the
+    version provided, otherwise False will be returned.
+    """
+    installed_version = get_qemu_version(params)
+    if installed_version is None:
+        logging.error("Cannot get local qemu version, return False directly.")
+        return False
+    if is_rhev != installed_version['is_rhev']:
+        return False
+    installed_version_value = installed_version['major'] * 1000000 + \
+        installed_version['minor'] * 1000 + \
+        installed_version['update']
+    compared_version_value = int(major) * 1000000 + \
+        int(minor) * 1000 + \
+        int(update)
+    if compared_version_value > installed_version_value:
+        return False
+    return True
+
+
 def check_if_vm_vcpu_match(vcpu_desire, vm):
     """
     This checks whether the VM vCPU quantity matches
@@ -3010,6 +3067,11 @@ def get_image_snapshot(image_file):
     """
     try:
         cmd = "qemu-img snapshot %s -l" % image_file
+        if compare_qemu_version(2, 10, 0):
+            # Currently the qemu lock is only introduced in qemu-kvm-rhev,
+            # if it's introduced in qemu-kvm, will need to update it here.
+            # The "-U" is to avoid the qemu lock.
+            cmd += " -U"
         snap_info = process.run(cmd, ignore_status=False).stdout.strip()
         snap_list = []
         if snap_info:
@@ -3063,6 +3125,11 @@ def get_image_info(image_file):
     """
     try:
         cmd = "qemu-img info %s" % image_file
+        if compare_qemu_version(2, 10, 0):
+            # Currently the qemu lock is only introduced in qemu-kvm-rhev,
+            # if it's introduced in qemu-kvm, will need to update it here.
+            # The " -U" is to avoid the qemu lock.
+            cmd += " -U"
         image_info = process.run(cmd, ignore_status=False).stdout.strip()
         image_info_dict = {}
         vsize = None


lento-sun:
Hi @kylazhang and @chunfuwen , if this PR can be merged, we can ignore https://github.com/autotest/tp-libvirt/pull/1262
This is a more common fix to avoid qemu lock for "qemuj-img info" and "qemu-img snapshot -l"

virttest/utils_misc.py
```diff
@@ -2370,6 +2370,59 @@ def get_qemu_best_cpu_model(params):
...
+    update_version = 0
+    is_rhev = False
+    version = {}
+    regex = r'\s*[Ee]mulator [Vv]ersion\s*(\d+)\.(\d+)\.(\d+)'
+
+    qemu_binary = get_qemu_binary(params)
+    version_raw = commands.getoutput("%s -version" % qemu_binary).splitlines()
+    for line in version_raw:
+        search_result = re.search(regex, line)
+        if bool(search_result):
```
kylazhang: I think no need bool here

virttest/utils_misc.py
```diff
@@ -2370,6 +2370,59 @@ def get_qemu_best_cpu_model(params):
     return params.get("default_cpu_model", None)
 
 
+def get_qemu_version(params):
+    """
+    Get the qemu-kvm(-rhev) version info
+    """
```
kylazhang: what's the params? need explain here

virttest/utils_misc.py
```diff
@@ -2370,6 +2370,59 @@ def get_qemu_best_cpu_model(params):
...
+    version['major'] = major_version
+    version['minor'] = minor_version
+    version['update'] = update_version
+    version['is_rhev'] = is_rhev
+    return version
+
+
+def compare_qemu_version(major, minor, update, is_rhev=True, params={}):
+    """
+    Check if local install qemu versions is newer than provided version.
```
kylazhang: 1 blank line here needed

virttest/utils_misc.py
```diff
@@ -2370,6 +2370,59 @@ def get_qemu_best_cpu_model(params):
...
+    :param major: The major version to be compared.
+    :param minor: The minor version to be compared.
+    :param update: The update version to be compared.
+    :param is_rhev: If the qemu is a rhev version.
+    :param params: Other params.
+    :return: True means local installed version is equal or newer than the
+    version provided, otherwise False will be returned.
+    """
+    installed_version = get_qemu_version(params)
+    if bool(is_rhev) != bool(installed_version['is_rhev']):
```
kylazhang: Is there any reason that need to use bool here?

virttest/utils_misc.py
```diff
@@ -2370,6 +2370,59 @@ def get_qemu_best_cpu_model(params):
...
+        return False
+    installed_version_value = int(installed_version['major']) * 1000000 + \
+        int(installed_version['minor']) * 1000 + \
+        int(installed_version['update'])
+    compareed_version_value = int(major) * 1000000 + \
```
kylazhang: compared

kylazhang:
others seems ok for me

virttest/utils_misc.py
lento-sun: yes, can be removed

virttest/utils_misc.py
```diff
@@ -2370,6 +2370,59 @@ def get_qemu_best_cpu_model(params):
     return params.get("default_cpu_model", None)
 
 
+def get_qemu_version(params):
+    """
+    Get the qemu-kvm(-rhev) version info
+    """
+    major_version = 0
+    minor_version = 0
```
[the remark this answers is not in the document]
lento-sun: update

virttest/utils_misc.py
lento-sun: update

virttest/utils_misc.py
lento-sun: update

virttest/utils_misc.py
lento-sun: Just thought this could be more readable, but actually this will cause some potential problem since
bool("abc") == bool("def") will return True.
So fixed.

virttest/utils_misc.py
lento-sun: fixed

virttest/utils_misc.py
```diff
@@ -2370,6 +2370,59 @@ def get_qemu_best_cpu_model(params):
...
+    :param update: The update version to be compared.
+    :param is_rhev: If the qemu is a rhev version.
+    :param params: Other params.
+    :return: True means local installed version is equal or newer than the
+    version provided, otherwise False will be returned.
+    """
+    installed_version = get_qemu_version(params)
+    if bool(is_rhev) != bool(installed_version['is_rhev']):
+        return False
+    installed_version_value = int(installed_version['major']) * 1000000 + \
```
chunfuwen: here installed['major'] directly return int type, so no need forced call int(xx)

virttest/utils_misc.py
```diff
@@ -3009,7 +3062,12 @@ def get_image_snapshot(image_file):
     1         1460096943             3.1M 2016-04-08 14:29:03   00:00:00.110
     """
     try:
+        avoid_qemu_lock_param = " -U"
         cmd = "qemu-img snapshot %s -l" % image_file
+        if compare_qemu_version(2, 10, 0):
+            # Currently the qemu lock only introduced in qemu-kvm-rhev,
```
chunfuwen: only introduced --> is introduced in qemu-kvm-rhev only

virttest/utils_misc.py
```diff
@@ -3009,7 +3062,12 @@ def get_image_snapshot(image_file):
...
+            # if it's introduced in qemu-kvm, we'll need to update it here.
```
chunfuwen: need to update --->need update
Moreover, better avoid use we/you/I  in technologies comments

virttest/utils_misc.py
```diff
@@ -3062,7 +3120,12 @@ def get_image_info(image_file):
                            'csize' : '65536'}
     """
     try:
+        avoid_qemu_lock_param = " -U"
         cmd = "qemu-img info %s" % image_file
+        if compare_qemu_version(2, 10, 0):
+            # Currently the qemu lock only introduced in qemu-kvm-rhev,
```
chunfuwen: same issue as previous ones

virttest/utils_misc.py
```diff
@@ -3062,7 +3120,12 @@ def get_image_info(image_file):
                            'csize' : '65536'}
     """
     try:
+        avoid_qemu_lock_param = " -U"
```
chunfuwen: the variable never changed and used only once, so maybe remove it.

lento-sun: This will make it more clear about what's "-U" for in code, but no need to check qemu manual

lento-sun: well, a comment will do this, too

virttest/utils_misc.py
```diff
@@ -2370,6 +2370,64 @@ def get_qemu_best_cpu_model(params):
...
+    for line in version_raw:
+        search_result = re.search(regex, line)
+        if search_result:
+            version['major'] = int(search_result.group(1))
+            version['minor'] = int(search_result.group(2))
+            version['update'] = int(search_result.group(3))
+        if "rhev" in str(line).lower():
+            version['is_rhev'] = True
+    for key, value in version.iteritems():
+        if value is None:
```
balamuruhans: `if None in version.values():`

virttest/utils_misc.py
```diff
@@ -2370,6 +2370,64 @@ def get_qemu_best_cpu_model(params):
...
+    :param is_rhev: If the qemu is a rhev version.
+    :param params: Other params.
+    :return: True means local installed version is equal or newer than the
+    version provided, otherwise False will be returned.
+    """
+    installed_version = get_qemu_version(params)
+    for key, value in installed_version.iteritems():
+        if value is None:
+            logging.error("Qemu version not clear, return False directly.")
+            return False
```
balamuruhans: This can be done in get_qemu_version(), return None in L2397 after logging will be more appropriate.

virttest/utils_misc.py
```diff
@@ -2370,6 +2370,64 @@ def get_qemu_best_cpu_model(params):
...
+    if is_rhev != installed_version['is_rhev']:
+        return False
+    installed_version_value = installed_version['major'] * 1000000 + \
+        installed_version['minor'] * 1000 + \
+        installed_version['update']
+    compared_version_value = int(major) * 1000000 + \
+        int(minor) * 1000 + \
+        int(update)
+    if compared_version_value > installed_version_value:
```
balamuruhans: `compared_version` can be a dict similar to `installed_version` dict returned by get_qemu_version()
```python
if compared_version != installed_version:
    return False
```

lento-sun:
test

virttest/utils_misc.py
lento-sun: fixed

virttest/utils_misc.py
lento-sun: fixed

virttest/utils_misc.py
lento-sun: fixed

virttest/utils_misc.py
lento-sun: fixed

virttest/utils_misc.py
lento-sun: fixed

virttest/utils_misc.py
lento-sun: fixed

virttest/utils_misc.py
lento-sun: fixed

virttest/utils_misc.py
lento-sun: fixed

virttest/utils_misc.py
lento-sun: fixed

virttest/utils_misc.py
lento-sun: 1. Not pass a dict but 3 params it to follow: https://github.com/autotest/tp-libvirt/blob/master/provider/libvirt_version.py
Just want to make them compliant. 
2. The purpose of the compare_qemu_version is to compare if local installed versions is newer than compared version, if so, that means some function should BE SUPPORTED. So here use a "<" but not a "!="

PR #1227, opened 2017-11-16
Support query locked image info in qemu-img

qiankehan:
Since qemu 2.10(refer to redhat BZ1509167), qemu-img info is required -U
option to query info of a locking images. Add this option for 2.10 or
higher while without this for former qemu versions.


qiankehan:
Duplicated with #1228. Closed

PR #1228, opened 2017-11-16
qemu_storage: Support '-U' for QemuImg

luckyh:
QEMU commit 244a566 introduced the image locking feature which brought
a new option '-U' to the qemu-img command as the result. This patch
adds the support to pass the new option to qemu-img when user needs
to do so.

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -283,7 +283,7 @@ def postprocess_image(test, params, image_name, vm_process_status=None):
             logging.debug("Guest is still running, skip the image check.")
             check_image_flag = False
         else:
-            image_info_output = image.info()
+            image_info_output = image.info(force_share=True)
             image_info = {}
             if image_info_output is not None:
                 for image_info_item in image_info_output.splitlines():
@@ -304,10 +304,10 @@ def postprocess_image(test, params, image_name, vm_process_status=None):
     if check_image_flag:
         try:
             if clone_master is None:
-                image.check_image(params, base_dir)
+                image.check_image(params, base_dir, force_share=True)
             elif clone_master == "yes":
                 if image_name in params.get("master_images_clone").split():
-                    image.check_image(params, base_dir)
+                    image.check_image(params, base_dir, force_share=True)
             # Allow test to overwrite any pre-testing  automatic backup
             # with a new backup. i.e. assume pre-existing image/backup
             # would not be usable after this test succeeds. The best
--- a/virttest/qemu_storage.py
+++ b/virttest/qemu_storage.py
@@ -38,6 +38,7 @@ class QemuImg(storage.QemuImg):
         q_result = process.run(self.image_cmd + ' -h', ignore_status=True,
                                shell=True, verbose=False)
         self.help_text = q_result.stdout
+        self.cap_force_share = '-U' in self.help_text
 
     @error_context.context_aware
     def create(self, params, ignore_errors=False):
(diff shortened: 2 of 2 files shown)


yanan-fu:
Do we need to add `force_share` for  `utils_test.get_image_version`  ?
From my checking, `get_image_version` was used to change the qcow2 version, copy image from nfs currently,  these images won't be accessed at the same time, no locking issue now.
But, do we need to think about it in advance ?

yanan-fu:
For qemu-img,  not only "info", "compare", "check" has `-U` option. 
bench, convert, dd, map, snapshot, rebase support this option option too.
For auto code,  function `convert`, `rebase`, `snapshot` are involved.  i am afraid these functions need be considered too.

yanan-fu:
```
     def get_format(self):
         """
         Get the fimage file format.
         """
         image_info = self.info() 
```
here  ?

luckyh:
@yanan-fu ok, get you point, but let's consider in which scenario we can only use `get_format`/`get_image_version` to get the information we want when the target image is using by a vm. Why not preform it like `fmt = vm.drives.get_by_filename("foo").format` or something else?

Let's also consider about the test scenarios of combining `-U` with `convert`, `rebase` and `snapshot`. They are talking about doing some risky things to the images, right? What are the expected results? I'd like to keep them as they are until we realize it is worth to do so.

pingl2016:
According to your implementation, users must know exactly what scenarios they could use qemu-img check/info. If they don't, for example when they try to get information of image which is writing by another process, they would get the wrong things as avocado-vt use force shared as default.

vivianQizhu:
Agree with @pingl2016 It might be better to let users(or caller) decide whether use force mode or not by themselves. There are two different scenarios for calling qemu-img info(compare): 1. for running image; 2. for not running image. Only scenario 1 need -U. So set default `force_share` from `help_text` might not be appropriate. The user(e.g. drive_mirror) might wish to decide pass `force_share=True` or find another way for their implementation by itself.

pingl2016:
Sorry for my mistake. Actually the force shared option is disabled by default. Then the patch seems ok to me. 
@yanan-fu coli will make a rework about the function about copying images from nfs server when installation failed.
Acked-by: Ping Li <pingl@redhat.com>

luckyh:
@vivianQizhu it is what I did in this patch: user can alter the behavior by parameter `force_share`, if he/she wants force sharing (`force_share=True`) and the binary supports this feature (`force_share &= self.cap_force_share`) then enable it, otherwise disable it.

yanan-fu:
@luckyh  we can use other method to get image format, but why not add `-U` option to make `get_format` work for it too.  IMO, this condition has no risk  for the image.

For `convert`, `rebase` and `snapshot`, since qemu-img support `-U`, i think the risk is acceptable, there should has some condition allow to use -U.   But, it is ok for me to keep the current implementation,  add it in the future when we meet some condition need to use it.

vivianQizhu:
@luckyh That's right.
Test passed,
14:11:06 INFO | PASS 1-smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.myprovider.drive_mirror.simple_test.job_complete.on_nfs.2qcow2

pingl2016:
@yanan-fu  Only when we booting a vm as readonly, then we could convert or rebase the image. But It makes no sense for the test scenario. As to snapshot, it is discouraged to use from qemu side. So it is also unnecessary to support

luckyh:
@yanan-fu because I don't encourage user using such the method to get the format information of an image since we (can) have a better approach.

yanan-fu:
@luckyh  thanks, make sense to me.

luckyh:
OK, let's merge this, thanks all.

PR #1233, opened 2017-11-23
qemu_vm: update for minimum guest memory

hereischen:
Using vm_mem_minimum can specify minimum guest memory.
If params["mem"] is provided, it will use this value.
Otherwise, it will use an auto calculated usable mem.
All these two should be within the range between
vm_mem_limit and vm_mem_minimum, if these max and min
values are provided.

id: 1514381

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -909,6 +909,9 @@ class VM(virt_vm.BaseVM):
             # if not provided, use automem
             else:
                 usable_mem_m = utils_misc.get_usable_memory_size(align=512)
+                if not usable_mem_m:
+                    raise exceptions.TestError("Insufficient memory to"
+                                               " start a VM.")
                 logging.info("Auto set guest memory size to %s MB" %
                              usable_mem_m)
                 mem_size_m = usable_mem_m
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -908,7 +908,7 @@ class VM(virt_vm.BaseVM):
                 mem_size_m = float(normalize_data_size(mem_size_m))
             # if not provided, use automem
             else:
-                usable_mem_m = utils_misc.get_usable_memory_size(align=1024)
+                usable_mem_m = utils_misc.get_usable_memory_size(align=512)
                 logging.info("Auto set guest memory size to %s MB" %
                              usable_mem_m)
                 mem_size_m = usable_mem_m
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -894,33 +894,43 @@ class VM(virt_vm.BaseVM):
 
         def add_memorys(devices, params):
             """
-            Add memory controller by params
+            Add memory controller by params.
 
             :param devices: VM devices container
             """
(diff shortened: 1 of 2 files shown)


xutian:
Looks Ok to me, please drop 'automem' param if it has a default values in configuration files.

virttest/qemu_vm.py
```diff
@@ -894,33 +894,43 @@ def set_dev_params(dev, dev_params,
...
-            automem = mem_params["automem"] == "yes"
+            # if params["mem"] is provided, use the value provided
+            if mem:
+                mem_size_m = "%sM" % mem_params["mem"]
+                mem_size_m = float(normalize_data_size(mem_size_m))
+            # if not provided, use automem
+            else:
+                usable_mem_m = utils_misc.get_usable_memory_size(align=1024)
+                logging.debug("Auto set guest memory size to %s MB" %
+                              usable_mem_m)
```
luckyh: This should be a 'info' message.

virttest/qemu_vm.py
```diff
@@ -894,33 +894,43 @@ def set_dev_params(dev, dev_params,
...
 
-            normalize_data_size = utils_misc.normalize_data_size
-            mem_size_m = "%sM" % mem_params["mem"]
-            mem_size_m = float(normalize_data_size(mem_size_m))
+            # vm_mem_limit(max) and vm_mem_minimum(min) take control here
             if mem_params.get("vm_mem_limit"):
                 max_mem_size_m = params.get("vm_mem_limit")
                 max_mem_size_m = float(normalize_data_size(max_mem_size_m))
                 if mem_size_m >= max_mem_size_m:
                     logging.debug("Guest max memory is limited to %s"
```
luckyh: The same as above.

virttest/qemu_vm.py
```diff
@@ -894,33 +894,43 @@ def set_dev_params(dev, dev_params,
...
-                    logging.debug("Host no enough free memory, reset guest"
-                                  " memory size to %s MB" % usable_mem_m)
-                    params["mem"] = str(int(usable_mem_m))
+
+            if mem_params.get("vm_mem_minimum"):
+                min_mem_size_m = params.get("vm_mem_minimum")
+                min_mem_size_m = float(normalize_data_size(min_mem_size_m))
+                if mem_size_m < min_mem_size_m:
+                    logging.debug("Guest min memory has to be %s"
+                                  % min_mem_size_m)
```
luckyh: The same as above.

virttest/qemu_vm.py
```diff
@@ -894,33 +894,43 @@ def set_dev_params(dev, dev_params,
...
 
             :param devices: VM devices container
             """
             options, devs = [], []
-            params.setdefault("mem", "512")
+            normalize_data_size = utils_misc.normalize_data_size
+            mem = params.get("mem", None)
             mem_params = params.object_params("mem")
-            mem_params.setdefault("automem", "no")
-            automem = mem_params["automem"] == "yes"
```
luckyh: Agreed with @xutian, we should also clean them out from the cfg files.

PR #1236, opened 2017-11-24
qemu_vm.py: add "is_panicked()" in qemu vm class

kimi1978wy:
define is_panicked() to check if the guest is panicked.

Signed-off-by: Yu Wang <wyu@redhat.com>

id:1457076

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -205,6 +205,18 @@ class VM(virt_vm.BaseVM):
         except virt_vm.VMStatusError:
             return False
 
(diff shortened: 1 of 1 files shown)


virttest/qemu_vm.py
```diff
@@ -205,6 +205,18 @@ def is_paused(self):
         except virt_vm.VMStatusError:
             return False
 
+    def is_panicked(self):
+        """
+        Return True if the qemu process is panicked
```
hereischen: VM is panicked?

hereischen:

PR #1238, opened 2017-11-27
This patch clears the targetcli configuration.

prudmirya:
iscsi: clear iscsi targetcli configuration on cleanup

"AttributeError: 'NoneType' object has no attribute 'export_flag'" as the
iscsi target configuration is not cleared,
this patch handles it by clearing during cleanup.
This fix also reduces the test execution time for forthcoming cases.


balamuruhans:
Thanks @prudmirya, let us merge it.

PR #1245, opened 2017-12-01
qemu_vm: fixed destination vm process handling issue

hereischen:
During migration, destination vms were not closed if error
occurs and  not_wait_for_migration=False. This patch fixed it.

id: 1510286

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -3990,7 +3990,7 @@ class VM(virt_vm.BaseVM):
             # If we're doing remote migration and it's completed successfully,
             # self points to a dead VM object
             if not not_wait_for_migration:
-                if self.is_alive():
+                if self.is_alive() and self.is_paused():
                     self.monitor.cmd("cont")
                 clone.destroy(gracefully=False)
                 if env:
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1368,9 +1368,6 @@ class VM(virt_vm.BaseVM):
         if root_dir is None:
             root_dir = self.root_dir
 
-        have_ahci = False
-        have_virtio_scsi = False
-        virtio_scsi_pcis = []
         pci_bus = {'aobject': params.get('pci_bus', 'pci.0')}
         spice_options = {}
 
@@ -1381,11 +1378,6 @@ class VM(virt_vm.BaseVM):
         # Clone this VM using the new params
         vm = self.clone(name, params, root_dir, copy_state=True)
 
-        # global counters
-        ide_bus = 0
-        ide_unit = 0
-        vdisk = 0
-        scsi_disk = 0
         self.last_boot_index = 0
         if params.get("kernel"):
             self.last_boot_index = 1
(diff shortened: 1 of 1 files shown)


hereischen:
@yanan-fu can you help review? Thanks

xutian:
Can you catch the exception which raised after migration failed, then  kill destination vms and re-raised them again ?

hereischen:
@xutian It will do， as long as `not_wait_for_migration=False`.
Thanks.

virttest/qemu_vm.py
```diff
@@ -1368,9 +1368,6 @@ def sort_key(dev):
         if root_dir is None:
             root_dir = self.root_dir
 
-        have_ahci = False
-        have_virtio_scsi = False
-        virtio_scsi_pcis = []
```
yanan-fu: @hereischen     What is this change for ?  From my understanding, it is not related with "destination vm does not be destroyed", right ?

virttest/qemu_vm.py
```diff
@@ -1381,11 +1378,6 @@ def sort_key(dev):
         # Clone this VM using the new params
         vm = self.clone(name, params, root_dir, copy_state=True)
 
-        # global counters
-        ide_bus = 0
-        ide_unit = 0
-        vdisk = 0
-        scsi_disk = 0
```
yanan-fu: same with above

virttest/qemu_vm.py
hereischen: just delete some vars that were not used

virttest/qemu_vm.py
hereischen: same here

yanan-fu:
LGTM

virttest/qemu_vm.py
```diff
@@ -3828,22 +3820,23 @@ def migrate(self, timeout=virt_vm.BaseVM.MIGRATE_TIMEOUT, protocol="tcp",
...
         try:
+            if (local and not (migration_exec_cmd_src and
+                               "gzip" in migration_exec_cmd_src)):
+                error_context.context("creating destination VM")
+                if stable_check:
+                    # Pause the dest vm after creation
+                    extra_params = clone.params.get("extra_params", "") + " -S"
+                    clone.params["extra_params"] = extra_params
+
+                clone.create(migration_mode=protocol, mac_source=self,
```
aliang123: When destination vm creation failed, it skips to line 3989 and execute code in line 3994. It'll send "cont" to the source vm which is in "running" status already. So a better solution is preferred.

hereischen: @aliang123 What is the better solution for this? When destination vm creation failed, the test result will be error or fail. It will be destroyed later, no matter source vm is "paused" or "running",

hereischen: @aliang123  I can understand your point. I think that is for the destination vm when there is no errors.

aliang123: @hereischen, suggest to check vm status after line 3993, if vm.status==paused, then execute line 3994.

hereischen: @aliang123 updated, thanks.

aliang123:
Acked-by: Aihua Liang <aliang@redhat.com>

luckyh:
Thanks all.

PR #1247, opened 2017-12-04
add one method to determine whether current qemu-img command

chunfuwen:
support locking feature or not. If yes, additional -U suboption
need be added for some subcommands:info, compare, check, bench,
convert, dd, map, snapshot, rebase

### DIFF
--- a/virttest/libvirt_storage.py
+++ b/virttest/libvirt_storage.py
@@ -544,3 +544,23 @@ class PoolVolume(object):
             logging.info("Volume '%s' does not exist or '%s' has been exist."
                          % (old_name, new_name))
             return False
+
+
+def check_qemu_image_lock_support():
+    """
+    QEMU commit 244a566(qemu-2.10) introduced the image locking feature which brought
+    a new option '-U' to the qemu-img command:info, compare, check, bench, convert,
+    dd, map, snapshot, rebase. This method provides one way to determine whether
+    current qemu-img support or not.
+
+    :return: True if current qemu-img command support
+    """
+    cmd = "qemu-img"
+    try:
+        binary_path = process.system_output("which %s" % cmd)
+    except process.CmdError:
+        raise process.CmdError(cmd, binary_path,
+                               "qemu-img command is not found")
+    cmd_result = process.run(binary_path + ' -h', ignore_status=True,
+                             shell=True, verbose=False)
+    return '-U' in cmd_result.stdout


chunfuwen:
@kylazhang 
@balamuruhans 

please take a look soon. 
Quite a few libvirt cases hard code qemu-img sub commands,it could be impacted by locking feature afterwards qemu-2.10

virttest/libvirt_storage.py
```diff
@@ -544,3 +544,17 @@ def clone_volume(self, old_name, new_name):
...
+    QEMU commit 244a566(qemu-2.10) introduced the image locking feature which brought
+    a new option '-U' to the qemu-img command:info, compare, check, bench, convert,
+    dd, map, snapshot, rebase. This method provides one way to determine whether
+    current qemu-img support or not.
+
+    :return: True if current qemu-img command support
+    """
+    cmd_result = process.run("qemu-img" + ' -h', ignore_status=True,
+                             shell=True, verbose=False)
+    return '-U' in cmd_result.stdout
```
balamuruhans: we can re-use `class QemuImg(storage.QemuImg)` in virttest/qemu_storage.py

chunfuwen: @balamuruhans , I observed some change introduced in qemu_storage.py.
But actually in our libvirt test code (tp-libvrit), it  rarely use this module to create or info image operations,but hard code qemu-img command. 
This patch was used to provide one solution for those test case,which use qemu-img commands directly.

virttest/libvirt_storage.py
```diff
@@ -544,3 +544,17 @@ def clone_volume(self, old_name, new_name):
             logging.info("Volume '%s' does not exist or '%s' has been exist."
                          % (old_name, new_name))
             return False
+
+
+def check_qemu_image_locking_feature_support():
```
balamuruhans: let us have method name like `def check_qemu_image_lock_support():`

virttest/libvirt_storage.py
```diff
@@ -544,3 +544,17 @@ def clone_volume(self, old_name, new_name):
...
+    """
+    QEMU commit 244a566(qemu-2.10) introduced the image locking feature which brought
+    a new option '-U' to the qemu-img command:info, compare, check, bench, convert,
+    dd, map, snapshot, rebase. This method provides one way to determine whether
+    current qemu-img support or not.
+
+    :return: True if current qemu-img command support
+    """
+    cmd_result = process.run("qemu-img" + ' -h', ignore_status=True,
```
balamuruhans: It is better to check for the binary and proceed
```python
cmd = "qemu-img"
try:
    binary_path = process.system_output("which %s" % cmd)
except process.CmdError:
    raise ...
cmd_result = process.run("%s %s -h" % (binary_path, cmd), ignore_status=True,
                         shell=True, verbose=False)
return '-U' in cmd_result.stdout
```

virttest/libvirt_storage.py
balamuruhans: @chunfuwen okay, agreed

virttest/libvirt_storage.py
chunfuwen: updated!
thx

chunfuwen: updated
thx

virttest/libvirt_storage.py
chunfuwen: updated

virttest/libvirt_storage.py
```diff
@@ -544,3 +544,23 @@ def clone_volume(self, old_name, new_name):
...
+
+    :return: True if current qemu-img command support
+    """
+    cmd = "qemu-img"
+    try:
+        binary_path = process.system_output("which %s" % cmd)
+    except process.CmdError:
+        raise process.CmdError(cmd, binary_path,
+                               "qemu-img command is not found")
+    cmd_result = process.run("qemu-img" + ' -h', ignore_status=True,
```
balamuruhans: sometimes it might not be in global path, that's why it is better to use with binary_path.

chunfuwen: updated
thx

PR #1252, opened 2017-12-12
iscsi: limit to cleanup all

hsmj1412:
Suppose there are 2 iscsi, but the case just want to cleanup one
of them. To use confirm=True will break the other iscsi config.

### DIFF
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -225,7 +225,7 @@ class _IscsiComm(object):
                              'G': (1024, 1024),
                              'T': (1024, 1048576),
                              }
-            if emulated_size.has_key(self.unit):
+            if self.unit in emulated_size:
                 block_size = emulated_size[self.unit][1]
                 size = int(self.emulated_size) * emulated_size[self.unit][0]
                 self.emulated_expect_size = block_size * size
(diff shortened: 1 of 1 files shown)


kylazhang:
Can we add some flag to enable/disable the targetcli clear?

kylazhang:
@prudmirya what's your suggestion

chunfuwen:
@kylazhang ,this fail more than 100 automation cases, we may merge them firstly.

balamuruhans:
> Can we add some flag to enable/disable the targetcli clear?

@chunfuwen I would prefer @kylazhang's suggestion instead of removing the targetcli clear part.
by default let it be disabled.

chunfuwen:
@hsmj1412 , please add default parameter to cleanup(self) in iscsi.py with below:
cleanup(self, comfirmed = False).
and comfirmed == True,then do targetcli clear part operation

hsmj1412:
@chunfuwen Done

virttest/iscsi.py
```diff
@@ -335,7 +335,7 @@ def logout(self):
         if self.logged_in():
             iscsi_logout(self.target)
 
-    def cleanup(self):
+    def cleanup(self, comfirmed=False):
         """
         Clean up env after iscsi used.
```
chunfuwen: comfirmed --> confirmed
Moreover, add doc string for this parameter

hsmj1412: Done

kylazhang:
LGTM

kylazhang:
@chunfuwen @balamuruhans please help to confirm, thanks

chunfuwen:
@balamuruhans , please help merge it since it is urgent

balamuruhans:
Let us merge this.

PR #1253, opened 2017-12-12
utils_misc: get_mem_info() api to retrieve meminfo attributes

balamuruhans:
get_mem_info api can be used get memory information from /proc/meminfo
based on attribute regex in host/guest

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2708,14 +2708,31 @@ def get_free_mem(session, os_type):
     :return string: freespace M-bytes
     """
     if os_type != "windows":
-        cmd = "grep 'MemFree:' /proc/meminfo"
-        output = session.cmd_output(cmd)
-        free = re.findall(r"\d+\s\w", output)[0]
+        free = "%s kB" % get_mem_info(session, 'MemFree')
     else:
         output = session.cmd_output("wmic OS get FreePhysicalMemory")
         free = "%sK" % re.findall("\d+", output)[0]
     free = float(normalize_data_size(free, order_magnitude="M"))
     return int(free)
+
+
+def get_mem_info(session=None, attr='MemTotal'):
+    """
+    Get memory information attributes in host/guest
+
+    :param session: VM session object
+    :param attr: memory information attribute
+
+    :return: memory information of attribute in kB
+    """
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -2708,16 +2708,33 @@ def get_free_mem(session, os_type):
...
-        free = re.findall(r"\d+\s\w", output)[0]
+        free = "%s kB" % get_mem_info(session, 'MemFree')
     else:
         output = session.cmd_output("wmic OS get FreePhysicalMemory")
         free = "%sK" % re.findall("\d+", output)[0]
     free = float(normalize_data_size(free, order_magnitude="M"))
     return int(free)
 
 
+def get_mem_info(session=None, attr='MemTotal'):
```
aliang123: get_mem_info only applied to linux? And redundant code: "line 2734~2735" vs "line 2715~2716". I think an optimization is needed.

balamuruhans: @aliang123 Thank you for reviewing.

> get_mem_info only applied to linux? 

yes, It only applied for Linux, I haven't worked / have Windows env to implement for it.

> And redundant code: "line 2734-2735" with "line 2715-2716". I think an optimization is needed.

As `get_free_mem()` only get's free memory, `get_mem_info()` would help us to retrieve any memory information from /proc/meminfo as it is, ie in kB and `get_free_mem()` should not be disturbed as testcases are using it. so IMHO this 2 line will help us for better compatibility.

aliang123: @balamuruhans, got it. It looks good to me.
Acked-by: Aihua Liang<aliang@redhat.com>

balamuruhans: @aliang123 Thank you.

balamuruhans:
Tested `get_free_mem()` with and without this patch for ensuring it doesn't regress,
**without this patch**
```
>>> utils_misc.get_free_mem(session, "linux")
100468
```

**with this patch**
```
>>> utils_misc.get_free_mem(session, "linux")
100468
```

```
>>> utils_misc.get_mem_info(session, "MemTotal")
129803584
```

```
\# cat /proc/meminfo | grep 'MemTotal'
MemTotal:       129803584 kB

```

balamuruhans:
Thanks @aliang123, let us merge it.

PR #1267, opened 2017-12-22
virsh.memtune_get: return -1 if got "unlimited"

czhang0:
memtune_get() require return one integar. If there is no limit for one
memtune type, "virsh memtune" will output "unlimited", so change return
value to -1 in this case.

Signed-off-by: czhang0 <47047802@qq.com>

### DIFF
--- a/virttest/libvirt_xml/devices/serial.py
+++ b/virttest/libvirt_xml/devices/serial.py
@@ -10,7 +10,8 @@ from virttest.libvirt_xml.devices.character import CharacterBase
 
 class Serial(CharacterBase):
 
-    __slots__ = ('protocol_type', 'target_port', 'target_type', 'sources')
+    __slots__ = ('protocol_type', 'target_port', 'target_type',
+                 'target_model', 'sources')
 
     def __init__(self, type_name='pty', virsh_instance=base.virsh):
         # Additional attribute for protocol type (raw, telnet, telnets, tls)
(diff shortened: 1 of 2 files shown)


virttest/virsh.py
```diff
@@ -2524,7 +2524,7 @@ def memtune_get(name, key):
     memtune_output = memtune_list(name)
     memtune_value = re.findall(r"%s\s*:\s+(\S+)" % key, str(memtune_output))
     if memtune_value:
-        return int(memtune_value[0])
+        return int(memtune_value[0] if memtune_value[0] != "unlimited" else -1)
```
chunfuwen: suggest move  memtune_value[0] != "unlimited" to if memtune_value,just something like;
if memtune_value and memtune_value[0] != "unlimited":
          return int(memtune_value[0])
else:
        return -1

virttest/libvirt_xml/devices/serial.py
```diff
@@ -20,6 +21,8 @@ def __init__(self, type_name='pty', virsh_instance=base.virsh):
                                tag_name='target', attribute='port')
         accessors.XMLAttribute('target_type', self, parent_xpath='/',
                                tag_name='target', attribute='type')
+        accessors.XMLAttribute('target_model', self, parent_xpath='/target',
+                               tag_name='model', attribute='name')
```
balamuruhans: please have a separate commit for adding new attribute to serial xml

PR #1269, opened 2017-12-22
utils_misc: get uptime of system/VM

balamuruhans:
Tests uses uptime of the system to validate, like migration uses to
know whether reboot occurred during migration or not.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2808,6 +2808,25 @@ def set_winutils_letter(session, cmd, label="WIN_UTILS"):
     return cmd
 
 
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -2808,6 +2808,25 @@ def set_winutils_letter(session, cmd, label="WIN_UTILS"):
     return cmd
 
 
+def get_uptime(session=None):
+    """
+    Get the uptime of system in secs
+
+    :param session: VM object or remote session object
```
sathnaga: you meant `VM session object or None for host`?

PR #1271, opened 2017-12-25
utils_misc:Update version compare function to recognize qemu ma

chloerh:
To make sure the version comparing will go on to give compare result
if it's qemu-*-ma, not to return 'False' immediately.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2387,7 +2387,7 @@ def get_qemu_version(params):
 
     :param params: Passed to get_qemu_binary, can set to {} if not needed.
     :return: A dict contain qemu versoins info as {'major': int, 'minor': int,
-    'update': int, 'is_rhev': bool}
+    'update': int, 'is_rhev': bool, 'is_ma': bool}
     """
     version = {'major': None, 'minor': None, 'update': None, 'is_rhev': False}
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -2402,6 +2402,8 @@ def get_qemu_version(params):
             version['update'] = int(search_result.group(3))
         if "rhev" in str(line).lower():
             version['is_rhev'] = True
+        if '-ma' in str(line).lower():
+            version['is_ma'] = True
```
balamuruhans: https://github.com/avocado-framework/avocado-vt/blob/58cae3af968a4be113720a586a97b08bb3f017e6/virttest/utils_misc.py#L2338-L2341

Let us have the new key 'is_ma' in version init and in doc string.
Code looks good to me otherwise

chloerh: Thanks for the review. Updated.

virttest/utils_misc.py
```diff
@@ -2426,7 +2428,8 @@ def compare_qemu_version(major, minor, update, is_rhev=True, params={}):
         logging.error("Cannot get local qemu version, return False directly.")
         return False
     if is_rhev != installed_version['is_rhev']:
-        return False
+        if not installed_version.get('is_ma'):
```
chunfuwen: better value is_ma pass in by parameter, maybe add one parameter is_ma = False

PR #1272, opened 2017-12-27
vm_xml: Add manipulations for guest boot options

hahakiki2010:
1) add boot manipulations to VMXML class
2) modify VMOSXML to add more os options
3) modify VMFeaturesXML to add smm option
4) fixup the bad first argument given to super()

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -1472,6 +1472,42 @@ class VMXML(VMXMLBase):
             pass  # Element already doesn't exist
         self.xmltreefile.write()
 
+    def set_boot_order_by_target_dev(self, target, order):
+        """
+        Set boot order by target dev
+
+        :param target: The target dev on host
+        :param order: The boot order number
+        """
+        devices = self.get_devices()
+        for device in devices:
+            if device.device_tag == "disk":
+                if device.target.get("dev") == target:
+                    device.boot = order
+        self.set_devices(devices)
+
+    def set_os_attrs(self, **attr_dict):
+        """
+        Set attributes of VMOSXML
+
+        :param attr_dict: The key words of os attributes
+        """
+        try:
+            os_xml = getattr(self, "os")
+            if attr_dict:
+                for name, value in attr_dict.items():
+                    setattr(os_xml, name, value)
+            setattr(self, "os", os_xml)
+            self.xmltreefile.write()
+        except (AttributeError, TypeError, ValueError) as detail:
+            raise xcepts.LibvirtXMLError("Invalid os tag or attribute: %s" % detail)
+
+    def remove_all_disk(self):
+        """
+        Remove all disk devices.
+        """
+        self.remove_all_device_by_type('disk')
+
 
 class VMCPUXML(base.LibvirtXMLBase):
 
@@ -1852,7 +1888,7 @@ class VMCPUTuneXML(base.LibvirtXMLBase):
                         'emulator_quota'):
                 accessors.XMLElementInt(slot, self, parent_xpath='/',
                                         tag_name=slot)
-        super(self.__class__, self).__init__(virsh_instance=virsh_instance)
+        super(VMCPUTuneXML, self).__init__(virsh_instance=virsh_instance)
         self.xml = '<cputune/>'
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


hahakiki2010:
https://github.com/avocado-framework/avocado-vt/pull/12

hahakiki2010:
autotest/tp-libvirt#1234

virttest/libvirt_xml/vm_xml.py
```diff
@@ -1462,6 +1462,42 @@ def remove_all_boots(self):
...
+        :param order: The boot order number
+        """
+        devices = self.get_devices()
+        for device in devices:
+            if device.device_tag == "disk":
+                if device.target.get("dev") == target:
+                    device.boot = order
+        self.set_devices(devices)
+
+    def set_os_attrs(self, attr_dict={}):
```
waynesun09: **attr_dict

virttest/libvirt_xml/vm_xml.py
```diff
@@ -1950,7 +1987,17 @@ def __init__(self, virsh_instance=base.virsh):
...
         accessors.XMLElementText('init', self, parent_xpath='/',
                                  tag_name='init')
-        super(self.__class__, self).__init__(virsh_instance=virsh_instance)
+        accessors.XMLAttribute('loader_readonly', self, parent_xpath='/',
+                               tag_name='loader', attribute='readonly')
+        accessors.XMLAttribute('loader_type', self, parent_xpath='/',
+                               tag_name='loader', attribute='type')
+        accessors.XMLAttribute('nvram_template', self, parent_xpath='/',
+                               tag_name='nvram', attribute='template')
+        accessors.XMLElementText('nvram', self, parent_xpath='/',
```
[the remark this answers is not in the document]
hahakiki2010: yes, the secure attribute for loader has been added already

virttest/libvirt_xml/vm_xml.py
```diff
@@ -1472,6 +1472,42 @@ def remove_all_boots(self):
...
+        self.set_devices(devices)
+
+    def set_os_attrs(self, **attr_dict):
+        """
+        Set attributes of VMOSXML
+
+        :param attr_dict: The key words of os attributes
+        """
+        try:
+            os_xml = getattr(self, "os")
```
kylazhang: I think it is better to use like os_xml = self.os instead of getattr to make it easy to understand

virttest/libvirt_xml/vm_xml.py
```diff
@@ -1472,6 +1472,42 @@ def remove_all_boots(self):
...
+            if attr_dict:
+                for name, value in attr_dict.items():
+                    setattr(os_xml, name, value)
+            setattr(self, "os", os_xml)
```
kylazhang: Also here, use self.os = os_xml

virttest/libvirt_xml/vm_xml.py
```diff
@@ -1852,7 +1888,7 @@ def __init__(self, virsh_instance=base.virsh):
                         'emulator_quota'):
                 accessors.XMLElementInt(slot, self, parent_xpath='/',
                                         tag_name=slot)
-        super(self.__class__, self).__init__(virsh_instance=virsh_instance)
+        super(VMCPUTuneXML, self).__init__(virsh_instance=virsh_instance)
```
kylazhang: Is there any difference?

virttest/libvirt_xml/vm_xml.py
hahakiki2010: kyla, it will cause inspekt lint failed

virttest/libvirt_xml/vm_xml.py
hahakiki2010: same comments as above

virttest/libvirt_xml/vm_xml.py
hahakiki2010: seemingly this is required by inspekt check... The previous way will cause inspekt style check failed

virttest/libvirt_xml/vm_xml.py
hahakiki2010: [root@localhost virttest]# inspekt lint libvirt_xml/vm_xml.py
Pylint disabled: W,R,C,E1002,E1101,E1103,E1120,F0401,I0011
Pylint enabled : W0611
************* Module avocado-vt.virttest.libvirt_xml.vm_xml
E0203:1497,21: VMXML.set_os_attrs: Access to member 'os' before its definition line 1502
Lint: libvirt_xml/vm_xml.py FAIL 
Syntax check FAIL
[root@localhost virttest]#

kylazhang: Change the operation to another func can resolve the syntax check issues, anyway, it is also a method to get attr, so it is ok.

virttest/libvirt_xml/vm_xml.py
kylazhang: here should be no problem for syntax check as the check fail because the definition after attr get, anyway, it is also a method to set attr, so it's ok

PR #1274, opened 2017-12-28
virsh.memtune_get: return -1 if got "unlimited"

czhang0:
memtune_get() require return one integar. If there is no limit for one
memtune type, "virsh memtune" will output "unlimited", so change return
value to -1 in this case.

Signed-off-by: czhang0 <47047802@qq.com>


kylazhang:
@balamuruhans @chunfuwen please help to review, thanks

balamuruhans:
Thanks @czhang0 @kylazhang.

PR #1275, opened 2017-12-28
utils_net: remove the serial session close after serial login

YanhuiMa:
When login vm via serial, the serial session will be closed after
get mac ip map, then serial output cann't be logged. So the session
close is too early and remove it.

ID: 1485780
Signed-off-by: yama <yama@redhat.com>

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3418,7 +3418,6 @@ def update_mac_ip_address(vm, timeout=240):
     try:
         session = vm.wait_for_serial_login(timeout=timeout)
         addr_map = get_guest_address_map(session)
-        session.close()
         if not addr_map:
             logging.warn("No VM's NIC got IP address")
             return


YanhuiMa:
1-smp_16.16384m.repeat1.run_test.Host_RHEL.m7.u1.raw.virtio_blk.up.virtio_net.RHEL.7.5.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_reboot' PASS

sitoliu:
LGTM.

xutian:
NACK, here is nothing wrong with closing this connection. please debug to find root cause.

YanhuiMa:
@xutian Why do you say it is not the root cause? Could you pls explain more?  Because this is the root cause I analyzed. Pls see the latest two comments in https://github.com/autotest/tp-qemu/pull/1179. I have explained my analysis.

YanhuiMa:
@luckyh Because now session and serial console are the same thing, the previous design about it has problems. Pls modify it, I will close it after you add  a new serial session representation.

vivianQizhu:
@YanhuiMa bz 1572436 is closed as verified. Would you close this pull request? Thanks.

PR #1279, opened 2018-01-04
Fix ci check

luckyh:
The SandboxCommandBase class now has a name conflict of method
`__set_name__`, it is treated as a special method from python 3.6,
let's rename it and its related methods since they are just
"normal" methods.

Reference:
  https://docs.python.org/3/reference/datamodel.html#object.__set_name__


clebergnu:
[Avocado-VT: Travis-CI failing](https://trello.com/c/z757Cvlv/1197-avocado-vt-travis-ci-failing)

PR #1282, opened 2018-01-09
virsh: migrate_postcopy() API for triggering postcopy migration

balamuruhans:
when migration is enabled with --postcopy (without --postcopy-after-precopy)
in virsh command line, postcopy logic is enabled but doesn't actually start
to perform postcopy unless migrate-postcopy command is sent from virsh. This
migrate_postcopy() can be used to serve the purpose.

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -1410,6 +1410,18 @@ def domain_exists(name, **dargs):
         return False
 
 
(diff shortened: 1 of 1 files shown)


balamuruhans:
Thanks @chunfuwen, merging it.

PR #1283, opened 2018-01-09
utils_test/libvirt: API to check migration triggered and vm state

balamuruhans:
For executing commands during migration of guests, ensure the migration is really started and execute it. supporting APIs are included as part of this patch.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1410,6 +1410,39 @@ class MigrationTest(object):
         # Set connect uri back to local uri
         vm.connect_uri = srcuri
 
+    def check_vm_state(self, vm, state='paused', uri=None):
+        """
+        checks whether state of the vm is as expected
+
+        :param vm: VM Object
+        :param state: expected state of the VM
+        :param uri: connect uri
+
+        :return: True if state of VM is as expected, False otherwise
+        """
+        if not virsh.domain_exists(vm.name, uri=uri):
+            return False
+        vm_state = virsh.domstate(vm.name, uri=uri).stdout.strip()
+        return vm_state.lower() == state.lower()
+
+    def wait_for_migration_start(self, vm, state='paused', uri=None, timeout=60):
+        """
+        checks whether migration is started or not
+
+        :param vm: VM object
+        :param state: expected VM state in destination host
+        :param uri: connect uri
+        :param timeout: time in seconds to wait for migration to start
+
+        :return: True if migration is started False otherwise
+        """
+        def check_state():
+            try:
+                return self.check_vm_state(vm, state, uri)
+            except Exception:
+                return False
+        return utils_misc.wait_for(check_state, timeout)
+
 
 def check_result(result, expected_fails=[], skip_if=[], any_error=False):
     """
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1333,17 +1333,26 @@ class MigrationTest(object):
                                                           extra_opts))
                 migration_thread.start()
                 eclipse_time = 0
+                stime = int(time.time())
                 if func:
-                    stime = int(time.time())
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1333,17 +1333,25 @@ def do_migration(self, vms, srcuri, desturi, migration_type,
...
                 migration_thread.start()
                 eclipse_time = 0
+                stime = int(time.time())
                 if func:
-                    stime = int(time.time())
-                    if func == process.run:
-                        try:
-                            func(args['func_params'], shell=args['shell'])
-                        except KeyError:
+                    # we execute command once the migration is started
```
chunfuwen: Better remove 'we' this word

virttest/utils_test/libvirt.py
```diff
@@ -1333,17 +1333,25 @@ def do_migration(self, vms, srcuri, desturi, migration_type,
...
+                    if self.check_migrate_started(vm):
+                        if func == process.run:
+                            logging.info("Migration started for %s", vm.name)
+                            try:
```
chunfuwen: maybe move try ... except to outer space to just under  check_migrate_started(vm)

virttest/utils_test/libvirt.py
```diff
@@ -1333,17 +1333,25 @@ def do_migration(self, vms, srcuri, desturi, migration_type,
...
+                        else:
                             func(args['func_params'])
                     else:
-                        func(args['func_params'])
-                    eclipse_time = int(time.time()) - stime
-                    logging.debug("start_time:%s, eclipse_time:%s", stime, eclipse_time)
+                        logging.error("Migration failed to start for %s",
+                                      vm.name)
+                eclipse_time = int(time.time()) - stime
+                logging.debug("start_time:%s, eclipse_time:%s", stime, eclipse_time)
```
chunfuwen: stime is int, so %d is more suitable.

virttest/utils_test/libvirt.py
```diff
@@ -1410,6 +1418,39 @@ def cleanup_dest_vm(self, vm, srcuri, desturi):
...
+        :return: True if state of VM is as expected, False otherwise
+        """
+        if not virsh.domain_exists(vm.name, uri=uri):
+            return False
+        vm_state = virsh.domstate(vm.name, uri=uri).stdout.strip()
+        return vm_state.lower() == state.lower()
+
+    def check_migrate_started(self, vm, state='paused', uri=None, timeout=60):
+        """
+        checks whether migration is started of not
```
chunfuwen: of to or

virttest/utils_test/libvirt.py
```diff
@@ -1410,6 +1418,39 @@ def cleanup_dest_vm(self, vm, srcuri, desturi):
...
+        :param timeout: time in seconds to wait for migration to start
+
+        :return: True if migration is started False otherwise
+        """
+        def check_state():
+            try:
+                return self.check_vm_state(vm, state, uri)
+            except Exception:
+                return False
+        return utils_misc.wait_for(check_state, timeout)
```
chunfuwen: can we just use check_vm_state() directly, rather than wrap one duplicate func?

balamuruhans: we have to pass few parameters to the check_vm_state(), so we use duplicate func given input to wait_for()

virttest/utils_test/libvirt.py
balamuruhans: it is better to be in same place, as for process.run user may or may not give shell=True, it was meant to handle that.

balamuruhans:
@chunfuwen Thanks for reviewing, made the changes as suggested.

balamuruhans:
@sathnaga made the changes as suggested.

balamuruhans:
Thanks @sathnaga @chunfuwen, let us merge it.

PR #1284, opened 2018-01-09
utils_test/libvirt.py: enable param source_file_list and disk name configurable

balamuruhans:
when multiple VMs are used for migration, this method can be used for
maintaining temp disk by making the name configurable which is hardcoded
currently, also source_file_list param is enabled and used for setup/cleanup


balamuruhans:
@chunfuwen Thanks for reviewing

PR #1287, opened 2018-01-10
libvirt_xml/vm_xml: use readlines() to get all output lines

czhang0:
Change in check_feature_name():
readline() only get oneline, not whole output lines, so use
readlines().


balamuruhans:
@czhang0 Thanks for the patch!

PR #1289, opened 2018-01-10
test_setup: add several functions to handle hugepages

czhang0:
Add 3 functions to achieve following:
1. set kernel hugepages
2. get the number of certain hugepages totally allocated by kernel
3. get the number of certain hugepages for given NUMA node

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -451,6 +451,63 @@ class HugePageConfig(object):
             raise ValueError("Root hugepage control sysfs directory %s did not"
                              " exist" % self.pool_path)
 
+    def get_kernel_hugepages(self, pagesize):
+        """
+        Get specific hugepages number allocated by kernel at runtime
+
+        read page number from
+        /sys/kernel/mm/hugepages/hugepages-${pagesize}kB/nr_hugepages
+
+        :param pagesize: hugepages size
+        :return: page number
+        """
+        pagedir = "%s/hugepages-%skB" % (self.pool_path, pagesize)
+        pagefile = "%s/nr_hugepages" % pagedir
+
+        if not os.path.isdir(pagedir) or not os.path.isfile(pagefile):
+            raise ValueError("Root hugepage control sysfs directory %s or"
+                             " file %s did not exist" % (pagedir, pagefile))
+        with open(pagefile, "r") as f:
+            hp = f.readline().strip()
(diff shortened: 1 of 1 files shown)


czhang0:
@luckyh ,

below is relevant commands used by similar fuctions in avocado.utils.memory 
grep Hugepagesize /proc/meminfo --> get default hugepage size
/sbin/sysctl vm.nr_hugepages --> get number of default hugepage size

we can get that the fuctions in avocado.utils.memory is only to handle default hugepages size, such as 2048K, so it can't be used to set/get for other huge pages, such as 1G.

So provide more common function to handle each hugepages size.

luckyh:
OK, thanks for your explanation, then I would suggest extending the original interfaces by adding the optional argument `pagesize` instead of creating the new ones.

virttest/test_setup.py
```diff
@@ -451,6 +451,63 @@ def get_multi_supported_hugepage_size(self):
...
+
+        if not isinstance(int(pagenum), type(0)):
+            raise ValueError("pagenum should be integer, while got "
+                             "%s" % pagenum)
+        if not os.path.isdir(pagedir) or not os.path.isfile(pagefile):
+            raise ValueError("Root hugepages control sysfs directory %s or"
+                             " file %s did not exist" % (pagedir, pagefile))
+        process.system("echo %s > %s" % (pagenum, pagefile), shell=True)
+
+    def get_free_hugepages(self, node, pagesize):
```
balamuruhans: test_setup already have this API as `get_node_num_huge_pages(self, node, pagesize)`

virttest/test_setup.py
```diff
@@ -451,6 +451,63 @@ def get_multi_supported_hugepage_size(self):
             raise ValueError("Root hugepage control sysfs directory %s did not"
                              " exist" % self.pool_path)
 
+    def get_kernel_hugepages(self, pagesize):
```
balamuruhans: please check HugePageConfig() in test_setup and it can be enhanced if required

sathnaga:
@czhang0 Based on above comments, I closing this PR for now, pls feel free to open it to
based on the requirement to implement a new API/reuse and enhance the old ones as suggested.
Thanks in advance.

kylazhang:
Please keep this pr open as we need it for current testing, and I will update later according to the comments

PR #1293, opened 2018-01-10
Update qemu_compare_version for qemu-kvm-ma package

hsmj1412:
qemu-kvm-ma also update for some new feature

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -1049,7 +1049,7 @@ def safe_rmdir(path, timeout=10):
             shutil.rmtree(path)
             success = True
             break
-        except OSError, err_info:
+        except OSError as err_info:
             # We are only going to try if the error happened due to
             # directory not empty (errno 39). Otherwise, raise the
             # original exception.
@@ -1876,7 +1876,7 @@ def get_dev_major_minor(dev):
     try:
         rdev = os.stat(dev).st_rdev
         return (os.major(rdev), os.minor(rdev))
-    except IOError, details:
+    except IOError as details:
         raise exceptions.TestError("Fail to get major and minor numbers of the "
                                    "device %s:\n%s" % (dev, details))
 
@@ -2104,8 +2104,7 @@ def get_host_cpu_models():
         """
         Update the cpu flags get from host to a certain order and format
         """
-        flag_list = re.split("\s+", cpu_flags.strip())
-        flag_list.sort()
+        flag_list = sorted(re.split("\s+", cpu_flags.strip()))
         cpu_flags = " ".join(flag_list)
         return cpu_flags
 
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -2338,7 +2338,8 @@ def get_qemu_version(params):
     :return: A dict contain qemu versoins info as {'major': int, 'minor': int,
     'update': int, 'is_rhev': bool}
     """
-    version = {'major': None, 'minor': None, 'update': None, 'is_rhev': False}
+    version = {'major': None, 'minor': None, 'update': None, 'is_rhev': False,
+               'is_ma': False, 'either_rhev_or_ma': False}
```
chunfuwen: I don't think that either_rhev_or_ma need explicitly parameterized since it can be dertermined by existing parameter:is_rhev,and is_ma

hsmj1412: Done

virttest/utils_misc.py
```diff
@@ -2374,7 +2376,8 @@ def compare_qemu_version(major, minor, update, is_rhev=True, params={}):
     if installed_version is None:
         logging.error("Cannot get local qemu version, return False directly.")
         return False
-    if is_rhev != installed_version['is_rhev']:
+    if not ((is_rhev and installed_version['is_rhev']) or
+            (is_ma and installed_version['is_ma'])):
```
lento-sun: If we want to compare qemu-kvm (not ma, not rhev) builds, that would be:
is_rhev = False
installed_version['is_rhev'] = False
is_ma = False 
installed_version['is_ma'] = False
And here the statement will be ' not ((False and False) or (False and False))' = True, and we'll return a False directly.

lento-sun: Could this be:
def compare_qemu_version(major, minor, update, **is_rhev=False, is_ma=False**, params{}:
...
if not ((is_rhev == installed_version['is_rhev']) and (is_ma == installed_version['is_ma'])):
    return False

hsmj1412: @lento-sun Thx, done

virttest/utils_misc.py
```diff
@@ -2374,8 +2377,10 @@ def compare_qemu_version(major, minor, update, is_rhev=True, params={}):
     if installed_version is None:
         logging.error("Cannot get local qemu version, return False directly.")
         return False
-    if is_rhev != installed_version['is_rhev']:
-        return False
+    if is_rhev or is_ma:
+        if not ((is_rhev and installed_version['is_rhev']) or
+                (is_ma and installed_version['is_ma'])):
+            return False
```
balamuruhans: when we use qemu-kvm that is not part of RHEV ie not ma and not rhev, then this API should not return false directly which is inconsistent.
let us have is_rhev and is_ma param in compare_qemu_version() as False by default,
`def compare_qemu_version(major, minor, update, is_rhev=False, is_ma=False, params={}):`

so that, compared qemu even if it is not part of rhev can be compared appropriately. The calling function should set the param is_rhev or is_ma for comparing.

PR #1297, opened 2018-01-11
RFE: Add basic support for luks and luks-inside-qcow2 in avocado-vt

hereischen:
id: 1528107
Signed-off-by: Haotong Chen hachen@redhat.com

### DIFF
--- a/shared/cfg/guest-hw.cfg
+++ b/shared/cfg/guest-hw.cfg
@@ -114,6 +114,9 @@ variants:
         # placeholder
 
 variants:
+    - luks:
+        image_format = luks
+        image_cipher = redhat
     - qcow2v3:
         image_format = qcow2
         image_extra_params = "compat=1.1"
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -1207,7 +1207,8 @@ class DevContainer(object):
                                    blk_extra_params=None, scsi=None,
                                    drv_extra_params=None,
                                    num_queues=None, bus_extra_params=None,
-                                   force_fmt=None):
+                                   force_fmt=None,
+                                   encryption=None, img_cipher=None,):
         """
         Creates related devices by variables
         :note: To skip the argument use None, to disable it use False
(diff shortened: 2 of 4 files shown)


shared/cfg/guest-hw.cfg
```diff
@@ -114,6 +114,9 @@ variants:
         # placeholder
 
 variants:
+    - luks:
+        image_format = luks
```
xutian: Hi guy
it looks that LUKS is a disk encryption specification, not a image format, right?

virttest/qemu_devices/qcontainer.py
```diff
@@ -1584,6 +1600,14 @@ def images_define_by_params(self, name, image_params, media=None,
                 scsi_hba = "virtio-scsi-device"
             elif "s390" in image_params.get("machine_type"):
                 scsi_hba = "virtio-scsi-ccw"
+        # support for luks for .luks and .qcow2
+        # .qcow2 requires "encryption = luks"
+        encryption = image_params.get("encryption")
+        img_cipher = image_params.get("image_cipher")
+        img_fmt = image_params.get("image_format")
+        if img_fmt == "luks":
```
xutian: I think it's not a image format here.

virttest/qemu_storage.py
```diff
@@ -90,51 +125,58 @@ def create(self, params, ignore_errors=False):
...
+                    _extend_option(cmd, "-o", [("encrypt.format", "luks"),
+                                               ("encrypt.key-secret", sec_id)])
+                # TODO: base + luks encryption
+                # for now you can ether use luks or base, not both
+                if self.base_tag:
+                    _add_option(cmd, "-b", [self.base_image_filename])
+                    if self.base_format:
+                        _add_option(cmd, "-F", [self.base_format])
+
+            if self.image_format == "luks":
```
xutian: I think here should be
if  encryption  == "luks":
     .....

shared/cfg/guest-hw.cfg
hereischen: No, there is .luks image format.
Please refer to
https://qemu.weilnetz.de/doc/qemu-doc.html#disk_005fimages_005fformats

eg:
>     #qemu-img create -f luks --object secret,id=sec0,data=`printf %s
> "redhat" |base64`,format=base64 -o key-secret=sec0 demo.luks 5G
>     #qemu-img info demo.luks 
>     image: demo.luks
>     file format: luks
>     virtual size: 5.0G (5368709120 bytes)
>     disk size: 256K
>     encrypted: yes

hereischen:
This is a work around for luks, please don't review it.

XueqiangWei:
tested with virti-scsi and rhel7.6, hotplug related cases are failed.

hereischen:
That is expected, as the patch does not support hot-plug.

PR #1298, opened 2018-01-11
RFC: Improvement of test setup development [v2]

luckyh:
Currently, the problem of test setup development is, without using `env`, it is hard to keep a persistent context for the `setup` and `cleanup` method of the same setuper instance, that makes the code buggy.

To solve this problem, we can introduce a manager which is alive through whole test phases for the setupers. The manager can register setupers, do all the setup procedures at the beginning of the test and do all the cleanup procedures at the end of the test.

---

Changes from v1 (https://github.com/avocado-framework/avocado-vt/pull/1276):

* Rename some objects
* Initialize `setuper` on `register` time

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -11,6 +11,9 @@ import shutil
 import platform
 import netaddr
 
+from abc import ABCMeta
+from abc import abstractmethod
+
 from avocado.utils import process
 from avocado.utils import archive
 from avocado.utils import wait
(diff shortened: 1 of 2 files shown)


hereischen:

virttest/test_setup.py
```diff
@@ -97,6 +100,106 @@ class PolkitConfigCleanupError(PolkitConfigError):
...
+
+    def initialize(self, test, params, env):
+        """
+        Initialize the setup manager.
+
+        :param test: VirtTest instance.
+        :param params: Dictionary with the test parameters.
+        :param env: Dictionary with test environment.
+        """
+        self.__setup_args = (test, params, env)
```
balamuruhans: should we also have to register the class here,
```python
self.register(ABCMeta)
```

luckyh: No, it's meaningless to register the metaclass.

balamuruhans: Okay, but we have to register before performing `setup_manager.do_setup()` ?

luckyh: Yes, it requires that. The reason why it is not designed to register `setuper`s statically is because we have the scenarios to register them conditionally, and some of the check must be performed on runtime.
```python
if params.get("setup_ksm", "no") == "yes":
    manager.register(KSMSetuper)
```

balamuruhans: fine, got it. Thanks for explaning.

luckyh:
OK, let's merge this, thanks!

PR #1301, opened 2018-01-12
utils_misc: Add function to align values to a particular multiples

nasastry:
Add function to align a value to a particular multiples
example: When given '512000' to align with '1024' multiples then
'524288' will be returned.

### DIFF
--- a/virttest/utils_numeric.py
+++ b/virttest/utils_numeric.py
@@ -1,0 +1,12 @@
+import math
+
+
+def align_factor(value, factor=1024):
+    """
+    Value will be factored to the mentioned number.
+
+    :param value: value to be checked or changed to factor
+    :param factor: value used for align
+    :return: factored value
+    """
+    return int(math.ceil(float(value) / factor) * factor)


virttest/utils_misc.py
```diff
@@ -196,6 +196,20 @@ def get_usable_memory_size(align=None):
...
+def align_value(value_to_be_aligned, align=1024):
+    """
+    Align the 'value_to_be_aligned' to 1024 multiples.
+
+    :param value_to_be_aligned: value to be checked or changed
+                                to 1024 multiples
+    :return: modified input value
+    """
+    value_aligned = int(value_to_be_aligned)
+    if value_aligned % 1000 == 0:
```
balamuruhans: so this would fail if,
`value_to_be_aligned = 999`

```python
if value_aligned % align != 0:
    value_aligned = ((value_aligned / align) + 1) * align
return value_aligned
```

nasastry: @balamuruhans 
With the above we get a return of '2023', which is wrong.
The following will give the correct answer but still this is not what I wanted. 
```
if value_aligned % align != 0:
    value_aligned = ((value_aligned // align) + 1) * align
return value_aligned
```

Memory value what I wanted is multiples of 1024 and should be exponet of 2.
example: 512000 is multiples of 1024 but,
512000/1024 = 500 (this is not the exponent of 2).
My expectation is 512*1024=524288 instead of 512000.

nasastry: @balamuruhans 
I need to convert the following values to 
```
512000 to 524288
1024000 to 1048576
1536000 to 1572864
2560000 to 2621440
```
This is how libvirt is doing. If I choose multiples of 1024 or the power 2 is not helping.
because:
checking multiples of 1024 and power 2 works for first two values 
i.e. 512000, 1024000
but not for third and fourth values
i.e. 1536000, 2560000

so for all the values
(value/1000)*1024 is working and giving values matching with libvirt.

nasastry:
@balamuruhans 
Figured out how libvirt is doing this. They are using 256*256 (i.e. 65536) multiples. Modified it accordingly. Please find the current working patch.

virttest/utils_misc.py
```diff
@@ -196,6 +196,23 @@ def get_usable_memory_size(align=None):
     return usable_mem
 
 
+def align_value(value_to_be_aligned, align=1024):
```
chunfuwen: parameter align rename to align_factor to be more meaningful

virttest/utils_misc.py
```diff
@@ -196,6 +196,23 @@ def get_usable_memory_size(align=None):
...
+
+    :param value_to_be_aligned: value to be checked or changed
+                                to 'align' multiples
+    :param align: value used for align
+    :return: aligned value
+    """
+    value_aligned = int(value_to_be_aligned)
+    align = int(align)
+    if value_aligned % align != 0:
+        # // get the divisor only in int
```
chunfuwen: remove  // from the comment line

virttest/utils_misc.py
luckyh: `align_value(value, factor=1024)`

virttest/utils_misc.py
```diff
@@ -196,6 +196,23 @@ def get_usable_memory_size(align=None):
...
+        value_aligned = ((value_aligned // align) + 1) * align
+    return value_aligned
```
luckyh: An one line implementation could be:
`return int(math.ceil(float(value) / factor) *  factor)`

luckyh:
Also, I suggest to introduce an independent module for this (e.g. `utils_numeric`).

virttest/utils_misc.py
balamuruhans: agree

nasastry: @luckyh Thanks for one liner, I thought of not using any modules, but I am fine with this approach.

virttest/utils_misc.py
nasastry: @chunfuwen Sure, will change.

nasastry:
@luckyh @balamuruhans @chunfuwen
Updated the patch, please review. Thanks in advance.

balamuruhans:
Thanks @nasastry for the patch, let us merge it.

balamuruhans:
Thanks @luckyh @chunfuwen for reviewing.

PR #1302, opened 2018-01-15
Update get_image_info function

hsmj1412:
Try to use the better way to get the info

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -3061,6 +3061,18 @@ def get_image_snapshot(image_file):
                                    (image_file, detail))
 
 
+def check_qemu_image_lock_support():
+    """
+    Check qemu-img whether supporting image lock or not
+    :return: The boolean of checking result
+    """
+    cmd = "qemu-img"
+    binary_path = utils_path.find_command(cmd)
+    cmd_result = process.run(binary_path + ' -h', ignore_status=True,
+                             shell=True, verbose=False)
+    return '-U' in cmd_result.stdout
+
+
 def get_image_info(image_file):
     """
     Get image information and put it into a dict. Image information like this:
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -3102,9 +3103,8 @@ def get_image_info(image_file):
     """
     try:
         cmd = "qemu-img info %s" % image_file
-        if compare_qemu_version(2, 10, 0):
-            # Currently the qemu lock is only introduced in qemu-kvm-rhev,
-            # if it's introduced in qemu-kvm, will need to update it here.
+        if libvirt_storage.check_qemu_image_lock_support():
```
xutian: image lock is not libvirt special feature, condication here is not suitable.

chunfuwen: Agree with xutian, so it need self-define one function to query whether lock feature is support or not

hsmj1412: @xutian Where do you think it should be?utils_misc?

hsmj1412: Done

chunfuwen:
@hsmj1412 ,after this PR is merged, probably create new PR to remove check_qemu_image_lock_support method from previous libvirt_storage.py

virttest/utils_misc.py
```diff
@@ -3100,11 +3100,19 @@ def get_image_info(image_file):
                            'dsize' : '931135488',
                            'csize' : '65536'}
     """
+    def check_qemu_image_lock_support():
+        cmd = "qemu-img"
+        try:
+            binary_path = process.system_output("which %s" % cmd)
```
lmr: You should use avocado.utils.path.find_command, which does the same and does not require calling a subprocess.

lmr: Or virttest.utils_misc.find_command, which amounts to the same.

hsmj1412:
@chunfuwen @lmr Updated

virttest/utils_misc.py
```diff
@@ -3100,11 +3100,16 @@ def get_image_info(image_file):
                            'dsize' : '931135488',
                            'csize' : '65536'}
     """
+    def check_qemu_image_lock_support():
```
chunfuwen: Move this implementation to individual method, rather than wrapper in this method

hsmj1412: @chunfuwen Updated

PR #1304, opened 2018-01-16
utils_test/libvirt: add API for pinging VM before and after migration

balamuruhans:
As part of migration test, guest should be pingable before and after
migration, so introduced ping_vm() API to be leveraged by the
migration tests

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1210,6 +1210,55 @@ class MigrationTest(object):
         # The CmdResult returned from virsh migrate command
         self.ret = None
 
+    def ping_vm(self, vm, test, params, uri=None, ping_count=10,
+                ping_timeout=60):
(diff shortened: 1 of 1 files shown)


balamuruhans:
> Any reason to distinguish ‘error’ for from source ping and ‘fail’ for from remote ping.
if VM is not up properly prior to migration, I consider it as test error as it might have setup/configuration issue, but if a proper VM is up, pingable before migration and not responding after migration I consider it as test fail.

balamuruhans:
> ‘Params[“vm_ip_dict”]’ Not getting used anywhere?

we use vm_ip_dict to retrieve all the VM's IP in single or multi VM environment before migration, so that same IPs can be used to ping and validate after migration.

balamuruhans:
@sathnaga Thanks for reviewing, let us merge this patch.

PR #1306, opened 2018-01-17
virttest/utils_env.py:params.get('remote_node_address')

weikunzz:
Modify params.get['remote_node_address'] to params.get('remote_node_ddress')
TypeError: 'instancemethod' object has no attribute '__getitem__'

Signed-off-by: weikunzz kuwei@redhat.com

### DIFF
--- a/virttest/utils_env.py
+++ b/virttest/utils_env.py
@@ -249,7 +249,7 @@ class Env(UserDict.IterableUserDict):
             session = None
             if remote_pp:
                 client = params.get('remote_shell_client', 'ssh')
-                remote_opts = (params.get['remote_node_address'],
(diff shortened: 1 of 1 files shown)


weikunzz:
I have set remote_node_address, it is : remote_node_address = ${ovirt_node_address} in avocado-vt/backends/v2v/cfg/convert_destination.cfg

luckyh:
@weikunzz ah, sorry, I just read it wrong.

I think it should be:
```patch
- remote_opts = (params.get['remote_node_address'],
+ remote_opts = (params['remote_node_address'],
```

luckyh:
@weikunzz they are almost the same, except the former will raise error but the latter will not when the given param is not set.

PR #1311, opened 2018-01-19
ip_sniffing: Fix bug of tshark sniffer

luckyh:
Some packet does not contain sufficient information and it will
cause an `IndexError`, we can ignore such the problematic packet.

ID: 1536309

### DIFF
--- a/virttest/ip_sniffing.py
+++ b/virttest/ip_sniffing.py
@@ -291,9 +291,13 @@ class TSharkSniffer(Sniffer):
 
         # BootP/DHCP (RFC 951/2131)
         if re.match(r"\d+\.\d+\.\d+\.\d+", packet[0]):
-            chaddr = packet[4]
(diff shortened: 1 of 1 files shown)


virttest/ip_sniffing.py
```diff
@@ -291,9 +291,13 @@ def _output_handler(self, line):
...
-            chaddr = packet[4]
-            yiaddr = packet[5]
-            m_type = packet[6]
+            try:
+                chaddr = packet[4]
+                yiaddr = packet[5]
+                m_type = packet[6]
+            except IndexError:
+                # Ignore problematical packets
+                return True
```
yanan-fu: @luckyh    why it return "True" here ?  Don't need to check anything for "problematical packet"  ?

yanan-fu:
LGTM

PR #1312, opened 2018-01-22
utils_net: Fix issue NoneType object has no attribute list_br

hahakiki2010:
Call ovs_br_exists when ovs init failed, we will meet following error:

2018-01-19 17:08:22,842 stacktrace       L0042 ERROR| Reproduced
traceback from:
/var/tmp/libvirt-ci/runtest/avocado-vt/avocado/avocado/core/test.py:596
2018-01-19 17:08:22,842 stacktrace       L0045 ERROR| Traceback (most
recent call last):
2018-01-19 17:08:22,842 stacktrace       L0045 ERROR|   File
"/var/tmp/libvirt-ci/runtest/avocado-vt/avocado-vt/avocado_vt/test.py",
line 270, in runTest
2018-01-19 17:08:22,842 stacktrace       L0045 ERROR|     raise
self.__status  # pylint: disable=E0702
2018-01-19 17:08:22,842 stacktrace       L0045 ERROR| AttributeError:
'NoneType' object has no attribute 'list_br'

Add ovs None check can help skip this error

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -1894,7 +1894,10 @@ def ovs_br_exists(brname, ovs=None):
     if ovs is None:
         ovs = __ovs
 
-    return brname in ovs.list_br()
+    if ovs is not None:
+        return brname in ovs.list_br()
+    else:
+        raise exceptions.TestError("Host does not support OpenVSwitch")
 
 
 @__init_openvswitch
@@ -3161,7 +3164,7 @@ def gen_ipv4_addr(network_num="10.0.0.0", network_prefix="24", exclude_ips=[]):
         exclude_ips.add('.'.join(network_num.split('.')[0:3]) + ".%s" %
                         str(1))
         exclude_ips.add(('.'.join(network_num.split('.')[0:3]) + ".%s" %
-                        str(255)))
+                         str(255)))
     network = netaddr.IPNetwork("%s/%s" % (network_num, network_prefix))
     for ip_address in network:
         if str(ip_address) not in exclude_ips:


virttest/utils_net.py
```diff
@@ -1894,7 +1894,10 @@ def ovs_br_exists(brname, ovs=None):
     if ovs is None:
         ovs = __ovs
 
-    return brname in ovs.list_br()
+    if ovs is not None:
```
xutian: return false and loging warning looks better if not a OVS system.

hahakiki2010: @xutian Given so, I'm afraid it will hard to distinguish the False is due to brname not exist or OVS not support. And raise exception here shall have no impact on non ovs functions, since this func is dedicated to ovs check.

kylazhang: Agree with @hahakiki2010

kylazhang:
@chunfuwen please help to check and merge, thanks

virttest/utils_net.py
```diff
@@ -1894,7 +1894,10 @@ def ovs_br_exists(brname, ovs=None):
...
+        return brname in ovs.list_br()
+    else:
+        raise exceptions.TestError("Host does not support OpenVSwitch")
```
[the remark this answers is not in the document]
hahakiki2010: But the root cause is we got a None OVS instance here but not BRNotExist? I think maybe "Host does not support OpenVSwitch" is more accurate for debugging.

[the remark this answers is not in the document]
balamuruhans: Also, I checked iface_ovs.py in libvirt test case which uses this API and @xutian review comment seems to make sense, because the testcase checks for ovs_br_exists() and decide to configure ovs.

virttest/utils_net.py
balamuruhans: we should consider using different way to check whether ovs is supported or not instead of using ovs_br_exists(). This API is designed to return None if ovs doesn't exist.

hahakiki2010: @balamuruhans, agreed with you to return False here, to keep this API dedicated to ovs bridge check
Then what if we raise ovs not support exception in add_ovs_bridge, because this func has the same issue as ovs_br_exists. When libvirt test case iface_ovs.py detect ovs bridge not exist, and try to add_ovs_bridge, exception like  "NoneType' object has no attribute 'add_br'" will be raised too.

virttest/utils_net.py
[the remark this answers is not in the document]
kylazhang: I think they are not the same situation. This patch is to deal with ovs not support issues, but not br not exist. As if __ovs is None, list_br will raise AttributeError. I think the change will not affect the iface_ovs test

PR #1313, opened 2018-01-22
domcapability_xml: add two cpu related methods

czhang0:
domcapability_xml: add two CPU related methods
    
    Since libvirt3.9, libvirt query qemu/kvm to get all supported features
    and store them in domcapabilities.
    
    Since libvirt3.9, domcapabilities give the overall features supported
    by current host/qemu/kvm via CPU model and a bunch of additional features. CPU
    model is one set of features defined by libvirt in '/usr/share/libvirt/cpu_map.xml'
    Specific features described in domcapabilities are addtional features, the valid
    policy values are 'require' and 'disable', 'require' means the feature can be emulated
    by host/qemu/kvm, 'disable' which means feature is disabled by current system due to
    host CPU doesn't support or qemu/kvm has no ability to emulate it or some other
    reasons like the feature block migration and so on.
    
    Added:
    1. get_hostmodel_name() to get CPU modelname in domcapabilities
    2. get_additional_feature_list(mode) to get CPU additional feature
       list when mode is 'host-model'
    
### DIFF
--- a/virttest/libvirt_xml/domcapability_xml.py
+++ b/virttest/libvirt_xml/domcapability_xml.py
@@ -2,6 +2,7 @@
 Module simplifying manipulation of XML described at
 http://libvirt.org/formatdomaincaps.html
 """
+import logging
 
 from virttest import xml_utils
 from virttest.libvirt_xml import base, accessors, xcepts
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/domcapability_xml.py
```diff
@@ -32,6 +33,106 @@ def __init__(self, virsh_instance=base.virsh):
...
+                        item = {}
+                        item[feature.get('name')] = feature.get('policy')
+                        if 'invtsc' not in item:
+                            feature_list.append(item)
+            return feature_list
+        except AttributeError, elem_attr:
+            logging.warn("Failed to find attribute %s" % elem_attr)
+            return []
+
+    def get_modelname(self):
```
kylazhang: I think the name should be get_hostmodel_name, as you only get host_model name here, or you can make it more general.

virttest/libvirt_xml/domcapability_xml.py
```diff
@@ -32,6 +33,106 @@ def __init__(self, virsh_instance=base.virsh):
...
+                    for feature in mode_node.findall('feature'):
+                        # Feature invtsc doesn't support migration, so not add it to feature list
+                        item = {}
+                        item[feature.get('name')] = feature.get('policy')
+                        if 'invtsc' not in item:
+                            feature_list.append(item)
+            return feature_list
+        except AttributeError, elem_attr:
+            logging.warn("Failed to find attribute %s" % elem_attr)
+            return []
```
kylazhang: you can only return feature_list at last outside the try..except

czhang0: I'll add finally clause to return feature_list. Thanks!

virttest/libvirt_xml/domcapability_xml.py
czhang0: Your name is readable more. Thanks!

virttest/libvirt_xml/domcapability_xml.py
```diff
@@ -32,6 +33,106 @@ def __init__(self, virsh_instance=base.virsh):
         super(DomCapabilityXML, self).__init__(virsh_instance)
         self['xml'] = self.__dict_get__('virsh').domcapabilities().stdout.strip()
 
+    def get_additional_feature_list(self, cpumodename):
```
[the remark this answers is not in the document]
czhang0: Adopt. Thanks!

virttest/libvirt_xml/domcapability_xml.py
```diff
@@ -32,6 +33,106 @@ def __init__(self, virsh_instance=base.virsh):
...
+            <vendor>Intel</vendor>
+            <feature policy="require" name="ss"/>
+            <feature policy="require" name="hypervisor"/>
+            <feature policy="require" name="tsc_adjust"/>
+            <feature policy="require" name="clflushopt"/>
+            <feature policy="require" name="pdpe1gb"/>
+            <feature policy="require" name="invtsc"/>
+        </mode>
+
+        :param cpumodename: cpu mode name, muste be 'host-model' since libvirt3.9
```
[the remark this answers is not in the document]
czhang0: Adopt. Thanks!

virttest/libvirt_xml/domcapability_xml.py
```diff
@@ -32,6 +33,106 @@ def __init__(self, virsh_instance=base.virsh):
...
+        :return: list of features, feature is dict-like, feature name is set to dict key,
+                 feature policy is set to dict value.
+                 returen is like [{'ss': 'require'}, {'pdpe1gb', 'require'}]
+        """
+        feature_list = []  # [{feature1, policy}, {feature2, policy}, ...]
```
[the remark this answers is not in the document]
czhang0: You are right, adopt.

virttest/libvirt_xml/domcapability_xml.py
```diff
@@ -32,6 +33,107 @@ def __init__(self, virsh_instance=base.virsh):
...
+            <feature policy="require" name="ss"/>
+        </mode>
+
+        :return: modelname string
+        """
+        xmltreefile = self.__dict_get__('xml')
+        try:
+            for mode_node in xmltreefile.findall('/cpu/mode'):
+                if mode_node.get('name') == 'host-model':
+                    return mode_node.find('model').text
```
balamuruhans: If xml have something like this
 ```
<cpu>
    <mode name='host-passthrough' supported='yes'/>
    <mode name='host-model' supported='no'/>
    <mode name='custom' supported='no'/>
  </cpu>
```
then we would get,

> WARNING:root:Failed to find attribute 'NoneType' object has no attribute 'text'

can we handle this too ?

balamuruhans:
Thanks @czhang0 for the patch, let us merge it. Thanks @kylazhang @chunfuwen

PR #1315, opened 2018-01-22
utils_misc: add function to get feature list for one CPU model

czhang0:
/usr/lib/libvirt/cpu_map.xml give each model definition.
One model is one set of features.

Add one function to get features for one specific CPU model

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -27,6 +27,8 @@ import platform
 import traceback
 import math
 
+import xml.etree.ElementTree as ET
+
 from aexpect.utils.genio import _open_log_files
 
 from avocado.core import status
@@ -3987,3 +3989,51 @@ class SELinuxBoolean(object):
         logging.debug("To check remote boolean value: %s", boolean_curr)
         if boolean_curr != self.remote_bool_value:
             raise exceptions.TestFail(result.stderr.strip())
+
+
+def get_model_features(model_name):
+    """
+    /usr/share/libvirt/cpu_map.xml defines all CPU models.
+    One CPU model is a set of features.
+    This function is to get features of one specific model.
+
+    :params model_name: CPU model name, valid name is given in cpu_map.xml
+    :return: feature list, like ['apic', 'ss']
+
+    """
+    features = []
+    conf = "/usr/share/libvirt/cpu_map.xml"
+
+    try:
+        output = open(conf, 'r').read()
+        root = ET.fromstring(output)
+        # Find model
+        for model_n in root.findall('arch/model'):
+            if model_n.get('name') == model_name:
+                model_node = model_n
+                for feature in model_n.findall('feature'):
+                    features.append(feature.get('name'))
+                break
+        # Handle nested model
+        nested_model = model_node.find('model')
+        if nested_model is not None:
+            nested_model_name = nested_model.get('name')
+            for model_n in root.findall('arch/model'):
+                if model_n.get('name') == nested_model_name:
+                    for feature in model_n.findall('feature'):
+                        features.append(feature.get('name'))
+                    break
+    except ET.ParseError, error:
+        logging.warn("Configuration file %s has wrong xml format" % conf)
+        raise
+    except AttributeError, elem_attr:
+        logging.warn("No attribute %s in file %s" % (str(elem_attr), conf))
+        raise
+    except Exception:
+        # Other excptions like IOError when open/read configuration file,
+        # capture here
+        logging.warn("Some other exceptions, like configuration file is not "
+                     "found or not file: %s" % conf)
+        raise
+
+    return features


virttest/utils_misc.py
```diff
@@ -3987,3 +3989,30 @@ def setup_remote(self):
...
         if boolean_curr != self.remote_bool_value:
             raise exceptions.TestFail(result.stderr.strip())
+
+
+def get_model_features(modelname):
+    """
+    /usr/share/libvirt/cpu_map.xml defines all CPU models.
+    One CPU model is a set of features.
+    This funtion is to get features of one specific model.
+    """
```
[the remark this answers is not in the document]
kylazhang: Also the return value is better to explain here

virttest/utils_misc.py
```diff
@@ -3987,3 +3989,30 @@ def setup_remote(self):
...
+    features = []
+    conf = "/usr/share/libvirt/cpu_map.xml"
+    if not os.path.exists(conf) or not os.path.isfile(conf):
```
kylazhang: why need 2 check? isfile is not enough?

kylazhang: After saw your following codes, I think you can add try in front of the func and then for the situation that file do not exist, you can catch FileNotFoundError like your following catch

virttest/utils_misc.py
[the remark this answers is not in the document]
czhang0: Adopt.

virttest/utils_misc.py
czhang0: isfile is enough.

virttest/utils_misc.py
```diff
@@ -3987,3 +3989,34 @@ def setup_remote(self):
         logging.debug("To check remote boolean value: %s", boolean_curr)
         if boolean_curr != self.remote_bool_value:
             raise exceptions.TestFail(result.stderr.strip())
+
+
+def get_model_features(model_name):
+    """
+    /usr/share/libvirt/cpu_map.xml defines all CPU models.
+    One CPU model is a set of features.
+    This funtion is to get features of one specific model.
```
balamuruhans: %s/funtion/function

virttest/utils_misc.py
```diff
@@ -3987,3 +3989,34 @@ def setup_remote(self):
...
+    :return: feature list, like ['apic', 'ss']
+
+    """
+    features = []
+    conf = "/usr/share/libvirt/cpu_map.xml"
+    if not os.path.isfile(conf):
+        logging.warn("Configuration file is not found or not file: %s" % conf)
+        return []
+
+    output = process.run("cat /usr/share/libvirt/cpu_map.xml")
```
balamuruhans: `conf` variable can be used here

virttest/utils_misc.py
czhang0: Thanks for your careful.

virttest/utils_misc.py
```diff
@@ -3987,3 +3989,34 @@ def setup_remote(self):
...
+    :return: feature list, like ['apic', 'ss']
+
+    """
+    features = []
+    conf = "/usr/share/libvirt/cpu_map.xml"
+    if not os.path.isfile(conf):
+        logging.warn("Configuration file is not found or not file: %s" % conf)
+        return []
+
+    output = process.run("cat %s" % conf)
```
sathnaga: use `open(conf, 'r').read()` instead...

```
try:
    output = open(conf, 'r').read()
except IOError:
    logging.warn("Configuration file is not found or not file: %s" % conf)
    output = []
```

chunfuwen: suggest use with ... statement since it help you handle connection exception gracefully

virttest/utils_misc.py
```diff
@@ -3987,3 +3989,34 @@ def setup_remote(self):
...
+    try:
+        for model_node in root.findall('arch/model'):
+            if model_node.get('name') == model_name:
+                for feature in model_node.findall('feature'):
+                    features.append(feature.get('name'))
+    except ET.ParseError, error:
+        logging.warn("Configuration file %s has wrong xml format" % conf)
+    except AttributeError, elem_attr:
+        logging.warn("No attribute %s in file %s" % (str(elem_attr), conf))
+    finally:
```
chunfuwen: suggest not use return statement in finally block, so you may just remove finally statement

virttest/utils_misc.py
czhang0: Adopt. Thanks!

virttest/utils_misc.py
czhang0: Adopt. Thanks!

balamuruhans:
Thanks @czhang0 for the patch, Thanks @sathnaga @kylazhang @chunfuwen for review.
Let us merge it.

PR #1318, opened 2018-01-23
Remove test-produced images as default

luckyh:
Some test could produce additional images to be used by itself, and
these images would be less likely to be used by other tests. Based
on this fact, we can be able to remove test-produced images as the
default operation.
    
### DIFF
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -387,6 +387,8 @@ drive_index_cd1 = 1
 drive_rerror_image1 =
 # What to do whether a write error is detected, such as 'stop'
 drive_werror_image1 =
+# Remove image after test finished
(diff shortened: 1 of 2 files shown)


hereischen:
Acked by: Haotong Chen hachen@redhat.com

shared/cfg/base.cfg
```diff
@@ -387,6 +387,8 @@ drive_index_cd1 = 1
 drive_rerror_image1 =
 # What to do whether a write error is detected, such as 'stop'
 drive_werror_image1 =
+# Remove image after test finished
+remove_image_image1 = no
```
balamuruhans: is there a typo %s/remove_image_image1/remove_image ?

PR #1319, opened 2018-01-23
utils_disk: add method to explicitly close guestfs handle

sitoliu:
The handle is closed implicitly when its reference count goes to
zero, eg. when it goes out of scope or the program ends, and we
need a way to force close the handle when it is not out of scope.

ID: 1537363

### DIFF
--- a/virttest/utils_disk.py
+++ b/virttest/utils_disk.py
@@ -578,3 +578,10 @@ class GuestFSModiDisk(object):
                 raise exceptions.TestError(err_msg % file_name)
         finally:
             self.umount_all()
+
+    def close(self):
+        """
+        Explicitly close the guestfs handle.
+        """
+        if self.g:
+            self.g.close()


virttest/utils_disk.py
```diff
@@ -144,7 +144,8 @@ def umount(src, dst, fstype=None, verbose=False, session=None):
             session.cmd_output_safe(fuser_cmd)
             return session.cmd_status(umount_cmd, safe=True) == 0
         process.system(fuser_cmd, ignore_status=True, verbose=True, shell=True)
-        return process.system(umount_cmd, ignore_status=True, verbose=True) == 0
+        return process.system(umount_cmd, ignore_status=True,
+                              verbose=True) == 0
```
vivianQizhu: Unintended?

sitoliu: Thanks, modified.

hereischen:

balamuruhans:
Looks good to me. Thanks all, let us merge it.

PR #1321, opened 2018-01-23
tests-shared.cfg: Move machine.cfg prior to subtests.cfg

vivianQizhu:
Currently machine.cfg works after subtests.cfg, which prevents q35
controllers topology redefinition in subtests, so move it upper.

Remove unattended_install..cdrom, unattended_install..extra_cdrom_ks
 configuration from machine.cfg, it will not take effect after above
modification. Actually it is not proper to define such parameters
here, it will be moved to unattended_install itself in tp-qemu.

id: 1537394

### DIFF
--- a/backends/qemu/cfg/tests-shared.cfg
+++ b/backends/qemu/cfg/tests-shared.cfg
@@ -11,9 +11,9 @@ connect_uri = default
 
 # Include the base config files.
 include base.cfg
+include machines.cfg
 include subtests.cfg
 include host-os.cfg
-include machines.cfg
 include guest-os.cfg
 include guest-hw.cfg
 include cdkeys.cfg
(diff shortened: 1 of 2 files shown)


vivianQizhu:
@yanan-fu Could you please help review? Thanks.

Zhengtong:
I haven't finished checking all the configuration content now. But I found some configurations was changes after the order modification. Beside subtests.cfg, guest-hw.cfg ... files, some internal configuration files will also cover the values in machines.cfg. for example,  the `vga = std` will be changed to be `vga = currius` for ppc.  Is there any plan to fix these param values up soon afterwards?

yanan-fu:
LGTM

qunfang:
LGTM
Acked-by: qzhang@redhat.com

vivianQizhu:
To make the multiform q35 topology configurable(which heavily influence the next generation qemu-kvm test), the machine.cfg has to be changed this way. I supposed that what --vt-list-guests shows could be slightly changed, with no machine type as suffix, so I tried to fix it with https://github.com/avocado-framework/avocado-vt/pull/1338

PR #1324, opened 2018-01-25
libvirt_xml/devices/interface.py: Add coalesce support

czhang0:
coalesce is newly added tag, format like below:
   ```
<interface>
      ...
      <coalesce>
         <rx>
            <frames max='32'/>
         </rx>
      </coalesce>
      ...
   </interface>
```

This commit is to add new attribute 'coalesce' to manipulate
coalesce operatation.


czhang0:
After add this, we can use InterfaceObj.coalesce to set or get coalesce value directly.

PR #1325, opened 2018-01-25
Introduced function to fetch the controller type for SR-IOV Mellanox Adapter PFs.

santwana:
Introduced function to fetch the controller type for SR-IOV Mellanox Adapter PFs.
Depending on the controller type, the output returned can be "Infiniband"/"Ethernet".

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -1443,6 +1443,17 @@ class PciAssignable(object):
         expected_count = int((re.findall("(\d+)", self.driver_option)[0])) * len(self.get_pf_ids())
         return (self.get_vfs_count() == expected_count)
 
+    def get_controller_type(self):
+        """
+        Get the Controller Type for SR-IOV Mellanox Adapter PFs
+        """
+        try:
+            cmd = "lspci | grep '%s'| grep -o '\s[A-Z].*:\s'" % self.pf_filter_re
+            return process.system_output(cmd, shell=True).split("\n")[-1].strip().strip(':')
+        except IndexError:
+            logging.debug("Unable to fetch the controller details")
+            return None
+
     def is_binded_to_stub(self, full_id):
         """
         Verify whether the device with full_id is already binded to driver.


virttest/test_setup.py
```diff
@@ -1438,6 +1438,17 @@ def check_vfs_count(self):
         expected_count = int((re.findall("(\d+)", self.driver_option)[0])) * len(self.get_pf_ids())
         return (self.get_vfs_count() == expected_count)
 
+    def get_controller_type(self):
+        """
+        Get the Controller Type for SR-IOV Mellanox Adapter PFs
```
bssrikanth: I dont think the check is specific to mellanox here.. with given command and pf_filter we can determine any other vendor card type on power.. please double check and update

balamuruhans: right!

virttest/test_setup.py
```diff
@@ -1438,6 +1438,17 @@ def check_vfs_count(self):
...
+        """
+        cmd = "lspci | grep 'Mellanox' | awk '/%s/ {print $2}'" % self.pf_filter_re
+        controller_type = process.system_output(cmd, shell=True).strip().split("\n")[-1]
+        if not controller_type:
+            raise exceptions.TestSetupFail("Unable to fetch the controller details : %s"
```
balamuruhans: API get_controller_type() should probably return None instead of raising setup fail, as it would check for the controller type if available else returns None.
```python
try:
    cmd = "lspci | awk '/%s/ {print $2}'" % self.pf_filter_re
    controller_type = process.system_output(cmd, shell=True).strip().split("\n")[-1]
except IndexError:
    logging.debug("Unable to fetch the controller details")
    return None
```
raising the setup failure should be taken care during actual setup.

virttest/test_setup.py
```diff
@@ -1438,6 +1438,17 @@ def check_vfs_count(self):
         expected_count = int((re.findall("(\d+)", self.driver_option)[0])) * len(self.get_pf_ids())
         return (self.get_vfs_count() == expected_count)
 
+    def get_controller_type(self):
+        """
+        Get the Controller Type for SR-IOV Mellanox Adapter PFs
+        """
+        cmd = "lspci | grep 'Mellanox' | awk '/%s/ {print $2}'" % self.pf_filter_re
```
balamuruhans: `cmd = "lspci | awk '/%s/ {print $2}'" % self.pf_filter_re`

virttest/test_setup.py
santwana: @bssrikanth @balamuruhans I agree , this isn't specific to mellanox. Have made the changes, please review.

virttest/test_setup.py
santwana: @balamuruhans Thanks for reviewing, have made the necessary changes.

virttest/test_setup.py
```diff
@@ -1438,6 +1438,17 @@ def check_vfs_count(self):
         expected_count = int((re.findall("(\d+)", self.driver_option)[0])) * len(self.get_pf_ids())
         return (self.get_vfs_count() == expected_count)
 
+    def get_controller_type(self):
+        """
+        Get the Controller Type for SR-IOV Mellanox Adapter PFs
+        """
+        try:
+            cmd = "lspci | awk '/%s/ {print $2}'" % self.pf_filter_re
+            return process.system_output(cmd, shell=True).strip().split("\n")[-1]
```
balamuruhans: This would make the API to retrieve better output

```python
cmd = "lspci | grep '%s'| grep -o '\s[A-Z].*:\s'" % self.pf_filter_re
return process.system_output(cmd, shell=True).split("\n")[-1].strip().strip(':')
```

balamuruhans:
Thanks @santwana @bssrikanth, let us merge it.

PR #1327, opened 2018-01-26
Add support of new virsh command domfsinfo

hahakiki2010:

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2570,6 +2570,25 @@ class VM(virt_vm.BaseVM):
                 vcpuinfo_list.append(vcpuinfo_dict)
         return vcpuinfo_list
 
+    def domfsinfo(self):
+        """
+        Return a dict's list include domain mounted filesystem information
+        via virsh command
+        """
+        result = virsh.domfsinfo(self.name, ignore_status=False,
+                                 uri=self.connect_uri).stdout.strip()
+        lines = result.splitlines()
+        domfsinfo_list = []
+        if len(lines) > 2:
+            head = lines[0]
+            lines = lines[2:]
+            names = head.split()
+            for line in lines:
+                values = line.split()
+                domfsinfo_list.append(dict(zip(names, values)))
+
+        return domfsinfo_list
+
     def get_used_mem(self):
         """
         Get vm's current memory(kilobytes).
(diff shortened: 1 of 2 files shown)


hahakiki2010:
autotest/tp-libvirt#1391

virttest/libvirt_vm.py
```diff
@@ -2570,6 +2570,28 @@ def vcpuinfo(self):
...
+        via virsh command
+        """
+        result = virsh.domfsinfo(self.name,
+                                 uri=self.connect_uri).stdout.strip()
+        lines = result.splitlines()
+        domfsinfo_list = []
+        if len(lines) > 2:
+            head = lines[0]
+            lines = lines[2:]
+        else:
```
chunfuwen: remove line 2585 and 2586,and move code line 2588 to 2591 to block if len(lines) > 2:

hahakiki2010: good point, thanks~

virttest/libvirt_vm.py
```diff
@@ -2570,6 +2570,25 @@ def vcpuinfo(self):
                 vcpuinfo_list.append(vcpuinfo_dict)
         return vcpuinfo_list
 
+    def domfsinfo(self):
+        """
+        Return a dict's list include domain mounted filesystem information
+        via virsh command
+        """
+        result = virsh.domfsinfo(self.name,
+                                 uri=self.connect_uri).stdout.strip()
```
kylazhang: What about the command failure result? As you also do not deal with the failures in virsh.py

hahakiki2010: I think the exceptions can be covered by function "command" in virsh.py, which will help raise exception if cmd failed to execute. Pls correct me if I'm wrong. thanks

kylazhang: ignore_status = dargs.get('ignore_status', True) default value is True, which means will ignore status, if you do not set it will not raise error, please check command function in virsh.py

hahakiki2010: I think you are right, the error is ignored by default. So how about change code as below, by setting  ignore_status=False?  in the func 
```
result = virsh.domfsinfo(self.name, ignore_status=False
+                                    uri=self.connect_uri).stdout.strip()
```

Or we pass ignore_status=False in dargs when invoking func

```
        command_dargs = {"readonly": readonly_mode, "quiet_mode": quiet_mode,
                         "debug": True, "ignore_status": False}
        result = virsh.domfsinfo(vm_name, **command_dargs)

```

Which one you think it's better? I checked other func in virsh.py, seemingly most commands are implemented as below, without setting ignore status implicitly 
```
def suspend(name, **dargs):
    return command("suspend %s" % (name), **dargs) 
```

kylazhang: it is ok

PR #1329, opened 2018-01-29
remote_access: Update cases for multiple sasl mechanism

hsmj1412:
From libvirt 3.2.0, the default sasl change from
digest-md5 to gssapi. Although the digest-md5 mechanism
is now considered insecure and should not be used.
In response to manual request, we update framework and scripts
for double sasl machanisms.

### DIFF
--- a/virttest/utils_conn.py
+++ b/virttest/utils_conn.py
@@ -11,9 +11,10 @@ import aexpect
 from avocado.utils import path
 from avocado.utils import process
 
-from . import propcan, remote, utils_libvirtd
-from . import data_dir
-from .compat_52lts import results_stderr_52lts
+from virttest import propcan, remote, utils_libvirtd
+from virttest import data_dir
+from virttest import utils_package
+from virttest.compat_52lts import results_stderr_52lts
 
 
 class ConnectionError(Exception):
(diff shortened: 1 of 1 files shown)


virttest/utils_conn.py
```diff
@@ -1177,6 +1209,16 @@ def server_setup(self, on_local=False):
                             'tls_no_sanity_certificate=1'}
             operate_libvirtdconf.sub_else_add(pattern2repl)
 
+        # edit the /etc/sasl2/libvirt.conf to change sasl method
+        if self.sasl_type == 'gssapi':
+            keytab = "keytab: /etc/libvirt/krb5.tab"
+        else:
+            keytab = ""
+        pattern2repl = {r".*mech_list\s*:\s*.*":
```
chunfuwen: pattern2repl rename to pattern_to_repl

virttest/utils_conn.py
```diff
@@ -1187,9 +1229,11 @@ def server_setup(self, on_local=False):
                     session = remote.wait_for_login('ssh', server_ip, '22',
                                                     server_user, server_pwd,
                                                     r"[\#\$]\s*$")
+                    rrunner = remote.RemoteRunner(session=session)
```
chunfuwen: rrunner rename to remote_runner

virttest/utils_conn.py
```diff
@@ -1545,13 +1617,37 @@ def conn_setup(self):
                             'unix_sock_rw_perms="%s"' % unix_sock_rw_perms}
             self.client_libvirtdconf.sub_else_add(pattern2repl)
 
+        # edit the /etc/sasl2/libvirt.conf to change sasl method
```
chunfuwen: it sound like it also update /etc/hosts, so please add this information too in your comment

kylazhang:
The code seems good, please give the related tp-libvirt PR testing result so we can make sure no regression, thanks

PR #1332, opened 2018-02-01
utils_numeric: rename align_factor to align_value

nasastry:
Rename align_factor to align_value
align_value is having a better meaning than align_factor

### DIFF
--- a/virttest/utils_numeric.py
+++ b/virttest/utils_numeric.py
@@ -1,12 +1,12 @@
 import math
 
 
-def align_factor(value, factor=1024):
+def align_value(value, factor=1024):
     """
     Value will be factored to the mentioned number.
 
     :param value: value to be checked or changed to factor
     :param factor: value used for align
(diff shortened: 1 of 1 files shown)


virttest/utils_numeric.py
```diff
@@ -1,7 +1,7 @@
 import math
 
 
-def align_factor(value, factor=1024):
+def align_value(value, factor=1024):
```
balamuruhans: https://github.com/avocado-framework/avocado-vt/blob/7d3356e4f29af386c60123b83749eb1d77d8781b/virttest/utils_numeric.py#L10

we can also mention the return as aligned value.

nasastry: @balamuruhans Done. Changed the doc string of 'return'. Please have a look.

balamuruhans:
Thanks @nasastry for the patch, let us merge it.

PR #1334, opened 2018-02-05
lv_utils: Add manipulations to mpath

hahakiki2010:
Provide mpath naming convention between lvname and mpath name

### DIFF
--- a/virttest/staging/lv_utils.py
+++ b/virttest/staging/lv_utils.py
@@ -399,3 +399,51 @@ def lv_reactivate(vg_name, lv_name, timeout=10):
                        "nuke the process that uses it first."), lv_name)
         raise exceptions.TestError(
             "The logical volume %s is still active" % lv_name)
+
+
+def gen_lvmap_mpath(session=None):
+    """
+    Return the name mapping between logical volume and device
+    mapper multipathed device.
+    :param session: When session is given, it means VM session
+                    Otherwise default value means it work for host
+    :return: None or a dict like:
+            {"lv_name": dm_name}
+    """
+    names_map = None
+    cmd = ("lvdisplay|awk '/LV Name/{n=$3} "
+           "/Block device/{d=$3; sub(\".*:\", \"dm-\", d); print n,d;}'")
+
+    if session:
+        output = session.cmd_output(cmd).strip()
+    else:
+        output = process.system_output(cmd, shell=True)
+    lines = output.splitlines()
+    if len(lines) >= 1:
+        names_map = {}
+        for line in lines:
+            values = line.split()
+            lv_name, mpath_name = values[0], values[1]
+            names_map.update({lv_name: mpath_name})
+    return names_map
+
+
+def get_vg_mapped_blk_target(vg_name, session=None):
+    """
+    Get vg mapped blk target with pvs Command
+    :param vg_name: Volume group Name
+    :param session: When session is given, it means VM session
+                    Otherwise default value means it work for host
+    :return: The mapped blk target or None
+    """
+    cmd = "pvs | grep %s" % vg_name
+    blk_target = None
+    if session:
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -1562,6 +1562,49 @@ def add_identities_into_ssh_agent():
     process.run("ssh-add")
 
 
+def gen_lvmap_mpath(session):
```
chunfuwen: I think host also need this function also, so probably remove session parameter

virttest/utils_misc.py
```diff
@@ -1562,6 +1562,49 @@ def add_identities_into_ssh_agent():
...
+        if len(lines) >= 1:
+            names_map = {}
+            for line in lines:
+                values = line.split()
+                lv_name, mpath_name = values[0], values[1]
+                names_map.update({lv_name: mpath_name})
+    return names_map
+
+
+def get_vg_mapped_blk_target(session, vg_name):
```
chunfuwen: see previous similar comments

virttest/utils_misc.py
balamuruhans: instead of removing, we can have session with default param as None, and handle it for host, when session is given let is work for guest.

hahakiki2010: thank you both, I have updated code to both support host and session scenarios. Pls help review again~

virttest/utils_misc.py
hahakiki2010: updated

hahakiki2010:
autotest/tp-libvirt#1391

hahakiki2010:
autotest/tp-libvirt#1391

virttest/utils_misc.py
```diff
@@ -1562,6 +1562,54 @@ def add_identities_into_ssh_agent():
...
+    if len(lines) >= 1:
+        names_map = {}
+        for line in lines:
+            values = line.split()
+            lv_name, mpath_name = values[0], values[1]
+            names_map.update({lv_name: mpath_name})
+    return names_map
+
+
+def get_vg_mapped_blk_target(vg_name, session=None):
```
chunfuwen: parameter order is not consistent with your doc string

hahakiki2010: fixed:)

virttest/utils_misc.py
```diff
@@ -1562,6 +1562,54 @@ def add_identities_into_ssh_agent():
     process.run("ssh-add")
 
 
+def gen_lvmap_mpath(session=None):
+    """
+    Return the name mapping between logical volume and device
+    mapper multipathed device.
+
+    :param session: Guest session
```
chunfuwen: may update doc string as below:
:param session: default is None, it means it access host,otherwise Guest Vm session

hahakiki2010: doc string updated

hahakiki2010:
@balamuruhans could u help review this PR? thanks

virttest/utils_misc.py
```diff
@@ -1562,6 +1562,56 @@ def add_identities_into_ssh_agent():
...
+def get_vg_mapped_blk_target(vg_name, session=None):
+    """
+    Get vg mapped blk target with pvs Command
+
+    :param vg_name: Volume group Name
+    :param session: When session is given, it means VM session
+                    Otherwise default value means it work for host
+    :return: The mapped blk target or None
+    """
+    cmd = "pvs | grep " + vg_name
```
balamuruhans: `cmd = "pvs | grep %s" % vg_name`

hahakiki2010: code changed

hahakiki2010:
@balamuruhans I think your suggestion is better,  that lv manipulations shall put into lv_utils. I have modified code as suggested. Pls help review, thanks~

PR #1337, opened 2018-02-06
nfs: set ssh_auto_recover as False by default in nfs cleanup

balamuruhans:
It is safe to have ssh_auto_recover as False, during env cleanup
of multiple setups in remote host, with default as True may cleanup
ssh obj and cause forthcoming remote cleanup to error out.

### DIFF
--- a/virttest/nfs.py
+++ b/virttest/nfs.py
@@ -353,7 +353,7 @@ class NFSClient(object):
         except process.CmdError:
             raise exceptions.TestFail("Failed to run: %s" % umount_cmd)
 
-    def cleanup(self, ssh_auto_recover=True):
+    def cleanup(self, ssh_auto_recover=False):
         """
         Cleanup NFS client.
         """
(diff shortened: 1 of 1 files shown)


virttest/nfs.py
```diff
@@ -336,8 +336,9 @@ def cleanup(self, ssh_auto_recover=True):
             raise exceptions.TestFail("Failed to umount %s" % self.mount_dir)
 
         # Recover SSH connection
-        self.ssh_obj.auto_recover = ssh_auto_recover
-        del self.ssh_obj
+        if ssh_auto_recover:
+            self.ssh_obj.auto_recover = ssh_auto_recover
+            del self.ssh_obj
```
sathnaga: Looks like this needs a relook as `self.ssh_obj` is getting deleted why update it?

balamuruhans: based on the update `__del__()` will work after `del self.ssh_obj` also it is suggested to set auto_recover as False by default which is taken care as well

PR #1341, opened 2018-02-07
guest-hw: Correct cd_format when install vm from cdrom

yanan-fu:
When install vm from cdrom, to aviod vm can not identify cdrom
as no driver, use ide for i440fx and ahci for q35 instead.

id: 1542520

### DIFF
--- a/shared/cfg/guest-hw.cfg
+++ b/shared/cfg/guest-hw.cfg
@@ -88,6 +88,11 @@ variants:
         # NOTE:  scsi-block and scsi-generic are pass-through physical dev only
         drive_format=scsi-hd
         cd_format=scsi-cd
+        unattended_install.cdrom, unattended_install..extra_cdrom_ks:
+            i440fx:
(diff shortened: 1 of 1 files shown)


CongLi:
All 'unattended_install.cdrom, check_block_size.4096_512, check_block_size.512_512, with_installation' cases should be worked as well.

CongLi:
And better to specify the reason for the change in the commit log.

yanan-fu:
@luckyh  good, a more robust solution.

Checked, the image iso can not be find when use scsi_cd too, so we need to add "cd_format_cd1=ide" too.
And same for winutils, close this PR and send a new one, thanks,

yanan-fu:
@luckyh good, a more robust solution.

Checked, to ensure different isos use corresponding drive letter that be used in answer file, add winutils and cd1, and remove virtio.
close this PR and send a new one https://github.com/autotest/tp-qemu/pull/1233, thanks!

PR #1342, opened 2018-02-08
Revert "libvirt_xml/devices/interface.py: Add coalesce support"

czhang0:
This reverts commit 87829ef924495da5d9011f70d77531d8a7c8d7a0.

The original commit make all interface cases failed during delete_device when there is no coalesce tag in interface, so revert it tentatively.

Signed-off-by: cuzhang <cuzhang@redhat.com>

PR #1344, opened 2018-02-08
Fix del_device accessor error

kylazhang:
As when parent_xpath does not exist, it will raise accessor error during `__eq__`
operation so add exception catch for this error


kylazhang:
Should fixed problem in https://github.com/avocado-framework/avocado-vt/pull/1342 and no need to reverse the PR

hsmj1412:
LGTM

kylazhang:
As it  block our more than 50 cases, so I will merge it. Thanks

PR #1349, opened 2018-02-11
vm_xml: Don't operate non-topology tags when handle topology element

czhang0:
When operate topology, we can't use one new CPUXML with topology tag
to update existing cpu element, otherwise other tags in cpu element
will be erased.

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -757,7 +757,12 @@ class VMXML(VMXMLBase):
             # is set true and not have all topology definition, it would fail
             # let us leave it to the user give proper values
             if (topology or add_topology) and (sockets or cores or threads):
-                vmcpu_xml = VMCPUXML()
+                # Only operate topology tag, other tags doesn't change
+                try:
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -743,7 +743,8 @@ def set_vm_vcpus(vm_name, vcpus, current=None, sockets=None, cores=None,
             # is set true and not have all topology definition, it would fail
             # let us leave it to the user give proper values
             if (topology or add_topology) and (sockets or cores or threads):
-                vmcpu_xml = VMCPUXML()
+                # Only operate topology tag, other tags doesn't change
+                vmcpu_xml = vmxml['cpu']
```
sathnaga: @cuzhang This will fail  if the guest xml does not have cpu attribute in the first place,  pls add check and if not lets fallback to `vmcpu_xml = VMCPUXML()`

kylazhang: @sathnaga please help to check again for this change

PR #1353, opened 2018-02-19
utils_misc: initialize original value of selinux in init

balamuruhans:
As original value for selinux boolean for local and remote
configuration is used in both setup and in clean up it is better
to initialize from __init__


balamuruhans:
Thanks @chunfuwen let us merge this.

PR #1354, opened 2018-02-20
libvirt_vm: use utils_package for package installation

balamuruhans:
reuse utils_package for install/remove packages and remove existing
redundant code.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -30,6 +30,7 @@ from . import data_dir
 from . import xml_utils
 from . import utils_selinux
 from . import test_setup
+from . import utils_package
 
 
 def normalize_connect_uri(connect_uri):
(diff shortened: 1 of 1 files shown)


virttest/libvirt_vm.py
```diff
@@ -2887,10 +2812,10 @@ def prepare_guest_agent(self, prepare_xml=True, channel=True, start=True):
         if not self.is_alive():
             self.start()
 
-        self.install_package('pm-utils', ignore_status=True)
-        self.install_package('qemu-guest-agent')
-
         session = self.wait_for_login()
+        packages = ['pm-utils', 'qemu-guest-agent']
```
sathnaga: we can keep `vm.install_package()` method as is and just use `utils_package.package_install(packages, session) ` inside that method, will help us in backward compatibility aswell and we can avoid taking session in all usage places...

balamuruhans: @sathnaga I agree and it is the better way. Thanks you!

virttest/libvirt_vm.py
```diff
@@ -2795,73 +2796,28 @@ def install_package(self, name, ignore_status=False):
...
-                if status:
-                    if not ignore_status:
-                        raise virt_vm.VMError("Update of package %s failed:"
-                                              "\n%s" % (name, output))
-                    else:
-                        logging.error("Update of package %s failed:"
-                                      "\n%s", name, output)
-        finally:
-            session.close()
+        if not utils_package.package_install(name, session):
```
chunfuwen: suggest add one check whether session is None before call it in utils_package.package_install

virttest/libvirt_vm.py
```diff
@@ -2795,73 +2796,28 @@ def install_package(self, name, ignore_status=False):
...
+            if not ignore_status:
+                session.close()
```
chunfuwen: session.close() call twice in line 2801 and 2805,actually can merge call once if put it before line 2800

virttest/libvirt_vm.py
balamuruhans: I believe wait_for_login will raise error if it fails and checking session is not necessary here

virttest/libvirt_vm.py
balamuruhans: we cant keep session.close() in 2800, then session will close only on utils_package fails. I think it is in right place.

hsmj1412:
There are some test error when to destroy the guest, because the "virsh destroy" operation will damage the guest repo's database. So we need try to clean the repo and re-try to install.

PR #1356, opened 2018-02-23
libvirt_xml/devices/interface.py: Use Dict for coalesce support

czhang0:
coalesce is newly added tag, format like below:
  ``` 
   <interface>
      ...
      <coalesce>
         <rx>
            <frames max='32'/>
         </rx>
      </coalesce>
      ...
   </interface>
```
This commit is to add new dict xml 'coalesce' to manipulate
coalesce operatation.

Usage:
InterfaceObj.coalesce = {'max':coalesce_value}

### DIFF
--- a/virttest/libvirt_xml/devices/interface.py
+++ b/virttest/libvirt_xml/devices/interface.py
@@ -87,12 +87,11 @@ class Interface(base.TypedDeviceBase):
                                parent_xpath='/',
                                tag_name='model',
                                attribute='type')
-        accessors.XMLAttribute(property_name="coalesce",
-                               libvirtxml=self,
-                               forbidden=None,
-                               parent_xpath='/coalesce/rx',
-                               tag_name='frames',
-                               attribute='max')
+        accessors.XMLElementDict(property_name="coalesce",
+                                 libvirtxml=self,
+                                 forbidden=None,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: virttest/libvirt_xml/devices/interface.py [hunk 1]
```diff
@@ -87,12 +87,11 @@ def __init__(self, type_name, virsh_instance=base.base.virsh):
                                parent_xpath='/',
                                tag_name='model',
                                attribute='type')
-        accessors.XMLAttribute(property_name="coalesce",
```
balamuruhans: do we have interface type as coalesce, I see it is part of network interface type from,
https://libvirt.org/formatdomain.html#coalesce

czhang0: Yes, libvirt has support coalesce. 
This commit is merely to extend coalesce/rx/frame to support more attributes not only 'max'. InterfaceObj.coalesce = {'max':coalesce_value}

Until now, only max attribute is defined in coalesce/rx/frame. Further future, if new attribute is defined, coalesce in avocado-vt doesn't need to change, only use dict directly in tp-libvirt, like
InterfaceObj.coalesce = {'max':coalesce_value, 'newattribute': value}

PR #1360, opened 2018-02-27
[Freeze]Python 3: Update print to print()

hsmj1412:

### DIFF
--- a/scripts/github/cache_populate.py
+++ b/scripts/github/cache_populate.py
@@ -6,16 +6,18 @@ import getpass
 from github import Github
 from github_issues import GithubIssues
 
+from __future__ import print_function
+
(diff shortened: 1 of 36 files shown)


hsmj1412:
There are some print() to print(()), wait for fixed

scripts/github/cache_populate.py
```diff
@@ -10,12 +10,12 @@
             password=getpass.getpass('Enter github password: '),
             user_agent='PyGithub/Python')
 
-print "Enter location (<user>/<repo>)",
+print("Enter location (<user>/<repo>)", end=' ')
```
chunfuwen: use  __future__ import print_function to resolve compatibility with python 2.6

PR #1364, opened 2018-02-28
selinux: avoid using selinux on ubuntu by default

balamuruhans:
configuring selinux in ubuntu cause unnecessary test errors/fails, it is
better not to configure selinux by default for ubuntu and use selinux_force
param to force it if required.


sathnaga:
Merging as ubuntu regressions will fail...

PR #1366, opened 2018-03-01
Python 3: Update print to print()

hsmj1412:

### DIFF
--- a/shared/scripts/telnet.py
+++ b/shared/scripts/telnet.py
@@ -3,7 +3,7 @@ import sys
 import os
 
 if len(sys.argv) != 5:
-    print "Usage: %s host_ip user password prompt" % sys.argv[0]
+    print("Usage: %s host_ip user password prompt" % sys.argv[0])
     sys.exit(1)
 
 host_ip = sys.argv[1]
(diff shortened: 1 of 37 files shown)


shared/scripts/telnet.py
```diff
@@ -1,34 +1,34 @@
...
-output = tn.read_until('Password:', 10)
-tn.write(password)
-tn.write(os.linesep)
-output += tn.read_until(prompt, 5)
-if "Login incorrect" in output:
-    print "Login incorrect"
-    sys.exit(1)
-tn.write("quit")
-tn.write(os.linesep)
-print output
```
chunfuwen: it sound like there are some misaction here that all lines in this python had been changed, but actually not

shared/scripts/virtio_console_guest.py
```diff
@@ -934,26 +934,26 @@ def recv(self, port, length=1, bfr=1024, mode=True):
...
+                  (length, len(recvs)))
 
     def clean_port(self, port, bfr=1024):
         in_f = self._open([port])
         ret = select.select([in_f[0]], [], [], 1.0)
         buf = ""
         if ret[0]:
             buf = os.read(in_f[0], bfr)
-        print ("PASS: Rest in socket: ") + str(buf[:10])
+        print("PASS: Rest in socket: ") + str(buf[:10])
```
kylazhang: Here need modify

PR #1367, opened 2018-03-01
Python 3: To translate raise

hsmj1412:
Converts raise E, V to raise E(V),
and raise E, V, T to six.reraise(E, V, T)

### DIFF
--- a/avocado_vt/test.py
+++ b/avocado_vt/test.py
@@ -62,6 +62,8 @@ if 'AUTOTEST_PATH' in os.environ:
 
 from autotest.client.shared import error
 
+import six
+
 
 def cleanup_env(env_filename, env_version):
     """
@@ -206,7 +208,7 @@ class VirtTest(test.Test):
         except Queue.Empty:
             pass
         else:
-            raise exc[1], None, exc[2]
+            six.reraise(exc[1], None, exc[2])
 
     def __safe_env_save(self, env):
         """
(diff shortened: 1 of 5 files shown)


avocado_vt/test.py
```diff
@@ -203,7 +203,7 @@ def verify_background_errors(self):
         except Queue.Empty:
             pass
         else:
-            raise exc[1], None, exc[2]
+            raise exc[1].with_traceback(exc[2])
```
chunfuwen: need introduce six module to keep compatable:
-raise E, V, T
+mport six
+six.reraise(E, V, T)

avocado_vt/test.py
```diff
@@ -60,6 +60,9 @@
 from autotest.client.shared import error
```
chunfuwen: You may take this opportunity to remove dependency for  from autotest.client.shared import error

avocado_vt/test.py
```diff
@@ -263,22 +239,15 @@ def runTest(self):
...
         """
         if self.__status != "PASS":
-            if isinstance(self.__status, error.TestNAError):
-                self.cancel(str(self.__status))
-            elif isinstance(self.__status, error.TestWarn):
-                self.log.warn(str(self.__status))
-            elif isinstance(self.__status, error.TestFail):
-                self.fail(str(self.__status))
-            else:
-                raise self.__status  # pylint: disable=E0702
```
kylazhang: why remove the judgement here?

chunfuwen: runtest()actually not called in this py file,but _runtest() was used instead, not quite sure whether it is useful or not. So better change them to exceptions.TestError,exceptions.TestFail, exceptions.TestWarn respectively

hsmj1412: As autotest has been obsoleted, so this section is deleted. But some tp-libvirt tests still need autotest, freeze this pr until tp-libvirt update or freeze the avocado-vt tag

PR #1375, opened 2018-03-05
vt_console: add set_linesep

hahakiki2010:

### DIFF
--- a/virttest/vt_console.py
+++ b/virttest/vt_console.py
@@ -98,6 +98,7 @@ class ConsoleSession(object):
     For more detailed function call information,
     you may need to refer to aexpect.client.ShellSession().
     """
+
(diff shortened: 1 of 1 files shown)


virttest/vt_console.py
```diff
@@ -167,6 +168,11 @@ def send_ctrl(self, *args, **kwargs):
         return self.__manager.proxy_call(self.send_ctrl.__name__,
                                          *args, **kwargs)
 
+    def set_linesep(self, *args, **kwargs):
```
chunfuwen: better add comment to describe this function

hahakiki2010: I don't think it's necessary, because none of func in this py file added comments, and also you can get the meaning from the name.

PR #1376, opened 2018-03-05
unattended_install: does nothing if not .ks file

zhengxiang93:
If unattended file is not a kickstart file, dest_fname would be used but not declared.
So add this patch to fix it.


xutian:
looks OK to me

zhengxiang93:
Create another new PR.

PR #1378, opened 2018-03-06
vt_console: add status_test_command and read_up_to_prompt

czhang0:

### DIFF
--- a/virttest/vt_console.py
+++ b/virttest/vt_console.py
@@ -43,6 +43,7 @@ class ConsoleManager(object):
 
     def __init__(self):
         self._console = None
+        self.status_test_command = None
         self._lock = threading.Lock()
 
     def __getstate__(self):
@@ -82,6 +83,8 @@ class ConsoleManager(object):
 
     def set_console(self, console):
         self._console = console
+        if self._console is not None:
+            self.status_test_command = self._console.status_test_command
 
     @lock
     def proxy_call(self, func, *args, **kwargs):
@@ -101,6 +104,7 @@ class ConsoleSession(object):
 
     def __init__(self, manager):
         self.__manager = manager
+        self.status_test_command = manager.status_test_command
         self.__closed = False
 
     def __repr__(self):
@@ -192,3 +196,8 @@ class ConsoleSession(object):
         self.__verify_session_status()
         return self.__manager.proxy_call(
             self.read_until_any_line_matches.__name__, *args, **kwargs)
+
+    def read_up_to_prompt(self, *args, **kwargs):
+        self.__verify_session_status()
+        return self.__manager.proxy_call(self.read_up_to_prompt.__name__,
+                                         *args, **kwargs)


czhang0:
Test result is following:
#avocado run --vt-type libvirt type_specific.local.virtual_network.iface_network.net_forward
JOB ID     : 3864fecace7c619e0015b0bcfa518f1b9a162538
JOB LOG    : /root/avocado/job-results/job-2018-03-06T13.32-3864fec/job.log
 (1/4) type_specific.local.virtual_network.iface_network.net_forward.net_isolated: PASS (34.67 s)
 (2/4) type_specific.local.virtual_network.iface_network.net_forward.net_nat: PASS (35.04 s)
 (3/4) type_specific.local.virtual_network.iface_network.net_forward.net_route: PASS (34.54 s)
 (4/4) type_specific.local.virtual_network.iface_network.net_forward.route_test: PASS (25.34 s)
RESULTS    : PASS 4 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 132.35 s

virttest/vt_console.py
```diff
@@ -43,6 +43,7 @@ class ConsoleManager(object):
 
     def __init__(self):
         self._console = None
+        self.status_test_command = None
```
chunfuwen: I assume that status_test_command can be one attribute of ConsoleSession, nothing with ConsoleManager

hereischen: Do you want to change `status_test_command` after the `ConsoleSession` was created?
If so, I think you should change it through `vm.serial_console`.
Please note on line 63 `self._console.set_status_test_command(status_test_command)`, you can set this command during the session creation.

czhang0: @chunfuwen , if no define status_test_command for ConsoleManager, we need to use statement like below to intialize it in __init__() definition of ConsoleSession:

     def __init__(self, manager):
        self.__manager = manager
        self.status_test_command = manager._console.status_test_command

_console looks like one private attribute for ConsoleManager, so accessing it directly outside of ConsoleManager looks ugly. So I define status_test_command for ConsoleManger in this case.

czhang0: @hereischen , vm.serial_login() call ConsoleManager.create_session() with status_test_command parameter, and latter call self._console.set_status_test_command() to set status_test_command.
So I think it works.

hereischen: @czhang0  what is the problem you want to solve?
It's not about the whether it  does work or not. It's the reason why you want do it this way.
You can set `status_test_command` by calling `vm.serial_login()` without this patch.

czhang0: There are some access like 'session.status_test_command' in avocado-vt codes.
For example, line 804 in virttest/utils_net.py:
     804: session.sendline(session.status_test_command)
For serial login, ConsoleSession has no status_test_command.
Without this patch, complains will emit like ' ConsoleSession has no attribute status_test_command'.
So it's reason I add it.
Thanks for your comment!

PR #1381, opened 2018-03-07
Python 3: Fix relative import issues

luckyh:
Fix encountered relative import issues.

### DIFF
--- a/virttest/passfd.py
+++ b/virttest/passfd.py
@@ -39,7 +39,7 @@ For more information, see one of the R. Stevens' books:
 
 import socket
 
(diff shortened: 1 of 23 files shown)


virttest/passfd.py
```diff
@@ -39,7 +39,7 @@
 
 import socket
 
-import _passfd
+from . import _passfd
```
hsmj1412: The Travis ci failed, due to this module is a C-language file, please double check

luckyh: @hsmj1412 updated, thanks.

PR #1391, opened 2018-03-08
qemu_guest_agent: Update shutdown()

xiangchunfu:
guest_agent: Update shutdown()

Currently, shutdown() returns "{}" or no response after sending the shutdown command,
remove success_resp=False option to raise error message,and remove useless
"return True" and ignore VAgentProtocolErro at the same time.

ID: 1468530, 1476648

Signed-off-by: Xiangchun Fu xfu@redhat.com

### DIFF
--- a/virttest/guest_agent.py
+++ b/virttest/guest_agent.py
@@ -509,8 +509,10 @@ class QemuAgent(Monitor):
         if mode in [self.SHUTDOWN_MODE_POWERDOWN, self.SHUTDOWN_MODE_REBOOT,
                     self.SHUTDOWN_MODE_HALT]:
             args = {"mode": mode}
-        self.cmd(cmd=cmd, args=args, success_resp=False)
-        return True
+        try:
+            self.cmd(cmd=cmd, args=args)
+        except VAgentProtocolError:
+            pass
 
     @error_context.context_aware
     def sync(self, sync_mode="guest-sync"):


xiangchunfu:
(1/2) smp_8.8192m.repeat1.run_test.Host_RHEL.m7.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.4.x86_64.myprovider.qemu_guest_agent.virtio_serial.check_reset_reboot_shutdown_fsfreeze: PASS (98.68 s)
(2/2) smp_8.8192m.repeat1.run_test.Host_RHEL.m7.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.4.x86_64.myprovider.qemu_guest_agent.isa_serial.check_reset_reboot_shutdown_fsfreeze: PASS (88.30 s)

xiangchunfu:
tp-qemu patch
https://github.com/autotest/tp-qemu/pull/1032

virttest/guest_agent.py
```diff
@@ -509,7 +509,10 @@ def shutdown(self, mode=SHUTDOWN_MODE_POWERDOWN):
         if mode in [self.SHUTDOWN_MODE_POWERDOWN, self.SHUTDOWN_MODE_REBOOT,
                     self.SHUTDOWN_MODE_HALT]:
             args = {"mode": mode}
-        self.cmd(cmd=cmd, args=args, success_resp=False)
+        try:
+            self.cmd(cmd=cmd, args=args, success_resp=True)
+        except VAgentCmdError:
```
lijinlijin: better return the detailed error msg for debugging usage.

xiangchunfu: @lijinlijin  It will return  "The command xxx has been disabled for this instance"  error msg when executing shutdown.  This msg is from VAgentCmdError,   VAgentCmdError depend on qemu monitor.  Do you wish like this "except 'The command xxx has been disabled for this instance'": ?

xiangchunfu: @lijinlijin  In order to tp-qemu can get this error Msg,  raise this error Msg.

virttest/guest_agent.py
```diff
@@ -509,7 +509,10 @@ def shutdown(self, mode=SHUTDOWN_MODE_POWERDOWN):
...
+            raise
```
luckyh: The above approach has the same meaning as `self.cmd(cmd=cmd, args=args, success_resp=True)`, the problem is there are several types of error but only some of them would not be allowed.

virttest/guest_agent.py
```diff
@@ -509,7 +509,11 @@ def shutdown(self, mode=SHUTDOWN_MODE_POWERDOWN):
         if mode in [self.SHUTDOWN_MODE_POWERDOWN, self.SHUTDOWN_MODE_REBOOT,
                     self.SHUTDOWN_MODE_HALT]:
             args = {"mode": mode}
-        self.cmd(cmd=cmd, args=args, success_resp=False)
+        try:
+            self.cmd(cmd=cmd, args=args, success_resp=True)
+        except VAgentCmdError as e:
+            str(e)
+            return False
```
hereischen: Based on https://github.com/autotest/tp-qemu/pull/1032
```python
        except VAgentCmdError:
            return False
```
would be enough.

Can you explain what `str(e)` is for on line 515

xiangchunfu: @hereischen You are right,  Don't need "str(e)". (original idea was to use str(e) to call   __str__(self) method in Class VAgentCmdError).

virttest/guest_agent.py
luckyh: If you need to check the content of the error message in your test context, here should let `self.cmd()` raise the `VAgentCmdError` directly instead of catching it and returning `False`.

So finally the code would be:
```patch
- self.cmd(cmd=cmd, args=args, success_resp=False)
- return True
+ self.cmd(cmd=cmd, args=args)
```

xiangchunfu: @luckyh  You are right,  cmd() can raise this Error.   I have updated it and re-tested it. 

 (1/2) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_reset_reboot_shutdown_fsfreeze: PASS (124.79 s)
 (2/2) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.qemu_guest_agent.isa_serial.check_reset_reboot_shutdown_fsfreeze: PASS (143.99 s)

luckyh:
@xiangchunfu the code looks good to me now, but I still request a sensible commit message, thanks.

vivianQizhu:
@luckyh Would you please check if the commit message proper and help merge it if so? Thanks.

virttest/guest_agent.py
```diff
@@ -509,8 +509,7 @@ def shutdown(self, mode=SHUTDOWN_MODE_POWERDOWN):
         if mode in [self.SHUTDOWN_MODE_POWERDOWN, self.SHUTDOWN_MODE_REBOOT,
                     self.SHUTDOWN_MODE_HALT]:
             args = {"mode": mode}
-        self.cmd(cmd=cmd, args=args, success_resp=False)
-        return True
+        self.cmd(cmd=cmd, args=args)
```
luckyh: Just realized that it should ignore the error that there is no response after sending the command, so you should catch `VAgentProtocolError` and ignore it at here.

xiangchunfu: @luckyh  You are right,  I have updated.

PR #1394, opened 2018-03-09
Python 3: Replace long to int

hsmj1412:

### DIFF
--- a/virttest/staging/backports/simplejson/encoder.py
+++ b/virttest/staging/backports/simplejson/encoder.py
@@ -4,6 +4,12 @@ import re
 from decimal import Decimal
 
 
+try:
+    long
+except NameError:
+    long = int
+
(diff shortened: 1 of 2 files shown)


virttest/staging/backports/simplejson/encoder.py
```diff
@@ -4,6 +4,12 @@
 from decimal import Decimal
 
 
+try:
+    long
+except NameError:
+    long = int
```
luckyh: Seems the packages under directory `backports` are provided for old pythons (<=2.5), so we do not have to modify them if not necessary.

hsmj1412: @luckyh "vt-bootstrap" will not scan the backports directory? And I think this change will not affect old pythons

PR #1397, opened 2018-03-09
Python 3: Fixes code broken from the removal of members in the types …

hsmj1412:
…module

### DIFF
--- a/virttest/asset.py
+++ b/virttest/asset.py
@@ -2,7 +2,6 @@ import logging
 import os
 import re
 import string
-import types
 import glob
 import shutil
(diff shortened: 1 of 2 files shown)


virttest/asset.py
```diff
@@ -2,14 +2,17 @@
...
 import shutil
 from six import StringIO
 try:
     import configparser as ConfigParser
 except ImportError:
     import ConfigParser
+try:
+    from types import StringTypes
+except ImportError:
+    StringTypes = str
```
kylazhang: Do we need to change str to string_types?

PR #1400, opened 2018-03-09
Python 3: Update codes style

hsmj1412:
This optional fixer performs several transformations
that make Python code more idiomatic.


virttest/unittest_utils/mock.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, value):
     def _types_match(arg1, arg2):
         if isinstance(arg1, basestring) and isinstance(arg2, basestring):
             return True
-        return type(arg1) == type(arg2)
+        return isinstance(arg1, type(arg2))
```
chunfuwen: type still used here

PR #1401, opened 2018-03-09
Python 3: Converts the exec statement to the exec() function

hsmj1412:


kylazhang:
I think

PR #1408, opened 2018-03-12
Python 3: Fix xrange() and range()

hsmj1412:


avocado_vt/plugins/vt_joblock.py
```diff
@@ -28,6 +28,8 @@
 
 from ..test import VirtTest
 
+from six.moves import xrange
```
chunfuwen: since there is no xrange on python3, better use below import statement to make it easily understandable:
```
from six.moves import xrange as range
```

PR #1413, opened 2018-03-13
Fix nfs mount dir check code

lento-sun:
When mount dir not existing, the error is in cmd's stderr, which is
not captured by current code.

### DIFF
--- a/virttest/nfs.py
+++ b/virttest/nfs.py
@@ -408,9 +408,9 @@ class NFSClient(object):
         """
         check_mount_dir_cmd = self.ssh_cmd + "'ls -d %s'" % self.mount_dir
         logging.debug("To check if the %s exists", self.mount_dir)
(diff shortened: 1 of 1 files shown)


virttest/nfs.py
```diff
@@ -408,9 +408,9 @@ def setup_remote(self):
         """
         check_mount_dir_cmd = self.ssh_cmd + "'ls -d %s'" % self.mount_dir
         logging.debug("To check if the %s exists", self.mount_dir)
-        output = process.system_output(check_mount_dir_cmd, shell=True,
-                                       ignore_status=True)
-        if re.findall("No such file or directory", output, re.M):
+        cmd_result = process.run(check_mount_dir_cmd, shell=True,
+                                 ignore_status=True)
+        if re.findall("No such file or directory", cmd_result.stderr, re.M):
```
chunfuwen: cmd_result.stderr to cmd_result.stout

PR #1417, opened 2018-03-14
Python 3: Update commands module

hsmj1412:

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -16,7 +16,18 @@
 
 import os
 import sys
-import subprocess
+
+
+def getstatusoutput(cmd):
+    """Return (status, output) of executing cmd in a shell."""
+    pipe = os.popen('{ ' + cmd + '; } 2>&1', 'r')
+    text = pipe.read()
+    sts = pipe.close()
+    if sts is None:
+        sts = 0
+    if text[-1:] == '\n':
+        text = text[:-1]
+    return sts, text
 
 
 class DocBuildError(Exception):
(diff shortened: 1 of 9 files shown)


hsmj1412:
Depend on https://github.com/avocado-framework/avocado/pull/2510

virttest/nfs.py
```diff
@@ -285,8 +285,9 @@ def is_mounted(self):
                                                                 self.mount_dir)
         cmd = self.ssh_cmd + "'%s'" % find_mountpoint_cmd
         logging.debug("The command: %s", cmd)
-        if process.system(cmd, verbose=True, ignore_status=True, shell=True):
-            return False
+        status, output = process.getstatusoutput(cmd)
+        if status:
+            logging.debug("The command result: <%s:%s>", status, output)
```
kylazhang: Why change the logical here? No need to return False?

scripts/regression.py
```diff
@@ -19,9 +18,16 @@
...
+def getstatusoutput(cmd):
+    """Return (status, output) of executing cmd in a shell."""
+    pipe = os.popen('{ ' + cmd + '; } 2>&1', 'r')
+    text = pipe.read()
+    sts = pipe.close()
+    if sts is None:
+        sts = 0
+    if text[-1:] == '\n':
+        text = text[:-1]
+    return sts, text
```
kylazhang: If no use for the status, just define with getoutput and only return output is ok.

hsmj1412:
1. commands module redirect stderr to stdout, but subprocess could not do it
2. the output of subprocess is bytes in Python3

kylazhang:
And it is better to add the explanation in commit msg

docs/source/conf.py
```diff
@@ -16,7 +16,18 @@
 
 import os
 import sys
-import subprocess
+
+
+def getstatusoutput(cmd):
```
chunfuwen: I don't think we need create new method, moreover, no need redirect stderror to stdoutput

virttest/nfs.py
hsmj1412: Fixed

scripts/regression.py
hsmj1412: Updated

docs/source/conf.py
hsmj1412: I can not confirm if redirection is necessary or not.

PR #1421, opened 2018-03-15
Fix the regex incompleteness

hsmj1412:
We use the vm name "avocado-vt-vm1" with "-"

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -2917,7 +2917,7 @@ def snapshot_info(name, snapshot, **dargs):
         raise process.CmdError(cmd, sc_output, "Failed to get snapshot info")
 
     for val in values:
-        data = re.search("(?<=%s:) *\w*" % val, sc_output.stdout)
+        data = re.search("(?<=%s:) *\w.*" % val, sc_output.stdout)
         if data is None:
             continue
         ret[val] = data.group(0).strip()


kylazhang:
```
# virsh snapshot-info avocado-vt-vm1 --current
Name:           1521087681
Domain:         avocado-vt-vm1
Current:        yes
State:          running
Location:       internal
Parent:         -
Children:       0
Descendants:    0
Metadata:       yes
```
With your change, the parent if is "-", the result will be different from '' to '-', is it accessable?

hsmj1412:
@kylazhang Due to line 2925, the parent is '-' is ok.

hsmj1412:
@kylazhang Fixed it

virttest/virsh.py
```diff
@@ -2908,14 +2908,17 @@ def snapshot_info(name, snapshot, **dargs):
...
-    values = ["Name", "Domain", "Current", "State", "Parent",
+    values = ["Domain", "Current", "State", "Parent",
               "Children", "Descendants", "Metadata"]
 
     cmd = "snapshot-info %s %s" % (name, snapshot)
     sc_output = command(cmd, **dargs)
     if sc_output.exit_status != 0:
         raise process.CmdError(cmd, sc_output, "Failed to get snapshot info")
 
+    data = re.search("(?<=Name:) *[\w-]*", sc_output.stdout)
```
chunfuwen: try use regexp:"(?<=Name:) *\w.*"  ,which make it compatable with avocado-vt-vm1

hsmj1412: @chunfuwen Done

PR #1424, opened 2018-03-15
Python 3: Fixes dictionary iteration methods(2nd)

hsmj1412:


hereischen:
I would suggest change your commit message to something else.
Eg: fixed dictionary iteration introduced by [commit sha] .

hsmj1412:
@hereischen The file is created in #1411 , whatever I will change it commit

PR #1426, opened 2018-03-16
virsh.snapshot-info: Fix match issue

czhang0:
This pathc is fix to domain mathch issue of output of 'virsh snapshot-info' cmd.
Use of '\S' in re expression can match all contiguous non-whitespace characters.
While '\w' can only match alphanumeric characters. For case "avocado-vt-vm1", '\S'
match get "avocado-vt-vm1", '\w' match get "avocaodo".


czhang0:
Bug has been fixed by https://github.com/avocado-framework/avocado-vt/pull/1426.
So withdraw this PR.

PR #1432, opened 2018-03-16
Fix vcpu issues after refactor from tp-libvirt

kylazhang:
1. Adding missing checking point, not only for config, but also live
2. When not current vcpu in xml, set it to max

It also related to https://github.com/autotest/tp-libvirt/pull/1345

### DIFF
--- a/virttest/utils_hotplug.py
+++ b/virttest/utils_hotplug.py
@@ -51,7 +51,8 @@ def get_cpu_xmldata(vm, options=""):
     try:
         cpu_xmldata['current_vcpu'] = int(vm_xml.current_vcpu)
     except LibvirtXMLNotFoundError:
-        logging.debug("current vcpu value not present in xml")
+        logging.debug("current vcpu value not present in xml, set as max value")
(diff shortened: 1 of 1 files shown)


virttest/utils_hotplug.py
```diff
@@ -51,7 +51,8 @@ def get_cpu_xmldata(vm, options=""):
     try:
         cpu_xmldata['current_vcpu'] = int(vm_xml.current_vcpu)
     except LibvirtXMLNotFoundError:
-        logging.debug("current vcpu value not present in xml")
+        logging.debug("current vcpu value not present in xml, set as max value")
+        cpu_xmldata['current_vcpu'] = int(vm_xml.vcpu)
     cpu_xmldata['vcpu'] = int(vm_xml.vcpu)
```
chunfuwen: this statement can be removed

balamuruhans:
@kylazhang, does this PR is still valid ?, does autotest/tp-libvirt#1345 require this change

PR #1436, opened 2018-03-20
Python 3: Fix the __future__ line

hsmj1412:

### DIFF
--- a/virttest/unittest_utils/mock.py
+++ b/virttest/unittest_utils/mock.py
@@ -1,13 +1,14 @@
-__author__ = "raphtee@google.com (Travis Miller)"
-
+from __future__ import print_function
 import collections
 import re
 import sys
-from __future__ import print_function
 
 import six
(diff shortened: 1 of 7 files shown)


virttest/unittest_utils/mock.py
```diff
@@ -1,15 +1,16 @@
...
-from __future__ import print_function
 
 import six
 from six import StringIO
 from six import string_types as basestring
 
 
+__author__ = "raphtee@google.com (Travis Miller)"
+
+
```
yanan-fu: yes, move `from __future__ import print_function` to the first line is must, and could you explain why move `__author__` ?

hsmj1412: @yanan-fu I don't want to intersperse other commands in the import site

yanan-fu:
LGTM  and test pass (`python setup.py install `has no error)

PR #1437, opened 2018-03-20
add alias attribute into disk xml

chunfuwen:
alias attribute can be added into disk manually since libvirt
3.9.0


kylazhang:
I find there are some existing device xml already add alias as ElementDict, so what about make it uniform?

PR #1445, opened 2018-03-21
This patch will instaall targetcli-fb package to bionic version.

prudmirya:
This patch will instaall targetcli-fb package to bionic version. Existing tagetcli package name changed to targetcli-fb in bionic

### DIFF
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -843,7 +843,10 @@ class Iscsi(object):
             raise exceptions.TestError("Failed to install iscsi initiator"
                                        " packages")
         # Install linux iscsi target software targetcli
-        iscsi_package = ["targetcli"]
+        if distro.detect().version == '18' and distro.detect().name == 'Ubuntu':
+            iscsi_package = ["targetcli-fb"]
+        else:
+            iscsi_package = ["targetcli"]
         if not utils_package.package_install(iscsi_package):
             logging.error("Failed to install targetcli trying with scsi-"
                           "target-utils or tgt package")


virttest/iscsi.py
```diff
@@ -843,7 +843,10 @@ def create_iSCSI(params, root_dir=data_dir.get_tmp_dir()):
             raise exceptions.TestError("Failed to install iscsi initiator"
                                        " packages")
         # Install linux iscsi target software targetcli
-        iscsi_package = ["targetcli"]
+        if distro.detect().version == '18' and distro.detect().name == 'Ubuntu':
+            iscsi_package = ["targetcli-fb"]
+        else:
+            iscsi_package = ["targetcli"]
```
sathnaga: have ```try...
block, ```
to handle it generic for all distros, future some other distro versions might also have this naming.
...

prudmirya: if i try to install targetcli on 18.04 version it will not error out because wait_for_login(), and some times guest got crashed.

virttest/iscsi.py
```diff
@@ -843,7 +843,10 @@ def create_iSCSI(params, root_dir=data_dir.get_tmp_dir()):
             raise exceptions.TestError("Failed to install iscsi initiator"
                                        " packages")
         # Install linux iscsi target software targetcli
-        iscsi_package = ["targetcli"]
+        if distro.detect().version == '18' and distro.detect().name == 'Ubuntu':
```
balamuruhans: This would break in future Ubuntu releases,
```python
if distro.detect().name == 'Ubuntu' and int(distro.detect().version) >= 18:
    iscsi_package = ["targetcli-fb"]
else:
    iscsi_package = ["targetcli"]
```

virttest/iscsi.py
balamuruhans: The better way for this would to implement a API in utils_package that would give you the package name by querying irrespective of distro, like `yum whatprovides targetcli` in this case for RHEL based package manager and similar in Ubuntu

PR #1448, opened 2018-03-22
virttest.qemu_devices: Improve memory hotplug verification

huangyum:
1. Support "merge", "dump", "discard-data" option for memory backend
2. Improve memdev verify process
ID:  1559278, 1559279, 1470549

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -1878,7 +1878,7 @@ class DevContainer(object):
         Create pc-dimm device from params.
         """
         params = params.object_params("dimm")
-        dimm_type = 'nvdimm' if params.get('share') else 'pc-dimm'
+        dimm_type = "nvdimm" if params.get("nv_backend") else "pc-dimm"
         attrs = qdevices.Dimm.__attributes__[:]
         params = params.copy_from_keys(attrs)
         dev = qdevices.Dimm(params=params, dimm_type=dimm_type)
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -889,12 +889,12 @@ class Memory(QObject):
     """
     QOM memory object, support for pinning memory on host NUMA nodes.
     The existing options 'prealloc', 'mem-path', 'host-nodes', 'size',
-    'share', and 'backend' are subsumed by the QOM objects 'memory-backend-ram',
-    and 'memory-backend-file'.
+    'share', 'merge', 'dump', 'discard-data' and 'backend' are subsumed
+    by the QOM objects 'memory-backend-ram', and 'memory-backend-file'.
     """
 
-    __attributes__ = ["size", "prealloc", "mem-path",
-                      "backend", "policy", "host-nodes", "share"]
+    __attributes__ = ["size", "prealloc", "mem-path", "backend", "policy",
+                      "host-nodes", "share", "merge", "dump", "discard-data"]
 
     def __init__(self, backend, params=None):
         super(Memory, self).__init__(backend, params)
@@ -910,23 +910,57 @@ class Memory(QObject):
         if params.get("host-nodes"):
             host_nodes = list(map(int, params["host-nodes"].split()))
             params["host-nodes"] = host_nodes
+        for k in params:
+            params[k] = True if params[k] == "yes" else params[k]
+            params[k] = False if params[k] == "no" else params[k]
         kwargs = {"qom-type": backend,
                   "id": params.pop("id"),
                   "props": dict(params)}
         return "object-add", kwargs
 
     def verify_unplug(self, out, monitor):
-        out = monitor.info("memory-devices", debug=False)
-        memdev = "/object/%s" % self.get_qid()
-        if memdev not in out:
-            return True
-        return False
+        """
+        :param out: Output of the unplug command
+        :param monitor: Monitor used for unplug
+        :return: True when successful, False when unsuccessful
+        """
+        out = monitor.info("memdev", debug=False)
+        memdev = self.get_qid()
+        for dev in out:
+            if dev["id"] == memdev:
+                return False
+        return True
 
     def verify_hotplug(self, out, monitor):
-        out = monitor.info("memory-devices", debug=False)
-        memdev = "/object/%s" % self.get_qid()
-        if memdev in out:
-            return True
+        """
+        :param out: Output of the hotplug command
+        :param monitor: Monitor used for hotplug
+        :return: True when successful, False when unsuccessful
+        """
+        out = monitor.info("memdev", debug=False)
+        memdev = self.get_qid()
+        params = self.params.copy()
+        for dev in out:
+            if dev["id"] == memdev:
+                if not params.get("host-nodes") and len(dev["host-nodes"]):
+                    return False
+                if params.get("host-nodes"):
+                    host_nodes = list(map(int, params["host-nodes"].split()))
+                    if dev["host-nodes"].sort() != host_nodes.sort():
+                        return False
+                args = (params["size"], "B", 1024)
+                size = int(float(utils_misc.normalize_data_size(*args)))
+                if dev["size"] != size:
+                    return False
+                dev.pop("size")
+                dev.pop("host-nodes")
+                for k in dev:
+                    if params.get(k):
+                        params[k] = True if params[k] == "yes" else params[k]
+                        params[k] = False if params[k] == "no" else params[k]
+                        if dev[k] != params[k]:
+                            return False
+                return True
         return False
 
 
@@ -949,8 +983,8 @@ class Dimm(QDevice):
         out = monitor.info("memory-devices", debug=False)
         if "unknown command" in out:       # Old qemu don't have info qtree
             return out
-        dev_id_name = 'id "%s"' % self.get_qid()
-        if dev_id_name in out:
+        dev_id_name = self.get_qid()
+        if dev_id_name in str(out):
             return True
         return False
 
@@ -958,8 +992,8 @@ class Dimm(QDevice):
         out = monitor.info("memory-devices", debug=False)
         if "unknown command" in out:       # Old qemu don't have info qtree
             return out
-        dev_id_name = 'id "%s"' % self.get_qid()
-        if dev_id_name not in out:
+        dev_id_name = self.get_qid()
+        if dev_id_name not in str(out):
             return True
         return False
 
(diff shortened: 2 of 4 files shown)


vivianQizhu:
@huangyum I would suggest split it into different commits and describe your modification in each commit log.

huangyum:
@vivianQizhu How about I describe the intentions one by one in commit log other than split the commits? The code change is simple, I think it is ok to put them in single one commit. Thanks!

virttest/qemu_devices/qdevices.py
```diff
@@ -907,23 +907,46 @@ def hotplug_qmp(self):
         if params.get("host-nodes"):
             host_nodes = list(map(int, params["host-nodes"].split()))
             params["host-nodes"] = host_nodes
+        for k in params.keys():
```
xutian: for k in params:

virttest/qemu_devices/qdevices.py
```diff
@@ -907,23 +907,46 @@ def hotplug_qmp(self):
...
+                if not params.get("host-nodes") and not len(dev["host-nodes"]):
+                    return False
+                host_nodes = list(map(int, params["host-nodes"].split()))
+                args = (params["size"], "B", 1024)
+                size = int(float(utils_misc.normalize_data_size(*args)))
+                if dev["size"] != size or dev["host-nodes"] != host_nodes:
+                    return False
+                dev.pop("size")
+                dev.pop("host-nodes")
+                for k in dev.keys():
```
xutian: same

virttest/qemu_devices/qdevices.py
```diff
@@ -907,23 +907,46 @@ def hotplug_qmp(self):
...
+            if dev["id"] == memdev:
+                return False
+        return True
 
     def verify_hotplug(self, out, monitor):
-        out = monitor.info("memory-devices", debug=False)
-        memdev = "/object/%s" % self.get_qid()
-        if memdev in out:
-            return True
+        out = monitor.info("memdev", debug=False)
```
xutian: doc string ??

virttest/qemu_devices/qdevices.py
```diff
@@ -907,23 +907,46 @@ def hotplug_qmp(self):
...
                   "props": dict(params)}
         return "object-add", kwargs
 
     def verify_unplug(self, out, monitor):
-        out = monitor.info("memory-devices", debug=False)
-        memdev = "/object/%s" % self.get_qid()
-        if memdev not in out:
-            return True
-        return False
+        out = monitor.info("memdev", debug=False)
```
xutian: doc string ??

huangyum: Thanks, all updated, please help review again.

virttest/qemu_devices/qdevices.py
```diff
@@ -907,23 +907,56 @@ def hotplug_qmp(self):
...
+        :param out: Output of the hotplug command
+        :param monitor: Monitor used for hotplug
+        :return: True when successful, False when unsuccessful
+        """
+        out = monitor.info("memdev", debug=False)
+        memdev = self.get_qid()
+        params = self.params.copy()
+        for dev in out:
+            if dev["id"] == memdev:
+                if not params.get("host-nodes") and not len(dev["host-nodes"]):
```
xutian: useless check here

xutian: line#949  check has cover this check, so  suggest to move  line#949 host node check in here

huangyum: @xutian  You are right that the check for len(dev["host-nodes"]) makes no sense here. I've removed it. But if we move L949 here, we need move L946 too, then the line will exceed 80 chars  even after splitting the if sentence into 2 lines.  I'd like to keep it still.  Thanks.

virttest/qemu_devices/qdevices.py
```diff
@@ -907,23 +907,56 @@ def hotplug_qmp(self):
...
+            if params[k] == "yes":
+                params[k] = True
+            if params[k] == "no":
+                params[k] = False
         kwargs = {"qom-type": backend,
                   "id": params.pop("id"),
                   "props": dict(params)}
         return "object-add", kwargs
 
     def verify_unplug(self, out, monitor):
```
yanan-fu: what's the param `out` used for ?

virttest/qemu_devices/qdevices.py
```diff
@@ -907,23 +907,56 @@ def hotplug_qmp(self):
...
+        :return: True when successful, False when unsuccessful
+        """
+        out = monitor.info("memdev", debug=False)
+        memdev = self.get_qid()
+        for dev in out:
+            if dev["id"] == memdev:
+                return False
+        return True
 
     def verify_hotplug(self, out, monitor):
```
yanan-fu: Same with `verify_unplug`

virttest/qemu_devices/qdevices.py
```diff
@@ -907,23 +907,56 @@ def hotplug_qmp(self):
...
+                args = (params["size"], "B", 1024)
+                size = int(float(utils_misc.normalize_data_size(*args)))
+                if dev["size"] != size or dev["host-nodes"] != host_nodes:
+                    return False
+                dev.pop("size")
+                dev.pop("host-nodes")
+                for k in dev:
+                    if params.get(k):
+                        params[k] = True if params[k] == "yes" else params[k]
+                        params[k] = False if params[k] == "no" else params[k]
```
yanan-fu: Duplicate with L910~L914 ?

virttest/qemu_devices/qdevices.py
huangyum: @yanan-fu  I checked the whole file,  all `verify_unplug()` functions have same parameters. This param seems useless in this function, but if we want to remove it, we need to update all the other functions, and where they are called.  Let's don't rock the boat. Thanks!

virttest/qemu_devices/qdevices.py
huangyum: @yanan-fu Same as above :)

virttest/qemu_devices/qdevices.py
huangyum: @yanan-fu Not exactly, except `if params.get(k):`.   I'm not sure if it is necessary to create a function for it.

yanan-fu: @huangyum  From my understanding, it is same, no need to handle it twice, thx

vivianQizhu: Agree with @yanan-fu You don't need #line 955-956

yanan-fu: @huangyum or, why not use  True, False in case cfg directly.

huangyum: @yanan-fu Because qemu cmd and qmp cmd accept different values for some options.  e.g. For option "dump",  qemu cmd accepts "on/yes/y/off/no/n" while qmp cmd only accepts "True/False".

virttest/qemu_devices/qdevices.py
yanan-fu: @huangyum   A long-standing problem, no effect for our testing but better to solve it. To avoid new function call with the same method, it is up to you.
Anyway, it is not related with this patch, current code is ok for me :)

virttest/qemu_devices/qdevices.py
huangyum: @yanan-fu @vivianQizhu  How about a function which changes a value from "yes" to "True" and "no" to "False"?  That's the only common part.

virttest/qemu_devices/qdevices.py
```diff
@@ -907,23 +907,56 @@ def hotplug_qmp(self):
...
+        :param monitor: Monitor used for hotplug
+        :return: True when successful, False when unsuccessful
+        """
+        out = monitor.info("memdev", debug=False)
+        memdev = self.get_qid()
+        params = self.params.copy()
+        for dev in out:
+            if dev["id"] == memdev:
+                if not params.get("host-nodes"):
+                    return False
```
yanan-fu: As we confirmed,  faulty logic here.  For cases `- no_policy` in hotplug-mem.cfg, no `host-nodes` are defined, always return False.

huangyum:
@vivianQizhu @yanan-fu Please help review again.  I made more changes to validate the verification for memory hotplug.  Thanks!

virttest/utils_test/qemu/__init__.py
```diff
@@ -610,7 +610,9 @@ def hotplug_memory(self, vm, name):
                 dev_type = "pc-dimm"
             step = "Hotplug %s '%s' to VM" % (dev_type, dev.get_qid())
             error_context.context(step, logging.info)
-            vm.devices.simple_hotplug(dev, vm.monitor)
+            _, ver_out = vm.devices.simple_hotplug(dev, vm.monitor)
+            if ver_out is False:
+                raise exceptions.TestError("Verify hotplug memory failed")
```
yanan-fu: `exceptions.TestFail` is more appropriate.

huangyum: @yanan-fu You are right, updated, thanks!

yanan-fu:
LGTM
Ackde-by: Yanan Fu <yfu@redhat.com>

huangyum:
@balamuruhans Hi, could you please help review this pr, many thanks!

virttest/qemu_devices/qdevices.py
```diff
@@ -907,23 +907,57 @@ def hotplug_qmp(self):
...
+        out = monitor.info("memdev", debug=False)
+        memdev = self.get_qid()
+        params = self.params.copy()
+        for dev in out:
+            if dev["id"] == memdev:
+                if not params.get("host-nodes") and len(dev["host-nodes"]):
+                    return False
+                if params.get("host-nodes"):
+                    host_nodes = list(map(int, params["host-nodes"].split()))
+                    if dev["host-nodes"] != host_nodes:
```
balamuruhans: if dev["host-nodes"] is single element then it is better to check as,
```python
if dev["host-nodes"] in host_nodes:
```
or if dev["host-nodes"] is a list, then it is better to check as,
```python
if dev["host-nodes"].sort() != host_nodes.sort():
```

virttest/qemu_devices/qcontainer.py
```diff
@@ -1810,7 +1810,7 @@ def dimm_device_define_by_params(self, params, name):
         Create pc-dimm device from params.
         """
         params = params.object_params("dimm")
-        dimm_type = 'nvdimm' if params.get('share') else 'pc-dimm'
+        dimm_type = "nvdimm" if params.get("nv_backend") else "pc-dimm"
```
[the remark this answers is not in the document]
huangyum: @balamuruhans Thanks, have updated it in base.cfg.

PR #1450, opened 2018-03-22
Adding alias property for different devices

kylazhang:


bssrikanth:
can we have this patch merged? https://github.com/autotest/tp-libvirt/pull/1462 which is dependent patch has got merged. We are seeing test failures

PR #1452, opened 2018-03-23
Python 3: division fixer

hereischen:
`from __future__ import division` was added into all modules that
use the division operator, so that the / operator to behave the
same in Python 2 as it does in Python 3.

Changed / to //, if the operator should be changed to //.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -4,6 +4,7 @@ Utility classes and functions to handle Virtual Machine creation using qemu.
 :copyright: 2008-2009, 2014 Red Hat Inc.
 """
 
+from __future__ import division
 import time
 import os
 import logging
--- a/virttest/remote.py
+++ b/virttest/remote.py
@@ -1,6 +1,7 @@
 """
 Functions and classes used for logging into guests and transferring files.
 """
+from __future__ import division
 import logging
 import time
 import re
--- a/virttest/remote_commander/remote_master.py
+++ b/virttest/remote_commander/remote_master.py
@@ -6,6 +6,8 @@ Created on Dec 6, 2013
 :author: jzupka, astepano
 :contact: Andrei Stepanov <astepano@redhat.com>
 '''
+
+from __future__ import division
 import sys
 import time
 import inspect
--- a/virttest/rss_client.py
+++ b/virttest/rss_client.py
@@ -6,6 +6,7 @@ Client for file transfer services offered by RSS (Remote Shell Server).
 :copyright: 2008-2010 Red Hat Inc.
 """
 
+from __future__ import division
 import socket
 import struct
 import time
--- a/virttest/storage.py
+++ b/virttest/storage.py
@@ -5,6 +5,7 @@ This exports:
   - two functions for get image/blkdebug filename
   - class for image operates and basic parameters
 """
+from __future__ import division
 import logging
 import os
 import shutil
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -4,6 +4,7 @@ Virtualization test utility functions.
 :copyright: 2008-2009 Red Hat Inc.
 """
 
+from __future__ import division
 import time
 import string
 import random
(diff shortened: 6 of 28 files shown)


virttest/qemu_vm.py
```diff
@@ -4,6 +4,7 @@
 :copyright: 2008-2009, 2014 Red Hat Inc.
 """
 
+from __future__ import division
```
hsmj1412: Any other update for this file?

virttest/remote.py
```diff
@@ -1,6 +1,7 @@
 """
 Functions and classes used for logging into guests and transferring files.
 """
+from __future__ import division
```
hsmj1412: As above

virttest/remote_commander/remote_master.py
```diff
@@ -6,6 +6,8 @@
 :author: jzupka, astepano
 :contact: Andrei Stepanov <astepano@redhat.com>
 '''
+
+from __future__ import division
```
hsmj1412: As above

virttest/rss_client.py
```diff
@@ -6,6 +6,7 @@
 :copyright: 2008-2010 Red Hat Inc.
 """
 
+from __future__ import division
```
hsmj1412: As above

virttest/storage.py
```diff
@@ -5,6 +5,7 @@
   - two functions for get image/blkdebug filename
   - class for image operates and basic parameters
 """
+from __future__ import division
```
hsmj1412: As above

virttest/utils_misc.py
```diff
@@ -4,6 +4,7 @@
 :copyright: 2008-2009 Red Hat Inc.
 """
 
+from __future__ import division
```
hsmj1412: As above

virttest/utils_numeric.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import division
```
hsmj1412: As above

virttest/qemu_vm.py
hereischen: Do you mean do I need to update the codes in this file?

hereischen: I have viewed every line (unless I accidentally missed some ) that uses / operator, if it produces same results for both Python 2 and 3. I **did not** change the original code, just added `from __future__ import division` to make sure the / operator to behave the
same in Python 2 as it does in Python 3

I changed some  / to //, that is I believe it was intended to get a integer out of the division operation. If i did not update it, it would get a float number in Python 3.

hsmj1412: @hereischen If the codes don't need to be updated, the old division works well in Python 2. And the module is not necessary to import

hereischen: How can you make sure they are same without running it?

hsmj1412: I mean the result is integer in Python 2 before you import the module, and the avocado works well now(with integer). And you are sure it will also work well when it get a float. So if you think it is necessary to make sure the behave the same, // operator is better because it keep the original result.

chunfuwen:
@hereischen 
This is good initiative.
But, the only change between python 2 and 3 is  integer-division behavior.
So the proper way may be  'replace '/' with '//'' in related python file ,where integer division integer, rather than generally use 'from __future__ import division'

hereischen:
In Python 3, 5 / 2 == 2.5 and not 2; all division between int values result in a float. This change has actually been planned since Python 2.2 which was released in 2002. Since then users have been encouraged to add `from __future__ import division` to any and all files which use the `/` and `//` operators or to be running the interpreter with the -Q flag. If you have not been doing this then you will need to go through your code and do two things:

1. Add `from __future__ import division` to your files
2. Update any division operator as necessary to either use `//` to use floor division or continue using `/` and expect a float

The reason that `/` isn’t simply translated to `//` automatically is that if an object defines a `__truediv__` method but not `__floordiv__` then your code would begin to fail (e.g. a user-defined class that uses `/` to signify some operation but not `//` for the same thing or at all).

Reference: https://docs.python.org/3/howto/pyporting.html#division

luckyh:
Okay, let's merge this.

PR #1454, opened 2018-03-23
virttest/qemu-vm: add a get_block_virtual_size() to get block virtual size

peixiu:

id: 1411207 1390485

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -4315,6 +4315,20 @@ class VM(virt_vm.BaseVM):
                     return block['locked']
         return False
 
+    def get_block_virtual_size(self, dev_dict={}):
+        """
+        Get block virtual-size of specified block device
+        from monitor's info block command.
+
+        :param dev_dict: define device dict from block info.
+        :return: Matched block virual size.
+        """
+        blocks_info = self.monitor.info("block")
+        for dict in blocks_info:
+            if dev_dict['device'] in dict['device']:
+                size = dict['inserted']['image']['virtual-size']
+        return size
+
     def live_snapshot(self, base_file, snapshot_file,
                       snapshot_format="qcow2"):
         """


virttest/qemu_vm.py
```diff
@@ -4315,6 +4315,21 @@ def check_block_locked(self, value):
...
+        """
+        Get block virtual-size of specified block device
+        from monitor's info block command.
+
+        :param dev_dict: define device dict from block info.
+        :return: Matched block virual size.
+        """
+        blocks_info = self.monitor.info("block")
+        for dict in blocks_info:
+            if dev_dict.values()[0] in dict.values():
```
hereischen: This will not work in Python3, If you want get a value from dict, please use it's key
`Python 3.6.2 (default, Sep  1 2017, 12:03:48) 
[GCC 7.1.1 20170802 (Red Hat 7.1.1-7)] on linux
Type "help", "copyright", "credits" or "license" for more information.`
`>>> d1 = {"a":1, "b":2, "c":3}`
`>>> d1.values()`
`dict_values([1, 2, 3])`
`>>> d1.values()[0]`
`Traceback (most recent call last):`
`  File "<stdin>", line 1, in <module>`
`TypeError: 'dict_values' object does not support indexing`

peixiu: ok, thanks, will update it~

peixiu:
@hereischen @aliang123 could you help to review this patch again?
thanks a lot~

virttest/qemu_vm.py
```diff
@@ -4315,6 +4315,20 @@ def check_block_locked(self, value):
                     return block['locked']
         return False
 
+    def get_block_virtual_size(self, dev_dict={}):
```
xutian: please move this function to qemu_monitor.py

peixiu: Thanks for your advice~

peixiu:
Will move this function to qemu_monitor.py, so close this one.

virttest/qemu_vm.py
peixiu: Hi xutian, 

I moved this function on qemu_monitor.pyhttps://github.com/avocado-framework/avocado-vt/pull/1465/, but some colleagues confused this change.
For this patch, I have some explanation:
1. I written in qemu_vm.py before due to there is a get_block_dev function, I think get_block_virtual_size is similar with get_block_dev; 
2. I written in qemu_monitor.py, look on functional, I'm not so clear for you considers, I just see this function used monitor.info("block") directly.
3. I asked for xuhan, he think it's a function related storage device, it's better to write on  block related files, like storage.py. What do you think for you?

Thanks~

peixiu: Hi xutian,
Also discussed with xuhan, he think should write in a block related file, so create a new block.py file on this patch  https://github.com/avocado-framework/avocado-vt/pull/1490, later any block related functions or classes can write in block.py.

Best Rgards~

PR #1464, opened 2018-03-28
Define a new class "IozoneTest"

aliang123:
utils_misc: define class "IozoneTest" to define functions needed
in iozone test for linux/windows guest.

bug id: 1550800

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -3046,6 +3046,88 @@ def get_windows_drive_letters(session):
     return drive_letters
 
 
+class IozoneTest(object):
+    """
+    Run iozone test on guest
+    """
+    def __init__(self, session, params):
+        """
+        run setup of iozone test
+        """
+        self.params = params
+        self.session = session
+        self.timeout = float(params.get("disk_op_timeout", 240))
+        self.iozone_version = params.get("iozone_version", "iozone3_458")
+        self.dst_dir = params.get("iozone_dst", "/home")
+        if self.params.get("os_type") == "linux":
+            iozone_status = self.check_linux_iozone()
+            if not iozone_status:
+                self.setup_linux_iozone()
+
+    def setup_linux_iozone(self):
+        """
+        Download iozone and complie it
+        """
+        logging.info("Download iozone tarball")
+        iozone_url = self.params.get("iozone_url", "http://www.iozone.org/src/stable/iozone3_458.tar")
+        download_cmd = "wget %s -P %s" % (iozone_url, self.dst_dir)
+        self.session.cmd(download_cmd, timeout=self.timeout)
+        logging.info("Decompress iozone tarball")
+        decompress_cmd = "cd %s && tar -xvf %s.tar" % (self.dst_dir, self.iozone_version)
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -3045,6 +3045,79 @@ def get_windows_drive_letters(session):
     return drive_letters
 
 
+class IozoneTest(object):
```
balamuruhans: I would suggest to have a separate file utils_iozone.

virttest/utils_misc.py
```diff
@@ -3045,6 +3045,79 @@ def get_windows_drive_letters(session):
...
+    Run iozone test on guest
+    """
+    def __init__(self, session, params):
+        """
+        run setup of iozone test
+        """
+        self.params = params
+        self.session = session
+        self.timeout = float(params.get("disk_op_timeout", 240))
+        self.iozone_version = params.get("iozone_version")
```
balamuruhans: can we have placeholder for all the params introduced for iozone in cfg

virttest/utils_misc.py
```diff
@@ -3045,6 +3045,79 @@ def get_windows_drive_letters(session):
...
+        if self.params.get("os_type") == "linux":
+            iozone_check = True
+            while iozone_check:
+                output = self.session.cmd_output("ps aux | grep iozone | grep -v grep")
+                if "iozone" not in output:
+                    iozone_check = False
+                    break
+                time.sleep(5)
+            cmd = "cd /home &&"
+            cmd += "rm -rf iozone*"
```
balamuruhans: does setup_linux_iozone uses /home to download ?, how does this path have iozone

virttest/utils_misc.py
```diff
@@ -3045,6 +3045,79 @@ def get_windows_drive_letters(session):
...
+            self.session.cmd(cmd, timeout=self.timeout)
```
balamuruhans: checking the status would be better

virttest/utils_misc.py
vivianQizhu: How about move it to utils_test? Since I need to remove iozone test from utils_test.VMStress, its load_stress() function will lose iozone part there. Refer to https://github.com/avocado-framework/avocado-vt/pull/1467

balamuruhans: @vivianQizhu, Agree with your suggestion it can be placed in utils_test. Also in utils_test can we have a common class that would have code for setup/install, check(), cleanup/uninstall() part where IozoneTest/VMStress can avoid redundancy by reusing it ?

vivianQizhu: @balamuruhans Yes, these two have similar functions, I will check with @aliang123 to see how we can reorganise it.

balamuruhans: sure :+1:

aliang123: Discussed with vivianQizhu, it will be realized in utils_test.VMStress. So will close this patch, if the new patch are available, thanks for reviewing.

aliang123:
New patch for this issue: https://github.com/autotest/tp-qemu/pull/1388, close the old one.

PR #1465, opened 2018-03-28
virttest/qemu-monitor: add a get_block_virtual_size() to get block virtual size

peixiu:

id: 1411207 1390485


virttest/qemu_monitor.py
```diff
@@ -2302,6 +2302,20 @@ def block_resize(self, device, size):
...
+    def get_block_virtual_size(self, dev_dict={}):
+        """
+        Get block virtual-size of specified block device
+        from monitor's info block command.
+
+        :param dev_dict: define device dict from block info.
+        :return: Matched block virual size.
+        """
+        blocks_info = self.info("block")
+        for dict in blocks_info:
```
balamuruhans: dict is a keyword, let us use different name.

virttest/qemu_monitor.py
```diff
@@ -2302,6 +2302,20 @@ def block_resize(self, device, size):
...
+            if dev_dict['device'] in dict['device']:
```
balamuruhans: dev_dict is a default param, handle KeyError

virttest/qemu_monitor.py
peixiu: thanks, has updated

virttest/qemu_monitor.py
peixiu: has updated, it's no need to set the default value for this param,thanks~

virttest/qemu_monitor.py
```diff
@@ -2302,6 +2302,20 @@ def block_resize(self, device, size):
...
+        from monitor's info block command.
+
+        :param dev_dict: define device dict from block info.
+        :return: Matched block virual size.
+        """
+        blocks_info = self.info("block")
+        for dict_block in blocks_info:
+            if dev_dict['device'] in dict_block['device']:
+                size = dict_block['inserted']['image']['virtual-size']
+        return size
```
balamuruhans: please initialize `size`, code would break if condition fails.

peixiu: yes, thanks for you advise, will update it~

PR #1466, opened 2018-03-29
Python3: Overriding get() method in Params

hereischen:
As IterableUserDict is not supported in Python3, Params inherits
UserDict from collections, whereas it inherits IterableUserDict
from UserDict in Python2.

Due it's different parent class, `get` method need to be override
to catch ParamNotFound error riased by `__getitem__` called by
`get` in Python3.

### DIFF
--- a/virttest/utils_params.py
+++ b/virttest/utils_params.py
@@ -23,11 +23,18 @@ class Params(IterableUserDict):
     def __getitem__(self, key):
         """ overrides the error messages of missing params[$key] """
         try:
-            return IterableUserDict.__getitem__(self, key)
+            return super(Params, self).__getitem__(key)
         except KeyError:
             raise ParamNotFound("Mandatory parameter '%s' is missing. "
                                 "Check your cfg files for typos/mistakes" %
                                 key)
+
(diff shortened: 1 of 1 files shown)


virttest/utils_params.py
```diff
@@ -23,12 +23,19 @@ class Params(IterableUserDict):
...
         try:
-            return IterableUserDict.__getitem__(self, key)
+            return super(Params, self).__getitem__(key)
         except KeyError:
             raise ParamNotFound("Mandatory parameter '%s' is missing. "
                                 "Check your cfg files for typos/mistakes" %
                                 key)
 
+    def get(self, key, default=None):
+        'D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None.'
```
luckyh: Please make it more meaningful.

hereischen:
@hsmj1412 Thanks, that's a good suggestion.
The reason it does not use `not in` is because it will need `def __contains__` define in parents of `self`. Because `self` which is `Params` does not define it.
Since It has two parents, we'd better use [] as it needs `__getitem__ ` which it has.

PR #1467, opened 2018-03-29
utils_test: Refactor VMStress to support stress tests and remove autotest dependency

vivianQizhu:
Reframe VMStress, to remove the autotest dependency, only support stress test, exclude iozone and unixbench part, which should be implemented independently.

id: 1552382

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -1755,46 +1755,48 @@ class StressError(Exception):
 class VMStress(object):
 
     """
-    Run Stress tool in vms, such as stress, unixbench, iozone and etc.
+    Run Stress tool in vms, such as stress, iozone and etc.
     """
 
-    def __init__(self, vm, stress_type):
+    def __init__(self, vm, stress_type, download_url="", make_cmds="",
+                 stress_cmds="", stress_args="", work_path="",
+                 uninstall_cmds="", download_type="tarball"):
         """
-        Set parameters for stress type
-        """
+        Set parameters for stress type, for the arguments have default value "",
+        they can be either passed here, or defined in vm.params
 
-        def _parameters_filter(stress_type):
-            """Set parameters according stress_type"""
-            _control_files = {'unixbench': "unixbench5.control",
-                              'stress': "stress.control",
-                              'iozone': "iozone.control"}
-            _check_cmds = {'unixbench': "pidof -s ./Run",
-                           'stress': "pidof -s stress",
-                           'iozone': "pidof -s iozone"}
-            _stop_cmds = {'unixbench': "killall ./Run",
-                          'stress': "killall stress",
-                          'iozone': "killall iozone"}
-            try:
-                control_file = _control_files[stress_type]
-                self.control_path = os.path.join(data_dir.get_root_dir(),
-                                                 "shared/control",
-                                                 control_file)
-                self.check_cmd = _check_cmds[stress_type]
-                self.stop_cmd = _stop_cmds[stress_type]
-            except KeyError:
-                self.control_path = ""
-                self.check_cmd = ""
-                self.stop_cmd = ""
+        :param vm: the vm to be loading stress
+        :param stress_type: the name of the stress tool
+        :param download_url: from where download the stress tool
+        :param make_cmds: make command of the stress tool
+        :param uninstall_cmds: uninstall command of the stress tool
+        :param work_path: the relative work path of the stress tool,
+        e.g. for iozone: "src/current"
+        :param stress_cmds: the command to launch stress,
+        use stress_type instead if not defined
+        :param stress_args: the arguments of the stress tool
+        :param download_type: currently support "git" or "tarball"
+        """
 
         self.vm = vm
         self.params = vm.params
         self.timeout = 60
         self.stress_type = stress_type
-        if stress_type not in ["stress", "unixbench", "iozone"]:
-            raise StressError("Stress %s is not supported now." % stress_type)
-
-        _parameters_filter(stress_type)
-        self.stress_args = self.params.get("stress_args", "")
+        stress_cmds = stress_cmds or stress_type
+        self.stress_cmds = self.params.get('stress_cmds_%s' % stress_type, stress_cmds)
+        self.stress_args = self.params.get("%s_args" % stress_type, stress_args)
+        self.download_url = self.params.get('download_url_%s' % stress_type,
+                                            download_url)
+        self.download_type = download_type
+        self.base_name = self.download_url.split("/")[-1]
+        self.make_cmds = self.params.get('make_cmds_%s' % stress_type, make_cmds)
+        self.make_cmds = self.make_cmds or './configure && make install'
+        self.uninstall_cmds = self.params.get('uninstall_cmds_%s' % stress_type, uninstall_cmds)
+        self.uninstall_cmds = self.uninstall_cmds or './configure && make uninstall'
+        self.work_path = self.params.get('%s_work_path' % stress_type, work_path)
+        self.check_cmd = "pidof -s %s" % stress_type
+        self.stop_cmd = "pkill -9 %s" % stress_type
+        self.dst_path = self.params.get('stress_dst_path', '/home')
 
     def get_session(self):
         try:
@@ -1808,16 +1810,21 @@ class VMStress(object):
         """
         load stress tool in guest
         """
+        self.install()
         session = self.get_session()
-        command = run_autotest(self.vm, session, self.control_path,
-                               None, None, self.params, copy_only=True,
-                               control_args=self.stress_args)
-        session.cmd("%s &" % command)
-        logging.info("Command: %s", command)
-        running = utils_misc.wait_for(self.app_running, first=0.5, timeout=60)
-        if not running:
-            raise StressError("Stress tool %s isn't running"
-                              % self.stress_type)
+        session.cmd_output_safe('cd %s' % os.path.join(self.dst_path,
+                                                       self.base_name, self.work_path))
+        launch_cmds = 'nohup %s %s &' % (self.stress_cmds, self.stress_args)
+        error_context.context("Launch stress with command: %s" % launch_cmds)
+        try:
+            session.cmd_output_safe(launch_cmds)
+            # The background process sometimes does not return to
+            # terminate, if timeout, send a blank line afterward
+        except aexpect.ShellTimeoutError:
+            session.cmd_output_safe('')
+        session.close()
+        if not self.app_running():
+            raise exceptions.TestError("Stress app does not running as expected")
 
     @error_context.context_aware
     def unload_stress(self):
@@ -1827,6 +1834,7 @@ class VMStress(object):
         def _unload_stress():
             session = self.get_session()
             session.sendline(self.stop_cmd)
+            session.close()
             if not self.app_running():
                 return True
             return False
@@ -1841,7 +1849,86 @@ class VMStress(object):
         """
         session = self.get_session()
         status = session.cmd_status(self.check_cmd, timeout=60)
+        session.close()
         return status == 0
+
+    def _git_download(self, url, destination):
+        """
+        git download
+        :param url: download url
+        :param destination: download destination
+        """
+        from avocado.utils import git
+        destination = os.path.join(destination, self.base_name)
+        git.get_repo(url, destination_dir=destination)
+
+    def _tarball_download(self, url, destination):
+        """
+        tarball download and uncompress
+        :param url: dowload url
+        :param destination: download destination
+        """
+        import tarfile
+        tarball_path = os.path.join(destination, self.base_name)
+        download.get_file(url, tarball_path)
+        tarball = tarfile.open(tarball_path)
+        tarball.extractall(destination)
+        self.base_name = tarball.members[0].name.split("/")[0]
+        tarball.close()
+
+    @error_context.context_aware
+    def download_stress(self):
+        """
+        Download stress tool
+        """
+        url = self.download_url
+        tmp_path = data_dir.get_tmp_dir()
+        error_context.context('Download stress tool from %s' % url,
+                              logging.info)
+        download_method = getattr(self, "_%s_download" % self.download_type)
+        download_method(url, tmp_path)
+        error_context.context('Copy stress tool to guest', logging.info)
+        self.vm.wait_for_login()
+        self.vm.copy_files_to(os.path.join(tmp_path, self.base_name), self.dst_path)
+
+    @error_context.context_aware
+    def install(self):
+        """
+        To download, abstract, build and install the stress tool
+        """
+        self.download_stress()
+        install_path = os.path.join(self.dst_path, self.base_name, self.work_path)
+        session = self.get_session()
+        session.cmd_output_safe('cd %s' % install_path)
+        error_context.context('make and install the %s' %
+                              self.stress_type, logging.info)
+        status, output = session.cmd_status_output(self.make_cmds)
+        session.close()
+        if status != 0:
+            raise exceptions.TestError(
+                "Installation failed with output:\n %s" % output)
+
+    @error_context.context_aware
+    def clean(self):
+        """
+        Uninstall stress application, and clean the source files
+        """
+        session = self.get_session()
+        install_path = os.path.join(self.dst_path, self.base_name)
+        if session.cmd_status('cd %s' % install_path) != 0:
+            error_context.context("No source files found in path %s" % path)
+            session.close()
+            return
+
+        error_context.context('Uninstall %s' % self.stress_type, logging.info)
+        status, output = session.cmd_status_output(self.uninstall_cmds)
+        if status != 0:
+            error_context.context('Uninstall stress failed with '
+                                  'error: %s' % output, logging.error)
+        error_context.context('Remove the source files', logging.info)
+        rm_cmd = 'cd && rm -rf %s' % install_path
+        session.cmd_output_safe(rm_cmd)
+        session.close()
 
 
 class HostStress(object):
@@ -1934,20 +2021,33 @@ class HostStress(object):
         return result.exit_status == 0
 
 
-def load_stress(stress_type, vms, params):
+def load_stress(stress_type, params, vms=None, download_url="", make_cmds="",
+                stress_cmds="", stress_args="", work_path="",
+                uninstall_cmds="", download_type="tarball"):
     """
     Load stress for tests.
 
     :param stress_type: The stress type you need
     :param params: Useful parameters for stress
     :param vms: Used when it's stress in vms
+    :param download_url: from where download the stress tool
+    :param make_cmds: make command of the stress tool
+    :param uninstall_cmds: uninstall command of the stress tool
+    :param work_path: the relative work path of the stress tool,
+    e.g. for iozone: "src/current"
+    :param stress_cmds: the command to launch stress,
+    use stress_type instead if not defined
+    :param stress_args: the arguments of the stress tool
+    :param download_type: currently support "git" or "tarball"
     """
     fail_info = []
     # Add stress/iozone tool in vms
     if stress_type in ['stress_in_vms', 'iozone_in_vms']:
         for vm in vms:
             try:
-                vstress = VMStress(vm, stress_type.split('_')[0])
+                vstress = VMStress(vm, stress_type.split('_')[0], download_url,
+                                   make_cmds, stress_cmds, stress_args, work_path,
+                                   uninstall_cmds, download_type="tarball")
                 vstress.load_stress_tool()
             except StressError as detail:
                 fail_info.append("Launch stress in %s failed: %s"
@@ -1998,7 +2098,7 @@ def load_stress(stress_type, vms, params):
     return fail_info
 
 
-def unload_stress(stress_type, vms):
+def unload_stress(stress_type, vms=None):
     """
     Unload stress loaded by load_stress(...).
     """
--- a/shared/cfg/guest-os/Linux.cfg
+++ b/shared/cfg/guest-os/Linux.cfg
@@ -19,6 +19,8 @@
     cdrom_check_cdrom_pattern = "/dev/cdrom-\w+|/dev/cdrom\d*"
     cdrom_test_cmd = "dd if=%s of=/dev/null bs=1 count=1"
     cdrom_info_cmd = "cat /proc/sys/dev/cdrom/info"
+    stress_args = '--cpu 4 --io 4 --vm 2 --vm-bytes 256M'
+    download_url_stress = http://people.seas.harvard.edu/~apw/stress/stress-1.0.4.tar.gz
     timedrift, timerdevice..boot_test:
         i386, x86_64:
             extra_params += " -no-kvm-pit-reinjection"
@@ -132,10 +134,7 @@
         s4_log_chk_cmd = 'dmesg | grep -E "ACPI: Waking up from system sleep state S4"\|"ACPI: PCI interrupt for device .* disabled"'
         s4_start_cmd = 'echo disk > /sys/power/state'
     with_stress:
-        stress_test = autotest_control
-        test_control_file = stress.control
-        # Use a low stress to make sure guest can response during stress
-        control_args = "--cpu 4 --io 4 --vm 2 --vm-bytes 256M"
+        stress_test = linux_stress
         test_timeout = 1800
         check_cmd = 'pgrep stress'
         stop_cmd = "killall -g stress"


vivianQizhu:
@huangyum Could you help review together with https://github.com/autotest/tp-qemu/pull/1280 ? Thanks.

virttest/utils_test/__init__.py
```diff
@@ -1808,40 +1782,146 @@ def load_stress_tool(self):
...
-            raise StressError("Stress tool %s isn't running"
-                              % self.stress_type)
+        options = self.params.get('stress_args',
+                                  '--cpu 2 --io 1 --vm 1 --vm-bytes 128M')
+        try:
+            session.cmd_output_safe('nohup stress %s &' % options)
+        # The background process sometimes does not return to
+        # terminate, if timeout, send a blank line afterward
+        except aexpect.ShellTimeoutError:
+            session.cmd_output_safe('')
```
huangyum: What's this for?

virttest/utils_test/__init__.py
```diff
@@ -1808,40 +1782,146 @@ def load_stress_tool(self):
...
+            error_context.context('stop stress failed', logging.info)
+        session.close()
+
+    def app_running(self, timeout=5):
+        """
+        check whether app really run in background
+        """
+        session = self.get_session()
+        check_alive_cmd = self.params.get('check_alive_cmd', 'pgrep stress')
+        time.sleep(timeout)
```
huangyum: Why need sleep before check the status?

virttest/utils_test/__init__.py
```diff
@@ -1808,40 +1782,146 @@ def load_stress_tool(self):
...
+        status, output = session.cmd_status_output(uninstall_cmd)
+        if status != 0:
+            error_context.context('Uninstall stress failed with '
+                                  'error: %s' % output, logging.error)
+            session.close()
+            return False
+
+        error_context.context('Uninstall stress success, now '
+                              'remove the source files', logging.info)
+        rm_cmd = 'cd && rm -rf %s' % self.dst_path
```
huangyum: Do we need to remove tarball_path as well?

sitoliu: Why add `cd` in this command?

vivianQizhu:
NACK, need further adjustment

virttest/utils_test/__init__.py
balamuruhans: it can be rather '\n'

vivianQizhu: Thanks @balamuruhans Inside session.cmd_output_safe('') it add up a '\n' as 'linesep', so send '\n' will be transferred to '\n\n'. So I suppose that a blank command will be okay.

vivianQizhu:
Integrating iozone test, not sure if it is proper to handle the mountpoint option here, I'd prefer let user update the args outside, since the args specification is not common for all stress type. @aliang123 What do you think?

virttest/utils_test/__init__.py
```diff
@@ -1765,36 +1765,45 @@ def __init__(self, vm, stress_type):
...
-                          'iozone': "killall iozone"}
+            _download_url = {
+                'stress': 'http://people.seas.harvard.edu/~apw/stress/stress-1.0.4.tar.gz',
+                'iozone': 'http://www.iozone.org/src/stable/iozone3_458.tar'}
+            _tarball_name = {'stress': 'stress-1.0.4.tar.gz',
+                             'iozone': 'iozone3_458.tar'}
+            _pkg_md5 = {'stress': '890a4236dd1656792f3ef9a190cf99ef',
+                        'iozone': 'de170c3ae60d48ebc331d16beaaa01cb'}
+            _make_cmds = {'stress': './configure && make && make install',
+                          'iozone': 'cd src/current && make linux'}
```
balamuruhans: I was thinking something like this,
```python
class VMStress(object):
    def __init__(self, vm, stress_type):
        self.download_url = self.params.get("%s_url" % stress_type, "")
        self.tarball_name = self.params.get("%s_tarball" % stress_type, "")
```
where the params are defined in cfg,
```
# params for stress tests
stress_args = "stress --io 5 --vm 5 --cpu 5"
stress_url = "http://people.seas.harvard.edu/~apw/stress/stress-1.0.4.tar.gz"
stress_tarball = "stress-1.0.4.tar.gz"
download_type = "wget"
check_cmd = "example"

# params for iozone tests
iozone_url = "http://www.iozone.org/src/stable/iozone3_458.tar"
iozone_tarball = "iozone3_458.tar"
download_type = "wget"
check_cmd = "example"
```
```python
self.download(download_type):
    """
    :param type: git, wget to be used to download the package
    """
self.install(make_cmd)
self.run()
self.check(check_cmd)
self.clean()
```
This should help adding new one in future by simple configuration changes and minimal code changes respective to that new test,
new_args = "new --option1 --option2"
new_test_object = VMStress(vm, new)

vivianQizhu: @balamuruhans Yes, that's better for extension, I will refine it. Thanks.

vivianQizhu: @balamuruhans I still prefer give a series default value. Since it seems not proper to put these values into somewhere like base.cfg, so if we define them in each specific test case cfg, we will have many same parameters everywhere in different cfg. These values are not like qemu test matrix parameters, they are not likely to change frequently for different test cases.
I agree that it might prevent the convenience for future extension. So I might implement it with @aliang123 's option, allowing pass default values when initialize the class, as well as define them in cfg. Something like:
```
class VMStress(object):
    def __init__(self, vm, stress_type, download_url="", tarball_name=""):
        self.download_url = self.params.get("%s_url" % stress_type, download_url)
        self.tarball_name = self.params.get("%s_tarball" % stress_type, tarball_name)
```

balamuruhans: @vivianQizhu @aliang123 I agree with you guys.

sathnaga: @vivianQizhu This PR will be good a feature for multiple tests, Thanks for it!

vivianQizhu:
@balamuruhans @aliang123 Updated, please help review, thanks.

vivianQizhu:
Test with stress-1.0.4.tar.gz:
16:07:57 INFO | Context: (load_stress_tool) --> (install) --> Download stress tool from http://people.seas.harvard.edu/~apw/stress/stress-1.0.4.tar.gz
16:07:57 INFO | Fetching http://people.seas.harvard.edu/~apw/stress/stress-1.0.4.tar.gz -> /var/tmp/avocado_z6P5US/stress-1.0.4.tar.gz
16:07:58 DEBUG| Sending command (safe): rm -f /var/tmp/avocado_z6P5US/stress-1.0.4.tar.gz
16:07:58 INFO | Context: (load_stress_tool) --> (install) --> Copy stress tool to guest
16:07:58 DEBUG| Trying to SCP with command 'scp -v -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PreferredAuthentications=password -r  -P 22 /var/tmp/avocado_z6P5US/stress-1.0.4 root@\[10.66.10.21\]:/home', timeout 600s
16:07:59 DEBUG| Got password prompt, sending 'kvmautotest'
16:07:59 DEBUG| SCP process terminated with status 0
16:07:59 INFO | Copy file from /var/tmp/avocado_z6P5US/stress-1.0.4 to 10.66.10.21:/home, elapsed time: 1.01753091812
16:07:59 DEBUG| Sending command (safe): cd /home/stress-1.0.4
16:07:59 INFO | Context: (load_stress_tool) --> make and install the stress
16:07:59 DEBUG| Sending command (safe): ./configure && make && make install
16:08:09 DEBUG| Sending command (safe): cd /home/
16:08:09 DEBUG| Sending command (safe): nohup stress --cpu 4 --io 4 --vm 2 --vm-bytes 256M &
16:08:10 DEBUG| Sending command: pidof -s stress
16:08:11 DEBUG| Sending command: echo $?
16:08:23 WARNI| IPv6 address sniffing is not supported yet by using TShark, please fallback to use other sniffers by uninstalling TShark when testing with IPv6
16:09:11 INFO | Context: stop stress app in guest
16:09:13 DEBUG| wait stress app quit (2.001626 secs)
16:09:13 DEBUG| Sending command: pidof -s stress
16:09:13 DEBUG| Sending command: echo $?
16:09:13 DEBUG| Sending command: cd /home/stress-1.0.4
16:09:13 DEBUG| Sending command: echo $?
16:09:13 INFO | Context: Uninstall stress
16:09:13 DEBUG| Sending command: ./configure && make uninstall
16:09:16 DEBUG| Sending command: echo $?
16:09:16 INFO | Context: Remove the source files
16:09:16 DEBUG| Sending command (safe): cd && rm -rf /home/stress-1.0.4

vivianQizhu:
Git download test:
16:09:18 INFO | Context: (load_stress_tool) --> (install) --> Download stress tool from https://github.com/stressapptest/stressapptest
16:09:18 DEBUG| Creating directory /var/tmp/avocado_z6P5US/stressapptest for git repo https://github.com/stressapptest/stressapptest
16:09:18 DEBUG| Initializing new git repo at /var/tmp/avocado_z6P5US/stressapptest for receiving git repo https://github.com/stressapptest/stressapptest
16:09:18 INFO | Running '/bin/git init'
16:09:18 DEBUG| [stdout] Initialized empty Git repository in /var/tmp/avocado_z6P5US/stressapptest/.git/
16:09:18 INFO | Command '/bin/git init' finished with 0 after 0.0015230178833s
16:09:18 INFO | Fetching git [REP 'https://github.com/stressapptest/stressapptest' BRANCH 'master'] -> /var/tmp/avocado_z6P5US/stressapptest
16:09:18 INFO | Running '/bin/git fetch -q -f -u -t https://github.com/stressapptest/stressapptest master:master'
16:09:22 INFO | Command '/bin/git fetch -q -f -u -t https://github.com/stressapptest/stressapptest master:master' finished with 0 after 4.67026400566s
16:09:22 DEBUG| Checking out branch master
16:09:22 INFO | Running '/bin/git checkout master'
16:09:22 DEBUG| [stderr] Already on 'master'
16:09:22 INFO | Command '/bin/git checkout master' finished with 0 after 0.00785994529724s
16:09:22 DEBUG| Specific commit not specified
16:09:22 INFO | Running '/bin/git log --pretty=format:%H -1'
16:09:22 INFO | Command '/bin/git log --pretty=format:%H -1' finished with 0 after 0.000777006149292s
16:09:22 DEBUG| [stdout] dd550fcf808a97a822a7d32ae18b1ee9a5213001
16:09:22 INFO | Running '/bin/git describe'
16:09:22 DEBUG| [stdout] v1.0.8-10-gdd550fc
16:09:22 INFO | Command '/bin/git describe' finished with 0 after 0.00102114677429s
16:09:22 INFO | git commit ID is dd550fcf808a97a822a7d32ae18b1ee9a5213001 (tag v1.0.8-10-gdd550fc)
16:09:22 INFO | Context: (load_stress_tool) --> (install) --> Copy stress tool to guest
16:09:22 DEBUG| Trying to SCP with command 'scp -v -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PreferredAuthentications=password -r  -P 22 /var/tmp/avocado_z6P5US/stressapptest root@\[10.66.10.21\]:/home', timeout 600s
16:09:23 DEBUG| Got password prompt, sending 'kvmautotest'
16:09:23 DEBUG| SCP process terminated with status 0
16:09:23 INFO | Copy file from /var/tmp/avocado_z6P5US/stressapptest to 10.66.10.21:/home, elapsed time: 0.988945007324
16:09:23 DEBUG| Sending command (safe): cd /home/stressapptest
16:09:24 INFO | Context: (load_stress_tool) --> make and install the stressapptest
16:09:24 DEBUG| Sending command (safe): ./configure && make && make install
16:09:43 DEBUG| Sending command (safe): cd /home/
16:09:44 DEBUG| Sending command (safe): nohup stressapptest -s 20 -M 256 -m 8 -C 8 -W &
16:09:44 DEBUG| Sending command: pidof -s stressapptest
16:09:45 DEBUG| Sending command: echo $?
16:10:45 INFO | Context: stop stressapptest app in guest
16:10:47 DEBUG| wait stressapptest app quit (2.002027 secs)
16:10:47 DEBUG| Sending command: pidof -s stressapptest
16:10:47 DEBUG| Sending command: echo $?
16:10:47 DEBUG| Sending command: cd /home/stressapptest
16:10:47 DEBUG| Sending command: echo $?
16:10:47 INFO | Context: Uninstall stressapptest
16:10:47 DEBUG| Sending command: ./configure && make uninstall
16:10:53 DEBUG| Sending command: echo $?
16:10:53 INFO | Context: Remove the source files
16:10:53 DEBUG| Sending command (safe): cd && rm -rf /home/stressapptest

virttest/utils_test/__init__.py
```diff
@@ -1808,25 +1811,28 @@ def load_stress_tool(self):
...
+                                                       self.base_name, self.work_path))
+        launch_cmds = 'nohup %s %s &' % (self.stress_cmds, self.stress_args)
+        error_context.context("Launch stress with command: %s" % launch_cmds)
+        try:
+            session.cmd_output_safe(launch_cmds)
+            # The background process sometimes does not return to
+            # terminate, if timeout, send a blank line afterward
+        except aexpect.ShellTimeoutError:
+            session.cmd_output_safe('')
+
```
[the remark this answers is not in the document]
vivianQizhu: @huangyum I won't close the session until clean() method is called, this whole class will use a same session for different functions, otherwise we will get too many sessions for different action, which is hard to debug.

virttest/utils_test/__init__.py
```diff
@@ -1839,10 +1845,85 @@ def app_running(self):
...
+        :param destination: download destination
+        """
+        import tarfile
+        tarball_path = os.path.join(destination, self.base_name)
+        download.get_file(url, tarball_path)
+        tarball = tarfile.open(tarball_path)
+        tarball.extractall(destination)
+        self.base_name = tarball.members[0].name.split("/")[0]
+        tarball.close()
+        self.session.cmd_output_safe("rm -f %s" % tarball_path)
```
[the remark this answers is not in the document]
vivianQizhu: @huangyum Yes, this is a mistake, it is intended to be removed from host. Thanks.

virttest/utils_test/__init__.py
```diff
@@ -1839,10 +1845,85 @@ def app_running(self):
...
+
+        uninstall_cmd = './configure && make uninstall'
+        error_context.context('Uninstall %s' % self.stress_type, logging.info)
+        status, output = session.cmd_status_output(uninstall_cmd)
+        if status != 0:
+            error_context.context('Uninstall stress failed with '
+                                  'error: %s' % output, logging.error)
+        error_context.context('Remove the source files', logging.info)
+        rm_cmd = 'cd && rm -rf %s' % install_path
+        session.cmd_output_safe(rm_cmd)
```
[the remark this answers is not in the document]
vivianQizhu: @huangyum I supposed we do not need, because I download it to the temporary path `data_dir.get_tmp_dir()`, I think it will be removed by the environment.

virttest/utils_test/__init__.py
```diff
@@ -1755,46 +1755,49 @@ def __str__(self):
...
+        self.stress_cmds = self.params.get('stress_cmds_%s' % stress_type, stress_cmds)
+        self.stress_args = self.params.get("%s_args" % stress_type, stress_args)
+        self.download_url = self.params.get('download_url_%s' % stress_type,
+                                            download_url)
+        self.download_type = download_type
+        self.base_name = self.download_url.split("/")[-1]
+        self.make_cmds = self.params.get('make_cmds_%s' % stress_type, make_cmds)
+        self.work_path = self.params.get('%s_work_path' % stress_type, work_path)
+        self.check_cmd = "pidof -s %s" % stress_type
+        self.stop_cmd = "killall %s" % stress_type
```
balamuruhans: can we use `"pkill -9 %s" % stress_type` to force the stress process to kill ? what would you say

virttest/utils_test/__init__.py
```diff
@@ -1839,10 +1845,84 @@ def app_running(self):
...
+    def install(self):
+        """
+        To download, abstract, build and install the stress tool
+        """
+        self.download_stress()
+        install_path = os.path.join(self.dst_path, self.base_name)
+        self.session.cmd_output_safe('cd %s' % install_path)
+        error_context.context('make and install the %s' %
+                              self.stress_type, logging.info)
+        self.session.cmd_output_safe(self.make_cmds)
```
balamuruhans: we need to handle exception here, if installation fails let us raise TestError

virttest/utils_test/__init__.py
```diff
@@ -1839,10 +1845,84 @@ def app_running(self):
...
+        """
+        Download stress tool
+        """
+        url = self.download_url
+        tmp_path = data_dir.get_tmp_dir()
+
+        error_context.context('Download stress tool from %s' % url,
+                              logging.info)
+        download_method = getattr(self, "_%s_download" % self.download_type)
+        download_method(url, tmp_path)
```
balamuruhans: :+1:

virttest/utils_test/__init__.py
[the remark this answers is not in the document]
balamuruhans: we can use `data_dir.clean_tmp_files()`

virttest/utils_test/__init__.py
```diff
@@ -1755,46 +1755,49 @@ def __str__(self):
...
+        self.download_type = download_type
+        self.base_name = self.download_url.split("/")[-1]
+        self.make_cmds = self.params.get('make_cmds_%s' % stress_type, make_cmds)
+        self.work_path = self.params.get('%s_work_path' % stress_type, work_path)
+        self.check_cmd = "pidof -s %s" % stress_type
+        self.stop_cmd = "killall %s" % stress_type
+        self.dst_path = self.params.get('stress_dst_path', '/home')
+        self.check_installed_cmds = self.params.get('check_installed_cmds_%s' %
+                                                    stress_type, check_installed_cmds)
+        self.session = self.get_session()
```
balamuruhans: why to use self.session here, self.get_session() returns Bool ?

virttest/utils_test/__init__.py
```diff
@@ -1808,25 +1811,28 @@ def load_stress_tool(self):
...
+        launch_cmds = 'nohup %s %s &' % (self.stress_cmds, self.stress_args)
+        error_context.context("Launch stress with command: %s" % launch_cmds)
+        try:
+            session.cmd_output_safe(launch_cmds)
+            # The background process sometimes does not return to
+            # terminate, if timeout, send a blank line afterward
+        except aexpect.ShellTimeoutError:
+            session.cmd_output_safe('')
+
+        return self.app_running()
```
balamuruhans: we need to assert whether stress is running or not, this would return Bool but caller doesn't know whether stress started properly or not.

virttest/utils_test/__init__.py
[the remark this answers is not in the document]
vivianQizhu: @balamuruhans Maybe other cases will use this temporary directory at the same time, I am afraid that I will delete the files that still under use. Avocado will execute it at the end of its run() function, I suppose that it would be better to clean them after all cases finished.

virttest/utils_test/__init__.py
vivianQizhu: @balamuruhans Same, since this class has no `test` as its property, It only notice user whether the function be executed properly by return a bool value, the caller should has its own `test` object, which can raise the error there by itself. I assume that it is a 'function' class instead of a 'test' class.

virttest/utils_test/__init__.py
vivianQizhu: @balamuruhans Every time I call self.get_session(), it will open up a new session and return it. I previously use separated sessions in each functions by calling get_session() and close it when the function finished its task. But this will lead to massive session logs in test result folder, which is hard to debug. So I only open up one session in the beginning, and use this one session for all functions, and close it until clean() is called. Actually I am not really sure which method is better...

virttest/utils_test/__init__.py
balamuruhans: @vivianQizhu, IMHO your API `utils_test.VMStress` should be independent of the test or caller, we can use `exceptions.TestError()` because even if the test code or caller is not aware of, API should be consistent and do what it suppose to.

virttest/utils_test/__init__.py
balamuruhans: I have mentioned in another comment to make use of `exceptions.TestError`

virttest/utils_test/__init__.py
vivianQizhu: @balamuruhans It return session in line 1805. But I agree that closing the session immediately after using it will be safer, I will modify it.

virttest/utils_test/__init__.py
vivianQizhu: @balamuruhans Yes, it makes sense. I will modify it. Thanks.

virttest/utils_test/__init__.py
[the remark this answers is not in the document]
balamuruhans: Agreed! Thank you.

virttest/utils_test/__init__.py
```diff
@@ -1755,46 +1755,48 @@ def __str__(self):
...
+        self.stress_cmds = self.params.get('stress_cmds_%s' % stress_type, stress_cmds)
+        self.stress_args = self.params.get("%s_args" % stress_type, stress_args)
+        self.download_url = self.params.get('download_url_%s' % stress_type,
+                                            download_url)
+        self.download_type = download_type
+        self.base_name = self.download_url.split("/")[-1]
+        self.make_cmds = self.params.get('make_cmds_%s' % stress_type, make_cmds)
+        self.work_path = self.params.get('%s_work_path' % stress_type, work_path)
+        self.check_cmd = "pkill -9 %s" % stress_type
+        self.stop_cmd = "killall %s" % stress_type
```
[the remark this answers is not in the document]
vivianQizhu: @balamuruhans Yes, my mistake...thanks.

virttest/utils_test/__init__.py
```diff
@@ -1755,46 +1755,48 @@ def __str__(self):
...
+        self.dst_path = self.params.get('stress_dst_path', '/home')
+        self.check_installed_cmds = self.params.get('check_installed_cmds_%s' %
```
aliang123: useless param "self.check_installed_cmds"

virttest/utils_test/__init__.py
```diff
@@ -1755,46 +1755,44 @@ def __str__(self):
...
 
     """
-    Run Stress tool in vms, such as stress, unixbench, iozone and etc.
+    Run Stress tool in vms, such as stress, iozone and etc.
     """
 
-    def __init__(self, vm, stress_type):
-        """
-        Set parameters for stress type
+    def __init__(self, vm, stress_type, download_url="", make_cmds="",
```
sathnaga: `utils_test.load_stress(stress_type, vms, params):` should be made aware of these function params for tests to make use of it incase need.

virttest/utils_test/__init__.py
```diff
@@ -1841,8 +1845,88 @@ def app_running(self):
...
+        Uninstall stress application, and clean the source files
+        """
+        session = self.get_session()
+        install_path = os.path.join(self.dst_path, self.base_name)
+        if session.cmd_status('cd %s' % install_path) != 0:
+            error_context.context("No source files found in path %s" % path)
+            session.close()
+            return
+
+        uninstall_cmd = './configure && make uninstall'
```
sathnaga: @vivianQizhu uninstall cmd does not require ./configure again and it is better to get from user like `make_cmds`

sathnaga:
@vivianQizhu PR looks good overall, I just tested it with my test, https://github.com/autotest/tp-libvirt/pull/1530 with this patch and it was seemless without any other changes to my testcase apart from params setting :+1: 

```
(1/1) guest_multiboot.qemu.qcow2.virtio_scsi.smp2.virtio_net.Guest.RHEL.7.devel.ppc64le.powerkvm-libvirt.multivm_cpustress.stress_suspend: PASS (316.55 s)
```
Here are few recommendations
1) My test config needed below changes to make it work, 
```
stress_args = "--cpu 4 --io 4 --vm 2 --vm-bytes 256M"
download_url_stress = http://people.seas.harvard.edu/~apw/stress/stress-1.0.4.tar.gz
make_cmds_stress = "./configure && make install"
stress_work_path = ''
```
it would be good if we can document this as a default config setting in `base.cfg` and usage of this method in commit msg, would be helpful for future reference for users.
2) lets make uninstall_cmd as a test param as comment above.
3) we can actually identify the download_url type from the link and choose the method [can be TODO not for now]

Thanks for improvement :+1:

balamuruhans:
For "stress_on_host", `vms` parameter is not necessary so we can make it default parameter, necessary changes in the order of parameter in caller functions have to be taken care
```python
def load_stress(stress_type, params, vms=None):
```

https://github.com/avocado-framework/avocado-vt/blob/707e465f8788731758b50987f19c5f2a58f15e59/virttest/utils_test/__init__.py#L2008-L2015

vivianQizhu:
@sathnaga Thanks for your comments, I feel that it is not very proper to put these params in base.cfg, iozone needs define different params for it anyway, and it might differ from Windows guest. How about we put it into guest-os/Linux.cfg?

virttest/utils_test/__init__.py
vivianQizhu: @sathnaga In my test, sometimes `make uninstall` will fail if don't `./configure` first, this might rely on the process between `make uninstall` and `make install`, I wish this function could be more independent, and I supposed that with `./configure` here is harmless, so allow me keep it here. But true it should be configurable, I will move it to init params. Thanks for review.

vivianQizhu:
@sathnaga @balamuruhans Thanks for your review, I updated it with below changes:
1. Make `uninstall_cmds` configurable, and set default value in `__init__` function.
2. Set default value for `make_cmds` in `__init__` function.
3. Set default `stress_args` and `download_url_stress` in guest-os/Linux.cfg
4. Redefine `load_stress()` as above comment described. I found no where call this function in current code, so no other related modification.

@sathnaga This way you don't need any extra configuration.

balamuruhans:
@vivianQizhu lastly could you fix,

```python
def unload_stress(stress_type, vms=None):
```
Thank you!

vivianQizhu:
@balamuruhans Updated, fixed `unload_stress()` with `vms=None`. Thanks.

virttest/utils_test/__init__.py
```diff
@@ -1841,8 +1849,87 @@ def app_running(self):
...
+        self.vm.wait_for_login()
+        self.vm.copy_files_to(os.path.join(tmp_path, self.base_name), self.dst_path)
+
+    @error_context.context_aware
+    def install(self):
+        """
+        To download, abstract, build and install the stress tool
+        """
+        self.download_stress()
+        install_path = os.path.join(self.dst_path, self.base_name)
```
aliang123: for iozone: install_path = os.path.join(self.dst_path, self.base_name, self.work_path)

virttest/utils_test/__init__.py
```diff
@@ -1755,46 +1755,48 @@ def __str__(self):
...
     """
-    Run Stress tool in vms, such as stress, unixbench, iozone and etc.
+    Run Stress tool in vms, such as stress, iozone and etc.
     """
 
-    def __init__(self, vm, stress_type):
-        """
-        Set parameters for stress type
+    def __init__(self, vm, stress_type, download_url="", make_cmds="",
+                 uninstall_cmds="", work_path="", stress_cmds="",
```
aliang123: for iozone, no "uninstall_cmds" needed. Suggest:
  def __init__(self, vm, stress_type, download_url="", make_cmds="", work_path="", stress_cmds="", stress_args="",  uninstall_cmds="", download_type="tarball") 
or something like that.

virttest/utils_test/__init__.py
vivianQizhu: @aliang123 That's right, thanks for correction, updated.

virttest/utils_test/__init__.py
vivianQizhu: @aliang123 Get your point, move it backward. And also move work_path backward, since stress does not need it. Thanks.

balamuruhans:
Thanks all, let us merge it.

PR #1470, opened 2018-03-30
Python 3 : fix pickle issue in __safe_env_savet

chunfuwen:
previously use hasattr(stacktrace, "str_unpickable_object") to condition judgement,
but actually can not dertermin whether env.data is really picklale or not
so in python3,it will raise ValueError when calling stacktrace.str_unpickable_object()

### DIFF
--- a/avocado_vt/test.py
+++ b/avocado_vt/test.py
@@ -20,6 +20,7 @@ import imp
 import logging
 import os
 import sys
+import pickle
 try:
     import queue as Queue
 except ImportError:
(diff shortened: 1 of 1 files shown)


avocado_vt/test.py
```diff
@@ -223,7 +224,9 @@ def __safe_env_save(self, env):
         try:
             env.save()
         except Exception as details:
-            if hasattr(stacktrace, "str_unpickable_object"):
+            try:
+                pickle.dumps(env.data)
+            except Exception:
                 self.log.warn("Unable to save environment: %s",
                               stacktrace.str_unpickable_object(env.data))
             else:    # TODO: Remove when 36.0 LTS is not supported
```
kylazhang: The "TODO" comments need to be removed as it will be used in the future. And what type of exception it catch during env.save?

chunfuwen: I assume 36.0 LTS is still used on RHEL6.9. The main logic here is that env.data may be one pickable object, maybe not in some cases,so need carefully manage this.

PR #1471, opened 2018-03-30
python 3 : fix bytes call string interface issue

chunfuwen:
process.system_output() return bytes instead of string on python 3
therefore need to decode() before call any string interfaces

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -793,13 +793,13 @@ def preprocess(test, params, env):
     if kvm_userspace_ver_cmd:
         try:
             kvm_userspace_version = avocado_process.system_output(
-                kvm_userspace_ver_cmd, shell=True).strip()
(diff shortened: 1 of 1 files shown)


virttest/env_process.py
```diff
@@ -793,13 +793,13 @@ def preprocess(test, params, env):
     if kvm_userspace_ver_cmd:
         try:
             kvm_userspace_version = avocado_process.system_output(
-                kvm_userspace_ver_cmd, shell=True).strip()
+                kvm_userspace_ver_cmd, shell=True).decode().strip()
```
luckyh: I think it should be handled by `process.system_output()`, otherwise, each place which invoked this function would acquire this change.

chunfuwen: that's true,but process.system_output was ever changed by someone to return bytes forcedly. Not sure whether there is special consideration.
But anyway, I will submit one patch to restore return value to text

PR #1473, opened 2018-03-30
Add support for vcpu hotplug through qemu monitor for powerpc

sathnaga:
Add support for vcpu hotplug through qemu monitor for powerpc
guests

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -25,6 +25,7 @@ import threading
 import time
 import sys
 import aexpect
+import platform
 
 from avocado.core import exceptions
 from avocado.utils import path as utils_path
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -3050,29 +3051,61 @@ def hotplug_domain_vcpu(domain, count, by_virsh=True, hotplug=True):
...
+                for item in range(0, int(count), threads):
+                    if item < vcpu_count:
+                        continue
+                    cmd = "device_add host-spapr-cpu-core,id=core%d,core-id=%d" % (item, item)
+                    result = virsh.qemu_monitor_command(vm.name, cmd, "--hmp",
+                                                        debug=True)
+            else:
+                for item in range(int(count), vcpu_count, threads):
+                    cmd = "device_del core%d" % item
+                    result = virsh.qemu_monitor_command(vm.name, cmd, "--hmp",
```
balamuruhans:
```python
if ppc:
    if hotplug:
        cmd = <>
    else:
        cmd = <>
else:
    if hotplug:
        cmd = <>
    else:
        cmd = <>
return virsh.qemu_monitor_command(cmd)
```

sathnaga: we could do that but problem is we need multiple `virsh.qemu_monitor_command` to be called for ppc... thats not the case with `cpu-add`

balamuruhans: we can form list of commands where for `cpu-add` it would be only one, also have to check the result for each iteration, current logic would return the result of last iteration it might miss if failure occurs before last iteration,
```python
for cmd in cmds:
    result = virsh.qemu_monitor_command(cmd)
    libvirt.check_exit_status(result)
return result
```

sathnaga: agree on the result checks, will do check exit_status for ppc and return last...

PR #1478, opened 2018-04-03
Python 3: Decodes bytes to str

hsmj1412:
1. Add a new method to do the change
2. Decodes the url bytes result to str

### DIFF
--- a/virttest/compat_52lts.py
+++ b/virttest/compat_52lts.py
@@ -16,6 +16,9 @@
 This module contains helpers that allows running Avocado-vt with Avocado
 master as well as with 52.x LTS release.
 """
+import sys
+
+from six import string_types
 
 
 def results_stdout_52lts(result):
@@ -40,3 +43,17 @@ def results_stderr_52lts(result):
         return result.stderr_text
     else:   # 52lts stores string
         return result.stderr
+
+
+def decode_to_text(stream, encoding=sys.getdefaultencoding()):
+    """
+    Decode decoding string
+    :param stream: string stream
+    :param encoding: encode_type
+    :return: encoding text
+    """
+    if hasattr(stream, 'decode'):
+        return stream.decode(encoding)
+    if isinstance(stream, string_types):
+        return stream
+    raise TypeError("Unable to decode stream into a string-like type")
--- a/virttest/http_server.py
+++ b/virttest/http_server.py
@@ -12,6 +12,8 @@ except ImportError:
     from BaseHTTPServer import HTTPServer
     from SimpleHTTPServer import SimpleHTTPRequestHandler
 
+from virttest.compat_52lts import decode_to_text
+
 
 class HTTPRequestHandler(SimpleHTTPRequestHandler):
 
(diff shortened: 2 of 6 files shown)


virttest/http_server.py
```diff
@@ -52,7 +54,7 @@ def copyfile_range(self, source_file, output_file, range_begin, range_end):
         output_file.write(buf)
 
     def send_head_range(self, range_begin, range_end):
-        path = self.translate_path(self.path)
+        path = self.translate_path(decode_to_text(self.path))
```
chunfuwen: this may not place-in to decode_to_text, translate_path() method may be more suitable place.

virttest/utils_2to3_wrap.py
```diff
@@ -0,0 +1,14 @@
+"""
+Port Python2 to Python3 utility functions.
+Please make sure do NOT import other modules from virttest
+
+"""
+from six import string_types
+
+
+def decode_to_text(stream, encoding='utf-8'):
```
chunfuwen: suggest to hard code encoding to uft-8, but use encoding = sys.getdefaultencoding()

virttest/virsh.py
```diff
@@ -983,7 +984,7 @@ def driver(**dargs):
     """
     # libvirt schme composed of driver + command
     # ref: http://libvirt.org/uri.html
-    scheme = urllib.parse.urlsplit(canonical_uri(**dargs))[0]
+    scheme = urllib.parse.urlsplit(canonical_uri(decode_to_text(**dargs)))[0]
```
chunfuwen: I don't see necessity to decode_to_text(**dargs) here

virttest/utils_2to3_wrap.py
```diff
@@ -0,0 +1,14 @@
...
+    if hasattr(stream, 'decode'):
+        return stream.decode(encoding)
+    if isinstance(stream, string_types):
+        return stream
```
kylazhang: The order for instance() and hasattr() should exchange?

virttest/virsh.py
hsmj1412: @chunfuwen I tried to make sure the type of parameter is text

virttest/http_server.py
hsmj1412: Fixed

virttest/utils_2to3_wrap.py
hsmj1412: Updated

hsmj1412:
@kylazhang @chunfuwen Updated

virttest/http_server.py
```diff
@@ -93,7 +95,7 @@ def translate_path(self, path):
 
         """
         # abandon query parameters
-        path = urlparse(path)[2]
+        path = urlparse(decode_to_text(path))[2]
```
chunfuwen: I don't see the necessity for decode_to_text (path)

virttest/compat_52lts.py
```diff
@@ -40,3 +43,17 @@ def results_stderr_52lts(result):
         return result.stderr_text
     else:   # 52lts stores string
         return result.stderr
+
+
+def decode_to_text(stream, encoding=sys.getdefaultencoding()):
+    """
+    Decode decoding string
+    :param stream: string stream
+    :param encoding: string type
```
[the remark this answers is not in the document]
hsmj1412: Updated

virttest/http_server.py
hsmj1412: The urlparse will also parse the bytes string.

PR #1483, opened 2018-04-05
libvirt_vm: get uptime of the vm instance

balamuruhans:
API to get the uptime of vm instance, this would be helpful
in migration testcases where asserting the vm uptime before
and after migration validates whether vm went down/restart
during migration


sathnaga:
lets merge this...

PR #1490, opened 2018-04-10
virttest: add a new block.py file and write in a function get_block_virtual_size()

peixiu:
Create a new block.py file, related block functions or classes can write in this file.

id: 1411207 1390485

### DIFF
--- a/virttest/block.py
+++ b/virttest/block.py
@@ -1,0 +1,19 @@
+"""Utility functions to handle block related actions."""
+
+from . import qemu_monitor
+
+
+def get_block_virtual_size(dev_name):
+    """
+    Get block virtual-size of specified block device with dev_name.
+
+    :param dev_name: block device name, like "drive-hotadd" etc.
+    :return: Matched block device virual size.
+             The default unit of the size is Byte.
+             eg: disk image is 30G, should return 32212254720.
+    """
+    blocks_info = qemu_monitor.info("block")
+    for dict_block in blocks_info:
+        if dev_name == dict_block['device']:
+            size = dict_block['inserted']['image']['virtual-size']
+            return int(size)


peixiu:
@XueqiangWei @xiagao could you help to review this patch? thanks a lot~

virttest/block.py
```diff
@@ -0,0 +1,22 @@
...
+"""
+
+from . import qemu_monitor
+
+
+def get_block_virtual_size(dev_dict):
+    """
+    Get block virtual-size of specified block device
+    from monitor's info block command.
+
```
xiagao: From the code, I believe you want to use qmp but not hmp, so pls describe it here.

virttest/block.py
```diff
@@ -0,0 +1,22 @@
...
+    :param dev_dict: define device dict from block info.
+    :return: Matched block virual size.
+    """
+    blocks_info = qemu_monitor.info("block")
+    for dict_block in blocks_info:
+        if dev_dict['device'] in dict_block['device']:
```
xiagao: why not use "==" ?

virttest/block.py
```diff
@@ -0,0 +1,22 @@
...
+            size = dict_block['inserted']['image']['virtual-size']
+    if size:
+        return size
```
xiagao: It's better to handle size=none situation

peixiu: Hi xiagao, has updated code, could you help to review it again?

virttest/block.py
xiagao: Get block virtual-size of specified block device with dev_name  ***from the qemu monitor***

peixiu: is you mean from qmp monitor?

virttest/block.py
```diff
@@ -0,0 +1,21 @@
...
+
+    :param dev_name: block device name, like "drive-hotadd" etc.
+    :return: Matched block device virual size.
+    """
+    blocks_info = qemu_monitor.info("block")
+    for dict_block in blocks_info:
+        if dev_name == dict_block['device']:
+            size = dict_block['inserted']['image']['virtual-size']
+            return int(size)
+    return None
```
XueqiangWei: it will return None by default, so I think no need to add \"return None\" in the end

peixiu: oo, thanks, will update it~

virttest/block.py
```diff
@@ -0,0 +1,20 @@
+"""
+Utility functions to handle block related actions.
+
```
hereischen: There is a ci check error, please fix it.
if it is a one line doc string, please put it this way,
`"""Utility functions to handle block related actions."""`

virttest/block.py
```diff
@@ -0,0 +1,20 @@
...
+from . import qemu_monitor
+
+
+def get_block_virtual_size(dev_name):
+    """
+    Get block virtual-size of specified block device with dev_name.
+
+    :param dev_name: block device name, like "drive-hotadd" etc.
+    :return: Matched block device virual size.
```
hereischen: Please add measurement unit of the size, eg. M, G

virttest/block.py
peixiu: ok, thanks, will upodate~

virttest/block.py
peixiu: ok, thanks, will update

miyachen:
@YongxueHong Hi Yongxue, please help review this pr from power perspective, thanks.

virttest/block.py
```diff
@@ -0,0 +1,19 @@
...
+             The default unit of the size is Byte.
+             eg: disk image is 30G, should return 32212254720.
+    """
+    blocks_info = qemu_monitor.info("block")
+    for dict_block in blocks_info:
+        if dev_name == dict_block['device']:
+            size = dict_block['inserted']['image']['virtual-size']
+            return int(size)
```
YongxueHong: Suggest this function should be move to tp-qemu/provider,  I think it is belong to case level. this function just only get the size of image, so single..

peixiu:
Hi all, this function can be replaced by follows commands, can reduce a new function here.
    img = QemuImg(data_image_params, data_dir.get_data_dir(), data_image)
    block_virtual_size = json.loads(img.info(force_share=True,
                                    output="json"))["virtual-size"]
I added these codes to patch https://github.com/autotest/tp-qemu/pull/1302.
Closed this patch~

PR #1493, opened 2018-04-12
Python 3: port autotest.client.utils.run_parallel()

cuixiangz:

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -32,6 +32,11 @@ try:
 except ImportError:
     from BytesIO import BytesIO
 
+try:
+    basestring
+except NameError:
+    basestring = (str, bytes)
+
 import xml.etree.ElementTree as ET
 
 from aexpect.utils.genio import _open_log_files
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -4139,7 +4141,7 @@ def __init__(self, command, stdout_tee=None, stderr_tee=None, verbose=True,
 
         # Allow for easy stdin input by string, we'll let subprocess create
         # a pipe for stdin input and we'll write to it in the wait loop
-        if isinstance(stdin, basestring):
+        if isinstance(stdin, string_types):
```
kylazhang: If stdin is byte, it will fail. I saw some usage with "basestring = (str,bytes)" and not use six func. @chunfuwen @hsmj1412  please help to confirm, thanks

chunfuwen: One way follow as:
```
try:
      basestring
except NameError:
      basestring = (str, bytes)
```

czhang0: Below change can match expectation. For u'a', b'a' and 'a' both in Python 2 or Python 3 are all type of str or bytes.
```
if isinstance(str(stdin), (str, bytes)):
```

czhang0:
Below change can match expectation. For u'a', b'a' and 'a' both in Python 2 or Python 3 are all type of str or bytes.
```
if isinstance(str(stdin), (str, bytes)):
```

virttest/utils_misc.py
kylazhang: I think chwen's suggestion should be better as str(unicode) should have some problem?

czhang0: Yes, I just modified according to your suggestion.

PR #1496, opened 2018-04-13
utils_test.libvirt: add function to check skip

dzhengfy:
We will have more and more test scripts designed for multiple arches
with different machine types. One arch can only support several special
machine types. So this is to determine if the case for special machine
type can run on the host.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -25,6 +25,7 @@ import threading
 import time
 import sys
 import aexpect
+import platform
 
 from avocado.core import exceptions
 from avocado.utils import path as utils_path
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -3390,3 +3391,25 @@ def change_boot_order(vm_name, device_tag, boot_order, index=0):
...
     vmxml.sync()
+
+
+def check_machine_type_arch(machine_type):
+    """
+    Determine if one case for special machine
+    type can be run on the host
+
+    :param machine_type: the machine type the case is designed for
+
```
chunfuwen: remove addition empty blank line

virttest/utils_test/libvirt.py
```diff
@@ -3390,3 +3391,25 @@ def change_boot_order(vm_name, device_tag, boot_order, index=0):
...
+    :raise test.cancel: Skip if the host arch does not match
```
chunfuwen: test.cancel to exceptions.skip

PR #1498, opened 2018-04-13
Enable libvirt vm start in preprocess

sathnaga:
This patch enables the libvirt vm start in preprocess
without explicit need of `import` test to called
provided `create_vm_libvirt` params is enabled by test.

This feature helps us incase of multi vm boot scenarios
where as framework already gives us support to create/clone images,
just an additional step of create vm is required incase of
libvirt which is available incase of qemu already,
here we are safely assuming import test considering
we have working guest images present.

Signed-off-by: Mallesh Koti <mallesh@linux.vnet.ibm.com>

### DIFF
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -421,6 +421,11 @@ skip_image_check_during_running = no
 # skip cluster leak warning message in image check
 skip_cluster_leak_warn = no
 
+# Create libvirt multi vms
+# To enable multi libvirt vms setup, make create_vm_libvirt="yes" in addition
+# to vms = vm1 vm2 vm3 and make sure `master_images_clone = img1` not be None
+# inorder to get three vms cloned, provided main_vm = vm1
+create_vm_libvirt = no
 # Some preprocessor params
 # If there is any conflict between 'start_vm' and 'kill_vm_before_test',
 # the final desicion is made by start_vm.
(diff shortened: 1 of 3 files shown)


virttest/storage.py
```diff
@@ -446,7 +446,7 @@ def clone_image(params, vm_name, image_name, root_dir):
                     logging.info("Clone master image for vms.")
                     process.run(params.get("image_clone_command") %
                                 (m_image_fn, image_fn))
-
+            params["image_name_%s" % vm_name] = vm_image_name
             params["image_name_%s_%s" % (image_name, vm_name)] = vm_image_name
```
balamuruhans: can we reuse the same param ?

shared/cfg/base.cfg
```diff
@@ -416,6 +416,11 @@ skip_image_check_during_running = no
 # skip cluster leak warning message in image check
 skip_cluster_leak_warn = no
 
+# Create libvirt multi vms
+# To enable multi libvirt vms setup,make create_vm_libvirt="yes" in addition
```
balamuruhans: typo to include a space
`, make create_vm_libvirt="yes" in`

shared/cfg/base.cfg
```diff
@@ -416,6 +416,11 @@ skip_image_check_during_running = no
...
+# to vms = vm1 vm2 vm3 and make sure `master_images_clone = img1` not be None
+#inorder to get three vms cloned, provided main_vm = vm1
```
balamuruhans: typo a space after `#`

virttest/storage.py
sathnaga: Few other tests uses that old param so, I do not want to disturb it for backward compatibility and the newly added param is required to keep the created image mapped to multi vm.

sathnaga:
@balamuruhans done :-)

PR #1502, opened 2018-04-16
virt_vm: use connect_uri to verify guest dmesg on remote host

balamuruhans:
Libvirt connect uri can be used to retrieve guest dmesg from
remote host incase we migrate the guest, this patch would
enable it for verify_dmesg().


sitoliu:
Looks good to me.

PR #1507, opened 2018-04-17
tar: Swing options around to suit new tar

dagrh:
New gnu tar (1.30 ish) objects to --exclude being passed after the
directory, e.g.

tar -cvzf /dev/null /tmp --exclude=frob
....
tar: The following options were used after any non-optional arguments in archive create or update mode.  These options are positional and affect only arguments that follow them.  Please, rearrange them properly.
tar: --exclude ‘frob’ has no effect
tar: Exiting with failure status due to previous errors

unfortunately we do that in a few places.

Swing the options around to always put the directory at the end.

Signed-off-by: Dr. David Alan Gilbert <dgilbert@redhat.com>

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -921,9 +921,10 @@ def run_avocado(vm, params, test, testlist=[], timeout=3600,
         # result info tarball to host result dir
         results_tarball = os.path.join(test_path, "results.tgz")
         compress_cmd = "cd %s && " % result_path
-        compress_cmd += "tar cjvf %s ./*" % results_tarball
+        compress_cmd += "tar cjvf %s" % results_tarball
         compress_cmd += " --exclude=*core*"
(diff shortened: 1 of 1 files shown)


virttest/utils_test/__init__.py
```diff
@@ -921,9 +921,10 @@ def get_results(test_path, result_path, session):
         # result info tarball to host result dir
         results_tarball = os.path.join(test_path, "results.tgz")
         compress_cmd = "cd %s && " % result_path
-        compress_cmd += "tar cjvf %s ./*" % results_tarball
+        compress_cmd += "tar cjvf%s" % results_tarball
```
xutian: Please leave a blank space between options 'cjvf' and results_tarball

xutian: Other changes looks Ok to me.

dagrh: Fixed.

PR #1522, opened 2018-04-23
Python 3: Force open file with text instead of  "bytes" to make python 2 and 3 compatible

hahakiki2010:
Change file open mode without "b" to get it compatible both
py2 and py3. The old way openning file with "rb" & "wb" return bytes
instead of string in Python3, and this will broke tp-libvirt testing
as tp-libvirt use string in most cases.

### DIFF
--- a/virttest/element_tree.py
+++ b/virttest/element_tree.py
@@ -578,7 +578,7 @@ class ElementTree(object):
 
     def parse(self, source, parser=None):
         if not hasattr(source, "read"):
-            source = open(source, "rb")
+            source = open(source, "r")
         if not parser:
             parser = XMLTreeBuilder()
         while 1:
@@ -657,7 +657,7 @@ class ElementTree(object):
     def write(self, file, encoding="us-ascii"):
         assert self._root is not None
         if not hasattr(file, "write"):
-            file = open(file, "wb")
+            file = open(file, "w")
         if not encoding:
             encoding = "us-ascii"
         elif encoding != "utf-8" and encoding != "us-ascii":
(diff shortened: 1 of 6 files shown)


virttest/http_server.py
```diff
@@ -69,7 +69,7 @@ def send_head_range(self, range_begin, range_end):
             # Always read in binary mode. Opening files in text mode may cause
             # newline translations, making the actual size of the content
             # transmitted *less* than the content-length!
```
chunfuwen: it need special care here,see above comments

hahakiki2010: Accepted, will not change to text

virttest/rss_client.py
```diff
@@ -199,7 +199,7 @@ def _receive_packet(self, timeout=60):
     def _send_file_chunks(self, filename, timeout=60):
         if self._log_func:
             self._log_func("Sending file %s" % filename)
-        f = open(filename, "rb")
+        f = open(filename, "r")
```
chunfuwen: This may deal with binary file, so suggest keep code as before.

virttest/rss_client.py
```diff
@@ -217,7 +217,7 @@ def _send_file_chunks(self, filename, timeout=60):
     def _receive_file_chunks(self, filename, timeout=60):
         if self._log_func:
             self._log_func("Receiving file %s" % filename)
-        f = open(filename, "wb")
+        f = open(filename, "w")
```
chunfuwen: see previous comment

virttest/rss_client.py
hahakiki2010: Ok, will change back to rb

virttest/rss_client.py
hahakiki2010: sam as above

PR #1523, opened 2018-04-24
Python 3: Minor fix for expat parser

hahakiki2010:
It's to fix the Attribute Error issue when with python3, and
get compatible with python2 at the same time

### DIFF
--- a/virttest/element_tree.py
+++ b/virttest/element_tree.py
@@ -1164,7 +1164,7 @@ class XMLTreeBuilder(object):
         except AttributeError:
             pass
         encoding = None
-        if not parser.returns_unicode:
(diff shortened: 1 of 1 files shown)


virttest/element_tree.py
```diff
@@ -1164,8 +1164,9 @@ def __init__(self, html=0, target=None):
         except AttributeError:
             pass
         encoding = None
-        if not parser.returns_unicode:
-            encoding = "utf-8"
+        if hasattr(parser, "returns_unicode"):
+            if not parser.returns_unicode:
```
chunfuwen: combine two if into one line

hahakiki2010: Accepted, thanks

PR #1526, opened 2018-04-24
Python 3: Fixup issue unable to save environment

hahakiki2010:
This issue is to fix following issue with Python3, as “Pickling” 
converts a object into a byte stream instead of text
Unable to save environment: write() argument must be str, not bytes
({'version': 1, 'address_cache': {}, 'vm__avocado-vt-vm1':
<virttest.libvirt_vm.VM object at 0x7fa58f432a90>})


chunfuwen:
@hahakiki2010 ,please update commit message

PR #1528, opened 2018-04-25
Undefine libvirt vm using param during postprocess

sathnaga:
This patch introduces a new param 'kill_vm_libvirt' to undefine
a libvirt vm during test postprocess given that param is set.

This would be helpful incase of cleaning the env properly for
next test with right set of params(create_vm_libvirt) we can
create a newer vm based on given params for next test.

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -385,6 +385,8 @@ def postprocess_vm(test, params, env, name):
         if kill_vm_timeout:
             utils_misc.wait_for(vm.is_dead, kill_vm_timeout, 0, 1)
         vm.destroy(gracefully=params.get("kill_vm_gracefully") == "yes")
+        if params.get("kill_vm_libvirt") == "yes" and params.get("vm_type") == "libvirt":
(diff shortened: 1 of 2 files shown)


virttest/env_process.py
```diff
@@ -385,6 +385,8 @@ def postprocess_vm(test, params, env, name):
         if kill_vm_timeout:
             utils_misc.wait_for(vm.is_dead, kill_vm_timeout, 0, 1)
         vm.destroy(gracefully=params.get("kill_vm_gracefully") == "yes")
+        if params.get("kill_vm_libvirt") == "yes" and params.get("vm_type") == "libvirt":
+            vm.undefine()
```
balamuruhans: This would undefine vm from main_vm, we should also consider if multi vms are in env

balamuruhans: got it. yes, it works for me too.

balamuruhans:
Thanks @sathnaga for patch, let's merge it.

PR #1532, opened 2018-04-26
Update for migration params

hsmj1412:


hsmj1412:
@balamuruhans These params are not nfs only, so move them out

hsmj1412:
Line 810: The server_ip is not defined when storage type is not nfs

PR #1536, opened 2018-04-27
Python 3: drop usage of raw unicode str

hereischen:

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -4208,17 +4208,30 @@ class VM(virt_vm.BaseVM):
                 if match:
                     return block.split(":")[0]
         else:
-            for block in blocks_info:
-                match = True
-                for key, value in six.iteritems(p_dict):
-                    if isinstance(value, bool):
-                        check_str = "u'%s': %s" % (key, value)
+            # handles QMP output
+            def traverse_nested_dict(d):
+                iters = [six.iteritems(d)]
+                while iters:
+                    item = iters.pop()
+                    try:
+                        k, v = item.next()
+                    except StopIteration:
+                        continue
+                    iters.append(item)
+                    if isinstance(v, dict):
+                        iters.append(six.iteritems(v))
                     else:
-                        check_str = "u'%s': u'%s'" % (key, value)
-                    if check_str not in str(block):
-                        match = False
-                        break
-                if match:
+                        yield k, v
+            for block in blocks_info:
+                matched = True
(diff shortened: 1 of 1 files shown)


hereischen:
passed on both Python 2 and 3

virttest/qemu_vm.py
```diff
@@ -4209,17 +4224,10 @@ def get_block_old(self, blocks_info, p_dict={}):
...
-                        check_str = "u'%s': %s" % (key, value)
-                    else:
-                        check_str = "u'%s': u'%s'" % (key, value)
-                    if check_str not in str(block):
-                        match = False
-                        break
-                if match:
-                    return block['device']
+                    for (k, v) in traverse_nested_dict(block):
+                        if (k == key) and (value == v or value in v):
```
luckyh: Should not have `value in v` because it may get the wrong item.
For example, the vm has the following two block devices, and you want to find device id by the filename `image1`,
```
block100: filename="image100"
block1: filename="image1"
```
finally, it would return `block100` because its filename does contain the pattern `image1`.

virttest/qemu_vm.py
```diff
@@ -4209,17 +4224,10 @@ def get_block_old(self, blocks_info, p_dict={}):
                     return block.split(":")[0]
         else:
```
luckyh: This `else` clause handles QMP output, so we can add comment to help understand it.

virttest/qemu_vm.py
```diff
@@ -4209,17 +4224,10 @@ def get_block_old(self, blocks_info, p_dict={}):
...
-                    else:
-                        check_str = "u'%s': u'%s'" % (key, value)
-                    if check_str not in str(block):
-                        match = False
-                        break
-                if match:
-                    return block['device']
+                    for (k, v) in traverse_nested_dict(block):
+                        if (k == key) and (value == v or value in v):
+                            return block['device']
```
luckyh: Should do this when all the conditions are matched, not one of them.

virttest/qemu_vm.py
```diff
@@ -36,6 +36,21 @@
 from virttest.qemu_devices import qdevices, qcontainer
 
 
+def traverse_nested_dict(d):
```
luckyh: Suggest to move this function into `qemu_monitor.py` as an utility, since users do have other needs to handle this kind of QMP data.

virttest/qemu_vm.py
hereischen: updated thanks.

PR #1542, opened 2018-04-28
qemu_vm: add more option support for balloon device

huangyum:
Add support for option "deflate-on-oom" and "guest-stats-polling-interval"
id: 1572866, 1573806

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1255,9 +1255,16 @@ class VM(virt_vm.BaseVM):
                 numa_cmd += ",nodeid=%s" % nodeid
             return numa_cmd
 
-        def add_balloon(devices, devid=None, bus=None, use_old_format=None):
+        def add_balloon(devices, devid=None, bus=None,
+                        use_old_format=None, options={}):
             """
             This function is used to add balloon device
+
+            :param devices: device container object
+            :param devid: device id
+            :param bus: the bus balloon device use
+            :param use_old_format: use old format or not
+            :param options: optional keyword arguments
             """
             if not devices.has_option("device") or use_old_format is True:
                 devices.insert(StrDev('balloon', cmdline=" -balloon virtio"))
@@ -1271,6 +1278,8 @@ class VM(virt_vm.BaseVM):
             dev = QDevice(model, parent_bus=bus)
             if devid:
                 dev.set_param("id", devid)
+            for key, value in options.items():
+                dev.set_param(key, value)
             devices.insert(dev)
 
         def add_virtio_option(option, value, devices, dev, dev_type):
@@ -2237,8 +2246,13 @@ class VM(virt_vm.BaseVM):
             if balloon_params.get("balloon_dev_add_bus") == "yes":
                 balloon_bus = _get_pci_bus(
                     devices, balloon_params, 'balloon', True)
+            options = {}
+            deflate_on_oom = balloon_params.get("balloon_opt_deflate_on_oom")
+            options["deflate-on-oom"] = deflate_on_oom
+            guest_polling = balloon_params.get("balloon_opt_guest_polling")
+            options["guest-stats-polling-interval"] = guest_polling
             add_balloon(devices, devid=balloon_devid, bus=balloon_bus,
-                        use_old_format=use_ofmt)
+                        use_old_format=use_ofmt, options=options)
 
         # Add qemu options
         if params.get("msg_timestamp"):


virttest/qemu_vm.py
```diff
@@ -1269,6 +1276,8 @@ def add_balloon(devices, devid=None, bus=None, use_old_format=None):
             dev = QDevice(model, parent_bus=bus)
             if devid:
                 dev.set_param("id", devid)
+            for k in kwargs:
+                dev.set_param(k, kwargs[k])
```
yanan-fu: How about following format:
```
for key, value in kwargs.items():
    dev.set_param(key, value)
```

huangyum: Yes, this could work too. But I'm not sure which one is better or both are ok.

virttest/qemu_vm.py
```diff
@@ -2245,8 +2254,14 @@ def sort_key(dev):
             if balloon_params.get("balloon_dev_add_bus") == "yes":
                 balloon_bus = _get_pci_bus(
                     devices, balloon_params, 'balloon', True)
+            kwargs = {}
+            if balloon_params.get("deflate-on-oom") == "yes":
+                kwargs["deflate-on-oom"] = "yes"
+            if balloon_params.get("guest-polling"):
+                val = balloon_params.get("guest-polling", 0)
+                kwargs["guest-stats-polling-interval"] = val
```
yanan-fu: In order to run L2261 and L2262,   balloon_params.get("guest-polling") can't be "0" in L2260.
But you set the default value for "guest-polling" to "0" in L2261. it is unnecessary from my understanding.

Does following code meet what you want to do ?
```
kwargs["guest-stats-polling-interval"] = balloon_params.get("guest-polling", 0)
```

virttest/qemu_vm.py
```diff
@@ -2245,8 +2254,14 @@ def sort_key(dev):
             if balloon_params.get("balloon_dev_add_bus") == "yes":
                 balloon_bus = _get_pci_bus(
                     devices, balloon_params, 'balloon', True)
+            kwargs = {}
+            if balloon_params.get("deflate-on-oom") == "yes":
+                kwargs["deflate-on-oom"] = "yes"
```
yanan-fu: same with 'guest-polling'

virttest/qemu_vm.py
huangyum: @yanan-fu This part is to complete the qemu cmdline. The default value for option "guest-stats-polling-interval" is 0 if it is not specified in cmdline. In another word, if the value is 0, it doesn't matter whether to add it in cmdline.

yanan-fu: @huangyum  This match with my previous understanding.  If so, one line is ok, or no need to set default value in L2261.

yanan-fu:
@huangyum  please add "id" .

yanan-fu:
LGTM

vivianQizhu:
ACK

jingzhao84:
@luckyh could you help to review this pr

virttest/qemu_vm.py
```diff
@@ -1271,6 +1278,8 @@ def add_balloon(devices, devid=None, bus=None, use_old_format=None):
             dev = QDevice(model, parent_bus=bus)
             if devid:
                 dev.set_param("id", devid)
+            for key, value in six.iteritems(kwargs):
```
hereischen: I suggest use `for key, value in kwargs.items():` directly.
The rest are ok to me.

hereischen:

virttest/qemu_vm.py
hereischen: It is different, but it does not matter in this case.
Because you didn't use the functionality provided by `iteritems()` here, and yet you used `six` instead of python build-ins.
So  you don't need to do any changes, if you use `kwargs.items()` which is python build-in.

virttest/qemu_vm.py
```diff
@@ -1255,9 +1255,16 @@ def add_numa_node(devices, memdev=None, mem=None,
                 numa_cmd += ",nodeid=%s" % nodeid
             return numa_cmd
 
-        def add_balloon(devices, devid=None, bus=None, use_old_format=None):
+        def add_balloon(devices, devid=None, bus=None,
+                        use_old_format=None, **kwargs):
```
luckyh: `**kwargs` -> `options={}`

virttest/qemu_vm.py
```diff
@@ -2237,8 +2246,12 @@ def sort_key(dev):
             if balloon_params.get("balloon_dev_add_bus") == "yes":
                 balloon_bus = _get_pci_bus(
                     devices, balloon_params, 'balloon', True)
+            kwargs = {}
+            kwargs["deflate-on-oom"] = balloon_params.get("deflate-on-oom", "no")
```
luckyh: `params.get("deflate-on-oom", "no")` -> `params.get("balloon_opt_deflate_on_oom")`

virttest/qemu_vm.py
```diff
@@ -2237,8 +2246,12 @@ def sort_key(dev):
...
+            guest_polling = balloon_params.get("guest-polling", 0)
```
luckyh: `params.get("guest-polling", 0)` -> `params.get("balloon_opt_guest_polling")`

huangyum: @luckyh , thanks, all updated, including related  tp-qemu patches.

PR #1545, opened 2018-05-02
Python 3: fixed a bug related to filter

hereischen:
There was a bug when `self.info("block-jobs")` returns `[]`:
Although the `query_block_job` should return `{}`,
it returns `[]` on Python 2.
In later condition check, `if []`is equivalent to `if {}`,
so some cases passed.
However on Python 3, it returns `<fiter object>`, `if <fiter object>`
is qeuivalent to `if True`, which leads to case failures.


hereischen:
passed on Python3

PR #1549, opened 2018-05-03
utils_misc: Add function get_pid()

hahakiki2010:
Add this common operation in avocado-vt, consider
it maybe useful for others

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -4538,3 +4538,23 @@ def _wait_for_commands(bg_jobs, start_time, timeout):
         bg_job.result.duration = time.time() - start_time
 
     return True
+
+
+def get_pid(name, session=None):
+    """
+    Get pid by process name
+
+    :param name: Name of the process to retrieve its pid
+    :param session: ShellSession object of VM or remote host
+    :return: Pid of the process or None in case of exceptions
+    """
+    cmd = "ps -ef | grep '%s' | grep -v grep" % name
+    if session:
+        status, output = session.cmd_status_output(cmd)
+    else:
+        ret = process.run(cmd, shell=True, ignore_status=True)
+        status, output = ret.exit_status, results_stdout_52lts(ret).strip()
+    if status:
+        return None
+    else:
+        return int(output.split()[1])


virttest/utils_test/libvirt.py
```diff
@@ -3419,3 +3419,17 @@ def check_machine_type_arch(machine_type):
                                        "supported on the host with "
                                        "arch '%s'" % (machine_type,
                                                       arch))
+
+
+def get_libvirtd_pid():
```
balamuruhans: we can place it in utils_misc and tt can be something like,

```python
def get_pid(process, session=None):
    """
    :param process: name of the process to retrieve its pid
    :param session: ShellSession object of VM or remote host
    :return:  pid of the process or return None if process is not available
    """
```

hahakiki2010:
@balamuruhans I updated the func to be more general. Pls help review again. Thanks

virttest/utils_misc.py
```diff
@@ -4538,3 +4538,23 @@ def _wait_for_commands(bg_jobs, start_time, timeout):
...
+
+def get_pid(name, session=None):
+    """
+    Get pid by process name
+
+    :param name: Name of the process to retrieve its pid
+    :param session: ShellSession object of VM or remote host
+    :return: Pid of the process or raise exception if process is not available
+    """
+    cmd = "ps -ef | grep %s | grep -v grep" % name
```
balamuruhans: let us wrap the name in quotes,
`cmd = "ps -ef | grep '%s' | grep -v grep" % name`

virttest/utils_misc.py
```diff
@@ -4538,3 +4538,23 @@ def _wait_for_commands(bg_jobs, start_time, timeout):
...
+    if session:
+        status, output = session.cmd_status_output(cmd)
+    else:
+        ret = process.run(cmd, shell=True, ignore_status=True)
+        status, output = ret.exit_status, results_stdout_52lts(ret).strip()
+    if status:
+        raise exceptions.TestError("Pls check if %s is available" % name)
```
balamuruhans: IMO, I think returning None would be better. if test needs to check for a negative scenario and proceed then returning None would help.

virttest/utils_misc.py
hahakiki2010: fixed

virttest/utils_misc.py
hahakiki2010: agreed and updated

hahakiki2010:
@balamuruhans All comments have updated. Thanks!

balamuruhans:
Thanks @hahakiki2010 @sathnaga. let's merge it.

PR #1550, opened 2018-05-03
Method to generate stress events to guest

sathnaga:
This patch introduces a class to generate
different stress events to guest like devices hotplug,
suspend etc, this can further be expanded with other
events and will help incase of multi vm stress tests

### DIFF
--- a/virttest/utils_stress.py
+++ b/virttest/utils_stress.py
@@ -1,0 +1,93 @@
+import random
+import time
+import threading
+
+from avocado.utils import cpu
+
+from virttest import virsh
+from virttest import utils_hotplug
+from virttest.utils_test import libvirt
+
+
+class VMStressEvents():
+    def __init__(self, params, env):
+        """
+        :param params: test param
+        """
+        self.host_cpu_list = cpu.cpu_online_list()
+        self.iterations = int(params.get("stress_itrs", 20))
+        self.event_sleep_time = int(params.get("event_sleep_time", 10))
+        self.current_vcpu = params.get("smp", 32)
+        self.max_vcpu = params.get("virsh_maxcpus", 32)
+        self.ignore_status = params.get("ignore_status", "no") == "yes"
+        self.vms = env.get_all_vms()
+        self.events = params.get("stress_events", "reboot").split(',')
+        self.threads = []
+
+    def run_threads(self):
+        for vm in self.vms:
+            for event in self.events:
+                self.threads.append(threading.Thread(target=self.vm_stress_events, args=(event, vm)))
+        for thread in self.threads:
+            thread.start()
+
+    def wait_for_threads(self):
+        for thread in self.threads:
+            thread.join()
+
+    def vm_stress_events(self, event, vm):
+        """
+        Stress events
+        :param event: event name
+        :param vm: vm object
+        """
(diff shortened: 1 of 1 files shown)


virttest/utils_stress.py
```diff
@@ -0,0 +1,65 @@
...
+import time
+
+from avocado.utils import cpu
+
+from virttest import virsh
+from virttest import utils_hotplug
+from virttest.utils_test import libvirt
+
+
+def vm_stress_events(event, vm, params):
```
balamuruhans: we can have stress associated with component like cpu, lifecycle, memory (todo) separately, based on the params set it in __init__ and trigger it.

sathnaga: it is better to have freeflow method like this and choose the events we want in testcase, we need not restrict components here, rather we can mix and match many events in testcase, what say?

balamuruhans: mix and match would work better if we segregate,
- instead of calling each time, we can set it once and trigger it once
- redundancy can be avoided
```python
class Stress_Events():
    def __init__(self, vm):
        self.vm = vm
        self.events = self.vm.params.get("events", ["reboot"])
        self.ignore_status = self.vm.params.get("event_ignore_status", "no") == "yes"

    def check_status(self):
        libvirt.check_exit_status(self.ret)

    def trigger(self):
        dargs = self.vm.params.get("dargs", None)
        args = {'name':self.vm.name, **dargs}
        for each_event in events:
            self.event(each_event, **args)

    def event(self, event_name, args):
        if virsh.__dict__.get(event_name):
            self.ret = virsh.event_name(args)
        elif self.vm.__dict__.get(event_name):
            self.vm.event_name(args)
        else:
            # handle appropriately
        if not self.ignore_status:
            self.check_status()

```

sathnaga: @balamuruhans yes, partly we can go with this for init , but the events has to be run in parallel using `threading`, let me see if I can introduce `threading` as well here, or I can use the class for common init.

sathnaga:
@balamuruhans Have addressed your suggestion and  have made testcase`multivm_stress.py` as completely generic and lets control events for different components cpu, etc through different testcase config  like `multivm_cpustress.cfg` etc...

```
 (1/1) guest_multiboot.qemu.qcow2.virtio_scsi.smp2.virtio_net.Guest.RHEL.7.devel.ppc64le.powerkvm-libvirt.multivm_cpustress.stress_suspend: PASS (559.20 s)
```

virttest/utils_stress.py
```diff
@@ -0,0 +1,94 @@
...
+                                'max_live': self.max_vcpu,
+                                'cur_config': self.current_vcpu,
+                                'cur_live': self.current_vcpu,
+                                'guest_live': self.current_vcpu}
+                    utils_hotplug.check_vcpu_value(vm, exp_vcpu, option="--live")
+            elif "reboot" in event:
+                vm.reboot()
+                vm.wait_for_login()
+            else:
+                pass
```
balamuruhans: please raise the error as event not implemented or not supported, else we it would simply pass the testcase.

sathnaga: Sure..

balamuruhans:
@sathnaga Thanks for the patch, let us merge it.

PR #1554, opened 2018-05-07
test_setup: Enable setting hugepages on target nodes

huangyum:
Improve set_hugepages() to enable setting hugepages on target numa nodes

### DIFF
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -195,6 +195,14 @@ mem_chk_re_str = ([0-9]+)
 # this domain, hugepage = "yes" is pre-requisite
 # hp_nosharepages = "no"
 
+# Set hugepages on specific host numa node
+# target_nodes = 0 1 2
+
+# Number of hugepages to be set on host node
+# target_num_node0 = 1024
+# target_num_node1 = 1024
+# target_num_node2 = 1024
+
 # memory pages belonging to the domain will be locked in host's memory
 # and the host will not be allowed to swap them out, hugepage = "yes" is
 # pre-requisite
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -439,6 +439,21 @@ class HugePageConfig(object):
             target_hugepages = int(target_hugepages)
 
         self.target_hugepages = target_hugepages
+        self.target_nodes = params.get("target_nodes")
+        if self.target_nodes:
+            self.target_node_num = {}
+            self.node_list = self.target_nodes.split()
+            self.node_len = len(self.node_list)
+            for node in self.node_list:
+                num = params.object_params("node%s" % node).get("target_num")
+                # let us distribute the hugepages if target_num_node is not set
+                if not num:
+                    num = self.target_hugepages / self.node_len
+                    if node == self.node_list[0]:
+                        # if target_hugepages is not divisible by node_len
+                        # add remainder it to first node
+                        num += self.target_hugepages % self.node_len
+                self.target_node_num[node] = num
 
     @error_context.context_aware
     def check_hugepage_support(self):
@@ -645,7 +660,11 @@ class HugePageConfig(object):
                       self.hugepage_size)
         logging.debug("Number of large memory pages needed for this test: %s",
                       self.target_hugepages)
-        self.set_hugepages()
+        if self.target_nodes:
+            for node, num in six.iteritems(self.target_node_num):
+                self.set_node_num_huge_pages(num, node, self.hugepage_size)
+        else:
+            self.set_hugepages()
         self.mount_hugepage_fs()
 
         return self.suggest_mem


vivianQizhu:
@huangyum I am wondering do we need this in the module test_setup. It seems the added functionality has no relationship with the rest of `set_hugepages()`, its basic function is accomplished by `set_node_num_huge_pages()`, why not get the `target_node` in the specific test case and call `set_node_num_huge_pages()` directly?

huangyum:
@vivianQizhu Thanks for your review.  Currently hugepage setting is handled in env_process.py by calling 
```
895         h = test_setup.HugePageConfig(params)                                     
896         suggest_mem = h.setup()                     
```
which seems to fine to me.  
It is true that `set_node_num_huge_pages()` has implemented what I want. But we never put it in use. 
The original `set_hugepages()` only set hugepages for whole system regardless of numa nodes.  I made the change to use `set_node_num_huge_pages()` when `target_nodes` is specified.  I put it here because I think they both mean to set hugepages.

vivianQizhu:
@huangyum Thanks for explaination, it makes sense, but what if we get `target_nodes` and execute the setting loop inside `set_node_num_huge_pages()`? Cause it seems a little wired to put this part at the beginning of `set_hugepages()` and return directly without executing any of the older code, that tells their functions are relatively independent.

huangyum:
@vivianQizhu  I just found tp-libvirt has used `set_node_num_huge_pages()` and `set_hugepages() `in their test.  So I would like to not modify the two functions. Instead, I put it in `setup()`. Please help review again, thanks.

vivianQizhu:
Code looks good to me.

virttest/test_setup.py
```diff
@@ -645,7 +646,12 @@ def setup(self):
                       self.hugepage_size)
         logging.debug("Number of large memory pages needed for this test: %s",
                       self.target_hugepages)
-        self.set_hugepages()
+        if self.target_nodes:
+            for node in self.target_nodes.split():
+                ret = self.set_node_num_huge_pages(self.target_hugepages, node,
```
balamuruhans: we are not using the return value.

huangyum: @balamuruhans Thanks, updated!

virttest/test_setup.py
```diff
@@ -645,7 +646,12 @@ def setup(self):
                       self.hugepage_size)
         logging.debug("Number of large memory pages needed for this test: %s",
                       self.target_hugepages)
-        self.set_hugepages()
+        if self.target_nodes:
+            for node in self.target_nodes.split():
+                self.set_node_num_huge_pages(self.target_hugepages, node,
```
balamuruhans: And, I think we should divide the target_hugepages across nodes rite ? 
we get overall hugepages number using get_target_hugepages or `target_hugepages` param, but here we try to set all the hugepages in each node instead of dividing across nodes.

virttest/test_setup.py
```diff
@@ -439,6 +439,12 @@ def __init__(self, params):
             target_hugepages = int(target_hugepages)
 
         self.target_hugepages = target_hugepages
+        self.target_nodes = params.get("target_nodes")
+        if self.target_nodes:
+            self.target_node_num = {}
+            for node in self.target_nodes.split():
+                num = params.object_params("node%s" % node).get("target_num")
+                self.target_node_num[node] = num
```
balamuruhans: Let us handle if user haven't provide target_num_node by distributing the target hugepages,
```python
if self.target_nodes:
    self.target_node_num = {}
    self.node_list = self.target_nodes.split()
    self.node_len = len(self.node_list)
    for node in node_list:
        num = params.object_params("node%s" % node).get("target_num")
        # let us distribute the hugepages if target_num_node is not set
        if not num:
            num = self.target_hugepages / self.node_len
            if node == node_list[0]:
                # if target_hugepages is not divisible by node_len
                # add remainder it to first node
                num += self.target_hugepages % self.node_len
        self.target_node_num[node] = num
```

virttest/test_setup.py
```diff
@@ -439,6 +439,12 @@ def __init__(self, params):
             target_hugepages = int(target_hugepages)
 
         self.target_hugepages = target_hugepages
+        self.target_nodes = params.get("target_nodes")
```
balamuruhans: let's have target_nodes, target_num_node0 params in base.cfg and explain its usage

huangyum: @balamuruhans Good idea!  Please help check if the explanation is clear. Thanks!

shared/cfg/base.cfg
```diff
@@ -195,6 +195,12 @@ mem_chk_re_str = ([0-9]+)
 # this domain, hugepage = "yes" is pre-requisite
 # hp_nosharepages = "no"
 
+# Set hugepages on specific host numa node
+# target_nodes = 0
+
+# Number of hugepages to be set on host node0
+# target_num_node0 = 1024
```
balamuruhans: Explanation looks good to me,

it is better to mention with multiple nodes,
```
# target_nodes = 0,1,2
# target_num_node0 = 1024
# target_num_node1 = 1024
# target_num_node2 = 1024
```

huangyum: @balamuruhans Thanks, updated as you suggested.

sathnaga:
@huangyum could you pls repush, looks like travis has a failure...

virttest/test_setup.py
```diff
@@ -439,6 +439,21 @@ def __init__(self, params):
...
+        if self.target_nodes:
+            self.target_node_num = {}
+            self.node_list = self.target_nodes.split()
+            self.node_len = len(self.node_list)
+            for node in self.node_list:
+                num = params.object_params("node%s" % node).get("target_num")
+                # let us distribute the hugepages if target_num_node is not set
+                if not num:
+                    num = self.target_hugepages / self.node_len
+                    if node == node_list[0]:
```
balamuruhans: self.node_list[0] would make travis happy :)

balamuruhans:
Thanks all. let us merge it.

PR #1561, opened 2018-05-10
Python 3: Fix a decode issue of ppm_utils

luckyh:
Fix the following decode issue of ppm_utils.

| VM 'avocado-vt-vm1' produced an invalid screendump
| 'utf-8' codec can't decode byte 0xaa in position 6166: invalid start byte


hereischen:

PR #1563, opened 2018-05-14
Python3 : bytes/str changes in rss_client

hereischen:

### DIFF
--- a/virttest/rss_client.py
+++ b/virttest/rss_client.py
@@ -170,7 +170,7 @@ class FileTransferClient(object):
         except socket.error as e:
             raise FileTransferSocketError("Error receiving data from server",
                                           e)
-        return "".join(strs)
+        return b"".join(strs)
 
     def _report_stats(self, sr):
         if self._log_func:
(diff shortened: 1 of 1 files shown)


virttest/rss_client.py
```diff
@@ -170,7 +170,7 @@ def _receive(self, size, timeout=60):
         except socket.error as e:
             raise FileTransferSocketError("Error receiving data from server",
                                           e)
-        return "".join(strs)
+        return b"".join(strs)
```
luckyh: This change would impact `_receive_packet()`, so need to its caller correspondingly.

virttest/rss_client.py
```diff
@@ -223,7 +224,8 @@ def _receive_file_chunks(self, filename, timeout=60):
             try:
                 end_time = time.time() + timeout
                 while True:
-                    data = self._receive_packet(end_time - time.time())
+                    data = self._receive_packet(end_time - time.time()
+                                                ).decode()
```
luckyh: Not need to change this since `f` has been opened in binary mode.

PR #1564, opened 2018-05-14
Add param to cleanup env

sathnaga:
This patch introduces a param `env_cleanup`, if set
will clear the env from environment, which is needed
while `create_vm_libvirt` and `kill_vm_libvirt` is enbled.

### DIFF
--- a/avocado_vt/test.py
+++ b/avocado_vt/test.py
@@ -430,7 +430,7 @@ class VirtTest(test.Test):
                                       "postprocessing: %s",
                                       sys.exc_info()[1])
                 finally:
-                    if self.__safe_env_save(env):
+                    if self.__safe_env_save(env) or params.get("env_cleanup", "no") == "yes":
                         env.destroy()   # Force-clean as it can't be stored
 
         except Exception as e:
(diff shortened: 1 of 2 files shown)


virttest/env_process.py
```diff
@@ -1407,6 +1407,10 @@ def postprocess(test, params, env):
     if err:
         raise RuntimeError("Failures occurred while postprocess:\n%s" % err)
 
+    if params.get("env_cleanup", "no") == "yes":
+        if env:
+            env.destroy()
```
balamuruhans: cleaning it here would override the framework currently being handled, so to avoid the conflict can we can have this condition in test.py,

https://github.com/avocado-framework/avocado-vt/blob/770c2a27f15da28016925d79af4380dfb1ce5a68/avocado_vt/test.py#L417-L434


```python
if self.__safe_env_save(env) or params.get("env_cleanup", "no") == "yes":
    env.destroy()   # Force-clean as it can't be stored
```

sathnaga: Sure, I can validate that.

sathnaga: That's a right place, Tnx. have modified it.

PR #1571, opened 2018-05-17
Python 3: Remove autotest requirements

hsmj1412:


requirements.txt
```diff
@@ -1,4 +1,3 @@
-autotest>=0.16.2
```
luckyh: How about:
```
autotest>=0.16.2; python_version < '3.0'
```

luckyh: Better to apply for `requirements-travis.txt` as well.

hsmj1412: Updated

PR #1577, opened 2018-05-21
unattended_install: does nothing if not .ks file

zhengxiang93:
If unattended file is not a kickstart file, dest_fname would be used
but not declared. So add this patch to fix it.

Signed-off-by: Zheng Xiang <xiang.zheng@linaro.org>


xutian:
LGTM

PR #1579, opened 2018-05-22
[Python3] Fix openvswitch issues

kylazhang:
1. Change str to int as in python3 str can not compare with int
2. Fix python3 __init__ parameters error when base class is object, which report
"TypeError: object.__init__() takes no parameters"
3. Fix existing format issues and error

### DIFF
--- a/virttest/openvswitch.py
+++ b/virttest/openvswitch.py
@@ -158,7 +158,7 @@ class OpenVSwitchControl(object):
             int_ver = ''.join(a)
         except Exception:
             raise ValueError("Wrong version format '%s'" % version)
-        return int_ver
+        return int(int_ver)
 
     @classmethod
     def get_version(cls):
@@ -356,9 +356,7 @@ class OpenVSwitchSystem(OpenVSwitchControlCli_CNT, OpenVSwitchControlDB_CNT):
         :param ovs_pidfile: Path of OVS ovs-vswitchd pid.
         :param install_prefix: Path where is openvswitch installed.
         """
-        sup = super(man[self.__class__, OpenVSwitchSystem], self)
-        sup.__init__(self, db_path, db_socket, db_pidfile, ovs_pidfile,
-                     dbschema, install_prefix)
+        super(man[self.__class__, OpenVSwitchSystem], self).__init__()
 
         self.cleanup = False
         self.pid_files_path = None
@@ -459,9 +457,7 @@ class OpenVSwitch(OpenVSwitchSystem):
         :param ovs_pidfile: Path of OVS ovs-vswitchd pid.
         :param install_prefix: Path where is openvswitch installed.
         """
-        super(man[self, OpenVSwitch], self).__init__(db_path, db_socket,
-                                                     db_pidfile, ovs_pidfile,
-                                                     dbschema, install_prefix)
+        super(man[self.__class__, OpenVSwitch], self).__init__()
         self.tmpdir = "/%s/openvswitch" % (tmpdir)
         try:
             os.mkdir(self.tmpdir)


virttest/openvswitch.py
```diff
@@ -459,9 +457,7 @@ def __init__(self, tmpdir, db_path=None, db_socket=None, db_pidfile=None,
         :param ovs_pidfile: Path of OVS ovs-vswitchd pid.
         :param install_prefix: Path where is openvswitch installed.
         """
-        super(man[self, OpenVSwitch], self).__init__(db_path, db_socket,
-                                                     db_pidfile, ovs_pidfile,
-                                                     dbschema, install_prefix)
+        super(man[self.__class__, OpenVSwitch], self).__init__()
```
chunfuwen: is any special reason to leave out those parameters ,which work as initialized value?

vivianQizhu:
@xiywang @quanwenli Please help review this, see if it resolve our issue, thanks.

quanwenli:
Tests are passed with this patch.

Acked-by: Wenli Quan wquan@redhat.com

quanwenli:
@luckyh could you help review it, it blocks openvswitch test. Thanks

virttest/openvswitch.py
```diff
@@ -356,9 +356,7 @@ def __init__(self, db_path=None, db_socket=None, db_pidfile=None,
         :param ovs_pidfile: Path of OVS ovs-vswitchd pid.
         :param install_prefix: Path where is openvswitch installed.
         """
-        sup = super(man[self.__class__, OpenVSwitchSystem], self)
-        sup.__init__(self, db_path, db_socket, db_pidfile, ovs_pidfile,
-                     dbschema, install_prefix)
+        super(man[self.__class__, OpenVSwitchSystem], self).__init__()
```
luckyh: It's interesting if calling `man[self.__class__, OpenVSwitchSystem]` got `object` in py3. I tried this in py2 and it got
```
<class 'virttest.openvswitch.managed_OpenVSwitchSystem_OpenVSwitchControlCli_140_OpenVSwitchControlDB_140'>
```
I believe that there must be something wrong in `man.__getitem__` with py3, so the code should not be changed like this.

vivianQizhu: @luckyh I got same return value in py3:
```
# python3
Python 3.6.6 (default, Jul 19 2018, 14:25:17) 
[GCC 8.1.1 20180712 (Red Hat 8.1.1-5)] on linux
>>> from virttest import versionable_class
>>> from virttest import openvswitch
>>> ovs = versionable_class.factory(openvswitch.OpenVSwitchSystem)()
<class 'virttest.openvswitch.managed_OpenVSwitchSystem_OpenVSwitchControlCli_140_OpenVSwitchControlDB_140'> <class 'type'>
```
And I get this for `super()`
```
<super: <class 'managed_OpenVSwitchSystem_OpenVSwitchControlCli_140_OpenVSwitchControlDB_140'>, <managed_OpenVSwitchSystem_OpenVSwitchControlCli_140_OpenVSwitchControlDB_140 object>>
```
I think `man.__getitem__ ` should not be blamed.
I suppose because the parent class `OpenVSwitchControlDB_140` does not define its `__init__()`, so `super().__init()`  will call `object.__init__()`. Py2 just not prompt the error when user call `object.__init__()` with parameters, while py3 does. 
As the parent class does not define `__init__()`, it inherits `object.__init__()`, so I think it does not make sense to pass parameters to `super().__init__()`. So the adjustment here is just right.

luckyh: then my question is, if `super().__init__()` would just call `object.__init__()`, why do we have to do that?

vivianQizhu: @luckyh Exactly, I think this `super().__init__()` could be dropped. And the change in L#460 should be dropped, it does need call parent's `__init__()` and need those parameters. Although the class `OpenVSwitch()` initialization is ill even in py2, that's actually nothing to do with this pull request, it should be handled separately.

vivianQizhu:
Hi @kylazhang Just saw this, we should be able to close it. :)

PR #1590, opened 2018-05-28
qemu_vm: Fix fd passing issue

luckyh:
After aexpect commit e6636d8 being merged, file descriptors should
be explicitly passed to the child process, so here edit `qemu_vm`
accordingly.


luckyh:
depends on https://github.com/avocado-framework/avocado-vt/pull/1575

luckyh:
passed my test both on py2 and py3.

PR #1591, opened 2018-05-28
env_process: archive sosreport of host/guest/remote host if enabled

balamuruhans:
check for the param that enables sosreport and archive it during the
postprocess of the test.


sathnaga:
@balamuruhans can you paste a result summary with these options enabled

balamuruhans:
@sathnaga sure, working on it.

PR #1592, opened 2018-05-29
Add guest config file for CentOS 7.5 ppc64le

sathnaga:
Add config file to support CentOS 7.5 ppc64le
guest.

### DIFF
--- a/shared/cfg/guest-os/Linux/CentOS/7.5/ppc64le.cfg
+++ b/shared/cfg/guest-os/Linux/CentOS/7.5/ppc64le.cfg
@@ -1,0 +1,10 @@
+- ppc64le:
+    grub_file = /boot/grub/grub.conf
+    vm_arch_name = ppc64le
+    image_name += -ppc64le
+    boot_path = "ppc/ppc64"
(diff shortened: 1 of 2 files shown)


shared/cfg/guest-os/Linux/CentOS/7.5/ppc64le.cfg
```diff
@@ -0,0 +1,10 @@
+- ppc64le:
+    grub_file = /boot/grub/grub.conf
+    vm_arch_name = ppc64le
+    image_name += -ppc64le
+    boot_path = "ppc/ppc64"
+    unattended_install.url:
+        unattended_file = unattended/CentOS-7-4.ks
```
balamuruhans: is it okay to reuse CentOS 74 ks file ? better can be have separate one.

sathnaga: It works for now, I do not have any specific additions for 7.5ks, and it will be just a copy of same file, if there is a need for specific additions in future we can change it,

PR #1607, opened 2018-06-06
remote: Reenable  multiple files transfer

rdkdd:
This customizes quoting (e.g. shell-escaping) for list of
strings/string and thus enables multiple file transfer.


rdkdd:
@ldoktor yes it is not necessary to process extra function, so I have changed destination part of `scp` to `pipes.quote` as you proposed. Could you merge this bit now please?

PR #1612, opened 2018-06-11
[python3] Fix for getattr type return to Enum

kylazhang:


kylazhang:
on python3
>>> getattr(signal, 'SIGINT')
<Signals.SIGINT: 2>
>>> getattr(signal, 'SIGINT').value
2

on python2
>>> getattr(signal, 'SIGINT')
2

so on python3 need to get the int value through \.value

PR #1614, opened 2018-06-12
env_process.py: use Thread.join

lolyu:
use join to wait for thread's completion.


hereischen:

PR #1617, opened 2018-06-14
test_setup: added UlimitConfig

hereischen:
UlimitConfig can be used to config ulimit according to vt_ulimit_xxx
params during preprocess.
And it will restore the ulimit in the postprocess.

id: 1543734


virttest/test_setup.py
```diff
@@ -2391,3 +2393,52 @@ def disable(self):
...
+                self.test.error(str(error))
+
+    def _restore_default_ulimit(self):
+        logging.info("Setting ulimit options back to defaults.")
+        for key in self.ulimit_options:
+            try:
+                setrlimit(self.ulimit_options[key], self.ulimit[key])
+            except Exception as error:
+                self.test.error("Failed to restore default ulimit, "
+                                "details: %s." % error)
```
luckyh: Better to let `setrlimit()` raises the error directly.

virttest/test_setup.py
```diff
@@ -2391,3 +2393,52 @@ def disable(self):
...
+    vt_ulimit_nofile is equivalent "ulimit -n"
+    """
+
+    ulimit_options = {"core": RLIMIT_CORE,
+                      "nofile": RLIMIT_NOFILE}
+
+    def _set_ulimit(self):
+        # get default ulimit values
+        self.ulimit = {key: getrlimit(
+            self.ulimit_options[key]) for key in self.ulimit_options}
```
luckyh: We can only restore the value for those changed ones.

hereischen:
@ldoktor please help review, thanks.

PR #1618, opened 2018-06-14
Update the default public key

hsmj1412:
OpenSSH 7.0 disables ssh-dss keys by default,
so set the default ssh-rsa

Ref: https://www.openssh.com/txt/release-7.0

### DIFF
--- a/virttest/ssh_key.py
+++ b/virttest/ssh_key.py
@@ -12,7 +12,7 @@ from virttest.compat_52lts import decode_to_text
 def get_public_key():
     """
     Return a valid string ssh public key for the user executing autoserv or
-    autotest. If there's no DSA or RSA public key, create a DSA keypair with
+    autotest. If there's no DSA or RSA public key, create a RSA keypair with
     ssh-keygen and return it.
 
     :returns: a ssh public key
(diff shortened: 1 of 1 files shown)


hsmj1412:
@sathnaga @balamuruhans @ldoktor @rdkdd @luckyh @hereischen Please help check and confirm this PR will not block your test

luckyh:
The change looks good to me, thanks.

virttest/ssh_key.py
```diff
@@ -32,19 +32,19 @@ def get_public_key():
...
-        print('DSA keypair found, using it')
-        public_key_path = dsa_public_key_path
-
-    elif has_rsa_keypair:
+    if has_rsa_keypair:
         print('RSA keypair found, using it')
         public_key_path = rsa_public_key_path
 
+    elif has_dsa_keypair:
+        print('DSA keypair found, using it')
```
balamuruhans: can you help to make use of logging instead here and elsewhere.

hsmj1412: Updated

virttest/ssh_key.py
```diff
@@ -32,31 +32,31 @@ def get_public_key():
...
-    else:
-        print('Neither RSA nor DSA keypair found, creating DSA ssh key pair')
-        process.system('ssh-keygen -t dsa -q -N "" -f %s' %
-                       dsa_private_key_path)
+    elif has_dsa_keypair:
+        logging.info('DSA keypair found, using it')
         public_key_path = dsa_public_key_path
 
+    else:
+        logging.info('Neither RSA nor DSA keypair found, creating DSA ssh key pair')
```
xutian: should be "creating RSA  ssh key pair ", right?

hsmj1412: Fixed, thx

balamuruhans:
Thanks all. let us merge it.

PR #1625, opened 2018-06-20
Fix private key access deny issue in spice graphics

chunfuwen:
openssl in new platform remove private key read permission for other users,
but qemu need access it in TLS enable spice
Fix it by granting read permission to other users

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -1400,6 +1400,7 @@ def create_x509_dir(path, cacert_subj, server_subj, passphrase,
          '%sca-cert.pem -subj "%s"' %
          (ssl_cmd, days, path, passphrase, path, cacert_subj)),
         ('%s genrsa -out %s %d' % (ssl_cmd, path + server_key, bits)),
+        ('chmod o+r %s' % path + server_key),
         ('%s req -new -key %s -out %s/server-key.csr -subj "%s"' %
          (ssl_cmd, path + server_key, path, server_subj)),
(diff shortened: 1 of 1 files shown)


chunfuwen:
just paste some error message here:
```
TestFail: Expect VM can be started, but failed with: VM 'avocado-vt-vm1' failed to start: error: Failed to start domain avocado-vt-vm1error: internal error: process exited while connecting to monitor: (process:26411): Spice-WARNING **: 05:23:23.375: reds.c:2872:reds_init_ssl: Could not use private key file2018-06-15T09:23:23.375945Z qemu-kvm: failed to initialize spice server&#10;
```

virttest/utils_misc.py
```diff
@@ -1400,6 +1400,7 @@ def create_x509_dir(path, cacert_subj, server_subj, passphrase,
          '%sca-cert.pem -subj "%s"' %
          (ssl_cmd, days, path, passphrase, path, cacert_subj)),
         ('%s genrsa -out %s %d' % (ssl_cmd, path + server_key, bits)),
+        ('chmod a+r %s' % path + server_key),
```
hahakiki2010: I think o+x is enough

hahakiki2010:
Please fix the travis ci failure first

chunfuwen:
@hahakiki2010 ,the failure are not brought by this PR , sound like travics-ci has some rules change

virttest/utils_misc.py
chunfuwen: updated

PR #1632, opened 2018-06-21
env_process: default to log libvirtd info

balamuruhans:
As the libvirtd is enabled by default to capture debug logs for
all the tests but archive it only for fail/error, it is better to
enable the log level to `info` so as to debug and get close to the
issue as well not polluting the archiving space.

Also back up of the libvirtd config should be done before enabling
and restarting the libvirtd is the right thing to do.

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -2367,13 +2367,24 @@ class LibvirtdDebugLog(object):
         self.test = test
         self.libvirtd = utils_libvirtd.Libvirtd()
         self.libvirtd_conf = utils_config.LibvirtdConfig()
-        self.backupfile = "%s.bkup.debuglog" % self.libvirtd_conf.conf_path
+        self.backupfile = "%s.backup" % self.libvirtd_conf.conf_path
 
     def enable(self):
         """ Enable libvirtd debug log """
         if not self.log_file or not os.path.isdir(os.path.dirname(self.log_file)):
             self.log_file = utils_misc.get_path(self.test.debugdir,
                                                 "libvirtd.log")
+        if os.path.isfile(self.backupfile):
+            os.remove(self.backupfile)
+
+        # backup libvirtd conf before restarting libvirtd
+        try:
+            with open(self.backupfile, "w") as fd:
+                fd.write(self.libvirtd_conf.backup_content)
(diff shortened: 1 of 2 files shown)


virttest/test_setup.py
```diff
@@ -2374,16 +2374,21 @@ def enable(self):
         if not self.log_file or not os.path.isdir(os.path.dirname(self.log_file)):
             self.log_file = utils_misc.get_path(self.test.debugdir,
                                                 "libvirtd.log")
+        if os.path.isfile(self.backupfile):
+            os.remove(self.backupfile)
+
+        # backup libvirtd conf before restarting libvirtd
+        fd = open(self.backupfile, "w")
```
hereischen: Maybe use `with` statement this time.

virttest/test_setup.py
```diff
@@ -2374,16 +2374,21 @@ def enable(self):
         if not self.log_file or not os.path.isdir(os.path.dirname(self.log_file)):
             self.log_file = utils_misc.get_path(self.test.debugdir,
                                                 "libvirtd.log")
+        if os.path.isfile(self.backupfile):
```
hereischen: After this change,it only keeps 'last' debug log when `enable()` is called.
What about make this optional?

virttest/test_setup.py
balamuruhans: sure, agree. I would make the change as suggested.

virttest/test_setup.py
balamuruhans: This is not the debug log, it is the libvirtd configuration file available in `/etc/libvirt/libvirtd.conf`
 - so clean any irrelevant file in same name if any before backing up
 - backup the current configuration file.
- change the configuration file and restart libvirtd.

balamuruhans:
@hereischen made the changes as suggested. please help to review.

sathnaga:
@balamuruhans confusion may coz of the filename, `        self.backupfile = "%s.bkup.debuglog" % self.libvirtd_conf.conf_path
`, it would be good if we can remove `debuglog` suffix..

balamuruhans:
@sathnaga Thank you for being careful, made the respective change.

PR #1639, opened 2018-06-28
Chardev: refactor chardev creation, for better multiple backend support

vivianQizhu:
1. qdevices: Add Serial Bus representation
2. qcontainer: Add a wrapper for serial device creation

id: 1588423

### DIFF
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -129,21 +129,24 @@ q35, arm64-pci, arm64-mmio:
 # Serial port support
 # You can assign more than one serial to guest with this parameter.
 # like:
-# serials = "S1 S2 S3"
-serials = "serial0"
-
-# virtio_serialport and virtio_console ports
-# virtio_ports = "vs1 vs2 vc1"
-# Port types
-# virtio_port_type = serialport
-# virtio_port_type_vc1 = console
-# Host socket type
-# virtio_port_chardev = spicevmc
-# virtio_port_chardev_vc1 = socket
-# chardev name prefix (port index will be appended)
+# serials = "serial0 vs1 vs2 vc1"
+# serial_type_vs1 = virtserialport
+# serial_type_vc1 = virtconsole
+# Chardev backend type: 'unix_socket' by default
+# chardev_backend_vs1 = unix_socket
+# chardev_backend_vc1 = tcp_socket
+# Chardev host for tcp/udp socket backend
+# chardev_host = $host_ip (127.0.0.1 by default)
+# chardev_ipv4 = yes
+# chardev_ipv6 = yes
+# chardev_name = vdagent/smartcard/usbredir (only for spicevmc and spiceport)
+# Virtio port name prefix (port index will be appended)
 # virtio_port_name_prefix = "com.redhat.spice."
 # By default the port name is used (vc1)
 # virtio_port_name_prefix_vc1 = ""
+
+serials = "serial0"
+serial_type = "isa-serial"
 
 # Emulated machine type, run following command to see supported machine type.
 # qemu-kvm -M ?
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -28,6 +28,7 @@ from virttest.qemu_devices.utils import (DeviceError, DeviceHotplugError,
                                          DeviceInsertError, DeviceRemoveError,
                                          DeviceUnplugError, none_or_int)
 from virttest.compat_52lts import results_stdout_52lts, decode_to_text
+from virttest.utils_params import Params
 
 #
 # Device container (device representation of VM)
@@ -1423,9 +1424,9 @@ class DevContainer(object):
         # bus: ahci, virtio-scsi-pci, USB
         #
         if not use_device:
-            if fmt and (fmt == "scsi" or (fmt.startswith('scsi') and
-                                          (scsi_hba == 'lsi53c895a' or
-                                           scsi_hba == 'spapr-vscsi'))):
+            if fmt and (fmt == "scsi" or (fmt.startswith('scsi')
+                                          and (scsi_hba == 'lsi53c895a'
+                                               or scsi_hba == 'spapr-vscsi'))):
                 if not (bus is None and unit is None and port is None):
                     logging.warn("Using scsi interface without -device "
                                  "support; ignoring bus/unit/port. (%s)", name)
@@ -1717,6 +1718,145 @@ class DevContainer(object):
                                                image_params.get(
                                                    "bus_extra_params"),
                                                image_params.get("force_drive_format"))
+
+    def serials_define_by_variables(self, serial_id, serial_type, chardev_id,
+                                    bus_type=None, serial_name=None,
+                                    bus=None, nr=None, reg=None):
+        """
+        Creates related devices by variables
+
+        :param serial_id: the id of the serial device
+        :param serial_type: the type of the serial device
+        :param chardev_id: the id of the chardev device
+        :param bus_type: bus type of the serial device(optional, virtio only)
+        :param serial_name: the name option of serial device(optional)
+        :param bus: the busid of parent bus(optional, virtio only)
+        :param nr: the nr(port) of the parent bus(optional, virtio only)
+        :param reg: reg option of isa-serial(optional)
+        :return: the device list that construct the serial device
+        """
+
+        devices = []
+        # For virtio devices, generate controller and create the port device
+        if serial_type.startswith('virt'):
+            if not bus:
+                if bus_type == 'virtio-serial-device':
+                    pci_bus = {'type': 'virtio-bus'}
+                else:
+                    pci_bus = {'aobject': 'pci.0'}
+                bus = self.get_first_free_bus(
+                    {'type': 'SERIAL', 'atype': bus_type}, [None, nr])
+                #  Multiple virtio console devices can't share a single bus
+                if bus is None or serial_type == 'virtconsole':
+                    _hba = bus_type.replace('-', '_') + '%s'
+                    bus = self.idx_of_next_named_bus(_hba)
+                    bus = self.list_missing_named_buses(
+                        _hba, 'SERIAL', bus + 1)[-1]
+                    logging.debug("list missing named bus: %s", bus)
+                    devices.append(
+                        qdevices.QDevice(bus_type,
+                                         {"id": bus},
+                                         bus,
+                                         pci_bus,
+                                         qdevices.QSerialBus(
+                                             bus, bus_type, bus)))
+                else:
+                    bus = bus.busid
+            devices.append(
+                qdevices.QDevice(serial_type,
+                                 {"id": serial_id},
+                                 parent_bus={'busid': bus}))
+            devices[-1].set_param('name', serial_name)
+        else:  # none virtio type, generate serial device directly
+            devices.append(qdevices.QDevice(serial_type, {"id": serial_id}))
+            devices[-1].set_param("reg", reg)
+        devices[-1].set_param('chardev', chardev_id)
+
+        return devices
+
+    def serials_define_by_params(self, serial_id, params,
+                                 file_name=None):
+        """
+        Wrapper for creating serial device from serial params.
+
+        :param serial_id: id of serial object
+        :param params: serial params
+        :param file_name: the file path of the serial device (optional)
+        :return: the device list that construct the serial device
+        """
+        serial_type = params['serial_type']
+        machine = params.get('machine_type')
+
+        # Arm lists "isa-serial" as supported but can't use it,
+        # fallback to "-serial"
+        legacy_cmd = " -serial unix:'%s',server,nowait" % file_name
+        legacy_dev = qdevices.QStringDevice('SER-%s' % serial_id,
+                                            cmdline=legacy_cmd)
+        arm_serial = (serial_type == 'isa-serial'
+                      and 'arm' in params.get("machine_type", ""))
+        if (arm_serial or not self.has_option("chardev")
+                or not self.has_device(serial_type)):
+            return legacy_dev
+
+        bus_type = None
+        if serial_type.startswith('virt'):
+            if '-mmio' in machine:
+                controller_suffix = 'device'
+            elif machine.startswith("s390"):
+                controller_suffix = 'ccw'
+            else:
+                controller_suffix = 'pci'
+            bus_type = 'virtio-serial-%s' % controller_suffix
+
+        prefix = params.get('virtio_port_name_prefix')
+        name = prefix + serial_id if prefix else serial_id
+        chardev_id = 'chardev_%s' % serial_id
+        chardev_device = self.chardev_define_by_params(chardev_id, params, file_name)
+        serial_devices = self.serials_define_by_variables(serial_id,
+                                                          serial_type,
+                                                          chardev_id,
+                                                          bus_type=bus_type,
+                                                          serial_name=name,
+                                                          bus=params.get(
+                                                              "serial_bus"),
+                                                          nr=params.get(
+                                                              "serial_nr"),
+                                                          reg=params.get(
+                                                              "serial_reg"))
+
+        return [chardev_device] + serial_devices
+
+    def chardev_define_by_params(self, chardev_id, params, file_name=None):
+        """
+        Wrapper for creating -chardev device from params.
+        :param chardev_id: chardev id
+        :param params: chardev params
+        :param file_name: file name of chardev (optional)
+        :return: CharDevice object
+        """
+        backend = params.get('chardev_backend', 'unix_socket')
+        # for tcp_socket and unix_socket, both form to 'socket'
+        _backend = 'socket' if 'socket' in backend else backend
+        # Generate -chardev device
+        chardev_param = Params({'backend': _backend, 'id': chardev_id})
+        if backend in ['unix_socket', 'file', 'pipe', 'serial',
+                       'tty', 'parallel', 'parport']:
+            chardev_param.update({'path': file_name})
+        elif backend in ['udp', 'tcp_socket']:
+            chardev_param.update(
+                {'host': params['chardev_host'],
+                 'port': params['chardev_port'],
+                 'ipv4': params.get('chardev_ipv4'),
+                 'ipv6': params.get('chardev_ipv6')})
+        if 'socket' in backend:  # tcp_socket & unix_socket
+            chardev_param.update(
+                {'server': params.get('chardev_server', 'yes'),
+                 'nowait': params.get('chardev_nowait', 'yes')})
+        elif backend in ['spicevmc', 'spiceport']:
+            chardev_param.update(
+                {'debug': params.get('chardev_debug'),
+                 'name': params.get('chardev_name')})
+        return qdevices.CharDevice(chardev_param, chardev_id)
 
     def cdroms_define_by_params(self, name, image_params, media=None,
                                 index=None, image_boot=None,
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -2039,9 +2039,9 @@ class QSCSIBus(QSparseBus):
 
 class QBusUnitBus(QDenseBus):
 
-    """ Implementation of bus-unit bus (ahci, ide) """
+    """ Implementation of bus-unit/nr bus (ahci, ide, virtio-serial) """
 
-    def __init__(self, busid, bus_type, lengths, aobject=None, atype=None):
+    def __init__(self, busid, bus_type, lengths, aobject=None, atype=None, unit_spec='unit'):
         """
         :param busid: id of the bus (mybus.0)
         :type busid: str
@@ -2056,8 +2056,9 @@ class QBusUnitBus(QDenseBus):
         """
         if len(lengths) != 2:
             raise ValueError("len(lenghts) have to be 2 (%s)" % self)
-        super(QBusUnitBus, self).__init__('bus', [['bus', 'unit'], lengths],
+        super(QBusUnitBus, self).__init__('bus', [['bus', unit_spec], lengths],
                                           busid, bus_type, aobject, atype)
+        self.unit_spec = unit_spec
 
     def _update_device_props(self, device, addr):
         """ Always set the properties """
@@ -2066,7 +2067,7 @@ class QBusUnitBus(QDenseBus):
     def _set_device_props(self, device, addr):
         """This bus is compound of m-buses + n-units, set properties """
         device.set_param('bus', "%s.%s" % (self.busid, addr[0]))
-        device.set_param('unit', addr[1])
+        device.set_param(self.unit_spec, addr[1])
 
     def _check_bus(self, device):
         """ This bus is compound of m-buses + n-units, check correct busid """
@@ -2093,9 +2094,26 @@ class QBusUnitBus(QDenseBus):
                     bus = int(busid[1])
         if isinstance(busid, int):
             bus = busid
-        if device.get_param('unit'):
-            unit = int(device.get_param('unit'))
+        if device.get_param(self.unit_spec):
+            unit = int(device.get_param(self.unit_spec))
         return [bus, unit]
+
+
+class QSerialBus(QBusUnitBus):
+
+    """ Serial bus representation """
+
+    def __init__(self, busid, bus_type, aobject=None, max_ports=32):
+        """
+        :param busid: bus id
+        :param bus_type: bus type(virtio-serial-device, virtio-serial-pci)
+        :param aobject: autotest object
+        :param max_ports: max ports, default 32
+        """
+
+        super(QSerialBus, self).__init__(busid, 'SERIAL', [1, max_ports],
+                                         aobject, bus_type, unit_spec='nr')
+        self.first_port = (0, 1)
 
 
 class QAHCIBus(QBusUnitBus):
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -142,7 +142,7 @@ class VM(virt_vm.BaseVM):
             self.process = None
             self.serial_ports = []
             self.serial_console_log = None
-            self.virtio_console = None
+            self.serial_session_device = None
             self.redirs = {}
             self.spice_options = {}
             self.vnc_port = 5900
@@ -339,12 +339,6 @@ class VM(virt_vm.BaseVM):
         return [self.get_serial_console_filename(_) for _ in
                 self.params.objects("serials")]
 
-    def get_virtio_port_filenames(self):
-        """
-        Get socket file of virtio ports
-        """
-        return [_.hostfile for _ in self.virtio_ports]
-
     def cleanup_serial_console(self):
         """
         Close serial console and associated log file
@@ -476,104 +470,6 @@ class VM(virt_vm.BaseVM):
             cmd += " -mon chardev=%s" % monitor_id
             cmd += _add_option("mode", "control")
             return cmd
-
-        def add_serial(devices, name, filename):
-            # Arm lists "isa-serial" as supported but can't use it,
-            # fallback to "-serial"
-            legacy_cmd = " -serial unix:'%s',server,nowait" % filename
-            if (not devices.has_option("chardev") or
-                    'arm' in params.get("machine_type", "")):
-                return legacy_cmd
-            for serial_dev in ("sclpconsole", "spapr-vty", "isa-serial"):
-                if devices.has_device(serial_dev):
-                    break
-            else:
-                return legacy_cmd
-
-            serial_id = "serial_id_%s" % name
-            cmd = " -chardev socket"
-            cmd += _add_option("id", serial_id)
-            cmd += _add_option("path", filename)
-            cmd += _add_option("server", "NO_EQUAL_STRING")
-            cmd += _add_option("nowait", "NO_EQUAL_STRING")
-            cmd += " -device %s" % serial_dev
-            if serial_dev == 'spapr-vty':
-                # Workaround for console issue, details:
-                # http://lists.gnu.org/archive/html/qemu-ppc/2013-10/msg00129.html
-                reg = 0x30000000 + 0x1000 * self.serial_ports.index(name)
-                cmd += _add_option("reg", hex(reg))
-            cmd += _add_option("chardev", serial_id)
-            return cmd
-
-        def add_chardev(devices, params, qid=None):
-            """
-            Generate qdevices.CharDevice object
-
-            :param devices: device container object
-            :param params: dict to create char device object
-            :param qid: char device ID
-            """
-            if not devices.has_option("chardev"):
-                logging.warn("'chardev' option not support")
-                return None
-
-            dynamic = False
-            chardev = qdevices.CharDevice(params=params)
-            if not qid:
-                qid = utils_misc.generate_random_id()
-                dynamic = True
-            chardev.set_param("id", qid, dynamic=dynamic)
-
-            return chardev
-
-        def add_virtio_port(devices, chardev, params, name, bus, index=None):
-            """
-            Appends virtio_serialport or virtio_console device to cmdline.
-            :param chardev: qdevices.CharDevice object
-            :param params: Space sepparated chardev params
-            :param name: Name of the port
-            :param bus: Which virtio-serial-pci device use
-            :param index: Index of the current virtio_port
-            """
-            def set_extra_options(virtio_port, params):
-                """Set extra params pairs"""
-                extra_params = params.get('virtio_port_params', '')
-                for _ in extra_params.split():
-                    try:
-                        if "=" not in _:
-                            key, value = _, "NO_EQUAL_STRING"
-                        else:
-                            key, value = _.split('=')
-                        virtio_port.set_param(key, value)
-                    except Exception:
-                        msg = ("Invaild params %s in " % _ +
-                               "'virtio_port_param' = %s" % extra_params)
-                        logging.error(msg)
-                return virtio_port
-
-            # used by spiceagent (com.redhat.spice.*)
-            if 'console' in params.get('virtio_port_type'):
-                port_type = 'virtconsole'
-            else:
-                port_type = 'virtserialport'
-            virtio_port = QDevice(port_type)
-
-            if not virtio_port.get_param("id"):
-                devid = utils_misc.generate_random_id()
-                virtio_port.set_param("id", devid, dynamic=True)
-
-            if params.get('virtio_port_name_prefix'):
-                prefix = params["virtio_port_name_prefix"]
-                name = "%s%d" % (prefix, index)
-            virtio_port.set_param("name", name)
-
-            if bus:
-                virtio_port.set_param("bus", bus)
-
-            virtio_port.set_param("chardev", chardev.get_qid())
-            set_extra_options(virtio_port, params)
-
-            return virtio_port
 
         def add_log_seabios(devices):
             if not devices.has_device("isa-debugcon"):
@@ -1632,83 +1528,46 @@ class VM(virt_vm.BaseVM):
                 devices.insert(pvpanic_dev)
 
         # Add serial console redirection
-        for serial in params.objects("serials"):
+        serials = params.objects('serials')
+        if serials:
+            self.serial_session_device = serials[0]
+            host = params.get('chardev_host', '127.0.0.1')
+            free_ports = utils_misc.find_free_ports(
+                5000, 6000, len(serials), host)
+            reg_count = 0
+        for index, serial in enumerate(serials):
             serial_filename = vm.get_serial_console_filename(serial)
-            cmd = add_serial(devices, serial, serial_filename)
-            devices.insert(StrDev('SER-%s' % serial, cmdline=cmd))
+            serial_params = params.object_params(serial)
+            # Workaround for console issue, details:
+            # http://lists.gnu.org/archive/html/qemu-ppc/2013-10/msg00129.html
+            if 'ppc' in params.get('vm_arch_name', arch.ARCH)\
+                    and serial_params.get('serial_type') == 'spapr-vty':
+                reg = 0x30000000 + 0x1000 * reg_count
+                serial_params['serial_reg'] = hex(reg)
+                reg_count += 1
+            backend = serial_params.get('chardev_backend',
+                                        'unix_socket')
+            if backend in ['udp', 'tcp_socket']:
+                serial_params['chardev_host'] = host
+                serial_params['chardev_port'] = free_ports[index]
+            serial_devices = devices.serials_define_by_params(
+                serial, serial_params, serial_filename)
 
-        # Add virtio_serial ports
-        if not devices.has_device("virtconsole"):
-            logging.warn("virt-console/serialport devices are not supported")
-        else:
-            no_virtio_serial_pcis = 0
-            no_virtio_ports = 0
-            virtio_port_spread = int(params.get('virtio_port_spread', 2))
-            for port_name in params.objects("virtio_ports"):
-                port_params = params.object_params(port_name)
-                parent_bus = _get_pci_bus(
-                    devices, port_params, "vio_port", True)
-                bus = params.get('virtio_port_bus', False)
-                if bus is not False:     # Manually set bus
-                    bus = int(bus)
-                elif not virtio_port_spread:
-                    # bus not specified, let qemu decide
-                    pass
-                elif not no_virtio_ports % virtio_port_spread:
-                    # Add new vio-pci every n-th port. (Spread ports)
-                    bus = no_virtio_serial_pcis
-                else:  # Port not overriden, use last vio-pci
-                    bus = no_virtio_serial_pcis - 1
-                    if bus < 0:     # First bus
-                        bus = 0
-                # Add virtio_serial_pcis
-                # Multiple virtio console devices can't share a
-                # single virtio-serial-pci bus. So add a virtio-serial-pci bus
-                # when the port is a virtio console.
-                if (port_params.get('virtio_port_type') == 'console' and
-                        params.get('virtio_port_bus') is None):
-                    if '-mmio:' in params.get('machine_type'):
-                        dev = QDevice('virtio-serial-device')
-                    elif params.get('machine_type').startswith("s390"):
-                        dev = QDevice("virtio-serial-ccw")
-                    else:
-                        dev = QDevice(
-                            'virtio-serial-pci', parent_bus=parent_bus)
-                    dev.set_param('id',
-                                  'virtio_serial_pci%d' % no_virtio_serial_pcis)
-                    devices.insert(dev)
-                    no_virtio_serial_pcis += 1
-                for i in range(no_virtio_serial_pcis, bus + 1):
-                    if '-mmio:' in params.get('machine_type'):
-                        dev = QDevice('virtio-serial-device')
-                    elif params.get('machine_type').startswith("s390"):
-                        dev = QDevice("virtio-serial-ccw")
-                    else:
-                        dev = QDevice(
-                            'virtio-serial-pci', parent_bus=parent_bus)
-                    dev.set_param('id', 'virtio_serial_pci%d' % i)
-                    devices.insert(dev)
-                    no_virtio_serial_pcis += 1
-                if bus is not False:
-                    bus = "virtio_serial_pci%d.0" % bus
-                # Add actual ports
-                char_params = port_params.copy()
-                backend = port_params.get("virtio_port_chardev", "socket")
-                port_file = self.get_virtio_port_filename(port_name)
-                char_params.update({"backend": backend,
-                                    "server": "yes",
-                                    "nowait": "yes",
-                                    "name": port_name,
-                                    "path": port_file})
-                char_dev = add_chardev(devices, char_params)
-                virtio_port = add_virtio_port(devices,
-                                              char_dev,
-                                              port_params,
-                                              port_name,
-                                              bus,
-                                              no_virtio_ports)
-                devices.insert([char_dev, virtio_port])
-                no_virtio_ports += 1
+            devices.insert(serial_devices)
+
+            # Create virtio_ports (virtserialport and virtconsole)
+            serial_type = serial_params['serial_type']
+            if serial_type.startswith('virt'):
+                if backend == "spicevmc":
+                    serial_filename = 'dev%s' % serial
+                port_name = serial_devices[-1].get_param('name')
+                if "console" in serial_type:
+                    self.virtio_ports.append(qemu_virtio_port.VirtioConsole(
+                        serial, port_name, serial_filename))
+                else:
+                    self.virtio_ports.append(qemu_virtio_port.VirtioSerial(
+                        serial, port_name, serial_filename))
+
         # Add virtio-rng devices
         for virtio_rng in params.objects("virtio_rngs"):
             rng_params = params.object_params(virtio_rng)
@@ -2442,43 +2301,23 @@ class VM(virt_vm.BaseVM):
         Let's consider the first serial port as serial console.
         Note: requires a version of netcat that supports -U
         """
-        try:
-            tmp_serial = self.serial_ports[0]
-        except IndexError:
+        if self.serial_session_device is None:
             logging.warning("No serial ports defined!")
             return
-        log_name = "serial-%s-%s.log" % (tmp_serial, self.name)
+        log_name = "serial-%s-%s.log" % (
+            self.serial_session_device, self.name)
         self.serial_console_log = os.path.join(utils_misc.get_log_file_dir(),
                                                log_name)
+        file_name = self.get_serial_console_filename(
+            self.serial_session_device)
         self.serial_console = aexpect.ShellSession(
-            "nc -U %s" % self.get_serial_console_filename(tmp_serial),
+            "nc -U %s" % file_name,
             auto_close=False,
             output_func=utils_misc.log_line,
             output_params=(log_name,),
             prompt=self.params.get("shell_prompt", "[\#\$]"),
             status_test_command=self.params.get("status_test_command",
                                                 "echo $?"))
-        del tmp_serial
-
-    def create_virtio_console(self):
-        """
-        Establish a session with the serial console.
-        """
-        for port in self.virtio_ports:
-            if isinstance(port, qemu_virtio_port.VirtioConsole):
-                logfile = "serial-%s-%s.log" % (port.name, self.name)
-                socat_cmd = "nc -U %s" % port.hostfile
-                self.virtio_console = aexpect.ShellSession(
-                    socat_cmd,
-                    auto_close=False,
-                    output_func=utils_misc.log_line,
-                    output_params=(logfile,),
-                    prompt=self.params.get("shell_prompt", "[\#\$]"))
-                return
-        if self.virtio_ports:
-            logging.warning(
-                "No virtio console created in VM. Virtio ports: %s", self.virtio_ports)
-        self.virtio_console = None
 
     def update_system_dependent_devs(self):
         # Networking
@@ -2840,8 +2679,10 @@ class VM(virt_vm.BaseVM):
                     raise virt_vm.VMPAError(pa_type)
 
             # Create serial ports.
-            for serial in params.objects("serials"):
-                self.serial_ports.append(serial)
+            for serial in params.objects('serials'):
+                if not params.object_params(serial).get(
+                        "serial_type").startswith('virt'):
+                    self.serial_ports.append(serial)
 
             if (name is None and params is None and root_dir is None and
                     self.devices is not None):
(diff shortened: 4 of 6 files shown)


vivianQizhu:
Generate correct command line:
```
    -chardev socket,id=qmp_id_qmpmonitor1,path=/var/tmp/avocado_7boIC5/monitor-qmpmonitor1-20180703-134454-w7aI0QFb,server,nowait \
    -chardev socket,id=qmp_id_catch_monitor,path=/var/tmp/avocado_7boIC5/monitor-catch_monitor-20180703-134454-w7aI0QFb,server,nowait \
    -device pvpanic,ioport=0x505,id=idVzkC1q \
    -chardev socket,id=chardev_serial0,host=127.0.0.1,port=5000,server,nowait \
    -device isa-serial,id=serial0,chardev=chardev_serial0 \
    -chardev socket,id=chardev_vc1,host=127.0.0.1,port=5001,server,nowait \
    -device virtio-serial-pci,id=virtio_serial_pci0,bus=pci.0,addr=0x3 \
    -device virtconsole,id=vc1,name=vc1,chardev=chardev_vc1,bus=virtio_serial_pci0.0,nr=1 \
    -chardev socket,id=chardev_vc2,host=127.0.0.1,port=5002,server,nowait \
    -device virtconsole,id=vc2,name=vc2,chardev=chardev_vc2,bus=virtio_serial_pci0.0,nr=2 \
    -chardev socket,id=chardev_vc3,host=127.0.0.1,port=5003,server,nowait \
    -device virtconsole,id=vc3,name=vc3,chardev=chardev_vc3,bus=virtio_serial_pci0.0,nr=3 \
    -chardev socket,id=chardev_vc4,host=127.0.0.1,port=5004,server,nowait \
    -device virtconsole,id=vc4,name=vc4,chardev=chardev_vc4,bus=virtio_serial_pci0.0,nr=4 \
    -chardev socket,id=chardev_vs1,host=127.0.0.1,port=5005,server,nowait \
    -device virtserialport,id=vs1,name=vs1,chardev=chardev_vs1,bus=virtio_serial_pci0.0,nr=5 \
    -chardev socket,id=chardev_vs2,host=127.0.0.1,port=5006,server,nowait \
    -device virtserialport,id=vs2,name=vs2,chardev=chardev_vs2,bus=virtio_serial_pci0.0,nr=6 \
    -chardev socket,id=chardev_vs3,host=127.0.0.1,port=5007,server,nowait \
    -device virtserialport,id=vs3,name=vs3,chardev=chardev_vs3,bus=virtio_serial_pci0.0,nr=7 \
    -chardev socket,id=chardev_vs4,host=127.0.0.1,port=5008,server,nowait \
    -device virtserialport,id=vs4,name=vs4,chardev=chardev_vs4,bus=virtio_serial_pci0.0,nr=8  \
```

virttest/qemu_devices/qcontainer.py
```diff
@@ -1666,6 +1666,163 @@ def images_define_by_params(self, name, image_params, media=None,
...
+                devices[-1].set_param('localaddr', kwargs.get('localaddr'))
+                devices[-1].set_param('localport', kwargs.get('localport'))
+            else:
+                devices[-1].set_param('to', kwargs.get('to'))
+                devices[-1].set_param('nodelay', kwargs.get('nodelay'))
+        if 'socket' in backend:  # tcp_socket & unix_socket
+            devices[-1].set_param('server', kwargs.get('server'))
+            devices[-1].set_param('nowait', kwargs.get('nowait'))
+            devices[-1].set_param('reconnect', kwargs.get('reconnect'))
+        devices[-1].format_params(devices[-1].params)
```
xutian: split lines from 1691 to 1729 into function  chardev_define_by_params.

virttest/utils_misc.py
```diff
@@ -440,14 +440,16 @@ def is_port_free(port, address):
     return free
 
 
-def find_free_port(start_port, end_port, address="localhost"):
+def find_free_port(start_port, end_port, address="localhost", reserved=[]):
```
xutian: don't used empty list as default value.

virttest/qemu_devices/qcontainer.py
vivianQizhu: @xutian Here I can not go through kwargs and set all params inside it for `device[-1]`, since the kwargs actually contains some other options for other devices created later, not only for this device. So that's very hard to distinguish them earlier in `define_by_params`

virttest/utils_misc.py
vivianQizhu: @xutian Updated default value to `None`, thanks.

virttest/qemu_devices/qcontainer.py
vivianQizhu: @xutian I have tried to move it to `serials_define_by_params`, only including chardev related options in **kwargs, and the options for other devices is laid outside as individual parameters. Thanks for your suggestions.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1703,6 +1703,154 @@ def images_define_by_params(self, name, image_params, media=None,
...
+        :param serial_params: serial params
+        :param name: the name of the serial device
+        :param file_name: the file path of the serial device (optional)
+        :param index: the index of the serial device (optional)
+        :return: the device list that construct the serial device
+        """
+        serial_type = serial_params.get('serial_type', 'virtio')
+        port_type = None
+        machine_type = serial_params.get('machine_type')
+        if serial_type == 'virtio':  # translate virtio to ccw/device
```
xutian: line 1790 -  1806 are useless,  it should be instead with configuration in machine.cfg

virttest/qemu_devices/qdevices.py
```diff
@@ -2064,11 +2065,28 @@ def _dev2addr(self, device):
...
         if isinstance(busid, int):
             bus = busid
-        if device.get_param('unit'):
-            unit = int(device.get_param('unit'))
+        if device.get_param(self.unit_spec):
+            unit = int(device.get_param(self.unit_spec))
         return [bus, unit]
 
 
+class QSerialBus(QBusUnitBus):
```
sitoliu: Hi @vivianzhu, would you mind explaining the difference between QBusUnitBus and QPCIBus? Why serial bus (do you mean `virtio-serial-pci`?) should be a QBusUnitBus not a QPCIBus? Thanks!

vivianQizhu:
@luckyh @yanan-fu I have updated it per your comments, please help review, thanks.

luckyh:
See also https://github.com/avocado-framework/avocado-vt/pull/1768

vivianQizhu:
> See also #1768

Yeah, the failure there is caused by the confounded 'id' and 'name',  the issue will be fixed by the wrapper defined here.

rayx:
So, is my PR a duplicate effort again? I wonder if someone in the maintainers can put the overall test status on a wiki, with information like which tests don't work and the related issue number or PR number. That would help people who are outside redhat to use the test and contribute to it more efficiently, I guess.

vivianQizhu:
@rayx there are serials of issue on virtio console test, we are working on it. FYI, https://github.com/autotest/tp-qemu/pull/1154 https://github.com/autotest/tp-qemu/pull/1037

rayx:
@vivianQizhu OK, thanks for information. I'll leave it to your folks and perhaps try other tests (if I still have the motivation :)

luckyh:
@rayx sorry for that. For tp-qemu issues, currently we are using redhat internal bugzilla tool do the management (because of some legacy reasons), and yes it is quite unfriendly to upstream users. Well, there are about hundreds of issues in the list (since vt tests support extremely large set of variants), it is not possible to mirror them all to github.

You just raised a good suggestion, I'll see if we can do some improvement for it. Thanks for your contributions!

balamuruhans:
@rayx you can submit an issue in this github repo when you start working on new feature implementation or fixing any issue observed. I helps us to discuss to find existing solution or PR sent, it can avoid this situation.
Your contribution will really helps us in making framework better.Thank you :)

rayx:
Thank @luckyh and @balamuruhans for your reply. I was probably a bit optimistic and underestimated the complexity and size of qemu test. My purpose is really about finding aarch64 specific failures. I'll focus on those issues and log the common issues I find on github.

luckyh:
@rayx ok, I see. @ldoktor often runs aarch64 specific tests, so I think he is the right person to tell about the status of that, thanks.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1720,6 +1720,147 @@ def images_define_by_params(self, name, image_params, media=None,
...
+        :param path: the path option of chardev device(optional)
+        :param name: the name option of serial device(optional)
+        :param bus: the busid of parent bus(optional, virtio only)
+        :param nr: the nr(port) of the parent bus(optional, virtio only)
+        :param reg: reg option of isa-serial(optional)
+        :keyword chardev options per backend
+        :return: the device list that construct the serial device
+        """
+        if (not self.has_option('chardev') or
+                not self.has_device(serial_type)):
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu I think it should get `True` for `not self.has_device(serial_type)`, for the default definition `serial_type = "isa-serial"`, so it get `-serial` by L1743

virttest/qemu_devices/qcontainer.py
```diff
@@ -1720,6 +1720,147 @@ def images_define_by_params(self, name, image_params, media=None,
...
+            cmd = " -serial unix:'%s',server,nowait" % file_name
+            return qdevices.StrDev('SER-%s' % serial_id, cmdline=cmd)
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Thanks, it should be `qdevices.QStringDevice`, `StrDev = qdevices.QStringDevice` is defined in `qemu_vm` but we are in `qcontainer`, will update.

virttest/qemu_devices/qcontainer.py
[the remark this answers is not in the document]
yanan-fu: @vivianQizhu  `I think it should get True for not self.has_device(serial_type)`  ?   do you mean are do not have `isa-serial` device ?

The old logic is,  arm have `isa_serial` but it can not work normally, so use `- serial` instead.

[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Really? Thanks for the information, so is that a workaround for the product issue? I will modify the code to handle it if so.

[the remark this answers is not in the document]
yanan-fu: @vivianQizhu  please refer the old comment, and if we do not test with other arch,  we should keep old logic.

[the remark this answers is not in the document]
vivianQizhu: @yanan-fu You are right, I will update it. Thanks.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1720,6 +1720,147 @@ def images_define_by_params(self, name, image_params, media=None,
...
+        prefix = serial_params.get('virtio_port_name_prefix')
+        name = serial_id
+        name = prefix + name if prefix else name
+        serial_params['name'] = name
+
+        chardev_args = {'logfile': serial_params.get('chardev_logfile'),
+                        'logappend': serial_params.get('chardev_logappend')}
+        backend = serial_params.get('chardev_backend', 'unix_socket')
+        if backend not in ['socket', 'vc', 'ringbuf',
+                           'spiceport', 'spicevmc']:
```
yanan-fu: `socket` and  `vc`  have option `mux`, why exclude them ?

virttest/qemu_devices/qcontainer.py
```diff
@@ -1720,6 +1720,147 @@ def images_define_by_params(self, name, image_params, media=None,
...
+        if backend in ['file', 'pipe', 'serial',
+                       'tty', 'parallel', 'parport']:
+            chardev_args.update({'path': file_name})
+        elif backend in ['spicevmc', 'spiceport']:
+            chardev_args.update({'debug': serial_params.get('chardev_debug'),
+                                 'name': name})
+        elif backend in ['stdio']:
+            chardev_args.update({'signal': serial_params.get('chardev_signal')})
+        elif 'unix' in backend:  # unix_socket
+            chardev_args.update({'path': file_name})
```
yanan-fu: Can move it together with L1827~1829 ( when use tcp_socket, unix_socket)

virttest/qemu_devices/qdevices.py
```diff
@@ -2091,11 +2092,28 @@ def _dev2addr(self, device):
...
+    def __init__(self, busid, bus_type, aobject=None, max_ports=32):
+        """
+        :param busid: bus id
+        :param bus_type: bus type(virtio-serial-device, virtio-serial-pci)
+        :param aobject: autotest object
+        :param max_ports: max ports, default 32
+        """
+
+        super(QSerialBus, self).__init__(busid, 'SERIAL', [1, max_ports],
+                                         aobject, bus_type, unit_spec='nr')
```
yanan-fu: Could you explain which device has option `nr` ?   I can't get it from qemu side.

vivianQizhu: @yanan-fu 
```
#  /usr/libexec/qemu-kvm -device virtserialport,?
virtserialport.name=str
virtserialport.chardev=str (ID of a chardev to use as a backend)
virtserialport.nr=uint32
```
That should the virtio-serial bus distributing to each device.

virttest/qemu_devices/qdevices.py
```diff
@@ -2039,7 +2037,7 @@ class QBusUnitBus(QDenseBus):
 
     """ Implementation of bus-unit bus (ahci, ide) """
```
yanan-fu: Need update this doc, since it is not only for 'unit' option now,  then not only ahci and ide.

virttest/qemu_devices/qcontainer.py
yanan-fu: Default value for `backend` is `unix_socket`,  but here it use `['socket', 'vc', 'ringbuf','spiceport', 'spicevmc']`.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1720,6 +1720,147 @@ def images_define_by_params(self, name, image_params, media=None,
...
+                chardev_args.update(
+                    {'localaddr': serial_params.get('chardev_localaddr'),
+                     'localport': serial_params.get('chardev_localport')})
+            else:
+                chardev_args.update({'to': serial_params.get('chardev_to'),
+                                     'nodelay': serial_params.get('chardev_nodelay')})
+        if 'socket' in backend:  # tcp_socket & unix_socket
+            chardev_args.update({'server': serial_params.get('chardev_server', 'yes'),
+                                 'nowait': serial_params.get('chardev_nowait', 'yes'),
+                                 'reconnect': serial_params.get('chardev_reconnect')})
```
yanan-fu: New design: param name is `chardev_$option_name`.
Why don't use class CharDevice directly to generate chardev_args ?  (improve it to match the new request)
In this way,  param name should be same with option name.
```
1012         options = self.get_supported_options(backend)                           
1013         params = params.copy_from_keys(options)  
```

vivianQizhu: @yanan-fu The method you mentioned is not recommended by other maintainer, and it is indeed not flexible enough for all conditions. I have no strong opinion on this, just follow the way how we create image.

virttest/qemu_devices/qcontainer.py
vivianQizhu: @yanan-fu That's the condition statement for the option `mux`, nothing to do with what is default value.

vivianQizhu: @yanan-fu Ah I get your point, `socket` should be `unix_socket`, but yes,  `socket` and `vc` have the option, should be removed from this list. I just copy it from CharDevice, not confirm all of them, my careless. Thank you for being so careful.

virttest/qemu_devices/qcontainer.py
yanan-fu: @vivianQizhu  In my option,  L1821~1852 are almost duplicate with qdevices.CharDevice. 
For image, there do not have similar function to do this,   but for chardev, it have.

I also think about delete this part (get_supported_options, copy_from_keys, format_params) from CharDevice, but we can since it is useful for hotplug/ hot-unplug. Some other cases use it.

If use the current design, i am afraid we have to change both of these two parts if qemu have change in the future.

virttest/qemu_devices/qdevices.py
vivianQizhu: @yanan-fu I would like to keep the description in QSerialBus instead of here, the class QBusUnitBus can not represent SerialBus directly, the inherited relationship is sufficient to explain their relation.

virttest/qemu_devices/qdevices.py
```diff
@@ -2054,8 +2052,9 @@ def __init__(self, busid, bus_type, lengths, aobject=None, atype=None):
         """
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Not sure what's wrong here, but I am not able to modify it, it looks normal in my local environment...

virttest/qemu_devices/qdevices.py
yanan-fu: @vivianQizhu  My meaning is `bus-unit` is not suitable, as it can be `nr` too.
And for `ahci, ide`,   there also have  `QAHCIBus` and `QIDEBus`.  I think there is not conflicts with `QBusUnitBus`  and `QSerialBus`.   We can add docs for both of them. What do you think ?

virttest/qemu_devices/qcontainer.py
vivianQizhu: @yanan-fu All the chardev device should be created from the new introduced wrappers in the further, so there will be no other case use it. chardev usually is a set of devices, it can not be created alone. That's why it is similar with image: a complete device is composed by a set of device(bus/backend/frontend). We need to handle other factors related to environment besides options, like machine types, and the relation between devices, which are not proper to handle in device object, which means, CharDevice is not able to handle all the options. Thus, we have to define a wrapper anyway. 
In conclusion, the options of the there part of a complete device should be handled together, so not in device level.

vivianQizhu: @yanan-fu And, for other device that use chardev as backend(like nic), they should handle the options in its device creation part, not inside CharDevice object either.  So here we take a real device as an entirety in our design, not a single command line.

yanan-fu: @vivianQizhu  L1835~L1863 have nothing to do with machine type and others, it is only related with chardev options,  and ChaDevice.get_supported_options do the same thing( together with  copy_from_keys).  Correct me if i am wrong.

For this question:  `so there will be no other case use it`:  CharDevice support hotplug/hot-unplug,  `qemu_guest_agent_hotplug` use it.  How can we run hotplug without CharDevice ?  

If no CharDevice, current implement is good for me,  but why don't we call it directly while we can  not delete it.

vivianQizhu: @yanan-fu 
My point is, they should be handled together with all related devices, not in QDevice object level.
For hotplug, it should be all replaced with simple_hotplug().

vivianQizhu: @yanan-fu Well, actually I was about to only use CharDevice at the beginning. @luckyh Would you help to share your thought here? Thanks.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1720,6 +1720,150 @@ def images_define_by_params(self, name, image_params, media=None,
...
```
yanan-fu: @vivianQizhu  Checked all the options for `-chardev`.
1. For backend `ringbuf`, miss `size`.
2. For backend `unix_socket`, miss `telnet`.   and miss `websocket` only for qemu-3.1.0.
3. For backend `tcp_socket`, miss `telnet`, `tls-creds`, and miss  `websocket` only for qemu-3.1.0.
4. For backend `vc`, miss `[[,width=width][,height=height]][[,cols=cols][,rows=rows]]`.

I do not know if there are cases test these options, ignore me if no need to add them specially.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1720,6 +1720,150 @@ def images_define_by_params(self, name, image_params, media=None,
...
+        if backend in ['unix_socket', 'file', 'pipe', 'serial',
+                       'tty', 'parallel', 'parport']:
+            chardev_args.update({'path': file_name})
+        elif backend in ['spicevmc', 'spiceport']:
+            chardev_args.update({'debug': serial_params.get('chardev_debug'),
+                                 'name': name})
+        elif backend == 'stdio':
+            chardev_args.update({'signal': serial_params.get('chardev_signal')})
+        elif backend in ['udp', 'tcp_socket']:
+            chardev_args.update({'host': serial_params.get('chardev_host'),
```
yanan-fu: these  3 `elif` can be `if` ?

virttest/qemu_devices/qcontainer.py
```diff
@@ -1720,6 +1720,150 @@ def images_define_by_params(self, name, image_params, media=None,
...
+        virtio_port_type = serial_params.get('virtio_port_type')
+        if virtio_port_type:
+            if "-mmio" in machine:
+                controller_suffix = 'device'
+            elif machine.startswith("s390"):
+                controller_suffix = 'ccw'
+            else:
+                controller_suffix = 'pci'
+            serial_type = '%s-%s' % (serial_type, controller_suffix)
+            virtio_port_type = "virt%s" % virtio_port_type
```
yanan-fu: how about use `virtserialport`, `virtconsole` as `virtio_port_type` ?  It is more easy to understand as it is same with qemu help output.  And then no need to add prefix `virt` here.

virttest/qemu_devices/qcontainer.py
vivianQizhu: @yanan-fu `elif` is more efficient than `if`

virttest/qemu_devices/qcontainer.py
vivianQizhu: @yanan-fu `virtio_port_type` is previously defined as `serialport` and `console`, just want to adopt previous usage.

virttest/qemu_devices/qcontainer.py
vivianQizhu: @yanan-fu `telnet` and `tls-creds` should be included, for others there are no requirements currently. I will update it for `telnet` and `tls-creds` at least. Thanks.

virttest/qemu_devices/qdevices.py
```diff
@@ -1009,8 +1009,6 @@ def __init__(self,
                  child_bus=None):
         backend = params.get("backend", "socket")
         self.verify_supported_backend(backend)
-        options = self.get_supported_options(backend)
-        params = params.copy_from_keys(options)
         params = self.format_params(params)
```
yanan-fu: After remove `params = params.copy_from_keys(options)`,  `params=None` here, so it is meaningless for `format_params`.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1720,6 +1720,150 @@ def images_define_by_params(self, name, image_params, media=None,
...
+        legacy_dev = qdevices.QStringDevice('SER-%s' % serial_id,
+                                            cmdline=legacy_cmd)
+        arm_serial = (serial_type == 'isa-serial'
+                      and 'arm' in serial_params.get("machine_type", ""))
+        if (arm_serial or not self.has_option("chardev")
+            or not self.has_device(serial_type)):
+            return legacy_dev
+
+        virtio_port_type = serial_params.get('virtio_port_type')
+        if virtio_port_type:
```
yanan-fu: In qemu-vm, we already use:
```
            virtio_port_type = serial_params.get('virtio_port_type')
            if virtio_port_type:
                serial_params['serial_type'] = 'virtio-serial'
```
and have got `serial_type` from above, so how about  `if serial_type.startswith('virtio')`  or `if serial_type == 'virtio-serial'` ?   and i prefer the second one.

yanan-fu:
Need replace `virtio_port_chardev` with `chardev_backend` in base.cfg, although it is a comment.

virttest/qemu_devices/qcontainer.py
yanan-fu: @vivianQizhu  I understand your point. 
Previously,  `console`, `virtio_console`, `serialport`, `virtio_serial` all can use,  since framwork accept all of them, and convert them to acceptable port type of qemu command line.
```
port_params.get('virtio_port_type') in ("console", "virtio_console"):
```
```
if 'console' in params.get('virtio_port_type'):
                port_type = 'virtconsole'
            else:
                port_type = 'virtserialport'
```

For me, it is confused.   If `console` , `virtio_console` can only be `virtconsole`,  and `serialport`, `virtio_serial` can only be `virtserialport`,  use the final port type is more clearly.
While in your another patch https://github.com/autotest/tp-qemu/pull/1434,  you have correct `virtio_serial`  to `serialport`, i think we can update them together.

vivianQizhu: @yanan-fu Yes, it makes sense, I will update them.

vivianQizhu: @yanan-fu There is variants named `virtserialport` and `virtconsole` in test case, I am afraid there is name conflict if modified them, I will need rename the variants and reorganize those `only` `no` keyword for them. I think leave them there is better, not really necessary to reorganize all params, which makes no improvement, just for save this postfix adding. Thanks.

virttest/qemu_devices/qcontainer.py
vivianQizhu: @yanan-fu And for here, qemu-vm is not the only user for the API, we must make sure `virtio_port_type` is defined for `virtio-serial` type anytime when user call it, e.g. define a new device for hotplug.

virttest/qemu_devices/qdevices.py
vivianQizhu: @yanan-fu Actually they are independent, `format_params()` translate params before `yes` `no` `Ture` `False`, which is still in required here.

virttest/qemu_devices/qcontainer.py
vivianQizhu: @yanan-fu I am not adding `tls-creds` for now, the test case related to it requires complex configuration and in low priority, we have no plan to automate it recently.

virttest/qemu_devices/qcontainer.py
yanan-fu: @vivianQizhu  oh~  i see those variants,  if so, it is ok to use `console` and `serialport`.

virttest/qemu_devices/qcontainer.py
yanan-fu: @vivianQizhu  I have different option,  the `serial_type` is depends on `virtio_port_type` all the time with the current implement, in this API,  it already use `L1798    serial_type = serial_params['serial_type']`.

I am not very clear about other usage now, but if it is a problem,  i think we need to update `serial_type` first.

vivianQizhu:
@yanan-fu Updated as we discussed offline, please help review, thanks.

vivianQizhu:
Updated with:
1. Remove param `virtio_port_type`, use single `serial_type` to represent virtconsole and virtserialport.
2. Removed most unless chardev options, keep the original filter in the object `CharDevice`, so pass `Params()` instead of general dictionary when initialize it.
3. Initialize a new bus for virtconsole since they can not share a single bus, which I missed previously.
Also update the tp-qemu PR together and generally test it generates correct command line.
@luckyh @yanan-fu Please help review again, thanks.

virttest/qemu_vm.py
```diff
@@ -2786,7 +2633,9 @@ def create(self, name=None, params=None, root_dir=None,
 
             # Create serial ports.
             for serial in params.objects("serials"):
-                self.serial_ports.append(serial)
+                if params.object_params(serial).get(
+                        "serial_type") != 'virtio-serial':
```
yanan-fu: We don't use `virtio-serial` now,  we use `virtserialport`, `virtconsole` and so on.

virttest/qemu_vm.py
```diff
@@ -2390,41 +2259,19 @@ def _create_serial_console(self):
...
             logging.warning("No serial ports defined!")
             return
-        log_name = "serial-%s-%s.log" % (tmp_serial, self.name)
+        log_name = "serial-%s-%s.log" % (
+            self.serial_session_device, self.name)
         self.serial_console_log = os.path.join(utils_misc.get_log_file_dir(),
                                                log_name)
         self.serial_console = aexpect.ShellSession(
-            "nc -U %s" % self.get_serial_console_filename(tmp_serial),
+            "nc -U %s" % self.serial_console_log,
```
yanan-fu: `serial_console_log` is the log path + log file name, it is used to save the serial log.
`get_serial_console_filename(serial_session_device)` is the `path` in qemu command line.
What we can use for 'nc' is the `path` in qemu command line,  not log file name.

virttest/qemu_vm.py
```diff
@@ -1613,83 +1516,49 @@ def sort_key(dev):
                 devices.insert(pvpanic_dev)
 
         # Add serial console redirection
-        for serial in params.objects("serials"):
+        serials = params.objects('serials')
+        if serials:
+            self.serial_session_device = serials[0]
+            host = params.get('serial_chardev_host', '127.0.0.1')
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Honestly I don't want support that, that's too complicated and meaningless from testing perspective, it is only about socket transmission, same host or multi-host has no significant difference, so I will update the example in base.cfg. Thanks.

vivianQizhu:
@yanan-fu Thanks for review, updated.

vivianQizhu:
@yanan-fu I have updated it with a new function `chardev_define_by_params`, please help review, thanks.

shared/cfg/base.cfg
```diff
@@ -129,22 +129,22 @@ q35, arm64-pci, arm64-mmio:
...
-# virtio_port_chardev_vc1 = socket
-# chardev name prefix (port index will be appended)
+# serials = "serial0 vs1 vs2 vc1"
+# serial_type_vs1 = virtserialport
+# serial_type_vc1 = virtconsole
+# Chardev backend type: 'unix_socket' by default
+# serial_chardev_backend_vs1 = unix_socket
+# serial_chardev_backend_vc1 = tcp_socket
+# Chardev host for tcp/udp socket backend
+# serial_chardev_host = $host_ip (127.0.01 by default)
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Updated, thanks.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1718,6 +1719,146 @@ def images_define_by_params(self, name, image_params, media=None,
...
+                 'port': params['chardev_port'],
+                 'ipv4': params.get('chardev_ipv4'),
+                 'ipv6': params.get('chardev_ipv6')})
+        if 'socket' in backend:  # tcp_socket & unix_socket
+            chardev_param.update(
+                {'server': params.get('chardev_server', 'yes'),
+                 'nowait': params.get('chardev_nowait', 'yes')})
+        elif backend in ['spicevmc', 'spiceport']:
+            chardev_param.update({'debug': params.get('chardev_debug'),
+                                 'name':  params.get('serial_name')})
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Should be `chardev_name`, will update, thanks.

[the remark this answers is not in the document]
yanan-fu: And no call for `chardev_name`  from `params['chardev_name'] = name`,  do you mean  `chardev_name` ?

virttest/qemu_devices/qcontainer.py
```diff
@@ -1718,6 +1719,146 @@ def images_define_by_params(self, name, image_params, media=None,
...
+                controller_suffix = 'device'
+            elif machine.startswith("s390"):
+                controller_suffix = 'ccw'
+            else:
+                controller_suffix = 'pci'
+            bus_type = 'virtio-serial-%s' % controller_suffix
+
+        prefix = params.get('virtio_port_name_prefix')
+        name = serial_id
+        name = prefix + name if prefix else name
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Updated, thanks.

shared/cfg/base.cfg
```diff
@@ -129,22 +129,22 @@ q35, arm64-pci, arm64-mmio:
...
+# serial_chardev_backend_vc1 = tcp_socket
+# Chardev host for tcp/udp socket backend
+# serial_chardev_host = $host_ip (127.0.0.1 by default)
+# Virtio port name prefix (port index will be appended)
 # virtio_port_name_prefix = "com.redhat.spice."
 # By default the port name is used (vc1)
 # virtio_port_name_prefix_vc1 = ""
 
+serials = "serial0"
+serial_type = "isa-serial"
```
yanan-fu: To avoid  other operation in case level affect the default serial session by mistake,  how do you think about use fixed `serial_type_serial0 = isa-serial` ?  
And in qemu_vm,  use `self.serial_session_device = serial0` instead of `serials[0]`
```
        if serials:
            self.serial_session_device = serials[0]
```

virttest/qemu_devices/qcontainer.py
```diff
@@ -1718,6 +1719,145 @@ def images_define_by_params(self, name, image_params, media=None,
...
+
+    def chardev_define_by_params(self, chardev_id, params, file_name=None):
+        """
+        Wrapper for creating -chardev device from params.
+        :param chardev_id: chardev id
+        :param params: chardev params
+        :param file_name: file name of chardev (optional)
+        :return: CharDevice object
+        """
+        backend = params.get('chardev_backend', 'unix_socket')
```
yanan-fu: No param `chardev_backend` now.
For serial, we have `serial_chardev_backend`,  if want to distinguish other device,  we may need make `backend` as a param of  `chardev_define_by_params` .

virttest/qemu_devices/qcontainer.py
```diff
@@ -1718,6 +1719,145 @@ def images_define_by_params(self, name, image_params, media=None,
...
+        chardev_param = Params({'backend': _backend, 'id': chardev_id})
+        if backend in ['unix_socket', 'file', 'pipe', 'serial',
+                       'tty', 'parallel', 'parport']:
+            chardev_param.update({'path': file_name})
+        elif backend in ['udp', 'tcp_socket']:
+            chardev_param.update(
+                {'host': params['chardev_host'],
+                 'port': params['chardev_port'],
+                 'ipv4': params.get('chardev_ipv4'),
+                 'ipv6': params.get('chardev_ipv6')})
```
yanan-fu: This 4 params have the same problem with `chardev_backend`.  Especially when use `params[***]`, no `get(***)`

virttest/qemu_devices/qcontainer.py
```diff
@@ -1718,6 +1719,145 @@ def images_define_by_params(self, name, image_params, media=None,
...
+        if 'socket' in backend:  # tcp_socket & unix_socket
+            chardev_param.update(
+                {'server': params.get('chardev_server', 'yes'),
+                 'nowait': params.get('chardev_nowait', 'yes')})
+        elif backend in ['spicevmc', 'spiceport']:
+            chardev_param.update({'debug': params.get('chardev_debug'),
+                                 'name':  params.get('chardev_name')})
```
yanan-fu: Failed as:   'qemu-kvm: -chardev spicevmc,id=chardev_vs3,name=com.redhat.spice.vs3: unsupported type name: com.redhat.spice.vs3
allowed spice char type names: vdagent, smartcard, usbredir

vivianQizhu:
@yanan-fu Updated,
1. Remove some `serial_` prefix.
2. Line https://github.com/avocado-framework/avocado-vt/pull/1639/files#diff-4045499c9c61cb3d07734c53a7dcbb26R1814 , corresponding to https://github.com/avocado-framework/avocado-vt/pull/1639/files#diff-3f2317089f10636f453dca6c4f880f6cL1701, to resolve the spice issue. It use same value of serial id(`port_name` previously) for `name`, let's just keep the original design, but also allowing customer it by defining `chardev_name`.

virttest/qemu_vm.py
```diff
@@ -1632,83 +1535,47 @@ def sort_key(dev):
...
-            devices.insert(StrDev('SER-%s' % serial, cmdline=cmd))
+            serial_params = params.object_params(serial)
+            # Workaround for console issue, details:
+            # http://lists.gnu.org/archive/html/qemu-ppc/2013-10/msg00129.html
+            if 'ppc' in params.get('vm_arch_name', arch.ARCH)\
+                    and serial_params.get('serial_type') == 'spapr-vty':
+                reg = 0x30000000 + 0x1000 * reg_count
+                serial_params['serial_chardev_reg'] = hex(reg)
+                reg_count += 1
+            backend = serial_params.get('serial_chardev_backend',
```
yanan-fu: `chardev_backend`

virttest/qemu_vm.py
```diff
@@ -1632,83 +1535,47 @@ def sort_key(dev):
...
             serial_filename = vm.get_serial_console_filename(serial)
-            cmd = add_serial(devices, serial, serial_filename)
-            devices.insert(StrDev('SER-%s' % serial, cmdline=cmd))
+            serial_params = params.object_params(serial)
+            # Workaround for console issue, details:
+            # http://lists.gnu.org/archive/html/qemu-ppc/2013-10/msg00129.html
+            if 'ppc' in params.get('vm_arch_name', arch.ARCH)\
+                    and serial_params.get('serial_type') == 'spapr-vty':
+                reg = 0x30000000 + 0x1000 * reg_count
+                serial_params['serial_chardev_reg'] = hex(reg)
```
yanan-fu: `chardev_reg`

virttest/qemu_vm.py
```diff
@@ -1632,83 +1535,47 @@ def sort_key(dev):
...
+                reg_count += 1
+            backend = serial_params.get('serial_chardev_backend',
+                                        'unix_socket')
+            if backend in ['udp', 'tcp_socket']:
+                serial_params['serial_chardev_host'] = host
+                serial_params['serial_chardev_port'] = free_ports[index]
```
yanan-fu: `chardev_host` and `chardev_port`

virttest/qemu_vm.py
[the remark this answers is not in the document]
yanan-fu: `chardev_host`

virttest/qemu_vm.py
vivianQizhu: @yanan-fu Updated, but this one is `serial_reg`, it's an option for serial device, not chardev.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1718,6 +1719,146 @@ def images_define_by_params(self, name, image_params, media=None,
...
+                 'ipv4': params.get('chardev_ipv4'),
+                 'ipv6': params.get('chardev_ipv6')})
+        if 'socket' in backend:  # tcp_socket & unix_socket
+            chardev_param.update(
+                {'server': params.get('chardev_server', 'yes'),
+                 'nowait': params.get('chardev_nowait', 'yes')})
+        elif backend in ['spicevmc', 'spiceport']:
+            chardev_param.update(
+                {'debug': params.get('chardev_debug'),
+                 'name': params.get('chardev_name')})
```
yanan-fu: same with before, test failed.

qemu command line:
```
    -chardev spicevmc,id=chardev_vs3,name=vs3 \
    -device virtserialport,id=vs3,name=com.redhat.spice.vs3,chardev=chardev_vs3,bus=virtio_serial_pci0.0,nr=3 \
```

failed reason:
```
qemu-kvm: -chardev spicevmc,id=chardev_vs3,name=vs3: unsupported type name: vs3
allowed spice char type names: vdagent, smartcard, usbredir'
```
I am not familiar with this,  `chardev_name` is must for `spicevmc` ?

vivianQizhu: @yanan-fu Please refer here: https://github.com/autotest/tp-qemu/pull/1434/files#diff-8d0a580773b347958522648c9657235eL10 should be `vdagent`

yanan-fu: @vivianQizhu  Got the point.  The value of `name` option for `chardev` can only be `vdagent, smartcard, usbredir`.

In this way,  i think `params['chardev_name'] = params.get('chardev_name', serial_id)` is unreasonable.
If user want to test backend `spicevmc`,  `chardev_name` should be necessary, user should give correct value,  othervise, case failed as `qemu-kvm: -chardev spicevmc,id=chardev_vs3: chardev: spice channel: no name given'`

vivianQizhu: @yanan-fu I think we better adopt the original usage(no `chardev_name`): https://github.com/avocado-framework/avocado-vt/pull/1639/files#diff-3f2317089f10636f453dca6c4f880f6cL1701, so by default user only need to define `serial_id` as `vdagent, smartcard, usbredir`, like previously.

vivianQizhu: @yanan-fu Removed `params['chardev_name'] = params.get('chardev_name', serial_id)`

virttest/qemu_devices/qcontainer.py
```diff
@@ -1718,6 +1719,145 @@ def images_define_by_params(self, name, image_params, media=None,
                                                    "bus_extra_params"),
                                                image_params.get("force_drive_format"))
 
+    def serials_define_by_variables(self, serial_id, serial_type, chardev_id,
+                                    bus_type=None, serial_name=None,
+                                    bus=None, nr=None, reg=None):
```
yanan-fu: Test pass now.
 Only one concern, why do not keep consistent with other serial option, i mean, `serial_name`, but not `serial_reg`, `serial_nr` ?

shared/cfg/base.cfg
```diff
@@ -129,22 +129,25 @@ q35, arm64-pci, arm64-mmio:
...
+# serials = "serial0 vs1 vs2 vc1"
+# serial_type_vs1 = virtserialport
+# serial_type_vc1 = virtconsole
+# Chardev backend type: 'unix_socket' by default
+# chardev_backend_vs1 = unix_socket
+# chardev_backend_vc1 = tcp_socket
+# Chardev host for tcp/udp socket backend
+# chardev_host = $host_ip (127.0.0.1 by default)
+# chardev_ipv4 = yes
+# chardev_ipv6 = yes
```
yanan-fu: By default,  no `ipv4` and `ipv6`. It is
```
                 'ipv4': params.get('chardev_ipv4'),
                 'ipv6': params.get('chardev_ipv6')})
```
qemu command line `-chardev udp,host=127.0.0.1,id=chardev_vc2,port=5746`

virttest/qemu_vm.py
```diff
@@ -1632,83 +1535,47 @@ def sort_key(dev):
...
+        serials = params.objects('serials')
+        if serials:
+            self.serial_session_device = serials[0]
+            host = params.get('chardev_host', '127.0.0.1')
+            free_ports = utils_misc.find_free_ports(
+                5000, 6000, len(serials), host)
+            reg_count = 0
+        self.virtio_ports = []
+        for index, serial in enumerate(serials):
             serial_filename = vm.get_serial_console_filename(serial)
```
yanan-fu: Since  isa-serial  and  virtio serial use same method to get `serial_filename` now. I wonder if we need to update  function `get_virtio_port_filename` in `virt_vm.py`.  Since it use  `virtio_port-%s-%s" % (port_name, self.instance)`. 

It may do not affect our test, as in qemu_vm  `get_virtio_port_filename`  is ok.

shared/cfg/base.cfg
vivianQizhu: @yanan-fu It's just an example to show how to use it, not mean default here.

virttest/qemu_vm.py
```diff
@@ -142,6 +142,7 @@ def __init__(self, name, params, root_dir, address_cache, state=None):
             self.process = None
             self.serial_ports = []
             self.serial_console_log = None
+            self.serial_session_device = None
             self.virtio_console = None
```
yanan-fu: We still have `self.virtio_ports` now, but  function `create_virtio_console` be totally removed.

vivianQizhu: @yanan-fu You mean `self.virtio_console`?

yanan-fu: @vivianQizhu  we have `self.virtio_ports`,  but no `self.virtio_console` as function `create_virtio_console`  be removed.   `_create_serial_console` is only for `serials[0]`

shared/cfg/base.cfg
yanan-fu: @vivianQizhu  ok, for usage, `yes` `on` `True` all can use:
```
        for opt in ["server", "telnet", "nowait",
                    "ipv4", "ipv6", "nodelay"]:
            if params.get(opt) in ["yes", "on", True]:
                params[opt] = "NO_EQUAL_STRING"
            elif opt in params:
                del params[opt]
```

virttest/qemu_vm.py
yanan-fu: As we discussed,  user can only connect to one console (the first serial) at the same time, use common `_create_serial_console`. So remove ` self.virtio_console = None` here is ok.

shared/cfg/base.cfg
vivianQizhu: @yanan-fu I want to keep it simple, so only list one usage here, to not disturb users.

yanan-fu: It is ok for me ~

vivianQizhu:
@yanan-fu Update, removed `self.virtio_console`, please help review again, thanks.

yanan-fu:
Code LGTM, and  test pass,  can get expect qemu command line.

Test with 
```
    serials += " vs1 vs2 vs3 vc1 vc2 vc3"
    serial_type_vs1 = virtserialport
    serial_type_vs2 = virtserialport
    serial_type_vs3 = virtserialport
    serial_type_vc1 = virtconsole
    serial_type_vc2 = virtconsole
    serial_type_vc3 = virtconsole
    chardev_backend_vs1 = unix_socket
    chardev_backend_vs2 = udp
    chardev_backend_vs3 = spicevmc
    chardev_backend_vc1 = unix_socket
    chardev_backend_vc2 = udp
    chardev_backend_vc3 = spicevmc
    chardev_name_vc3 = vdagent
    chardev_name_vs3 = vdagent
    virtio_port_name_prefix_vc3 = "com.redhat.spice."
    virtio_port_name_prefix_vs3 = "com.redhat.spice."
```

qemu command line:
```
    -chardev socket,server,nowait,path=/var/tmp/avocado_o8q7_l_f/serial-serial0-20190626-004342-PWQHBbub,id=chardev_serial0 \
    -device isa-serial,id=serial0,chardev=chardev_serial0 \

    -chardev socket,server,nowait,path=/var/tmp/avocado_o8q7_l_f/serial-vs1-20190626-004342-PWQHBbub,id=chardev_vs1 \
    -device virtio-serial-pci,id=virtio_serial_pci0,bus=pci.0,addr=0x3 \
    -device virtserialport,id=vs1,name=vs1,chardev=chardev_vs1,bus=virtio_serial_pci0.0,nr=1 \

    -chardev udp,host=127.0.0.1,port=5599,id=chardev_vs2 \
    -device virtserialport,id=vs2,name=vs2,chardev=chardev_vs2,bus=virtio_serial_pci0.0,nr=2 \

    -chardev spicevmc,name=vdagent,id=chardev_vs3 \
    -device virtserialport,id=vs3,name=com.redhat.spice.vs3,chardev=chardev_vs3,bus=virtio_serial_pci0.0,nr=3 \

    -chardev socket,server,nowait,path=/var/tmp/avocado_o8q7_l_f/serial-vc1-20190626-004342-PWQHBbub,id=chardev_vc1 \
    -device virtio-serial-pci,id=virtio_serial_pci1,bus=pci.0,addr=0x4 \
    -device virtconsole,id=vc1,name=vc1,chardev=chardev_vc1,bus=virtio_serial_pci1.0,nr=1 \

    -chardev udp,host=127.0.0.1,port=5398,id=chardev_vc2 \
    -device virtio-serial-pci,id=virtio_serial_pci2,bus=pci.0,addr=0x5 \
    -device virtconsole,id=vc2,name=vc2,chardev=chardev_vc2,bus=virtio_serial_pci2.0,nr=1 \

    -chardev spicevmc,name=vdagent,id=chardev_vc3 \
    -device virtio-serial-pci,id=virtio_serial_pci3,bus=pci.0,addr=0x6 \
    -device virtconsole,id=vc3,name=com.redhat.spice.vc3,chardev=chardev_vc3,bus=virtio_serial_pci3.0,nr=1  \
```


virttest/qemu_vm.py
```diff
@@ -2442,43 +2308,22 @@ def _create_serial_console(self):
...
-        except IndexError:
+        if self.serial_session_device is None:
             logging.warning("No serial ports defined!")
             return
-        log_name = "serial-%s-%s.log" % (tmp_serial, self.name)
+        log_name = "serial-%s-%s.log" % (
+            self.serial_session_device, self.name)
         self.serial_console_log = os.path.join(utils_misc.get_log_file_dir(),
                                                log_name)
+        file = self.get_serial_console_filename(self.serial_session_device)
```
xutian: bad idea to use key words as var name.

virttest/qemu_vm.py
```diff
@@ -1632,83 +1534,47 @@ def sort_key(dev):
...
+                serial_params['chardev_host'] = host
+                serial_params['chardev_port'] = free_ports[index]
+            serial_devices = devices.serials_define_by_params(
+                serial, serial_params, serial_filename)
+
+            devices.insert(serial_devices)
+
+            # Create virtio_ports (virtio_serialports and virtio_consoles)
+            serial_type = serial_params['serial_type']
+            if serial_type.startswith('virt'):
```
xutian: For making code more readable, please use "virtio" instead "virt"

virttest/qemu_vm.py
```diff
@@ -1632,83 +1534,47 @@ def sort_key(dev):
...
         # Add serial console redirection
-        for serial in params.objects("serials"):
+        serials = params.objects('serials')
+        if serials:
+            self.serial_session_device = serials[0]
+            host = params.get('chardev_host', '127.0.0.1')
+            free_ports = utils_misc.find_free_ports(
+                5000, 6000, len(serials), host)
+            reg_count = 0
+        self.virtio_ports = []
```
xutian: let's move this define in  function  '__init__'

virttest/qemu_vm.py
yanan-fu: @xutian  serial_type is `virtserialport` and `virtconsole`, so we need use `virt` here :)

vivianQizhu: @xutian They are `virtserialport` and `virtconsole`, do no start with `virtio`.

virttest/qemu_vm.py
vivianQizhu: @xutian Yes, updated, thanks.

virttest/qemu_vm.py
vivianQizhu: @xutian It actually has one in `__init__`, I removed this redundant one, thanks.

yanan-fu: @xutian  oh~   `__inint__`  already have, we can remove this one.

virttest/qemu_vm.py
xutian: OK, please fix in your comments in your code

vivianQizhu:
@xutian Updated the comments, thanks.

virttest/qemu_vm.py
yanan-fu: @vivianQizhu   correct my comment above,   qemu_vm use `get_virtio_port_filenames`  not `get_virtio_port_filename`.   only one `get_virtio_port_filename` in virt_vm.  It is not correct now.

vivianQizhu: @yanan-fu I see the two impact ones: https://github.com/autotest/tp-qemu/blob/master/qemu/tests/qemu_guest_agent.py#L199 and https://github.com/autotest/tp-qemu/blob/master/qemu/tests/qemu_guest_agent_hotplug.py#L32
What do you think if I replace them with `get_serial_console_filename`? Cause now we actually totally abandon the usage of `get_virtio_port_filename` in qemu_vm context.

vivianQizhu: @yanan-fu And for qemu_vm.py:
```
index 9fc849c9..b0c51f62 100644
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -3018,7 +3018,6 @@ class VM(virt_vm.BaseVM):
         # Generate the tmp file which should be deleted.
         file_list = [self.get_testlog_filename()]
         file_list += qemu_monitor.get_monitor_filenames(self)
-        file_list += self.get_virtio_port_filenames()
         file_list += self.get_serial_console_filenames()
         file_list += list(self.logs.values())
```

yanan-fu: @vivianQizhu  it is ok for me to use `get_serial_console_filename` in case level. as when generate qemu command line, we use same method to generate filename.   
And  remove `get_virtio_port_filename` and `get_virtio_file_names` directly ?   i have checked tp-libvirt, no use.

vivianQizhu: @yanan-fu Even it is not used by tp-libvirt now, I don't really want to change the base class, I only care how qemu_vm works.

yanan-fu: At least,  the one in qemu_vm need be deleted, `get_virtio_port_filenames`. 

For virt_vm, i prefer to delete the two function.  As my understanding, only qemu care it but do not need it.  

@xutian  do you have any idea about this issue ?

vivianQizhu: @yanan-fu Okay, all deleted, hope has no side effect.

yanan-fu:
gating test pass.

Ack

vivianQizhu:
Thanks all.

PR #1645, opened 2018-07-03
defaults: Check for arch for default asset

dzhengfy:

### DIFF
--- a/virttest/defaults.py
+++ b/virttest/defaults.py
@@ -16,7 +16,9 @@ def get_default_guest_os_info():
     Gets the default asset and variant information
     TODO: Check for the ARCH and choose corresponding default asset
     """
-    return {'asset': 'jeos-27-x86_64', 'variant': 'JeOS.27'}
+    default_asset = 'jeos-27-'
+    default_asset = '%s%s' % (default_asset, ARCH)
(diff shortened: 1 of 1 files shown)


virttest/defaults.py
```diff
@@ -16,7 +16,9 @@ def get_default_guest_os_info():
     Gets the default asset and variant information
     TODO: Check for the ARCH and choose corresponding default asset
     """
-    return {'asset': 'jeos-27-x86_64', 'variant': 'JeOS.27'}
+    default_asset = 'jeos-27-'
+    default_asset = '%s%s' % (default_asset, ARCH)
```
chunfuwen: Have you have opportunity to validate the change?

dzhengfy: @chunfuwen ,
For ppc64le, I did validate the change on Power8 host, the jeos-27-ppc64le is returned.
For ppc64, I did not validate the change as I did not find a suitable ppc64 host.
For x86_64, I think it is obvious same as before.

get_default_guest_os_info() is only referred in this defaults.py modual in avocado-vt repo, so it is low risk, I think.
Do you have other concern?

chunfuwen: Sound good, primarily I assume ARCH will return upper letter value.

PR #1647, opened 2018-07-03
virsh: Update command func

hahakiki2010:
Allow pass "allow_output_check" to process.run as needs of
redirect stdout and stderr into one


balamuruhans:
Thanks for the patch @hahakiki2010.

PR #1649, opened 2018-07-03
Add support for host stress events

sathnaga:
Add support for host stress events, which would help
us to run multiple events in host in parallel to events for
guest and stress.

### DIFF
--- a/virttest/utils_stress.py
+++ b/virttest/utils_stress.py
@@ -21,12 +21,14 @@ class VMStressEvents():
         """
         self.host_cpu_list = cpu.cpu_online_list()
         self.iterations = int(params.get("stress_itrs", 1))
+        self.host_iterations = int(params.get("host_event_itrs", 10))
         self.event_sleep_time = int(params.get("event_sleep_time", 10))
         self.current_vcpu = params.get("smp", 32)
         self.max_vcpu = params.get("vcpu_maxcpus", 32)
         self.ignore_status = params.get("ignore_status", "no") == "yes"
         self.vms = env.get_all_vms()
         self.events = params.get("stress_events", "reboot").split(',')
+        self.host_events = params.get("host_stress_events", "").split(',')
         self.threads = []
         self.iface_num = params.get("iface_num", '1')
         self.iface_type = params.get("iface_type", "network")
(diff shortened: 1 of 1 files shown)


sathnaga:
Depends on: https://github.com/avocado-framework/avocado/pull/2697

virttest/utils_stress.py
```diff
@@ -158,3 +163,27 @@ def vm_stress_events(self, event, vm):
...
                 raise NotImplementedError
+
+    def host_stress_event(self, event):
+        """
+        Host Stress events
+
+        :param event: event name
+        """
+        for itr in range(self.host_iterations):
+            if "cpufreqgovernor" in event:
```
balamuruhans: for better readability can we have the name like `cpu_freq_governor` for this and other events too.

virttest/utils_stress.py
```diff
@@ -158,3 +163,27 @@ def vm_stress_events(self, event, vm):
...
+                cpu.set_cpufreq_governor()
+                logging.debug("Current governor: %s", cpu.get_cpufreq_governor())
+                time.sleep(self.event_sleep_time)
+            elif "cpuidlestate" in event:
+                idlestate = cpu.get_cpuidle_state()
+                cpu.set_cpuidle_state()
+                time.sleep(self.event_sleep_time)
+                cpu.set_cpuidle_state(setstate=idlestate)
```
balamuruhans: how do we make sure the cpu idle state is set or not ? can we do get_cpuidle_state() after setting to validate ?

virttest/utils_stress.py
```diff
@@ -158,3 +163,27 @@ def vm_stress_events(self, event, vm):
...
+            elif "cpuoffline" in event:
+                processor = self.host_cpu_list[random.randint(0, cpu.online_cpus_count()-1)]
+                cpu.offline(processor)
+                time.sleep(self.event_sleep_time)
+                cpu.online(processor)
```
balamuruhans: same here

virttest/utils_stress.py
sathnaga: validation will make the stress events heavier, there is check in the `set_cpuidle_state` function itself and that would suffice.

virttest/utils_stress.py
sathnaga: same as above.

virttest/utils_stress.py
sathnaga: Sure, I can change it

PR #1650, opened 2018-07-03
remote: support recursive/non recursive scp to copy directory/file

balamuruhans:
currently APIs like copy_files_to(), copy_files_from() allow files to
be copied to, from remote <-> local. This patch add support to configure
scp to copy recursively/non recursively for directories/files to be copied 
with param.

### DIFF
--- a/virttest/remote.py
+++ b/virttest/remote.py
@@ -517,7 +517,8 @@ def remote_scp(command, password_list, log_filename=None, transfer_timeout=600,
 
 
 def scp_to_remote(host, port, username, password, local_path, remote_path,
-                  limit="", log_filename=None, timeout=600, interface=None):
+                  limit="", log_filename=None, timeout=600, interface=None,
+                  directory=True):
     """
     Copy files to a remote host (guest) through scp.
 
@@ -529,9 +530,10 @@ def scp_to_remote(host, port, username, password, local_path, remote_path,
     :param limit: Speed limit of file transfer.
     :param log_filename: If specified, log all output to this file
     :param timeout: The time duration (in seconds) to wait for the transfer
-            to complete.
-    :interface: The interface the neighbours attach to (only use when using
-                ipv6 linklocal address.)
+                    to complete.
+    :param interface: The interface the neighbours attach to (only use when using
+                      ipv6 linklocal address).
+    :param directory: True to copy recursively if the directory to scp
     :raise: Whatever remote_scp() raises
     """
     if (limit):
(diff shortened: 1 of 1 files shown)


virttest/remote.py
```diff
@@ -543,19 +545,23 @@ def scp_to_remote(host, port, username, password, local_path, remote_path,
...
-               "-o PreferredAuthentications=password -r %s "
-               "-P %s %s %s@\[%s\]:%s" %
-               (limit, port, quote_path(local_path), username, host,
-                pipes.quote(remote_path)))
+    command = "scp"
+    if directory:
+        command = "%s -r" % command
+    command += (" -v -o UserKnownHostsFile=/dev/null "
+                "-o StrictHostKeyChecking=no "
+                "-o PreferredAuthentications=password -r %s "
```
xutian: -l limit
             Limits the used bandwidth, specified in Kbit/s.
it looks need to instead  '-r' with '-l' in this line.

virttest/remote.py
```diff
@@ -580,34 +587,39 @@ def scp_from_remote(host, port, username, password, remote_path, local_path,
...
-               "-o PreferredAuthentications=password -r %s "
-               "-P %s %s@\[%s\]:%s %s" %
-               (limit, port, username, host, quote_path(remote_path),
-                pipes.quote(local_path)))
+    command = "scp"
+    if directory:
+        command = "%s -r" % command
+    command += (" -v -o UserKnownHostsFile=/dev/null "
+                "-o StrictHostKeyChecking=no "
+                "-o PreferredAuthentications=password -r %s "
```
xutian: same question here.

virttest/remote.py
```diff
@@ -624,12 +636,15 @@ def scp_between_remotes(src, dst, port, s_passwd, d_passwd, s_name, d_name,
...
-               "-o PreferredAuthentications=password -r %s -P %s"
-               " %s@\[%s\]:%s %s@\[%s\]:%s" %
-               (limit, port, s_name, src, quote_path(s_path), d_name, dst,
-                pipes.quote(d_path)))
+    command = "scp"
+    if directory:
+        command = "%s -r" % command
+    command += (" -v -o UserKnownHostsFile=/dev/null "
+                "-o StrictHostKeyChecking=no "
+                "-o PreferredAuthentications=password -r %s -P %s"
```
xutian: same question here.

virttest/remote.py
balamuruhans: @xutian Thank you for review, I will fix it.

balamuruhans:
@xutian I have re-factored it to make copying file/directory configurable, the `limit` is handled in right way earlier so I have handled `-r`. please help to review.

virttest/remote.py
```diff
@@ -543,19 +545,23 @@ def scp_to_remote(host, port, username, password, local_path, remote_path,
...
-               "-o PreferredAuthentications=password -r %s "
-               "-P %s %s %s@\[%s\]:%s" %
-               (limit, port, quote_path(local_path), username, host,
-                pipes.quote(remote_path)))
+    command = "scp"
+    if directory:
+        command = "%s -r" % command
+    command += (" -v -o UserKnownHostsFile=/dev/null "
+                "-o StrictHostKeyChecking=no "
+                "-o PreferredAuthentications=password %s "
```
xutian: it looks "limit" argument should be handled like
```
if limit:
    commands += "  -l %s" % limit
```

balamuruhans: @xutian, it is already handled in this way,

https://github.com/avocado-framework/avocado-vt/blob/edee6694742b60e39a8c7a667185c38fdb039ad7/virttest/remote.py#L519-L538

balamuruhans:
Thank you @xutian @sathnaga.

rdkdd:
@balamuruhans, @xutian  this PR broke my tests since `directory=False` by default in `copy_files_to/from` functions. 

I would like to modify this to `directory=True` by default since this breaks functionality of directory copying in `copy_files_to()` from `virt_vm.py`. What is your opinion?

balamuruhans:
@rdkdd  Thank you for reporting it. I have missed it to have recursive copy as default.
Fixed it as per your suggestion in https://github.com/avocado-framework/avocado-vt/pull/1666.

PR #1655, opened 2018-07-04
vm_xml:Add option 'debug' to method 'sync'

chloerh:
Lots of cases failed on 'sync' because the xml to be defined has some
kinds of error. We could not see the exact error message of the failed
'virsh define'. Add 'debug' as an option to enable us to recognize
problem without having to rerun the case or debug in other ways.

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -607,7 +607,7 @@ class VMXML(VMXMLBase):
         """
         return librarian.get(type_name)
 
-    def undefine(self, options=None, virsh_instance=base.virsh):
+    def undefine(self, options=None, virsh_instance=base.virsh, debug=False):
         """Undefine this VM with libvirt retaining XML in instance"""
         # add --nvram in undefine for aarch64
         if "aarch" in platform.machine():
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -607,7 +607,7 @@ def get_device_class(type_name):
         """
         return librarian.get(type_name)
 
-    def undefine(self, options=None, virsh_instance=base.virsh):
+    def undefine(self, options=None, virsh_instance=base.virsh, debug=False):
```
balamuruhans: we can use `**dargs` instead to have generic support.

chloerh: But this way is much simpler.

balamuruhans: but **dargs will make things generic and there wont be any future changes required, it will start supporting all the virsh parameters.

chloerh: But the virsh parameters are supported by 'options', and 'debug=True' is not a virsh parameter.

balamuruhans: I meant `readonly`, `uri`, `quiet`, `virsh_exec` (virsh binary) are some of the virsh parameters, along with that other parameters like debug, timeout, unprivileged_user will also get support with the change.

chloerh: Oh, I see. But this is not a virsh command, this is simply a support function to manipulate xml of vm: update xml of vm by executing undefine , then define.

PR #1657, opened 2018-07-05
libvirt:Update function create_hostdev_xml

chloerh:
- Add option to determine to return xml object or xml file path
- Replace outdated function invoking

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1684,11 +1684,14 @@ def check_iface(iface_name, checkpoint, extra="", **dargs):
     return check_pass
 
 
-def create_hostdev_xml(pci_id, boot_order=0):
+def create_hostdev_xml(pci_id, boot_order=0, xmlfile=True):
     """
     Create a hostdev configuration file.
 
     :param pci_id: such as "0000:03:04.0"
+    :param boot_order: boot order of hostdev device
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1708,8 +1708,10 @@ def create_hostdev_xml(pci_id, boot_order=0):
         hostdev_xml.boot_order = boot_order
     attrs = {'domain': device_domain, 'slot': device_slot,
              'bus': device_bus, 'function': device_function}
-    hostdev_xml.source_address = hostdev_xml.new_source_address(**attrs)
+    hostdev_xml.source = hostdev_xml.new_source(**attrs)
     logging.debug("Hostdev XML:\n%s", str(hostdev_xml))
+    if not xmlfile:
+        return hostdev_xml
```
balamuruhans: This API is meant to return the Hostdev xml, why do we return the object ?

chloerh: For our test will need a hostdev object, and the pci_id parsing in this function is convinient.

balamuruhans: @chloerh sure, request you to add doc string that would explain the new param, as it helps.

sathnaga: @chloerh even the name of method says only about xml, it is good to have a another method to have hostdev object,
if not otherwise a doc string helps...

PR #1661, opened 2018-07-09
libvirt_vm: Update migrate func to pass 'readonly' argument

hahakiki2010:

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2003,21 +2003,21 @@ class VM(virt_vm.BaseVM):
         return utils_misc.get_uptime(session)
 
     def migrate(self, dest_uri="", option="--live --timeout 60", extra="",
-                ignore_status=False, debug=False, virsh_opt=""):
+                **dargs):
         """
         Migrate a VM to a remote host.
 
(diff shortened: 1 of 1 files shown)


virttest/libvirt_vm.py
```diff
@@ -2003,7 +2003,7 @@ def uptime(self, connect_uri=None):
         return utils_misc.get_uptime(session)
 
     def migrate(self, dest_uri="", option="--live --timeout 60", extra="",
-                ignore_status=False, debug=False, virsh_opt=""):
+                ignore_status=False, debug=False, readonly=False, virsh_opt=""):
```
balamuruhans: better to use `**dargs` instead,
```python
def migrate(self, dest_uri="", option="--live --timeout 60", extra="",
            **dargs):
    ...
    result = virsh.migrate(self.name, dest_uri, option,
                           extra, uri=self.connect_uri,
                           **dargs)
    ...

hahakiki2010: yes, this one is better. fixed, thanks

balamuruhans:
Thanks @hahakiki2010, let's merge.

PR #1662, opened 2018-07-09
utils_test: extract XZ compressed stress tool

balamuruhans:
If it is git/wget file based clone/download proceed, the tool can be
compressed (gzip, XZ etc.,) or might not compressed in file based
download, so handle it appropriately using archive in avocado.utils.
Also allow to use already downloaded tool as fallback or raise if
configuration is not appropriate

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -37,6 +37,7 @@ from avocado.utils import aurl
 from avocado.utils import download
 from avocado.utils import crypto
 from avocado.utils import path
+from avocado.utils import archive
 
 from six.moves import xrange
 
@@ -1859,7 +1860,8 @@ class Stress(object):
 
     def __init__(self, stress_type, params, download_url="", make_cmds="",
                  stress_cmds="", stress_args="", work_path="",
-                 uninstall_cmds="", download_type="tarball"):
+                 uninstall_cmds="", download_type="file",
+                 downloaded_file_path=None):
         """
         Set parameters for stress type, for the arguments have default value "",
         they can be either passed here, or defined in vm.params
@@ -1874,7 +1876,9 @@ class Stress(object):
         :param stress_cmds: the command to launch stress,
         use stress_type instead if not defined
         :param stress_args: the arguments of the stress tool
-        :param download_type: currently support "git" or "tarball"
+        :param download_type: currently support "git" or "file" download
+        :param downloaded_file_path: Already downloaded / available stress
+                                     tool path.
         """
 
         self.vm = None
@@ -1902,6 +1906,9 @@ class Stress(object):
         self.uninstall_cmds = self.uninstall_cmds or './configure && make uninstall'
         self.work_path = self.params.get('%s_work_path' % stress_type,
                                          work_path)
+        self.downloaded_file_path = self.params.get("%s_downloaded_file_path" %
+                                                    self.stress_type,
+                                                    downloaded_file_path)
         check_cmd = self.stress_cmds.split(" ")[0]
         self.check_cmd = "pidof -s %s" % check_cmd
         self.stop_cmd = "pkill -9 %s" % check_cmd
@@ -1969,19 +1976,20 @@ class Stress(object):
         destination = os.path.join(destination, self.base_name)
         git.get_repo(url, destination_dir=destination)
 
-    def _tarball_download(self, url, destination):
+    def _file_download(self, url, destination):
         """
-        tarball download and uncompress
-        :param url: dowload url
+        download and uncompress tar / XZ file
+        :param url: download url
         :param destination: download destination
         """
-        import tarfile
-        tarball_path = os.path.join(destination, self.base_name)
-        download.get_file(url, tarball_path)
-        tarball = tarfile.open(tarball_path)
-        tarball.extractall(destination)
-        self.base_name = tarball.members[0].name.split("/")[0]
-        tarball.close()
+        file_path = os.path.join(destination, self.base_name)
+        download.get_file(url, file_path)
+        file_type = process.getoutput("file %s" % file_path)
+        if "directory" not in file_type:
+            self.base_name = archive.uncompress(file_path, destination)
+            # First member of the archive can be the extracted folder or
+            # a file in extracted folder
+            self.base_name = self.base_name.split(os.sep)[0]
 
     def download_stress(self):
         """
@@ -1990,8 +1998,21 @@ class Stress(object):
         url = self.download_url
         tmp_path = data_dir.get_tmp_dir()
         logging.info('Download stress tool from %s', url)
-        download_method = getattr(self, "_%s_download" % self.download_type)
-        download_method(url, tmp_path)
+        # If it is git/wget based download proceed, else fall back if user already
+        # have downloaded tool path, else raise
+        try:
+            download_method = getattr(self, "_%s_download" % self.download_type)
+            download_method(url, tmp_path)
+        except AttributeError:
+            if not self.downloaded_file_path:
+                raise exceptions.TestError("Tool is not downloaded or download"
+                                           " link for Tool is not provided")
+            file_type = process.getoutput("file %s" % self.downloaded_file_path)
+            if "directory" not in file_type:
+                self.base_name = archive.uncompress(self.downloaded_file_path,
+                                                    tmp_path)
+            else:
+                self.base_name = self.downloaded_file_path
         source = os.path.join(tmp_path, self.base_name)
         if self.remote_host:
             logging.info('Copy stress tool to remote host')
@@ -2015,8 +2036,8 @@ class Stress(object):
         install_path = os.path.join(self.dst_path, self.base_name,
                                     self.work_path)
         self.make_cmds = "cd %s;%s" % (install_path, self.make_cmds)
-        logging.info('make and install the %s', self.stress_type)
-        status, output = self.cmd_status_output(self.make_cmds)
+        logging.info('make and install %s', self.stress_type)
+        status, output = self.cmd_status_output(self.make_cmds, timeout=600)
         if status != 0:
             raise exceptions.TestError(
                 "Installation failed with output:\n %s" % output)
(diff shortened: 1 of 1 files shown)


virttest/utils_test/__init__.py
```diff
@@ -1969,29 +1969,65 @@ def _git_download(self, url, destination):
         destination = os.path.join(destination, self.base_name)
         git.get_repo(url, destination_dir=destination)
 
-    def _tarball_download(self, url, destination):
+    def _tar_extract(self, tar_file, destination):
```
sathnaga: @balamuruhans it would be good if we reuse, this from avocado `avocado/utils/archive.py`, looks like this supports most of archive types.

balamuruhans: @sathnaga sure, I will make the change. Thank you for review.

balamuruhans:
@aliang123 @vivianQizhu @bssrikanth @sathnaga request for your review.

virttest/utils_test/__init__.py
```diff
@@ -1990,8 +1989,12 @@ def download_stress(self):
         url = self.download_url
         tmp_path = data_dir.get_tmp_dir()
         logging.info('Download stress tool from %s', url)
-        download_method = getattr(self, "_%s_download" % self.download_type)
-        download_method(url, tmp_path)
+        # If it is git based download proceed, else fall back to _download_file()
+        try:
+            download_method = getattr(self, "_%s_download" % self.download_type)
+            download_method(url, tmp_path)
+        except AttributeError:
```
vivianQizhu: Thanks for branching it. Although this would work, but it looks a bit strange to me to treat tar/xz download as an exception, for me that should be a download method parallel to `_git_download`. I would prefer keep using `download_method` without exception and keep `_tarball_download()`, that only requires xz installation use `tarball` as it's `download_type`. Or replacing `tarball` with something like `compress`, `_compress_download()` and `self.download_type = 'compress'`.

vivianQizhu: Or `_file_download` since it might not be compressed?

virttest/utils_test/__init__.py
```diff
@@ -1859,7 +1860,8 @@ class Stress(object):
 
     def __init__(self, stress_type, params, download_url="", make_cmds="",
                  stress_cmds="", stress_args="", work_path="",
-                 uninstall_cmds="", download_type="tarball"):
+                 uninstall_cmds="", download_type="tarball",
```
vivianQizhu: `download_type` should be updated to `file`. Otherwise looks good to me.

virttest/utils_test/__init__.py
```diff
@@ -1990,8 +1995,21 @@ def download_stress(self):
...
+        # If it is git/wget based download proceed, else fall back if user already
+        # have downloaded tool path, else raise
+        try:
+            download_method = getattr(self, "_%s_download" % self.download_type)
+            download_method(url, tmp_path)
+        except AttributeError:
+            if not self.downloaded_file_path:
+                raise exceptions.TestError("Tool is not downloaded or download"
+                                           " link for Tool is not provided")
+            file_type = process.getoutput("file %s" % self.downloaded_file_path)
```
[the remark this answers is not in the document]
balamuruhans: @aliang123 Hi, Thank you for reviewing,

1. my thinking is, when both `downloaded_file_path` and `download_url` are given, download from url should have high priority and override (The user might give both because already downloaded tool might be older version and wants to try downloading latest). If `download_url` is not given, it falls back and check if user given already `downloaded_file_path` and similar in the other way.

2. `downloaded_file_path` is new option to user, if user have developed a new tool which is not hosted or the tool is already downloaded, using this option user can make the framework to deploy it. If the `download_url` is configured as `None` and `downloaded_file_path` is configured along with other stress params specific to the new/downloaded tool the framework adopts it.

Please let me know your thoughts.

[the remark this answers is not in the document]
aliang123: So @balamuruhans , another question: if the same stress version in url and downloaded_file_path, we still need to download from url?

And if we use downloaded_file_path, but there is no downloaded stress tools, what can we do? Don't we need to check the tools' existance in  downloaded_file_path first, then we can uncompress it?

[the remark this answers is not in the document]
balamuruhans: @aliang123, I have checked with help of your suggestion and API looks to be stable in those scenarios, please feel free to correct me.

> And if we use `downloaded_file_path`, but there is no downloaded stress tools, what can we do? 

If we check inside `downloaded_file_path` for stress tools we cannot be sure that the existing files are proper or valid or corrupted, so if there is empty folder and no downloaded stress tool we raise `TestError` as make cmds will fail below if tool is not proper or not available,

https://github.com/avocado-framework/avocado-vt/blob/fa0abe7396c7af147c866ee90af1e7b62a129419/virttest/utils_test/__init__.py#L2015-L2023

> Don't we need to check the tools' existance in downloaded_file_path first, then we can uncompress it?

Okay, the framework will check and handles in better way, if the `downloaded_file_path` is not archive or empty Archive exception is raised in,

https://github.com/avocado-framework/avocado/blob/7f4c6bcf47063567479ee7d135d2da95b32717f3/avocado/utils/archive.py#L119-L123

virttest/utils_test/__init__.py
```diff
@@ -1859,7 +1860,8 @@ class Stress(object):
 
     def __init__(self, stress_type, params, download_url="", make_cmds="",
                  stress_cmds="", stress_args="", work_path="",
-                 uninstall_cmds="", download_type="tarball"):
+                 uninstall_cmds="", download_type="file",
+                 downloaded_file_path=None):
```
aliang123: @balamuruhans , the last question:
   we define downloaded_file_path=None in parent class "Stress", and we don't set it in both class VMStress and HostStress, like
  class VMStress(Stress):
    """
    Run Stress tool on VMs, such as stress, unixbench, iozone and etc.
    """

    def __init__(self, vm, stress_type, params, download_url="", make_cmds="",
                 stress_cmds="", stress_args="", work_path="",
                 uninstall_cmds="", download_type="file"):
So is that a design or a missing?

virttest/utils_test/__init__.py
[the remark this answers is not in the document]
aliang123: @balamuruhans , i see, thanks for your patient explanation.

virttest/utils_test/__init__.py
balamuruhans: @aliang123 indeed, its a miss. Thank you for figuring it out, I will fix it.

virttest/utils_test/__init__.py
```diff
@@ -1969,19 +1976,17 @@ def _git_download(self, url, destination):
...
-        download.get_file(url, tarball_path)
-        tarball = tarfile.open(tarball_path)
-        tarball.extractall(destination)
-        self.base_name = tarball.members[0].name.split("/")[0]
-        tarball.close()
+        file_path = os.path.join(destination, self.base_name)
+        download.get_file(url, file_path)
+        file_type = process.getoutput("file %s" % file_path)
+        if "directory" not in file_type:
+            self.base_name = archive.uncompress(file_path, destination)
```
aliang123: @balamuruhans , archive.uncompress returns "None" here, so self.base_name=None

virttest/utils_test/__init__.py
```diff
@@ -1990,8 +1995,21 @@ def download_stress(self):
...
+        try:
+            download_method = getattr(self, "_%s_download" % self.download_type)
+            download_method(url, tmp_path)
+        except AttributeError:
+            if not self.downloaded_file_path:
+                raise exceptions.TestError("Tool is not downloaded or download"
+                                           " link for Tool is not provided")
+            file_type = process.getoutput("file %s" % self.downloaded_file_path)
+            if "directory" not in file_type:
+                self.base_name = archive.uncompress(self.downloaded_file_path,
```
aliang123: The same issue as above.

virttest/utils_test/__init__.py
balamuruhans: @aliang123 , I think you have missed to include Avocado patch - https://github.com/avocado-framework/avocado/pull/2705, that returns the extracted path.

virttest/utils_test/__init__.py
balamuruhans: same here.

virttest/utils_test/__init__.py
aliang123: @balamuruhans, test with patch avocado-framework/avocado#2705, self.base_name still get a wrong value when i run iozone test...
current:
    self.base_name = iozone3_458/src/current/Changes.txt
correct:
    self.base_name= iozone3_458

So, with patch #2750, we can't get a correct self.base_name

virttest/utils_test/__init__.py
aliang123: same as above.

virttest/utils_test/__init__.py
balamuruhans: @aliang123 yes, correct I have tested it for stress-ng but not for iozone and `tarfile` module return the `getnames()` differently for iozone and stress-ng.
I have fixed in PR https://github.com/avocado-framework/avocado/pull/2743.
Thank you for reporting it and making code better.

virttest/utils_test/__init__.py
balamuruhans: sure.

virttest/utils_test/__init__.py
aliang123: @balamuruhans, yes, with patch #2743, download-iozone can work normally.
So, after #2743 merged, i'll give my ack on this patch.

balamuruhans: yep :+1: I would wait for your ACK

balamuruhans: @aliang123, from avocado `archive.uncompress()` will return the first member of the extracted tar/zip file, the users have to handle it accordingly. so please see L1992 for it.

balamuruhans:
Thanks @vivianQizhu @aliang123 @sathnaga for reviewing and making it better.

balamuruhans:
I have tested the patch and it is working fine, let us merge as it we have PRs https://github.com/avocado-framework/avocado-vt/pull/1665 and https://github.com/avocado-framework/avocado-vt/pull/1677 which needs this to be in.

PR #1664, opened 2018-07-10
QPCIEBus: Update device's parent_bus when insert root port for it

vivianQizhu:
No real issue currently, just optimize it to avoid potential risk.

### DIFF
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -1882,7 +1882,9 @@ class QPCIEBus(QPCIBus):
         root_port = self.add_root_port(_addr, self.__root_port_type)
         if root_port is not None:
             added_devices.append(root_port)
-            device['bus'] = root_port.child_bus[0].busid
+            root_port_bus_id = root_port.child_bus[0].busid
+            device['bus'] = root_port_bus_id
+            self.reset_device_parent_bus(device, root_port_bus_id)
         return added_devices
 
     def _set_device_props(self, device, addr):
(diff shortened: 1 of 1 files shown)


vivianQizhu:
Not affect command line generation:
```
15:12:53 INFO | PASS 1-Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.5.x86_64.myprovider.boot.q35
```

vivianQizhu:
@ldoktor @jingzhao84 Would you please help review it? Many thanks.

vivianQizhu:
@ldoktor Thanks for helping review.
We hit a bug for case balloon_hotplug.with_migrate.migrate_after_hotplug.q35, before migration, the device is plugged into the empty `pcie_root_port_0`, while after migration, the device will go to `QPCIEBus.insert(device)` instead of `prepare_hotplug(device)`, so it get another root port like `pcie.0-root-port-8`. 
Although this patch can not resolve this issue, since migration clone vm from `params`. It just makes me though like it would be better to keep insert() act same as prepare_hotplug(), always change the `parent_bus` together with `params.bus`, to avoid the device been assigned to another `pcie-root-port`.
Anyway we have no real issue currently, so I am closing this pull request, and we might figure out better solution when we have certain issue really occurred.

vivianQizhu:
@ldoktor That's right, it should suffice, if migration clone from devices, the hotplugged device will have pcie-root-port as its parent_bus and won't be inserted into pci.0. Currently we have to edit params before migration: https://github.com/autotest/tp-qemu/pull/1437#issuecomment-403731363

PR #1669, opened 2018-07-16
Add readonly check func

kylazhang:

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -3550,3 +3550,18 @@ def check_qemu_cmd_line(content, err_ignore=False):
             raise exceptions.TestFail("Expected '%s' was not found in "
                                       "qemu command line" % content)
     return True
+
+
+def readonly_failure_check(cmd_result, msg="forbidden.* read only"):
+    """
+    Check if the cmd output show the readonly failure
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -3550,3 +3550,19 @@ def check_qemu_cmd_line(content, err_ignore=False):
             raise exceptions.TestFail("Expected '%s' was not found in "
                                       "qemu command line" % content)
     return True
+
+
+def readonly_failure_check(cmd_result):
```
balamuruhans: can we have `msg` as default arg,
```python
def readonly_failure_check(cmd_result, msg="forbidden.* read only"):
```

balamuruhans:
I think we can reuse,

https://github.com/avocado-framework/avocado-vt/blob/32f49b1a488a933880c86dd6a9214b77098e6751/virttest/utils_test/libvirt.py#L1514-L1526

virttest/utils_test/libvirt.py
kylazhang: Ok, thanks for remind

kylazhang:
@balamuruhans yes, thanks, I will close it. Thanks

PR #1672, opened 2018-07-18
unattended_install: Get autounattend file name for ovmf correctly

yanan-fu:
The full name of "unattended_file" may contain not only one dots
use strict match to get correct name for ovmf.
id: 1602780


yanan-fu:
According to maintainer's advise,  define unattended_file for ovmf in guest-os directly instead of generate it dynamically. 
Close this PR, and send https://github.com/avocado-framework/avocado-vt/pull/1673 to fix this issue.

PR #1677, opened 2018-07-25
utils_test: support different stress type on different VM

balamuruhans:
currently the stress type (stress or stress-ng or etc.,) is common across
all the VMs in multi vm environment. But to have different stress type
for different VM we do not have support. This patch enables it and the params
have to be suffixed with _VMNAME.

Example if we have VM1 name as virt-tests-vm1 and VM2 name as virt-tests-vm2
then the stress params should be like,

stress_args_virt-tests-vm1 = "--vm 2 --io 2 --cpu 2"
stress_args_virt-tests-vm2 = "-io 4 --cpu 3"

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -2069,7 +2069,13 @@ class VMStress(Stress):
         :param stress_args: the arguments of the stress tool
         :param download_type: currently support "git" or "tarball"
         """
-        super(VMStress, self).__init__(stress_type, params, download_url=download_url,
+        # This enables VM specific stress params like stress_cmds_virt-tests-vm1, etc.,
+        # to run different stress type on different VM
+        self.stress_type = params.get("stress_type_%s" % vm.name, stress_type)
+        self.params = params.object_params(vm.name)
+
+        super(VMStress, self).__init__(self.stress_type, self.params,
+                                       download_url=download_url,
                                        make_cmds=make_cmds, stress_cmds=stress_cmds,
                                        stress_args=stress_args, work_path=work_path,
                                        uninstall_cmds=uninstall_cmds,


vivianQizhu:
ACK

bssrikanth:
@balamuruhans good one. Can you please post the run result

balamuruhans:
@sathnaga right!, I will fix the commit msg.

balamuruhans:
@sathnaga @bssrikanth @vivianQizhu Thanks for reviewing. sure, I will post the result.

virttest/utils_test/__init__.py
```diff
@@ -2069,7 +2069,13 @@ def __init__(self, vm, stress_type, params, download_url="", make_cmds="",
         :param stress_args: the arguments of the stress tool
         :param download_type: currently support "git" or "tarball"
         """
-        super(VMStress, self).__init__(stress_type, params, download_url=download_url,
+        # This enables VM specific stress params like stress_cmds_virt-tests-vm1, etc.,
+        # to run different stress type on different VM
+        self.stress_type = params.get("stress_type_%s" % vm.name, stress_type)
```
sathnaga: @balamuruhans `stress_type` need not be appended with vm.name always, if it is given in that way fine, otherwise also it should work, right now it mandates it

balamuruhans: @sathnaga  it doesn't mandates currently, it just checks whether user wants to configure separate `stress_type_VM` for any specific VM, else it falls back to `stress_type` argument.

sathnaga: @balamuruhans sure, I was carried away with `load_stress(stress_type,` ` stress_type in ['stress_in_vms', 'iozone_in_vms']:` implementation and thought there was an `stress_type` in first place, looks like not, got it now,

LGTM.

sathnaga:
@balamuruhans I will wait for your test logs to merge it.

PR #1680, opened 2018-07-25
Minor fix for build pool

bssrikanth:
Added option param for the virsh pool build call


virttest/libvirt_storage.py
```diff
@@ -281,12 +281,12 @@ def set_pool_autostart(self, name, extra=""):
         logging.info("Set pool '%s' autostart.", name)
         return True
 
-    def build_pool(self, name):
+    def build_pool(self, name, options=""):
```
balamuruhans: even `**dargs` can be included and sent to `virsh.pool_build` which make it more generic

PR #1687, opened 2018-07-30
Get disk alias function fix

kylazhang:


virttest/utils_test/libvirt.py
```diff
@@ -3564,7 +3564,7 @@ def get_disk_alias(vm, source_file=None):
         ori_vmxml = vm_xml.VMXML.new_from_dumpxml(vm.name)
         disks = ori_vmxml.devices.by_device_tag('disk')
         for disk in disks:
-            find_source = disk.xmltreefile.find('source')
+            find_source = not (disk.xmltreefile.find('source') == None)
```
chunfuwen: better not use == to compare None,use not (disk.xmltreefile.find('source')  is None) instead

PR #1691, opened 2018-08-02
Python3: guest_agent: Correct password type conversion issue

vivianQizhu:
base64.b64encode() needs a byte input, while self.cmd() needs args
content being string type, so encode first for b64encode() and
then decode for self.cmd()

id: 1610713

### DIFF
--- a/virttest/guest_agent.py
+++ b/virttest/guest_agent.py
@@ -542,7 +542,7 @@ class QemuAgent(Monitor):
             password = decode_to_text(process.system_output(openssl_cmd)).strip('\n')
 
         args = {"crypted": crypted, "username": username,
-                "password": base64.b64encode(password)}
+                "password": base64.b64encode(password.encode()).decode()}
         return self.cmd(cmd=cmd, args=args)
 
     @error_context.context_aware


vivianQizhu:
Both python2 and python3 passed the test:
```
(1/2) Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.5.x86_64.myprovider.qemu_guest_agent.virtio_serial.check_set_user_password.no_crypto: PASS (52.88 s)
```
```
 (1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.my-provider.qemu_guest_agent.virtio_serial.check_set_user_password.no_crypto: PASS (50.69 s)
```

vivianQizhu:
@hereischen Would you please help review it? Thanks.

virttest/guest_agent.py
```diff
@@ -539,10 +538,10 @@ def set_user_password(self, password, crypted=False, username="root"):
 
         if crypted:
             openssl_cmd = "openssl passwd -crypt %s" % password
-            password = decode_to_text(process.system_output(openssl_cmd)).strip('\n')
+            password = process.system_output(openssl_cmd).strip('\n')
```
hereischen: `password = process.system_output(openssl_cmd).strip('\n')` is bytes
`password.decode()`  is str

virttest/guest_agent.py
```diff
@@ -539,10 +538,10 @@ def set_user_password(self, password, crypted=False, username="root"):
...
 
         args = {"crypted": crypted, "username": username,
-                "password": base64.b64encode(password)}
+                "password": base64.b64encode(password.encode()).decode()}
```
hereischen: `password = process.system_output(openssl_cmd).strip('\n')` is bytes, you don't need to encode it again.

vivianQizhu: @hereischen I though it should be, but without this encoding, b64encode() would still fail with `a bytes-like object is required, not 'str'`. I don't really understand why, but only this way it can pass.

virttest/guest_agent.py
hereischen: That is because this line is not correct.
```python
[root@dhcp-8-192 ~]# python
Python 3.5.5 |Anaconda, Inc.| (default, May 13 2018, 21:12:35) 
[GCC 7.2.0] on linux
>>> password = b"redhat"
>>> password.strip("\n")
TypeError: a bytes-like object is required, not 'str'
>>> import base64
>>> a = base64.b64encode(password)
>>> a
b'cmVkaGF0'
>>> 
```

vivianQizhu: @hereischen No, I am sure the fail one is b64encode():
```
01:53:23 ERROR|   File "/usr/local/lib/python3.6/site-packages/avocado_plugins_vt-63.0-py3.6.egg/virttest/guest_agent.py", line 547, in set_user_password
01:53:23 ERROR|     "password": base64.b64encode(password).decode()}
01:53:23 ERROR|   File "/usr/lib64/python3.6/base64.py", line 58, in b64encode
01:53:23 ERROR|     encoded = binascii.b2a_base64(s, newline=False)
01:53:23 ERROR| TypeError: a bytes-like object is required, not 'str'
```
I guess the real reason is `crypted` is `False`, so the password actually not comes from `process.system_output()`, it is passed by the caller, so it is text.

vivianQizhu: @hereischen So `decode_to_text()` should not be blamed, I will bring it back.

hereischen:

PR #1694, opened 2018-08-03
Add sanity checks

ligurio:


virttest/asset.py
```diff
@@ -609,5 +617,6 @@ def download_asset(asset, interactive=True, restore_image=False):
     """
     asset_info = get_asset_info(asset)
 
-    download_file(asset_info=asset_info, interactive=interactive,
-                  force=restore_image)
+    if asset_info != {}:
```
lmr: Here I'd prefer to be a bit more idiomatic (`if asset info:`), simply comparing a value to an empty data structure is not the way you'd do it in python.

PR #1698, opened 2018-08-07
Add methods to get/set the ntp state on guest/host

sathnaga:
Add methods to get/set the ntp on guest/host

### DIFF
--- a/virttest/utils_time.py
+++ b/virttest/utils_time.py
@@ -212,3 +212,28 @@ def update_clksrc(vm, clksrc=None):
     if clksrc and clksrc != 'kvm-clock':
         boot_option_added = "clocksource=%s" % clksrc
         utils_test.update_boot_option(vm, args_added=boot_option_added)
+
+
+def is_ntp_enabled(session):
+    """
+    Get current NTP state for guest/host
+    """
+    cmd = 'timedatectl |grep "NTP enabled"'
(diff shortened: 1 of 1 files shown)


virttest/utils_time.py
```diff
@@ -212,3 +212,32 @@ def update_clksrc(vm, clksrc=None):
...
+def ntp_switch(session, off=True):
+    """
+    Turn off/on ntp for guest/host
+    """
+    cmd = "timedatectl set-ntp"
+    if off:
+        cmd += " 0"
+    else:
+        cmd += " 1"
+    execute(cmd, session=session)
```
balamuruhans: let's get the return of execute() and log when we raise TestError in L241 and L243

virttest/utils_time.py
```diff
@@ -212,3 +212,32 @@ def update_clksrc(vm, clksrc=None):
...
         utils_test.update_boot_option(vm, args_added=boot_option_added)
+
+
+def is_ntp_enabled(session):
+    """
+    Get current NTP state for guest/host
+    """
+    cmd = 'timedatectl |grep "NTP enabled"'
+    output = execute(cmd, session=session)
+    if 'yes' in output.split(":")[1].strip():
```
balamuruhans:
```python
return 'yes' in output.split(":")[1].strip()
```

PR #1711, opened 2018-08-14
Do not cleanup export domain for rhv_upload

xiaodwan:
When vm is deleted, the disk will also be removed from data domain.
It's not necessary to delete disk from export domain(data domain)
for rhv_upload.

This fixes below error logs in CI.

DEBUG| VM not exist
ERROR| Failed to get kvm-win2008r2-x86_64 from iscsi_export:list index out of range
INFO | Remove VM kvm-win2008r2-x86_64 from export storage
ERROR| Failed to remove VM:'NoneType' object has no attribute 'id'

Signed-off-by: xiaodwan <xiaodwan@redhat.com>


virttest/utils_v2v.py
```diff
@@ -209,6 +209,7 @@ def __init__(self, test, params, env):
         self.delete_vm = 'yes' == params.get('vm_cleanup', 'yes')
         self.virsh_session_id = params.get("virsh_session_id")
         self.windows_root = params.get("windows_root", "C:\WINDOWS")
+        self.output_method = params.get("output_method")
```
balamuruhans: let us declare the param in cfg with the comments explaining the usage.

xiaodwan: I have updated the cfg file for describing the usage of output_method. Please have a look, Thanks!

backends/v2v/cfg/convert_destination.cfg
```diff
@@ -18,6 +18,28 @@ variants:
...
+        # rhev:
+        #     - In V2V command, it is "-o rhv" or "-o rhev".
+        #       When output_method is "rhev", the images will be uploaded
+        #       to export domain at first. You still need to import the VM
+        #       manually. When the VM is deleted, the disk in export domain
+        #       will not be deleted automatically. You need to delete it
+        #       by yourself.
+        #
+        # You can refer to man page of virt-v2v for details.
+        output_method = rhev
```
balamuruhans: @xiaodwan, thanks for nice explanation.

PR #1712, opened 2018-08-14
Add support for memory hotplug virt-install options

sathnaga:
Add support for memory hotplug virt-install options

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -459,7 +459,8 @@ class VM(virt_vm.BaseVM):
                 logging.warning("Unknown virt type hvm_or_pv, using default.")
                 return ""
 
-        def add_mem(help_text, mem, maxmem=None, hugepage=False):
+        def add_mem(help_text, mem, maxmem=None, hugepage=False,
+                    hotplugmaxmem=None, hotplugmemslots=1):
             if has_option(help_text, "memory"):
                 cmd = " --memory=%s" % mem
                 if maxmem:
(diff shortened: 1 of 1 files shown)


virttest/libvirt_vm.py
```diff
@@ -459,7 +459,8 @@ def add_hvm_or_pv(help_text, hvm_or_pv):
                 logging.warning("Unknown virt type hvm_or_pv, using default.")
                 return ""
 
-        def add_mem(help_text, mem, maxmem=None, hugepage=False):
+        def add_mem(help_text, mem, maxmem=None, hugepage=False,
+                    hotplugmaxmem=None, hotplugmemslots=1):
```
balamuruhans: why do we have to set `hotplugmemslot` default as 1 ?, it can be `None`

virttest/libvirt_vm.py
```diff
@@ -474,6 +475,17 @@ def add_mem(help_text, mem, maxmem=None, hugepage=False):
...
+                    if not has_sub_option('memory', 'hotplugmemorymax'):
+                        logging.warning("hotplugmemorymax option not supported"
+                                        "by virt-install")
+                    else:
+                        cmd += ",hotplugmemorymax=%s" % hotplugmaxmem
+                    if not has_sub_option('memory', 'hotplugmemoryslots'):
+                        logging.warning("hotplugmemoryslots option not "
+                                        "supported by virt-install")
+                    else:
+                        cmd += ",hotplugmemoryslots=%d" % int(hotplugmemslots)
```
balamuruhans: type cast is not necessary as you have already taken care of it during assignment.

virttest/libvirt_vm.py
sathnaga: `hotplugmaxmem` and `hotplugmemslots` are related, just to make sure the default value incase user not set `hotplugmemslots` explicitly, not sure of default slot value, let me check and set it accordingly.

virttest/libvirt_vm.py
sathnaga: sure, will update it.

virttest/libvirt_vm.py
sathnaga: @balamuruhans it expects both, so default as `1` in safest option for now, ```
Starting install...
ERROR    XML error: both maximum memory size and memory slot count must be specified
Domain installation does not appear to have been successful.
```

sathnaga:
@balamuruhans addressed the comments

PR #1714, opened 2018-08-15
Add virsh cmd hypervisor-cpu-compare

hahakiki2010:


virttest/virsh.py
```diff
@@ -3099,6 +3099,18 @@ def cpu_compare(xml_file, **dargs):
     return command("cpu-compare %s" % xml_file, **dargs)
 
 
+def hypervisor_cpu_compare(xml_file, options="", **dargs):
```
balamuruhans: request to have param for `virttype`, `emulator`, `arch` and `machine` and process it here instead in testcase. `options` would be used for `--error`

PR #1719, opened 2018-08-16
virtio_console_guest: Fix python3 related issues

vivianQizhu:
1. Use buffering=1 for os.fdopen:
python3.6 does not accept unbuffered text I/O for os.fdopen(), so
use buffering=1 instead

2. Encode text for os.write():
python 3.6 requires a bytes-like object instead of str for write()

id: 1611136

### DIFF
--- a/shared/scripts/virtio_console_guest.py
+++ b/shared/scripts/virtio_console_guest.py
@@ -1433,8 +1433,8 @@ class Daemon:
         so.close()
         se.close()
 
-        sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)
-        sys.stderr = os.fdopen(sys.stderr.fileno(), 'w', 0)
+        sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 1)
+        sys.stderr = os.fdopen(sys.stderr.fileno(), 'w', 1)
         return True
 
     def start(self):
--- a/shared/scripts/virtio_console_guest.py
+++ b/shared/scripts/virtio_console_guest.py
@@ -897,19 +897,19 @@ class VirtioGuestPosix(VirtioGuest):
         """
         in_f = self._open([port])
 
-        data = ""
+        data = b""
         writes = 0
 
         if not is_static:
             while len(data) < length:
-                data += "%c" % random.randrange(255)
+                data += b"%c" % random.randrange(255)
             try:
                 writes = os.write(in_f[0], data)
             except Exception as inst:
                 print(inst)
         else:
             while len(data) < 4096:
-                data += "%c" % random.randrange(255)
+                data += b"%c" % random.randrange(255)
         if mode:
             while (writes < length):
                 try:


vivianQizhu:
Test pass:
```
15:06:05 INFO | PASS 1-Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.myprovider.virtio_console.spread_linear.virtserialport.virtio_console_smoke.rw_host_offline
```
Log:
```
15:05:43 DEBUG| Executing 'virt.init([['vc1', 'yes'], ['vc2', 'yes'], ['vc3', 'yes'], ['vc4', 'yes'], ['vs1', 'no'], ['vs2', 'no'], ['vs3', 'no'], ['vs4', 'no']])' on virtio_console_guest.py, vm: avocado-vt-vm1, timeout: 10
15:05:44 DEBUG| Executing 'virt.recv('vs1', 0, 1024, False)' on virtio_console_guest.py, vm: avocado-vt-vm1, timeout: 10
15:05:44 DEBUG| Executing 'virt.send('vs1', 10, True)' on virtio_console_guest.py, vm: avocado-vt-vm1, timeout: 10
```
Session log of guest:
```
2018-08-16 15:05:41: PASS: Start
2018-08-16 15:05:41: PASS: Daemon start.
2018-08-16 15:05:43: PASS: Init and check virtioconsole files in system.
2018-08-16 15:05:44: PASS: Recv data length 0
2018-08-16 15:05:54: PASS: Send data length 12
2018-08-16 15:05:54: PASS: nothing
2018-08-16 15:05:54: PASS: Guest is ok no thread alive
2018-08-16 15:05:54: PASS: guest_exit
2018-08-16 15:05:54: PASS: virtio_guest finished
```

vivianQizhu:
Pass RHEL8 too:
```
 (1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.my-provider.virtio_console.spread_linear.virtserialport.virtio_console_smoke.rw_host_offline: PASS (64.51 s)
```

shared/scripts/virtio_console_guest.py
```diff
@@ -1425,6 +1425,7 @@ def daemonize(self):
         so = open(self.stdout, 'w')
         se = open(self.stderr, 'w')
 
+        time.sleep(0.1)
```
balamuruhans: the parent process is exited in L1417, so how do we observe this conflict ?

vivianQizhu: @balamuruhans Sorry to confuse you, it forks twice, so the parent here I mean is actually the parent of its parent, which `return False` in L1404 and continue in run() function.
So I have got this in my debugging:
```
2018-08-13 23:32:47: Return false for parent, children pid: 1670
2018-08-13 23:32:47: Exit parent, children pid: 1672
2018-08-13 23:32:47: open files success, pid is: 1672
2018-08-13 23:32:47: dup stdin success, returned fd2 is: None, pid: 1672
2018-08-13 23:32:47: pid: 1669, file: /tmp/guest_daemon_pi
2018-08-13 23:32:47: pid: 1669, file: /tmp/guest_daemon_po
2018-08-13 23:32:47: pid: 1669, file: /tmp/guest_daemon_pe
2018-08-13 23:32:47: pid: 1672, file: /tmp/guest_daemon_pi
2018-08-13 23:32:47: Traceback (most recent call last):
2018-08-13 23:32:47:   File "/tmp/virtio_console_guest.py", line 1591, in <module>
2018-08-13 23:32:47:     main()
2018-08-13 23:32:47:   File "/tmp/virtio_console_guest.py", line 1539, in main
2018-08-13 23:32:47:     pid = list(pid)[0][0]
2018-08-13 23:32:47: IndexError: list index out of range
```

And after sleep(0.1)

```
2018-08-14 02:13:56: Return false for parent, children pid: 1628
2018-08-14 02:13:56: main, pid is: 1627
2018-08-14 02:13:56: Exit parent, children pid: 1630
2018-08-14 02:13:56: open files success, pid is: 1630
2018-08-14 02:13:56: children 1630 sleep for a while
2018-08-14 02:13:56: pid: 1627, file: /tmp/guest_daemon_pi, current pid: 1627
2018-08-14 02:13:56: pid: 1627, file: /tmp/guest_daemon_po, current pid: 1627
2018-08-14 02:13:56: pid: 1627, file: /tmp/guest_daemon_pe, current pid: 1627
2018-08-14 02:13:56: pid: 1630, file: /tmp/guest_daemon_pi, current pid: 1627
2018-08-14 02:13:56: pid: 1630, file: /tmp/guest_daemon_po, current pid: 1627
2018-08-14 02:13:56: pid: 1630, file: /tmp/guest_daemon_pe, current pid: 1627
2018-08-14 02:13:56: PASS: main Start, pid is: 1627
2018-08-14 02:13:57: dup stdin success, returned fd2 is: None, pid: 1630
2018-08-14 02:13:57: dup stdout success, returned fd2 is: None, pid: 1630
2018-08-14 02:13:57: dup stderr success, returned fd2 is: None, pid: 1630
```

balamuruhans: @vivianQizhu Thanks for the nice explanation, 

can we do something using flag like this, It will make the fix more robust. Please let me know your thoughts

```patch
diff --git a/shared/scripts/virtio_console_guest.py b/shared/scripts/virtio_console_guest.py
index b14d4a9c..c53ecf40 100755
--- a/shared/scripts/virtio_console_guest.py
+++ b/shared/scripts/virtio_console_guest.py
@@ -1363,12 +1363,17 @@ class Daemon:
         self.stdout = stdout
         self.stderr = stderr
 
+        # Flag to check is_file_open() completes
+        self.check_is_file_open = False
+
     @staticmethod
-    def is_file_open(path):
+    def is_file_open(path, check=False):
         """
         Determine process which open file.
 
         :param path: Path to file.
+        :param check: if True toggles flag check_is_file_open to True
+
         :return: [[pid,mode], ... ].
         """
         opens = []
@@ -1393,6 +1398,10 @@ class Daemon:
                 if e.errno == 2:
                     continue
                 raise
+
+        if check and not self.check_is_file_open:
+            self.check_is_file_open = True
+
         return opens
 
     def daemonize(self):
@@ -1425,17 +1434,23 @@ class Daemon:
         so = open(self.stdout, 'w')
         se = open(self.stderr, 'w')
 
-        os.dup2(si.fileno(), sys.stdin.fileno())
-        os.dup2(so.fileno(), sys.stdout.fileno())
-        os.dup2(se.fileno(), sys.stderr.fileno())
+        def _check_is_file_open():
+            return self.check_is_file_open
+
+        # wait to avoid race with grand father process by closing
+        # the files
+        if utils_misc.wait_for(_check_is_file_open, 60):
+            os.dup2(si.fileno(), sys.stdin.fileno())
+            os.dup2(so.fileno(), sys.stdout.fileno())
+            os.dup2(se.fileno(), sys.stderr.fileno())
 
-        si.close()
-        so.close()
-        se.close()
+            si.close()
+            so.close()
+            se.close()
 
-        sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)
-        sys.stderr = os.fdopen(sys.stderr.fileno(), 'w', 0)
-        return True
+            sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)
+            sys.stderr = os.fdopen(sys.stderr.fileno(), 'w', 0)
+            return True
 
     def start(self):
         """
@@ -1515,7 +1530,7 @@ def main():
     s_stderr = sys.stderr.fileno()
 
     pid = filter(lambda x: x[0] != str(os.getpid()),
-                 daemon.is_file_open(stdout))
+                 daemon.is_file_open(stdout, check=True))
     pid = list(pid)[0][0]
 
     print("PASS: Start")

```

vivianQizhu: @balamuruhans Thanks for your advise, it is true that to simply sleep a while is not robust enough, but unfortunately using flag like this can not work as expected. 
Firstly is_file_open() is a `staticmethod` so no `self` object.
I tried to remove the `@staticmethod` and give it `self`, but it still not work, since the is_file_open() is executed in another process(grandfather), so the flag `self.check_is_file_open` will never being set to True(in the child process). 
And also the wait time before dup2 can not be that long, even I set it to `3`, I will hit another conflict:
```
2018-08-22 11:15:29: PASS: Start
2018-08-22 11:15:32: file open check time out, dup2 not executed
2018-08-22 11:15:32: self.check_is_file_open: False
2018-08-22 11:15:32: PASS: Daemon start.
2018-08-22 11:15:34: Traceback (most recent call last):
2018-08-22 11:15:34:   File "/tmp/virtio_console_guest.py", line 1609, in <module>
2018-08-22 11:15:34:     main()
2018-08-22 11:15:34:   File "/tmp/virtio_console_guest.py", line 1567, in main
2018-08-22 11:15:34:     os.write(d_stdin, os.read(s_stdin, 1))
2018-08-22 11:15:34: OSError: [Errno 32] Broken pipe
```
So it can not be too early and also can not be too late.
I will say that original design of this script is not robust itself.
I guess to make it strong we have to add many signals(like L1511) to ensure every steps in different process been executed in right order. But I have no confidence to schedule it very well...it is really hard to avoid dead lock if use signals between parent and children process.
So I use this sleep, it might not be a really fix, but a workaround here.

luckyh: It look the second patch fixed the issue, here is my test result:
```
# python3 virtio_console_guest.py 
  File "virtio_console_guest.py", line 1571, in <module>
    main()
  File "virtio_console_guest.py", line 1519, in main
    pid = list(pid)[0][0]
IndexError: list index out of range
```
After applied the following patch:
```patch
- sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)
- sys.stderr = os.fdopen(sys.stderr.fileno(), 'w', 0)
+ sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 1)
+ sys.stderr = os.fdopen(sys.stderr.fileno(), 'w', 1)
```
```
# python3 vim virtio_console_guest.py 
PASS: Start
PASS: Daemon start.
```

Would you please help double confirm this? thanks.

balamuruhans: @vivianQizhu, Thank you for experimenting. sorry, it was my mistake this flag logic would work on multi threaded case as it shares the same process address space. Here it will not work, let's try @luckyh's observation.

vivianQizhu: @luckyh Thanks for help, you are right, I am removing the first commit.

vivianQizhu:
@balamuruhans @luckyh Updated, please help review again, thanks.
Latest test result:
```
01:19:14 INFO | PASS 1-Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.virtio_console.spread_linear.virtserialport.virtio_console_smoke.rw_host_offline
```

shared/scripts/virtio_console_guest.py
```diff
@@ -902,18 +902,18 @@ def send(self, port, length=1, mode=True, is_static=False):
 
         if not is_static:
             while len(data) < length:
-                data += "%c" % random.randrange(255)
+                data += "%s" % random.randrange(255)
```
luckyh: Well, this change is incorrect, it wants to get length of bytestring data here not textstring data.
```python
Python 2.7.15 (default, May 16 2018, 17:50:09) 
[GCC 8.1.1 20180502 (Red Hat 8.1.1-1)] on linux2
>>> "%c" % 128
'\x80'
>>> "%s" % 128
'128'
```
It should be:
```python
data = b""
...
        data += b"%c" % random.randrange(255)
```

vivianQizhu: @luckyh With "%c", it send these bytes directly, I don't find a way to decode it after receive the data from socket. That' s why I use string and encode it before sending. b"%c" does not resolve this, still can not decode it after received:

qemu/tests/virtio_console.py 

if (int(port.sock.recv(1024)) < 10):
Will hit error:
```
04:52:34 ERROR| ValueError: invalid literal for int() with base 10: b'P}\xcdm\xc0f\xeb\x8d*\xf3'
```
With decode():
if (int(port.sock.recv(1024).decode()) < 10):
```
L0984 ERROR|   File "/var/lib/avocado/data/avocado-vt/test-providers.d/downloads/io-github-autotest-qemu/qemu/tests/virtio_console.py", line 255, in test_rw_host_offline
    if (int(port.sock.recv(1024).decode()) < 10):

2018-08-30 04:44:27,277 test             L0984 ERROR| UnicodeDecodeError: 'utf-8' codec can't decode byte 0x99 in position 0: invalid start byte
```
Any suggestions? Thanks.

vivianQizhu: And with current "%s" + encode() way, what I received is:
```
05:39:35 DEBUG| [stdout] b'13739159193'
```
I suppose that should be what intended.

luckyh: It is exactly a bug of `tp-qemu/qemu/tests/virtio_console.py`.
```python
    def test_rw_host_offline():
        """
        Try to read from/write to host on guest when host is disconnected.
        :param cfg: virtio_console_params - which type of virtio port to test
        :param cfg: virtio_port_spread - how many devices per virt pci (0=all)
        """
        (vm, guest_worker, port) = virtio_test.get_vm_with_single_port(
            params.get('virtio_console_params'))
        if port.is_open():
            port.close()

        guest_worker.cmd("virt.recv('%s', 0, 1024, False)" % port.name, 10)
        match, tmp = guest_worker._cmd("virt.send('%s', 10, True)" % port.name,
                                       10)
        #                            send 10 characters ^^
        if match is not None:
            test.fail("Write on guest while host disconnected "
                      "didn't time out.\nOutput:\n%s" % tmp)

        port.open()

        if (port.sock.recv(1024) < 10):
        # why compare the content of the received data with integer number '10'?
        # should be `len(port.sock.recv(1024)) < 10`, right?
```

vivianQizhu: @luckyh You are right, thanks for you patience, I have updated the code and add you to sign off list. Please help review again.

vivianQizhu:
Latest test result:
```
06:57:07 INFO | PASS 1-Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.virtio_console.spread_linear.virtserialport.virtio_console_smoke.rw_host_offline
```

PR #1729, opened 2018-08-27
kernel_interface: implement set/get APIs for proc/sys filesystem

balamuruhans:
we have to set or get the values of proc filesystem to configure test
environment in host, remote host or in VM so this patch introduces
getter/setter methods for it.

Example:
>>> obj = ProcFS("/proc/sys/vm/nr_hugepages")
>>> obj.proc_fs_value # To get the value
>>> obj.proc_fs_value = 1 # To set the value

>>> obj2 = SysFS("/sys/kernel/mm/transparent_hugepage/enabled")
>>> obj2.sys_fs_value # To get the value
>>> obj2.sys_fs_value = "madvise" # To set the value

### DIFF
--- a/virttest/kernel_interface.py
+++ b/virttest/kernel_interface.py
@@ -1,0 +1,149 @@
+import os
+import re
+import logging
+
+from avocado.utils import process
+
+
+class FS(object):
+    """
+    Base class for proc/sys FS set and get
+    """
+    def __init__(self, fs, session=None):
+        """
+        Initializes path, session and func to trigger cmd
+
+        :param fs: proc/sys filesystem path
+        :param session: ShellSession object of remote or VM
+        """
+        self.fs = fs
+        if not os.path.isfile(self.fs):
+            raise AttributeError("%s is not available: %s" % self.fs)
+        self.func = process.getstatusoutput
+        self.session = session
+        if self.session:
+            self.func = self.session.cmd_status_output
+
+    @property
+    def fs_value(self):
+        """
+        Getter method for FS
+
+        :return: String, current value in given filesystem
+        """
+        self.fs_val = str(self.func("cat %s" % self.fs)[1])
+        return self.fs_val
+
+    @fs_value.setter
+    def fs_value(self, value):
+        """
+        Setter method for FS
+
+        :param value: value to be set for given filesystem
+        :return: Boolean, True on successfully set False on failure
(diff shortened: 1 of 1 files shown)


sathnaga:
@balamuruhans looks good, lets make the class name generic, so that it can be used even for `sysfs` aswell.

balamuruhans:
@sathnaga I have worked for sysfs as well and pushed the changes, Thanks reviewing.

balamuruhans:
@luckyh request for your review.

luckyh:
@balamuruhans it's interesting, I'd like to know the story about this proposal. As far as I can see, if only manages the leaf nodes of sysfs/procfs, it may not have a significant benefit than reading/writing the node directly.

balamuruhans:
@luckyh Thank you for reviewing, I was thinking it would help in the following ways but not limited to as we can always extend its functionality,

1. I am working on enabling migration testcases which have combinations to perform allocating HP, enabling THP, defrag etc on source host, remote destination host and inside multiple guests, where I can use these APIs to reduce my test code.
2. Planning to have params for pre and post process to enable/disable some procfs/sysfs interfaces, so that we can run testcases in controlled environment without modifying the testcase.
3. refactor/cleanup framework and testcases by adopting this API.

clebergnu:
We should definitely look into an existing module if it's not a trivial
task.

luckyh:
> We should definitely look into an existing module if it's not a trivial task.

@clebergnu agree with you, thanks.

virttest/utils_misc.py
```diff
@@ -4695,3 +4695,147 @@ def get_sosreport(path=None, session=None, remote_ip=None, remote_pwd=None,
         if session:
             session.close()
         return host_path
+
+
+class FS(object):
```
chunfuwen: Suggest move those codes to separate file since utils_misc.py pile up too many logic,and hard to maintain.

balamuruhans: sure, I too have this concern. I will do it.

balamuruhans:
@luckyh @clebergnu I checked the `procfs` module suggested by you, it does not give us methods to set/get for the `procfs` interfaces like HP, THP, defrag etc.,
This patch would not be redundant of the module, also it extends methods for `sysfs` interface and support it on remote host / guest as well.

PR #1730, opened 2018-08-28
qemu_vm: Fix the bug of passing fd in migration

luckyh:
Since aexpect 1.5.1, we have to explicitly pass fd to the destination
qemu process during migration.


hereischen:

PR #1731, opened 2018-08-29
make timeout configurable during multi vm stress

bssrikanth:
currently timeout is being hardcoded, it is better to have
it configurable. In multivm scenario where we have more
vms, systems response time increases leading to shell time
outs.


virttest/utils_test/__init__.py
```diff
@@ -1887,7 +1887,7 @@ def __init__(self, stress_type, params, download_url="", make_cmds="",
         self.remote_host = None
         self.copy_files_to = remote.copy_files_to
         self.params = params
-        self.timeout = 60
+        self.timeout = int(self.params.get('timeout', 60))
```
sathnaga: @bssrikanth it would be good to have prefix added to timeout as it will be generic cfg param, timeout can be set by other configs too...e:g:- `cmd_timeout`, `stresscmd_timeout`, something like this

bssrikanth:
@sathnaga correct.. now I have changed that to diff param

virttest/utils_test/__init__.py
```diff
@@ -2038,7 +2038,7 @@ def install(self):
                                     self.work_path)
         self.make_cmds = "cd %s;%s" % (install_path, self.make_cmds)
         logging.info('make and install %s', self.stress_type)
-        status, output = self.cmd_status_output(self.make_cmds, timeout=600)
+        status, output = self.cmd_status_output(self.make_cmds, timeout=self.stress_shell_timeout)
```
balamuruhans: This could cause failure sometimes, as the make commands are expected to take time more than a minute, request to have separate timeouts for,
1.  `cmd_status()` -> as it wait for the shell to return and 
2. `wait_for()` -> it wait for `True` to be returned by the function passed

bssrikanth: @balamuruhans I have separated wait_for timeout and shell timeout :+1:

PR #1733, opened 2018-08-30
expand func check_result

juteman:
the func check_result called the name "check_result"
but it can only check stderr result. Add stdout check
so it can check stderr and stdout both

Signe-off-by: qiulzhan <qiulzhan@redhat.com>

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1511,7 +1511,11 @@ class MigrationTest(object):
         return utils_misc.wait_for(check_state, timeout)
 
 
-def check_result(result, expected_fails=[], skip_if=[], any_error=False):
+def check_result(result,
+                 expected_fails=[],
+                 skip_if=[],
+                 any_error=False,
+                 expected_match=[]):
     """
     Check the result of a command and check command error message against
     expectation.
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1511,7 +1511,11 @@ def check_state():
         return utils_misc.wait_for(check_state, timeout)
 
 
-def check_result(result, expected_fails=[], skip_if=[], any_error=False):
+def check_result(result,
+                 expected_fails=[],
+                 skip_if=[],
+                 any_error=False,
+                 expected_match=[]):
```
kylazhang: It is better to add all parameters explanation below

virttest/utils_test/libvirt.py
```diff
@@ -1524,36 +1528,44 @@ def check_result(result, expected_fails=[], skip_if=[], any_error=False):
     :param any_error: Whether expect on any error message. Setting to True will
                       will override expected_fails
     """
-    logging.debug("Command result:\n%s" % result)
     stderr = results_stderr_52lts(result)
+    stdout = results_stdout_52lts(result)
+    all_msg = '\n'.join([stdout, stderr])
+    logging.debug("Command result: %s" % all_msg)
```
chloerh: Here, could use ',' instead of '%', only in logging function.

chloerh:
The rest is ok to me

virttest/utils_test/libvirt.py
```diff
@@ -1511,7 +1511,11 @@ def check_state():
         return utils_misc.wait_for(check_state, timeout)
 
 
-def check_result(result, expected_fails=[], skip_if=[], any_error=False):
+def check_result(result,
```
chunfuwen: I noticed that you add one more parameter for check_result function, please make sure whether it will impact other use case where check_result were used.

juteman: I think this may will not make impact. The new parameter has default value and i add it as the last parameters.

virttest/utils_test/libvirt.py
kylazhang: As other parameters also missing explanation, could you please also add the left ones?

PR #1742, opened 2018-09-04
Add flag to cleanup libvirtd debug log

hahakiki2010:
commit 26fdd19 cleanup the libvirtd log if case is not fail
or error, but sometimes it is expected to keep the plain
libvirtd.log in all cases. So make this new param to support it

### DIFF
--- a/avocado_vt/test.py
+++ b/avocado_vt/test.py
@@ -259,27 +259,28 @@ class VirtTest(test.Test):
             self.__status = details
         finally:
             # Clean libvirtd debug logs if the test is not fail or error
-            if(self.params.get("vm_type") == 'libvirt' and
-               self.params.get("enable_libvirtd_debug_log", "yes") == "yes"):
-                libvirtd_log = self.params["libvirtd_debug_file"]
-                if("TestFail" not in str(sys.exc_info()[0]) and
-                   "TestError" not in str(sys.exc_info()[0])):
-                    if libvirtd_log and os.path.isfile(libvirtd_log):
-                        logging.info("cleaning libvirtd logs...")
-                        os.remove(libvirtd_log)
-                else:
-                    # tar the libvirtd log and archive
-                    logging.info("archiving libvirtd debug logs")
-                    from virttest import utils_package
-                    if utils_package.package_install("tar"):
-                        archive = os.path.join(libvirtd_log.strip(os.path.basename(libvirtd_log)),
-                                               "libvirtd.tar.gz")
-                        cmd = "tar -zcf %s -P %s" % (archive, libvirtd_log)
-                        if process.system(cmd) == 0:
+            if self.params.get("libvirtd_log_cleanup", "no") == "yes":
+                if(self.params.get("vm_type") == 'libvirt' and
+                   self.params.get("enable_libvirtd_debug_log", "yes") == "yes"):
+                    libvirtd_log = self.params["libvirtd_debug_file"]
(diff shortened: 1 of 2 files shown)


avocado_vt/test.py
```diff
@@ -259,27 +259,28 @@ def setUp(self):
...
+                        from virttest import utils_package
+                        if utils_package.package_install("tar"):
+                            archive = os.path.join(libvirtd_log.strip(os.path.basename(libvirtd_log)),
+                                                   "libvirtd.tar.gz")
+                            cmd = "tar -zcf %s -P %s" % (archive, libvirtd_log)
+                            if process.system(cmd) == 0:
+                                os.remove(libvirtd_log)
+                        else:
+                            logging.error("Unable to find tar to compress libvirtd "
+                                          "logs")
```
kylazhang: And if we do not cleanup the log, if we also need to archive them?

avocado_vt/test.py
```diff
@@ -259,27 +259,28 @@ def setUp(self):
...
-                        archive = os.path.join(libvirtd_log.strip(os.path.basename(libvirtd_log)),
-                                               "libvirtd.tar.gz")
-                        cmd = "tar -zcf %s -P %s" % (archive, libvirtd_log)
-                        if process.system(cmd) == 0:
+            if self.params.get("libvirtd_log_cleanup", "no") == "yes":
+                if(self.params.get("vm_type") == 'libvirt' and
+                   self.params.get("enable_libvirtd_debug_log", "yes") == "yes"):
+                    libvirtd_log = self.params["libvirtd_debug_file"]
+                    if("TestFail" not in str(sys.exc_info()[0]) and
+                       "TestError" not in str(sys.exc_info()[0])):
```
sathnaga: this check can be simply added here ```
if("TestFail" not in str(sys.exc_info()[0]) and
                "TestError" not in str(sys.exc_info()[0]) and (self.params.get("libvirtd_log_cleanup", "no") == "yes")):```

shared/cfg/base.cfg
```diff
@@ -630,6 +630,7 @@ virtinstall_extra_args = ""
 enable_libvirtd_debug_log = "yes"
 libvirtd_debug_level = "2"
 libvirtd_debug_file = ""
+libvirtd_log_cleanup = "no"
```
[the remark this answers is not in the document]
hahakiki2010: @kylazhang I just thought previous solution without 26fdd19 is better, to keep all types of logs but not cleanup them by default.

avocado_vt/test.py
sathnaga: ah, okies, you basically need a plain libvirtd.log in all cases and do not want a archive at all, fine,
got it, but let us keep it `yes` bydefault as we do not want our test env to be flooded(consume more space) with lot of libvirtd logs, that's the reason we introduced log files cleanup and and archive in first place.

we had hit issues due to space, as some test can get hung/run for hours can easily lead to GBs of logs.

shared/cfg/base.cfg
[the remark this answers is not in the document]
hahakiki2010: @kylazhang set the param to ```yes``` by default, as sathnaga said they have hit issues due to space

avocado_vt/test.py
hahakiki2010: see sathnaga comments below, actually a plain libvirtd.log is needed

balamuruhans:
Thank you all for fix and reviews, let's merge.

PR #1744, opened 2018-09-05
utils_misc:Add function of backup file and restore

chloerh:
To backup file before making changes, restore file after test finished


dzhengfy:
Please refer virttest/remote.py RemoteFile class. With that class, you can create an object with passing local IP. The class has implemented backup/restore functions and used in many modules already.

PR #1745, opened 2018-09-05
Replace 'brctl' with alternatives for perf scripts

luckyh:

ID: 1622460


luckyh:
Thanks all.

PR #1751, opened 2018-09-12
qemu_vm: add nic/net options

sitoliu:
ID: 1624808 1624809 1585571 1585582

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -611,7 +611,8 @@ class VM(virt_vm.BaseVM):
         def add_nic(devices, vlan, model=None, mac=None, device_id=None,
                     netdev_id=None, nic_extra_params=None, pci_addr=None,
                     bootindex=None, queues=1, vectors=None, pci_bus='pci.0',
-                    ctrl_mac_addr=None):
+                    ctrl_mac_addr=None, indirect_desc=None, mrg_rxbuf=None,
+                    ctrl_vq=None, ctrl_vlan=None, status=None):
             if model == 'none':
                 return
             if devices.has_option("device"):
@@ -662,6 +663,14 @@ class VM(virt_vm.BaseVM):
                     dev.set_param('mq', 'on')
                 if vectors:
                     dev.set_param('vectors', vectors)
+                check_opt = {"indirect_desc": indirect_desc,
+                             "mrg_rxbuf": mrg_rxbuf,
+                             "ctrl_vq": ctrl_vq,
+                             "ctrl_vlan": ctrl_vlan,
+                             "status": status}
+                for opt in check_opt:
+                    if check_opt[opt] and check_opt[opt] in ["on", "off"]:
+                        dev.set_param(opt, check_opt[opt])
             if devices.has_option("netdev"):
                 dev.set_param('netdev', netdev_id)
             else:
@@ -673,7 +682,7 @@ class VM(virt_vm.BaseVM):
                     netdev_extra_params=None, tapfds=None, script=None,
                     downscript=None, vhost=None, queues=None, vhostfds=None,
                     add_queues=None, helper=None, add_tapfd=None,
-                    add_vhostfd=None, vhostforce=None):
+                    add_vhostfd=None, vhostforce=None, poll_us=0):
             mode = None
             if nettype in ['bridge', 'network', 'macvtap']:
                 mode = 'tap'
@@ -713,6 +722,8 @@ class VM(virt_vm.BaseVM):
                         if add_vhostfd:
                             cmd += ",vhostfd=%(vhostfd)s"
                             cmd_nd += ",vhostfd=%(vhostfd)s"
+                    if poll_us > 0:
+                        cmd += ",poll-us=%d" % poll_us
                 if vhostforce in ["on", "off"]:
                     cmd += ",vhostforce=%s" % vhostforce
                     cmd_nd = cmd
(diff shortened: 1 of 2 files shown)


virttest/qemu_vm.py
```diff
@@ -711,6 +719,8 @@ def add_net(devices, vlan, nettype, ifname=None, tftp=None,
                         if add_vhostfd:
                             cmd += ",vhostfd=%(vhostfd)s"
                             cmd_nd += ",vhostfd=%(vhostfd)s"
+                    if poll_us > 1:
```
vivianQizhu: What if `poll_us=0`? Since you define the default value to `-1`, or you want it default to `0`?

virttest/qemu_vm.py
```diff
@@ -609,7 +609,8 @@ def add_smp(devices):
         def add_nic(devices, vlan, model=None, mac=None, device_id=None,
                     netdev_id=None, nic_extra_params=None, pci_addr=None,
                     bootindex=None, queues=1, vectors=None, pci_bus='pci.0',
-                    ctrl_mac_addr=None):
+                    ctrl_mac_addr=None, indirect_desc=None, mrg_rxbuf=None,
+                    ctrl_vq=None, ctrl_vlan=None):
```
xiywang: status can also be add here.
RHEL7-95926

sitoliu: @xiywang Thanks, updated it.

virttest/qemu_vm.py
sitoliu: poll-us should be a uint32, so set it as -1 as default.

sitoliu:
@vivianQizhu @xiywang Updated code, please help review it again, thanks!

virttest/qemu_vm.py
```diff
@@ -609,7 +609,8 @@ def add_smp(devices):
         def add_nic(devices, vlan, model=None, mac=None, device_id=None,
                     netdev_id=None, nic_extra_params=None, pci_addr=None,
                     bootindex=None, queues=1, vectors=None, pci_bus='pci.0',
-                    ctrl_mac_addr=None):
+                    ctrl_mac_addr=None, indirect_desc=None, mrg_rxbuf=None,
```
xiywang: Could you explain the difference between using nic_extra_params and adding params like ctrl_vlan?

sitoliu: I guess both are ok.

xiywang: If nic_extra_params can handle the new params, then we should not add them to the function. use "nic_extra_params = ",indirect_desc=on"" is enough.
better to keep vt concise

vivianQizhu: Agree with @xiywang

sitoliu:
Close this patch as Xiyue and Qianqian suggested, thanks.

virttest/utils_net.py
```diff
@@ -863,8 +863,6 @@ def ping(dest=None, count=None, interval=None, interface=None,
             command += " -l %s" % packetsize
         if ttl:
             command += " -i %s" % ttl
-        if interface:
-            command += " -S %s" % interface
```
pezhang: @sitoliu Is there any reason to remove this option?  It looks like "-S" a valid option.

virttest/qemu_vm.py
```diff
@@ -662,6 +663,14 @@ def add_nic(devices, vlan, model=None, mac=None, device_id=None,
                     dev.set_param('mq', 'on')
                 if vectors:
                     dev.set_param('vectors', vectors)
+                check_opt = {"indirect_desc": indirect_desc,
+                             "mrg_rxbuf": mrg_rxbuf,
+                             "ctrl_vq": ctrl_vq,
+                             "ctrl_vlan": ctrl_vlan,
+                             "status": status}
+                for opt in check_opt:
+                    if check_opt[opt] and check_opt[opt] in ["on", "off"]:
```
pezhang: @sitoliu  Seems this line equal to 'if check_opt[opt] in ["on", "off"]:', right?

virttest/qemu_vm.py
```diff
@@ -611,7 +611,8 @@ def add_smp(devices):
         def add_nic(devices, vlan, model=None, mac=None, device_id=None,
                     netdev_id=None, nic_extra_params=None, pci_addr=None,
                     bootindex=None, queues=1, vectors=None, pci_bus='pci.0',
-                    ctrl_mac_addr=None):
+                    ctrl_mac_addr=None, indirect_desc=None, mrg_rxbuf=None,
+                    ctrl_vq=None, ctrl_vlan=None, status=None):
```
vivianQizhu: Recommend use `**kwargs` instead, this way we don't need further modification when there is other new option.

virttest/qemu_vm.py
```diff
@@ -673,7 +682,7 @@ def add_net(devices, vlan, nettype, ifname=None, tftp=None,
                     netdev_extra_params=None, tapfds=None, script=None,
                     downscript=None, vhost=None, queues=None, vhostfds=None,
                     add_queues=None, helper=None, add_tapfd=None,
-                    add_vhostfd=None, vhostforce=None):
+                    add_vhostfd=None, vhostforce=None, poll_us=0):
```
vivianQizhu: Same here.

kimi1978wy:
LGTM
Acked by: Yu Wang <wyu@redhat.com>

sitoliu:
Hi @vivianQizhu , could you please help review this patch? Thanks so much

vivianQizhu:
> Hi @vivianQizhu , could you please help review this patch? Thanks so much

Please resolve the comments above first, thanks.

PR #1752, opened 2018-09-12
Add <reservations> tag for virtual disk xml

lento-sun:
Libvirt starts to support scsi-3 PR function by a <reservations> tag
since version 4.4.0. Add it to virtual disk's xml.

### DIFF
--- a/virttest/libvirt_xml/devices/disk.py
+++ b/virttest/libvirt_xml/devices/disk.py
@@ -61,12 +61,17 @@ class Disk(base.TypedDeviceBase):
             libvirt_xml.devices.Disk.DiskSource instance
         encryption:
             libvirt_xml.devices.Disk.Encryption instance.
-    """
+        auth:
+            libvirt_xml.devices.Disk.Auth instance.
+        reservations:
+            libvirt_xml.devices.Disk.Reservations instance.
+   """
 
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/devices/disk.py
```diff
@@ -61,12 +61,17 @@ class Disk(base.TypedDeviceBase):
...
+        reservations:
+            libvirt_xml.devices.Disk.Reservations instance.
+   """
 
     __slots__ = ('device', 'rawio', 'sgio', 'snapshot', 'driver', 'target', 'alias',
                  'address', 'boot', 'readonly', 'transient', 'share',
                  'mirror', 'ready', 'iotune', 'source', 'blockio', 'geometry',
-                 'wwn', 'serial', 'vendor', 'product', 'encryption', 'auth')
+                 'wwn', 'serial', 'vendor', 'product', 'encryption', 'auth',
+                 'reservations')
```
chunfuwen: I assume 'reservations' attribute is not for disk, but disk.source, so this field may not needed

lento-sun: @chunfuwen yes, it's in disk source. But since <auth> and <encryption> can be put both inside or outside <source>, so I added the <reservations> as them for future potential change, and for extending negative scenarios.

PR #1757, opened 2018-09-14
virt_vm: move uptime() of the vm instance from libvirt_vm

balamuruhans:
move API to get the uptime of vm instance to virt_vm, this would be
helpful in testcases to assert the vm uptime whether vm went
down/restart during and after test.

without this API caused regression with commit 76c2f8f in tp-qemu

Reported-by: qizhu <qizhu@redhat.com>

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2010,20 +2010,6 @@ class VM(virt_vm.BaseVM):
             fcntl.lockf(lockfile, fcntl.LOCK_UN)
             lockfile.close()
 
-    def uptime(self, connect_uri=None):
-        """
-        Get uptime of the vm instance.
-
-        :param connect_uri: Libvirt connect uri of vm
-        :return: uptime of the vm on success, None on failure
-        """
-        if connect_uri:
-            self.connect_uri = connect_uri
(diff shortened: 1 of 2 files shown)


balamuruhans:
@luckyh then reboot, wait_for_login, shutdown are also such case, it is better to have as part of vm because in a multivm env we can avoid taking sessions from testcase by wrap it here.

luckyh:
> @luckyh then reboot, wait_for_login, shutdown are also such case, it is better to have as part of vm because in a multivm env we can avoid taking sessions from testcase by wrap it here.

It depends on the context, for example, you can "reboot" or "shutdown" a vm even if no guest os runs on it. This isn't the point, I mean, the question is what it the right meaning of "uptime"? the elapsed time from the vm process been present, or the elapsed time from the guest os booted up? `utils_misc.get_uptime()` returns the latter which is your expect, and that's why I suggested here.

If you do want this api for vm objects, the "right place" to put it in is the base vm class in `virt_vm.py`, at least.

balamuruhans:
@luckyh okay, got it. Thanks for the suggestion.

vivianQizhu:
@huangyum Would you help test it for bz 1628792? Thanks.

PR #1759, opened 2018-09-15
qemu_vm: add net_dev_disable_option_rom

codomania:
In the obsense of romfile= or rombar=0, QEMU will load the iPXE option rom
from the ipxe-roms-qemu package into the network device NIC's ROM BAR.
In some cases you may need to disable the option rom loading.

Signed-off-by: Brijesh Singh <brijesh.singh@amd.com>

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -627,6 +627,12 @@ class VM(virt_vm.BaseVM):
                 if ctrl_mac_addr and ctrl_mac_addr in ["on", "off"]:
                     dev.set_param('ctrl_mac_addr', ctrl_mac_addr)
                 dev.set_param('mac', mac, dynamic=True)
+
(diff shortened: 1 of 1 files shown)


virttest/qemu_vm.py
```diff
@@ -627,6 +627,12 @@ def add_nic(devices, vlan, model=None, mac=None, device_id=None,
                 if ctrl_mac_addr and ctrl_mac_addr in ["on", "off"]:
                     dev.set_param('ctrl_mac_addr', ctrl_mac_addr)
                 dev.set_param('mac', mac, dynamic=True)
+
+                # passing romfile= tells QEMU to disable ROM entirely for
+                # this device
+                no_romfile = params.get("net_dev_disable_option_rom", None)
```
balamuruhans: request to default the params used here and document about the params, its value and its use in cfg

PR #1763, opened 2018-09-25
qemu_vm: added an "enable_debug_console" option

hereischen:
isa_debugcon is always added if the device is available to the QEMU,
added an "enable_debug_console" param to control whether add the
isa_debugcon device or not.

id: 1632587

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -552,9 +552,6 @@ class VM(virt_vm.BaseVM):
             return virtio_port
 
         def add_log_seabios(devices):
-            if not devices.has_device("isa-debugcon"):
-                return ""
-
             default_id = "seabioslog_id_%s" % self.instance
             filename = os.path.join(data_dir.get_tmp_dir(),
                                     "seabios-%s" % self.instance)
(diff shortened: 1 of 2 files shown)


virttest/qemu_vm.py
```diff
@@ -1676,7 +1673,9 @@ def sort_key(dev):
             add_virtio_rng(devices, rng_params, parent_bus)
 
         # Add logging
-        devices.insert(StrDev('isa-log', cmdline=add_log_seabios(devices)))
+        if (devices.has_device("isa-debugcon") and
```
xutian: better to add   "enable_debug_console"  parameter in  test.cfg  like:
``enable_debug_console = yes

hereischen: do you mean put `enable_debug_console = yes`  in `avocado-vt/backends/qemu/cfg`

xutian: yes

hereischen: updated, please have a look, thanks

hereischen:
@xutian is  this version ok to you?

PR #1764, opened 2018-09-25
test_setup: reuse params for remote host info

dzhengfy:
This is to make use of two sets of variables which is remote_* and
server_* as they are used in different scenarios. So this is to satisfy
different requirements of usages.

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -2225,9 +2225,9 @@ def remote_session(params):
     :param params: Test params dict for remote machine login details
     :return: remote session object
     """
-    server_ip = params["server_ip"]
-    server_user = params.get("server_user", "root")
-    server_pwd = params["server_pwd"]
+    server_ip = params.get(params.get("server_ip"), params['remote_ip'])
+    server_user = params.get(params.get("server_user"), params['remote_user'])
+    server_pwd = params.get(params.get("server_pwd"), params['remote_pwd'])
     return remote.wait_for_login('ssh', server_ip, '22', server_user,
                                  server_pwd, r"[\#\$]\s*$")
 


virttest/test_setup.py
```diff
@@ -2225,9 +2225,9 @@ def remote_session(params):
     :param params: Test params dict for remote machine login details
     :return: remote session object
     """
-    server_ip = params["server_ip"]
-    server_user = params.get("server_user", "root")
-    server_pwd = params["server_pwd"]
+    server_ip = params["remote_ip"]
```
balamuruhans: we use remote_ip, remote_pwd and remote_user in migration or in remote uri testcases for target credentials. But this `remote_session()` can be used as wrapper in a generic way to take sessions for any host/guest machine, so we can avoid conflict of params.

virttest/test_setup.py
```diff
@@ -2225,9 +2225,9 @@ def remote_session(params):
...
+    server_user = params.get("remote_user", "root")
```
balamuruhans: @dzhengfy, can we default to param `remote_*` something like,
```python
server_ip = params.get(params.get("server_ip"), params['remote_ip'])
server_user = params.get(params.get("server_user"), params['remote_user'])
server_pwd = params.get(params.get("server_pwd"), params['remote_pwd'])
```

dzhengfy: @balamuruhans , I like this idea, which can benefit both of scenarios. I will do the update.

balamuruhans:
could you help changing the commit msg with respect to the fix, so that we can merge it.

dzhengfy:
@balamuruhans commit message was updated. Thanks.

PR #1766, opened 2018-09-26
Drop ntp variant use.

sitoliu:
Since we will drop using ntpdate to get local time in rhel8,
so drop ntp variant as well.

ID: 1629672


yanan-fu:

PR #1767, opened 2018-09-26
Add NIC limitation for e1000 and e1000e

xiywang:
id: 1627659

1. e1000 is no longer supported since RHEL.8
2. e1000e is supported from qemu version 2.6.0


shared/cfg/guest-hw.cfg
```diff
@@ -7,6 +7,7 @@ variants:
         nic_model = rtl8139
     - e1000:
         no ppc64 ppc64le
+        no RHEL.8
```
vivianQizhu: How about `only RHEL.6 RHEL.7`, thus we don't need to add RHEL.9 or later in the further.

luckyh: Good catch, but it should be:
```
RHEL:
    only RHEL.6 RHEL.7
```

xiywang: Updated. Please help review again. Thanks.

shared/cfg/guest-hw.cfg
```diff
@@ -19,6 +21,7 @@ variants:
         nic_model = e1000-82545em
     - e1000e:
         only q35
```
sitoliu: Do you need to disable this line? As we confirmed before, e1000e should work on both pc and q35.

xiywang: I think it's ok. e1000e better works with q35 although it can run on pc.
The most common used nic model for virtual network is virtio-net. 
e1000 and e1000e are with lower priority. jinzhao's team always run e1000e with q35 so it's related to their requirements.
They can decide whether to enable pc in auto since I will not run the entire loop with e1000e.
@jingzhao84 What's your opinion?

jingzhao84: @sitoliu @xiywang  I think above resolution is ok. e1000e is the default nic on q35 machine type and e1000 is the default nic on i440fx machine type. and q35 will change to default machine type on RHEL8, For e1000,  can set it in specific case if needed and i think it's very low priority. (ps: 1. e1000 will not support rhel8, 2. e1000 only plug into pcie.0 or pcie-pci-bridge if you need)

Thanks
Jing

sitoliu: @jingzhao84 @xiywang Thanks for explanation. Do you agree to set e1000e - only q35 as the default resolution? If so, I need to re update this configuration in multi_nic_hotplug.cfg.

PR #1778, opened 2018-10-08
utils_misc: use pgrep in get_pid() to retrieve pid of process

balamuruhans:
if multiple process are with same process name, then a unique
match string in process's cmdline can be used to identify process
instead of process name.


xutian:
LGTM

balamuruhans:
Thank you @xutian

PR #1793, opened 2018-10-23
Add param driver_cache in attach_disks()

fangge1212:
Users should be able to pass driver_cache in params of attach_disks()


chunfuwen:
confirm it is not useful any more

PR #1800, opened 2018-10-31
nfs.py: use nfs-server instead of nfs

dzhengfy:
nfs-utils package does not provide /usr/lib/systemd/system/nfs.service
any more. So /usr/lib/systemd/system/nfs-server.service is our choice.

### DIFF
--- a/virttest/nfs.py
+++ b/virttest/nfs.py
@@ -162,7 +162,7 @@ class Nfs(object):
             self.nfs_setup = True
             path.find_command("service")
             path.find_command("exportfs")
-            if distro_details.name == 'Ubuntu':
+            if distro_details.name in ('Ubuntu', 'rhel'):
                 self.nfs_service = service.Factory.create_service("nfs-server")
             else:
                 self.nfs_service = service.Factory.create_service("nfs")


virttest/nfs.py
```diff
@@ -162,7 +162,8 @@ def __init__(self, params):
             self.nfs_setup = True
             path.find_command("service")
             path.find_command("exportfs")
-            if distro_details.name == 'Ubuntu':
+            distro_name = distro_details.name.lower()
+            if (distro_name == 'ubuntu' or distro_name == 'redhat'):
```
hereischen: `if distro_name in ( 'ubuntu', 'redhat'):`

dzhengfy: @hereischen ,thanks, and fix.

virttest/nfs.py
```diff
@@ -162,7 +162,8 @@ def __init__(self, params):
             self.nfs_setup = True
             path.find_command("service")
             path.find_command("exportfs")
-            if distro_details.name == 'Ubuntu':
+            distro_name = distro_details.name.lower()
```
chunfuwen: Suggest still keep distribution name as before, and no need to be converted to lower case, For RHEL, please check related distribution name

dzhengfy: fixed.

chloerh:
Great, this fixed my failed case

PR #1802, opened 2018-11-01
Move update_disk_driver_(with_iothread) here from tp-libvirt

fangge1212:
Function update_disk_driver_(with_iothread) is very useful and
common, so move it here from tp-libvirt.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -2321,6 +2321,48 @@ def set_guest_agent(vm):
             raise exceptions.TestFail("Fail to run qemu-ga in guest")
 
 
+def update_disk_driver(vm_name, disk_name, disk_type, disk_cache):
+    """
+    Update disk driver
+    """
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -2321,6 +2321,48 @@ def set_guest_agent(vm):
...
+    vmxml.sync()
+
+
+def update_disk_driver_with_iothread(vm_name, iothread):
+    """ Update disk driver with iothread."""
+    vmxml = vm_xml.VMXML.new_from_dumpxml(vm_name)
+    # Delete cputune/iothreadids section, it may have conflicts
+    # with domain iothreads.
+    del vmxml.cputune
+    del vmxml.iothreadids
```
dzhengfy: Directly removing other sections is a simple way, but not a proper way, because it will remove some useful configurations for testing. Could you think a better way to avoid of potential conflicts?

fangge1212: What do you mean by "because it will remove some useful configurations for testing"

PR #1803, opened 2018-11-05
qemu_storage: added a map method

hereischen:
Added a qemu-img map wrapper.

id: 1646843


hereischen:
@pingl2016 I will not support `--object` and `--image-opts` at this stage.
When start doing luks cases, I will refact the whole QemuImg, as almost all its methods don't support options like '--object' and '--image-opts'.

pingl2016:
Acked-by: Ping Li \<pingl@redhat.com\>

PR #1806, opened 2018-11-07
Add perf and event config in domain xml

jinqi7:


virttest/libvirt_xml/vm_xml.py
```diff
@@ -2446,6 +2453,85 @@ def __init__(self, virsh_instance=base.virsh):
         self.xml = '<memtune/>'
 
 
+class VMPerfXML(VMXML):
+
+    """
+    perf tag XML class
+
```
chunfuwen: suggest follow below style:
'''
xx
add Properties:
events:
      string,xx

'''

virttest/libvirt_xml/vm_xml.py
```diff
@@ -2446,6 +2453,85 @@ def __init__(self, virsh_instance=base.virsh):
...
+    Elements:
+        event
+    """
+
+    __slots__ = ('events',)
```
chunfuwen: remove additional ',' at the end

PR #1812, opened 2018-11-08
Add check_dumpxml

hahakiki2010:
Add this common func considering it may be
useful for other tests


virttest/utils_test/libvirt.py
```diff
@@ -3604,3 +3604,24 @@ def get_disk_alias(vm, source_file=None):
...
+    :return: True if exist, False otherwise
+    """
+    v_xml = vm_xml.VMXML.new_from_dumpxml(vm.name)
+    with open(v_xml.xml) as xml_f:
+        if content not in xml_f.read():
+            if err_ignore:
+                return False
+            else:
+                raise exceptions.TestFail("Expected '%s' was not found in "
+                                          "qemu command line" % content)
```
balamuruhans:
```python
```

hahakiki2010: updated

balamuruhans:
Thanks for the patch @hahakiki2010

PR #1818, opened 2018-11-13
qemu_storage: added a measure method

hereischen:
1. it supports to get the target_fmt's benchmark sizes by provide a
   size as an argument.
2. it can measure a image according to the target_fmt.

id: 1650490


hereischen:
@pingl2016 I know, if --size is set, -f $fmt will be overwrite, thanks

luckyh:
Thanks all.

PR #1819, opened 2018-11-13
Device should be connected to downstream port bus using device number 0

rayx:
QPCISwitchBus was designed to hide creation of downstream port from
user. To do that, it first creates a downstream port device and its bus,
connects the downstream port device to upstream port bus (represented by
QPCISwitchBus) at address specified by the device's addr parameter. Then
it connects the device to the downstream port bus (repsented by QPCIBus)
using the same address. This doesn't work when test wants to create a
topology like the following, because the device 2 would be connected to
downstream port 2 using a non-zero addr value, which cause the device
invisible to guest OS (QEMU doens't have a check for this so it starts
up happily).

   root port -> upstream port +-> downstream port 1 -> device 1
                              +-> downstream port 2 -> devive 2

The nature of the issue is the current code reuses device attr value for
two different purposes: the address to connect downstream port to
upstream port bus and the address to connect device to downstream port
bus. These two values are not necessarily same. They actually have to be
different in many cases.

The change fixes this issue by making sure device is always connected to
downstream port bus using device number 0. It does this by introducing a
new QPCIDownstreamPortBus class and provides its own copy of insert(),
which is almost same as QSparseBus's insert() except it hardcodes device
address to [0, 0]. This is not the best way to fix it, but is the
simplest way I can find without changing API of related classes.

Signed-off-by: Huan Xiong <huan.xiong@hxt-semitech.com>

### DIFF
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -1970,7 +1970,7 @@ class QPCISwitchBus(QPCIBus):
         if addr not in self.__downstream_ports:
             _addr = int(addr, 16)
             bus_id = "%s.%s" % (self.busid, _addr)
-            bus = QPCIBus(bus_id, 'PCIE', bus_id)
+            bus = QPCIDownstreamPortBus(bus_id, bus_id)
             self.__downstream_ports["0x%x" % _addr] = bus
             downstream = QDevice(self.__downstream_type,
                                  {'id': bus_id,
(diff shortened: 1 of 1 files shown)


rayx:
I observed this issue when I look into [this failure](https://github.com/autotest/tp-qemu/issues/1557). With this patch, the test generated correct command line (all xhci controllers' attr option are 0x0) and passed.

BTW, I didn't look into if rootport code has similar issue. I'd like to focus on downstream port in this PR.

PR #1822, opened 2018-11-14
Update v2v cmd running timeout value to 5200

xiaodwan:
For v2v '-o rhv' output, v2v write data to NFS server directly,
so 40min is usually enough, but in some special testing hosts,
if the hosts are plugged in with a telphone line, the transmission
speed will be very slow, 40min may still timeout.

Now v2v implements rhv-upload feature, it copies the data to an
imageio (HTTPS) endpoint, and it usually spends more time. 40 mins
is not enough now.

I tested the value in some different hosts, the time vairies from
20mins to 90min. And developors also said 2 hours is a safe value.

So set the timeout value to 5200s.


xiaodwan:
@chloerh @chunfuwen please help to review it, thanks.

PR #1824, opened 2018-11-16
qemu_storage: added a param `convert_tag` for QemuImg.convert()

hereischen:
The current implementation does not support converting an image multiple
times or converting multiple images once in one test case.
That is because `params["image_convert"]` can only be used once a case.

This patch works around this issue. However, it lose track which image
will be converted in the cfg.

id: 1650490


hereischen:
@pingl2016  You are right, thanks man~

PR #1837, opened 2018-11-23
vittest.utils_disk: bug fix when virtio_floppy is not specified

lijinlijin:
when virtio_floppy is not specified, copy drivers(both from floppy and cdrom) will be totally skipped which lead to guest os installation fail. Copy drivers from cdrom should be cont in this case.

ID:1652458

Signed-off-by: lijin <lijin@redhat.com>


lijinlijin:
@luckyh 
Could you help to review this PR?
Thanks

virttest/utils_disk.py
```diff
@@ -960,9 +960,11 @@ def _copy_virtio_drivers(self, virtio_floppy, cdrom_virtio):
         mnt_pnt = tempfile.mkdtemp(prefix='cdrom_virtio_', dir=self.tmpdir)
         mount(cdrom_virtio, mnt_pnt, options='loop,ro', verbose=DEBUG)
         try:
-            copytree(mnt_pnt, self.mount, ignore='*.vfd')
-            cmd = 'mcopy -s -o -n -i %s ::/* %s' % (virtio_floppy, self.mount)
-            process.run(cmd, verbose=DEBUG)
+            if os.path.isfile(cdrom_virtio):
+                copytree(mnt_pnt, self.mount, ignore='*.vfd')
+            if os.path.isfile(virtio_floppy):
```
xutian: both virtio_floppy and virtio_cdrom are exists, files copy from floppy will overwrite files copied from cdrom, I guest here should be

`
            if os.path.isfile(cdrom_virtio):
                  ...
            elif os.path.isfile(virtio_floppy):
                  ...
            else:
                 raise ...
`

lijinlijin:
@xutian 
Any further comment?

Thanks

PR #1844, opened 2018-11-27
Add pci address element and attributes to serial device

jinqi7:

### DIFF
--- a/virttest/libvirt_xml/devices/serial.py
+++ b/virttest/libvirt_xml/devices/serial.py
@@ -11,7 +11,7 @@ from virttest.libvirt_xml.devices.character import CharacterBase
 class Serial(CharacterBase):
 
     __slots__ = ('protocol_type', 'target_port', 'target_type',
-                 'target_model', 'sources', 'alias')
+                 'target_model', 'sources', 'address')
 
     def __init__(self, type_name='pty', virsh_instance=base.virsh):
         # Additional attribute for protocol type (raw, telnet, telnets, tls)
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/devices/serial.py
```diff
@@ -23,11 +24,19 @@ def __init__(self, type_name='pty', virsh_instance=base.virsh):
...
         accessors.XMLAttribute('target_model', self, parent_xpath='/target',
                                tag_name='model', attribute='name')
-        accessors.XMLElementDict('alias', self, parent_xpath='/',
-                                 tag_name='alias')
         accessors.XMLElementList('sources', self, parent_xpath='/',
                                  marshal_from=self.marshal_from_sources,
                                  marshal_to=self.marshal_to_sources)
+        accessors.XMLElementDict('address', self, parent_xpath='/',
+                                 tag_name='address')
+        accessors.XMLAttribute('address_type', self, parent_xpath='/',
```
chunfuwen: I assume those attributes configuration are not needed, just accessors.XMLElementDict is enough

jinqi7: Changed to use XMElementNest with Address class instead of XMLElementDict.

jinqi7: Changed to use ElementDict only for no special needed for sub attribute.

yanjing1993:
Suggest to reserve 'alias' for serial device, I think it is necessary to test it.

jinqi7:
> Suggest to reserve 'alias' for serial device, I think it is necessary to test it.

I think it can be added when it's needed.

yanjing1993:
> > Suggest to reserve 'alias' for serial device, I think it is necessary to test it.
> 
> I think it can be added when it's needed.

Reserving alias has no conflict with adding pci address in 'Serial' class, please reserve it.

PR #1845, opened 2018-11-27
[python3]openvswitch: fix errors on python3

vivianQizhu:
1. Change str to int as in python3 str can not compare with int
2. Drop super.init(). Since the parent classes do not define init, so
   it calles object.init(), while python3 does not allow pass parameters
   to object.init(), it will prompt error: "TypeError: object.init()
   takes no parameters". It is unnecessary to call object.init() even
   for python2, So drop it.

### DIFF
--- a/virttest/openvswitch.py
+++ b/virttest/openvswitch.py
@@ -158,7 +158,7 @@ class OpenVSwitchControl(object):
             int_ver = ''.join(a)
         except Exception:
             raise ValueError("Wrong version format '%s'" % version)
-        return int_ver
+        return int(int_ver)
 
     @classmethod
     def get_version(cls):
@@ -356,10 +356,6 @@ class OpenVSwitchSystem(OpenVSwitchControlCli_CNT, OpenVSwitchControlDB_CNT):
         :param ovs_pidfile: Path of OVS ovs-vswitchd pid.
         :param install_prefix: Path where is openvswitch installed.
         """
-        sup = super(man[self.__class__, OpenVSwitchSystem], self)
-        sup.__init__(self, db_path, db_socket, db_pidfile, ovs_pidfile,
-                     dbschema, install_prefix)
-
         self.cleanup = False
         self.pid_files_path = None
 
@@ -459,9 +455,6 @@ class OpenVSwitch(OpenVSwitchSystem):
         :param ovs_pidfile: Path of OVS ovs-vswitchd pid.
         :param install_prefix: Path where is openvswitch installed.
         """
-        super(man[self, OpenVSwitch], self).__init__(db_path, db_socket,
-                                                     db_pidfile, ovs_pidfile,
-                                                     dbschema, install_prefix)
         self.tmpdir = "/%s/openvswitch" % (tmpdir)
         try:
             os.mkdir(self.tmpdir)


vivianQizhu:
@kylazhang @luckyh @quanwenli @xiywang Please help review, thanks. Test passed on python3.

virttest/openvswitch.py
```diff
@@ -356,10 +356,6 @@ def __init__(self, db_path=None, db_socket=None, db_pidfile=None,
         :param ovs_pidfile: Path of OVS ovs-vswitchd pid.
         :param install_prefix: Path where is openvswitch installed.
         """
-        sup = super(man[self.__class__, OpenVSwitchSystem], self)
-        sup.__init__(self, db_path, db_socket, db_pidfile, ovs_pidfile,
-                     dbschema, install_prefix)
```
luckyh: I'm afraid that, by only removing these two lines, the above listed init args wouldn't be set correctly.

It looks like the `VersionableClass` facilities has the problem to handle `__init__()` well in multiple inheritance situation.

vivianQizhu: @luckyh If `OpenVSwitch.__init__()` requires a change, I suppose should be removing those parameters from its init and super.init.
Guess it wants pass the parameters to `OpenVSwitchControl`, but it should never passed to it before, since `OpenVSwitchControl` are not in the super list and it actually calls `object.init`. Where I can see does handle the parameters is `OpenVSwitch.init_new()`.  Or the parameters are initialized when `OpenVSwitchSystem()` called, but not via `init()`, since OpenVSwitchControl redefine its `new()` with those parameters, that should be where the instance created(before `init()` been called). 
Above is just guess，I will debug for it after I get a proper environment.

vivianQizhu:
@luckyh Updated, drop all redundent `init`. 
Since their ancestor class `OpenVSwitchControl` refine its `new` and handle all passed parameters, so all the actions that attempting to pass parameters by calling `super.init` are redundant and unnecessary.
The creation of both classes will start from `OpenVSwitchControl.new`:
Debug file:
```
diff --git a/virttest/openvswitch.py b/virttest/openvswitch.py
index 8f8dc99..d5b6425 100644
--- a/virttest/openvswitch.py
+++ b/virttest/openvswitch.py
@@ -113,6 +113,10 @@ class OpenVSwitchControl(object):
         :param install_prefix: Path where is openvswitch installed.
         """
         # if path is None set default path.
+        print("OpenVSwitchControl new is called."
+              " The parameters passed: db_path: %s, db_socket: %s, db_pidfile: %s,"
+              " ovs_pidfile: %s, dbschema: %s, install_prefix: %s"
+              % (db_path, db_socket, db_pidfile, ovs_pidfile, dbschema, install_prefix))
         if not install_prefix:
             install_prefix = "/"
         if not db_path:
@@ -356,6 +360,7 @@ class OpenVSwitchSystem(OpenVSwitchControlCli_CNT, OpenVSwitchControlDB_CNT):
         :param ovs_pidfile: Path of OVS ovs-vswitchd pid.
         :param install_prefix: Path where is openvswitch installed.
         """
+        print("OpenVSwitchSystem init")
         self.cleanup = False
         self.pid_files_path = None
 
@@ -455,9 +460,7 @@ class OpenVSwitch(OpenVSwitchSystem):
         :param ovs_pidfile: Path of OVS ovs-vswitchd pid.
         :param install_prefix: Path where is openvswitch installed.
         """
-        super(man[self, OpenVSwitch], self).__init__(db_path, db_socket,
-                                                     db_pidfile, ovs_pidfile,
-                                                     dbschema, install_prefix)
+        print("OpenVSwitch init")
```
Result:
```
Python 2.7.5 (default, Sep 12 2018, 05:31:16) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux2
>>> from virttest import openvswitch
>>> from virttest import versionable_class
>>> versionable_class.factory(openvswitch.OpenVSwitchSystem)("/tmp")
OpenVSwitchControl new is called. The parameters passed: db_path: /tmp, db_socket: None, db_pidfile: None, ovs_pidfile: None, dbschema: None, install_prefix: None
OpenVSwitchSystem init
<virttest.openvswitch.managed_OpenVSwitchSystem_OpenVSwitchControlCli_140_OpenVSwitchControlDB_140 object at 0x7f077f911e50>
>>> versionable_class.factory(openvswitch.OpenVSwitch)("/tmp")
OpenVSwitchControl new is called. The parameters passed: db_path: /tmp, db_socket: None, db_pidfile: None, ovs_pidfile: None, dbschema: None, install_prefix: None
OpenVSwitch init
<virttest.openvswitch.managed_OpenVSwitch_OpenVSwitchControlCli_140_OpenVSwitchControlDB_140 object at 0x7f079531e990>
```

vivianQizhu:
And I actually don't find any usage of the class `OpenVSwitch()`, in avocado-vt, tp-qemu and tp-libvirt. So only test it by the above unit test method.

vivianQizhu:
@kylazhang Request for your ack, thanks.

quanwenli:
LGTM. Thanks Qianqian.

PR #1852, opened 2018-11-29
env_process: make the firewalld param with `undefined` option

balamuruhans:
`firewalld_service` param can hold `undefined` (leave it in
whatever state it is in), set `yes` to turn on and `no` to
turn off firewalld.


shared/cfg/base.cfg
```diff
@@ -684,6 +684,7 @@ inputs = ""
 
 # firewalld_service param used as global param to abstract the firewalld configuration
 # the framework can control the firewalld settings with this param set/unset, by default
-# firewalld is enabled
+# firewalld is `undefined` (leave it in whatever state it is in), set `yes` to turn on
+# and `no` to turn off firewalld
 
-# firewalld_service = "yes"
+# firewalld_service = "undefined"
```
luckyh: How about the name `"default"`? It looks better.

luckyh: or `"preset"`, either one looks good to me.

balamuruhans: @luckyh Thanks for review, I have made the change as suggested.

PR #1858, opened 2018-11-30
install_package(): Don't raise exception when ignore_status is True

lento-sun:
When ignore_status=True, test will still be blocked if some
exceptions such as ShellProcessTerminatedError, ShellStatusError
raised. So change the code not to raise exceptions when
ignore_status=True, but just log an error.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2834,13 +2834,18 @@ class VM(virt_vm.BaseVM):
         :param name: Name of package to be installed
         """
         session = self.wait_for_login()
-        if not utils_package.package_install(name, session):
-            if not ignore_status:
-                session.close()
+        try:
+            if not utils_package.package_install(name, session):
                 raise virt_vm.VMError("Installation of package %s failed" %
                                       name)
-            logging.error("Installation of package %s failed", name)
-        session.close()
+        except Exception as exception_detail:
+            if ignore_status:
+                logging.error("When install: %s\nError happened: %s\n",
+                              name, exception_detail)
+            else:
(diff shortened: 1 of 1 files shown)


lento-sun:
we don't have pm-utils pkg in latest rhel version, and don't need it to test guest agent related cases. So the prepare_guest_agent() method in libvirt_vm.py always fail the whole case with an exception raised. This pr will fix these case failures.

virttest/libvirt_vm.py
```diff
@@ -2834,13 +2834,18 @@ def install_package(self, name, ignore_status=False):
         :param name: Name of package to be installed
         """
         session = self.wait_for_login()
-        if not utils_package.package_install(name, session):
-            if not ignore_status:
-                session.close()
+        try:
+            if not utils_package.package_install(name, session):
```
kylazhang: Also need to add condition judgment for ignore_status as before as if install fail and no exception, no need to raise error as before

chunfuwen: I assume if install fail , it need intentionedly raise error whether it throw exception itself or not.

lento-sun: if install failed without a exception raised, I'm still raising a virt_vm.VMError in following untouched code as before. 
So if install failed OR exception raised by utils_package.package_install, a exception will always caught by the except clause.

lento-sun: So you mean to remove the following
```
raise virt_vm.VMError("Installation of package %s failed" %
                                       name)
```
But instead to add `if ignore_status` in the try clause?

PR #1860, opened 2018-12-03
virttest.guest_agent: Fix issue that qga can't get response with '\xff' on python3

xiagao:
id: 1631240


luckyh:
Thanks all.

PR #1862, opened 2018-12-06
Fixup: Update numa cell based on modified vcpu number

sathnaga:
Currently, numa cell is not updated when there is a
modification for vcpu numbers due to which numa
check during vcpu hotplug tests fails, this patch tries to
update the numa config accordingly.

Libvirt does not complain currently about the mismatch in
numa cell vcpus vs maxvcpus, below bug is raised for the
same.
Reference: https://bugzilla.redhat.com/show_bug.cgi?id=1656762

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -735,7 +735,7 @@ class VMXML(VMXMLBase):
     @staticmethod
     def set_vm_vcpus(vm_name, vcpus, current=None, sockets=None, cores=None,
                      threads=None, add_topology=False, topology_correction=False,
-                     virsh_instance=base.virsh):
+                     update_numa=True, virsh_instance=base.virsh):
         """
         Convenience method for updating 'vcpu', 'current' and
         'cpu topology' attribute property with of a defined VM
@@ -748,6 +748,7 @@ class VMXML(VMXMLBase):
         :param threads: number of threads, default None
         :param add_topology: True to add new topology definition if not present
         :param topology_correction: Correct topology if wrong already
+        :param update_numa: Update numa
         :parma virsh_instance: virsh instance
         """
         vmxml = VMXML.new_from_dumpxml(vm_name, virsh_instance=virsh_instance)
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -772,6 +773,18 @@ def set_vm_vcpus(vm_name, vcpus, current=None, sockets=None, cores=None,
...
                                          'threads': threads}
                 vmxml['cpu'] = vmcpu_xml
+            if update_numa and vmxml.cpu.numa_cell:
+                vmcpu_xml = vmxml['cpu']
+                if len(vmxml.cpu.numa_cell) == 2 and vcpus >= 4:
+                    node0xml = vmcpu_xml.numa_cell[0]
+                    node0xml["cpus"] = "0-%s" % ((vcpus / 2) - 1)
+                    node1xml = vmcpu_xml.numa_cell[1]
+                    node1xml["cpus"] = "%s-%s" % (vcpus / 2, vcpus - 1)
+                    vmcpu_xml.set_numa_cell([node0xml, node1xml])
```
balamuruhans: we can make it more generic,
```python
total_numa = len(vmxml.cpu.numa_cell)
if vcpus < total_numa:
    vcpus_numa = total_numa // vcpus
    vcpus_rem = total_numa % vcpus
    index = 0
    for each_numa in range(total_numa):
        node = {}
        if vcpus_numa > 1:
            if each_numa == total_numa - 1:
                node["cpus"] = "%s-%s" % (index, index + vcpus_numa + vcpus_rem - 1)
            else:
                node["cpus"] = "%s-%s" % (index, index + vcpus_numa - 1)
        else:
            node["cpus"] = str(vcpus_numa)
        numa_list.append(node)
    vmcpu_xml.set_numa_cell(numa_list)

else:
    logging.warning("Guest numa could not be updated, expect "
                    "failures if guest numa is checked")
```

sathnaga: @balamuruhans make sense, i just wanted to rework on it based on bugzilla update, nevertheless we can implement the generic approach, will check on this and update it.

sathnaga: as the intention of this change is to have a numa cell matched with no. of vcpus lets try to change in generic way as you have suggested if not possible lets fall back to having two nodes with vcpus divided across,

sathnaga: we will have total_numa < vcpus and most of the times that is going to be the case.....

sathnaga:
@balamuruhans Thanks for suggestion. ve addressed, let me know your comments and help merge.

hsmj1412:
@sathnaga @balamuruhans Hi, in some cases , the vmcpu_xml is used before defined, so it will report “UnboundLocalError: local variable 'vmcpu_xml' referenced before assignment”.

sathnaga:
@hsmj1412 can you pls check if https://github.com/avocado-framework/avocado-vt/pull/2129 solves the issue you are facing, sorry for the inconvenience.

PR #1871, opened 2018-12-17
Add device type 'iommu'

chloerh:

### DIFF
--- a/virttest/libvirt_xml/devices/iommu.py
+++ b/virttest/libvirt_xml/devices/iommu.py
@@ -1,0 +1,26 @@
+"""
+Support for the iommu device XML
+
+http://libvirt.org/formatdomain.html#elementsDevices
+"""
+
+from virttest.libvirt_xml import accessors
+from virttest.libvirt_xml.devices import base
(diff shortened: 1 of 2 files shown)


virttest/libvirt_xml/devices/iommu.py
```diff
@@ -0,0 +1,32 @@
...
+
+    __slots__ = ('model', 'intremap', 'eim')
+
+    def __init__(self, virsh_instance=base.base.virsh):
+        accessors.XMLAttribute(property_name='model',
+                               libvirtxml=self,
+                               parent_xpath='/',
+                               tag_name='iommu',
+                               attribute='model')
+        accessors.XMLAttribute(property_name="intremap",
```
chunfuwen: I  assume initremap and eim is property under driver element, so suggest  expose as dict
        accessors.XMLElementDict('driver', self, parent_xpath='/',
                                 tag_name='driver')
```
<devices>
  <iommu model='intel'>
    <driver intremap='on'/>
  </iommu>
</devices>
```

PR #1875, opened 2018-12-25
test_setup: clear memory caches before set hugepages

huangyum:
ID: 1660816

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -429,7 +429,10 @@ class HugePageConfig(object):
         self.hugepage_force_allocate = params.get("hugepage_force_allocate",
                                                   "no")
         self.suggest_mem = None
-        self.lowest_mem_per_vm = int(params.get("lowest_mem", "256"))
+        normalize_data_size = utils_misc.normalize_data_size
+        vm_mem_minimum = params.get("vm_mem_minimum", "512M")
+        self.lowest_mem_per_vm = float(normalize_data_size(vm_mem_minimum))
+        utils_memory.drop_caches()
 
         target_hugepages = params.get("target_hugepages")
         if target_hugepages is None:


virttest/test_setup.py
```diff
@@ -429,7 +429,10 @@ def __init__(self, params):
         self.hugepage_force_allocate = params.get("hugepage_force_allocate",
                                                   "no")
         self.suggest_mem = None
-        self.lowest_mem_per_vm = int(params.get("lowest_mem", "256"))
+        normalize_data_size = utils_misc.normalize_data_size
+        vm_mem_minimum = params.get("vm_mem_limit", "512M")
```
yanan-fu: `vm_mem_limit` is the max value,  and `vm_mem_minimum` is minimum value.

huangyum: @yanan-fu Yeah, my mistake. Updated.

huangyum:
@jingzhao84 Would you please help review, thanks.

vivianQizhu:
@huangyum Could you help explain why increase 256 to 512M? Thanks

vivianQizhu:
And, would you like to include your analyse in the bz in the commit log: "lowest_mem_per_vm should be gotten from param "vm_mem_minimum" other than "lowest_mem", which is defined nowhere"? It helps later developing if we make commit log as detailed as possible.

huangyum:
@vivianQizhu I tried to boot guest with 256M, after guest boot up, it hardly gave any response. It's too small for a guest, while 512M seems work fine.  And updated the commit message. Thanks.

vivianQizhu:
@huangyum Okay, but hard to assume that no one actually use that small memory, do we have any officially announced min memory supported?

balamuruhans:
@vivianQizhu yes, minimum memory in virtualized guest is 512M,
https://access.redhat.com/articles/rhel-kvm-limits
https://access.redhat.com/articles/906543

vivianQizhu:
@balamuruhans Greate, thanks for the information! Let's merge it.

PR #1880, opened 2018-12-28
virttest.guest_agent: Add function of guest-exec and guest-exec-status

xiagao:
ID: 1564829

### DIFF
--- a/virttest/guest_agent.py
+++ b/virttest/guest_agent.py
@@ -733,22 +733,20 @@ class QemuAgent(Monitor):
             return ret
         return -1
 
-    def _guest_file_operation(self, operation, **kwargs):
+    def _cmd_args_update(self, cmd, **kwargs):
         """
-        Basic guest file command function.
+        Update qga commands' args.
 
-        :param operation: open,write,read,flush,seek or flush.
-        :param kwargs: optional keyword arguments including but not limited
-               to below.
+        :param cmd: command to send.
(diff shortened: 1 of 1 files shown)


lijinlijin:
Acked-by: Li Jin lijin@redhat.com

virttest/guest_agent.py
```diff
@@ -822,5 +825,36 @@ def guest_file_seek(self, handle, offset, whence):
...
         :return: a dict with position and eof.
         """
-        return self._guest_file_operation("seek", handle=handle, offset=offset,
-                                          whence=whence)
+        cmd = "guest-file-seek"
+        return self._cmd_args_update(cmd, handle=handle, offset=offset,
+                                     whence=whence)
+
+    def guest_exec(self, path, arg=None, env=None, input_data=None,
+                   capture_output=None):
```
hereischen: `capture_output=False`

xiagao: No, if capture_output =None, guest-exec cmd will use default value(false)

xiagao: if capture_output =None, `_cmd_args_update() ` function will handle it to delete this key, then guest-exec cmd will use the default value.
such as:  {"execute": "guest-exec", "arguments": {"path": "echo", "arg": ["This", "is", "a", "test."]}}
this qga cmd will adopt the default `capture_output` value.

luckyh:
Thanks all.

PR #1886, opened 2019-01-03
Replace method of checking 'type' with 'isinstance' 

chloerh:


virttest/utils_gdb.py
```diff
@@ -389,7 +389,7 @@ def send_signal(self, signal_name):
         if type(signal_name) is str:
```
hereischen: what about this line?

chloerh: edited as well

hereischen:

PR #1893, opened 2019-01-07
utils_net: Add methods to get/set promisc mode and mtu size

PaulYuuu:
A little of testcases need to modify promiscuous mode or mtu size on
the interface, I think they should be supported by Interface(), so add 
them.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -390,52 +390,69 @@ class Interface(object):
                                     self.name, id(self))
 
     @warp_init_del
+    def set_iface_flag(self, flag, active=True):
+        # Get existing device flags
+        ifreq = struct.pack('16sh', self.name.encode(), 0)
+        flags = struct.unpack('16sh',
+                              fcntl.ioctl(sockfd, arch.SIOCGIFFLAGS, ifreq))[1]
+
+        # Set new flags
+        if active:
+            flags = flags | flag
+        else:
+            flags = flags & ~flag
+
+        ifreq = struct.pack('16sh', self.name.encode(), flags)
+        fcntl.ioctl(sockfd, arch.SIOCSIFFLAGS, ifreq)
+
+    @warp_init_del
+    def is_iface_flag_on(self, flag):
+        ifreq = struct.pack('16sh', self.name.encode(), 0)
+        flags = struct.unpack('16sh',
+                              fcntl.ioctl(sockfd, arch.SIOCGIFFLAGS, ifreq))[1]
+
+        if flags & flag:
+            return True
+        else:
+            return False
+
     def up(self):
         '''
         Bring up the bridge interface. Equivalent to ifconfig [iface] up.
         '''
+        self.set_iface_flag(arch.IFF_UP)
 
-        # Get existing device flags
-        ifreq = struct.pack('16sh', self.name.encode(), 0)
-        flags = struct.unpack('16sh',
-                              fcntl.ioctl(sockfd, arch.SIOCGIFFLAGS, ifreq))[1]
-
-        # Set new flags
-        flags = flags | arch.IFF_UP
-        ifreq = struct.pack('16sh', self.name.encode(), flags)
-        fcntl.ioctl(sockfd, arch.SIOCSIFFLAGS, ifreq)
-
-    @warp_init_del
     def down(self):
         '''
-        Bring up the bridge interface. Equivalent to ifconfig [iface] down.
+        Bring down the bridge interface. Equivalent to ifconfig [iface] down.
         '''
+        self.set_iface_flag(arch.IFF_UP, active=False)
 
-        # Get existing device flags
-        ifreq = struct.pack('16sh', self.name.encode(), 0)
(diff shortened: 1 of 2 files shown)


virttest/utils_net.py
```diff
@@ -437,6 +437,56 @@ def is_up(self):
...
+        ifreq = struct.pack('16sh', self.name.encode(), 0)
+        flags = struct.unpack('16sh',
+                              fcntl.ioctl(sockfd, arch.SIOCGIFFLAGS, ifreq))[1]
+
+        # Set new flags
+        if flags & arch.IFF_PROMISC:
+            return True
+        else:
+            return False
+
```
vivianQizhu: This seems pretty much alike with the function `up` `down` `is_up`. Can we reorganize them like:
```
def set_device_flag(self, flag='IFF_UP', on=True):
    ifreq = struct.pack('16sh', self.name.encode(), 0)
    flags = struct.unpack('16sh',
                          fcntl.ioctl(sockfd, arch.SIOCGIFFLAGS, ifreq))[1]
    if on:
        flags = flags | getattr(arch, flag)
    else:
        flags = flags & ~getattr(arch, flag)
    ifreq = struct.pack('16sh', self.name.encode(), flags)
    fcntl.ioctl(sockfd, arch.SIOCSIFFLAGS, ifreq)

def is_device_flag_on(self, flag=IFF_UP):
    ifreq = struct.pack('16sh', self.name.encode(), 0)
    flags = struct.unpack('16sh',
                          fcntl.ioctl(sockfd, arch.SIOCGIFFLAGS, ifreq))[1]

    if flags & getattr(arch, flag):
        return True
    else:
        return False

def up(self):
    self.set_device_flag()

def down(self):
    self.set_device_flag(on=False)

def promisc_on(self)
    self.set_device_flag(IFF_PROMISC)

def promisc_off(self)
    self.set_device_flag(IFF_PROMISC, on=False)
```

vivianQizhu:
@xiywang Request for your review, thanks.

virttest/utils_net.py
xiywang: Agree with the opinion above.

virttest/utils_net.py
```diff
@@ -518,6 +568,29 @@ def set_netmask(self, netmask):
...
+                            socket.AF_INET, b'\x00' * 14)
+        try:
+            res = fcntl.ioctl(sockfd, arch.SIOCGIFMTU, ifreq)
+        except IOError:
+            return None
+
+        return struct.unpack('16sH14s', res)[1]
+
+    @warp_init_del
+    def set_mtu(self, newmtu):
```
xiywang: In common cases, if we want to set mtu in guest, we should set mtu on host tap0 either.
It will be like:
interface.set_mtu()
process.system_output("ifconfig tap0 mtu <value>")

Is that how you use the function? Please enlighten me with the usage, thanks.

PaulYuuu: @vivianQizhu @xiywang Thanks for your review.
If we want to set the mtu size of tap0, it will be like `Interface(tap0).set_mtu(1500)`
And I will update the codes according to your suggestions above.

vivianQizhu:
Code looks good to me and test passed, just one concern about mtu, I am not sure about its boundary:
```
>>> utils_net.Interface('virbr0').set_mtu(2000)
  File "/usr/lib/python2.7/site-packages/avocado_plugins_vt-69.0-py2.7.egg/virttest/utils_net.py", line 373, in new_func
    return func(*args, **argkw)
  File "/usr/lib/python2.7/site-packages/avocado_plugins_vt-69.0-py2.7.egg/virttest/utils_net.py", line 556, in set_mtu
    fcntl.ioctl(sockfd, arch.SIOCSIFMTU, ifreq)
IOError: [Errno 22] Invalid argument
```
```utils_net.Interface('enp0s29f0u2').set_mtu(3000)``` works
Seems it differs from device, anyway the OS will raise Exception when the value is invalid, so my concern is do we need handle those Exceptions?  Maybe we correct the value automatically or return something to let user know that their behavior is ambitious?

**Comment by vivianQizhu** [comment]
Or certainly we can let it be and let users handle the Exception by themselves, let me know your thought, thanks.

**Comment by PaulYuuu** [comment]
@vivianQizhu For different nic devices, their Maximum mtu values are different.
```python
IOError: [Errno 22] Invalid argument
```
This error is matches the system report when we set an invalid MTU size.
```bash
$ ip link set dev virbr0 mtu 2000
RTNETLINK answers: Invalid argument
$ ifconfig virbr0 mtu 2000
SIOCSIFMTU: Invalid argument
```
Users who want to use this func need to know the maximum value of their device support.
I can raise the Exception like the follows:
```python
Set MTU failed: [Errno 22] Invalid argument
```
or
```python
raise OSError("(SIOCSIFMTU: %s)" % e.strerror)
```
But I think they are too similar, so we don't need raise a clear Exception.

vivianQizhu:
@PaulYuuu That's okay for me, thanks.

PR #1894, opened 2019-01-07
Add vm name to vm_ip_dict in ping_vm() after migration

fangge1212:
Vm name may change after migration if --dname is used. So we also
need to add vm name to vm_ip_dict after migration, or ping_vm()
will fail after migraiton.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1240,6 +1240,7 @@ class MigrationTest(object):
             server_ip = params.get("server_ip")
             src_uri = "qemu:///system"
             vm.connect_uri = uri
+            vm_ip[vm.name] = vm.get_address()
             server_pwd = params.get("server_pwd")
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1240,6 +1240,7 @@ def ping_vm(self, vm, test, params, uri=None, ping_count=10,
             server_ip = params.get("server_ip")
             src_uri = "qemu:///system"
             vm.connect_uri = uri
+            vm_ip[vm.name] = vm.get_address()
```
chunfuwen: move vm_ip[vm.name] to ahead 1238

fangge1212: After checked the code, I found we can't move vm_ip[vm.name] outside if-else, because in else, it needs to start vm if vm is not up, then get vm address

PR #1896, opened 2019-01-07
utils_netperf: Specify a platform type to compile netperf

PaulYuuu:
Sometimes netperf can't guess the correct platform to compile, so
force specify one to compile.

error info: "configure: error: cannot guess build type; you must specify one"

ID: 1663945


PaulYuuu:
I think `--build=` is a required option when we compiling netperf, and usually we compile it on localhost, so I add it to `setup_cmd`

PR #1898, opened 2019-01-07
utils_misc: check for all acceptable qemu binary name

balamuruhans:
let `get_qemu_binary()` checks possible qemu binary name
instead of proceeding with `kvm`.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2246,12 +2246,19 @@ def get_qemu_binary(params):
     if not os.path.isfile(qemu_binary_path):
         logging.debug('Could not find params qemu in %s, searching the '
                       'host PATH for one to use', qemu_binary_path)
-        try:
-            qemu_binary = utils_path.find_command('qemu-kvm')
-            logging.debug('Found %s', qemu_binary)
-        except utils_path.CmdNotFoundError:
-            qemu_binary = utils_path.find_command('kvm')
-            logging.debug('Found %s', qemu_binary)
+        QEMU_BIN_NAMES = ['qemu-kvm', 'qemu-system-%s' % (ARCH),
+                          'qemu-system-ppc64', 'qemu-system-x86',
+                          'qemu_system', 'kvm']
+        for qemu_bin in QEMU_BIN_NAMES:
+            try:
+                qemu_binary = utils_path.find_command(qemu_bin)
+                logging.debug('Found %s', qemu_binary)
+                break
+            except utils_path.CmdNotFoundError:
+                continue
+        else:
+            raise exceptions.TestError("qemu binary names %s not found in "
+                                       "system" % ' '.join(QEMU_BIN_NAMES))
     else:
         library_path = os.path.join(
             _get_backend_dir(params), 'install_root', 'lib')


virttest/utils_misc.py
```diff
@@ -2240,18 +2240,25 @@ def get_qemu_binary(params):
...
         logging.debug('Could not find params qemu in %s, searching the '
                       'host PATH for one to use', qemu_binary_path)
-        try:
-            qemu_binary = utils_path.find_command('qemu-kvm')
-            logging.debug('Found %s', qemu_binary)
-        except utils_path.CmdNotFoundError:
-            qemu_binary = utils_path.find_command('kvm')
-            logging.debug('Found %s', qemu_binary)
+        QEMU_BIN_NAMES = ['kvm', 'qemu-kvm', 'qemu-system-ppc64',
+                          'qemu_system', 'qemu-system-x86']
```
vivianQizhu: Though it already extends a lot, do we consider also include other platform upstream build? Like check 'qemu-system-${arch}'(should be `qemu-system-x86_64` instead of `qemu-system-x86`?)

virttest/utils_misc.py
```diff
@@ -2240,18 +2240,26 @@ def get_qemu_binary(params):
...
                       'host PATH for one to use', qemu_binary_path)
-        try:
-            qemu_binary = utils_path.find_command('qemu-kvm')
-            logging.debug('Found %s', qemu_binary)
-        except utils_path.CmdNotFoundError:
-            qemu_binary = utils_path.find_command('kvm')
-            logging.debug('Found %s', qemu_binary)
+        QEMU_BIN_NAMES = ['kvm', 'qemu-kvm', 'qemu-system-ppc64',
+                          'qemu_system', 'qemu-system-x86',
+                          'qemu-system-%s' % (ARCH)]
```
vivianQizhu: I suppose 'kvm' should be the last one to try, so at the end of the list.

balamuruhans: `qemu-kvm` in rhel based systems and `kvm` in ubuntu based systems executes qemu-system-$ARCH with kvm enabled,
```shell
#!/bin/sh
exec qemu-system-$ARCH -enable-kvm "$@"
```
so we should be use them right if it is available ?

vivianQizhu: @balamuruhans Yes, it will work well with current implementation, I just thought that qemu-kvm and other bin name have higher change to be used in our project. If put them behind `kvm`, every time it will go through kvm first(which waste a bit time). I think that original logic is better, it assumes that mostly we will have `qemu-kvm`, we check `kvm` only when there is no `qemu-kvm`.
Having said that, I have no strong opinion on this, it's functional good.

balamuruhans: I have changed the order based on your suggestion, please check if it is fine.

virttest/utils_misc.py
```diff
@@ -2240,18 +2240,26 @@ def get_qemu_binary(params):
...
+        for qemu_bin in QEMU_BIN_NAMES:
+            try:
+                qemu_binary = utils_path.find_command(qemu_bin)
+                logging.debug('Found %s', qemu_binary)
+                break
+            except utils_path.CmdNotFoundError:
+                continue
+        if not qemu_binary:
+            raise exceptions.TestError("qemu binary names %s not found in "
+                                       "system" % ' '.join(QEMU_BIN_NAMES))
```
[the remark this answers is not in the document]
balamuruhans: Thank you @lolyu, I have fixed it.

balamuruhans:
Thanks @vivianQizhu @lolyu for reviewing, let's merge it.

PR #1909, opened 2019-01-14
qemu_vm: add vsock device

vivianQizhu:
1. The cmd line: -device vhost-vsock-pci,id=vhost-vsock0,guest-cid=3
2. guest-cid is between '3' and '0xffffffff - 1'(unsigned 32-bit
 integer. 0-2 and 0xffffffff are reserved)
3. It is a virtio device so should be a pcie device

id: 1665845

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -9,8 +9,10 @@ import time
 import os
 import logging
 import fcntl
+import struct
 import re
 import random
+import errno
 
 from functools import partial
 
@@ -18,6 +20,7 @@ import aexpect
 from avocado.core import exceptions
 from avocado.utils import process
 from avocado.utils import crypto
+from avocado.utils import linux_modules
 
 import six
 from six.moves import xrange
@@ -1896,6 +1899,46 @@ class VM(virt_vm.BaseVM):
                               device_driver=device_driver,
                               pci_bus=pci_bus)
                 iov += 1
+
+        def get_cid(cid):
+            """ Get an unused guest cid from system """
+            while cid:
+                cid_c = struct.pack('L', cid)
+                try:
+                    fcntl.ioctl(
+                        vsock_fd, arch.VHOST_VSOCK_SET_GUEST_CID, cid_c)
+                except IOError as e:
+                    if e.errno == errno.EADDRINUSE:
+                        cid += 1
+                        continue
+                    else:
+                        raise e
+                else:
+                    return cid
+
+        # Add vsock device, cid 0-2 are reserved by system
+        vsocks = params.objects('vsocks')
+        if vsocks:
+            vsock_path = "/dev/vhost-vsock"
+            if not os.path.exists(vsock_path):
+                logging.info("vsock module was not loaded, loading it...")
+                if not linux_modules.load_module('vhost_vsock'):
+                    raise exceptions.TestError(
+                        "Failed on loading module vhost_vsock.")
+            vsock_fd = os.open(vsock_path, os.O_RDWR)
+            min_cid = 3
+            for vsock in vsocks:
+                guest_cid = get_cid(min_cid)
+                vsock_params = {"id": vsock, "guest-cid": guest_cid}
+                if '-mmio:' in params.get('machine_type'):
+                    dev_vsock = QDevice('vhost-vsock-device', vsock_params)
+                elif params.get('machine_type').startswith("s390"):
+                    dev_vsock = QDevice("vhost-vsock-ccw", vsock_params)
+                else:
+                    dev_vsock = QDevice('vhost-vsock-pci', vsock_params)
+                devices.insert(dev_vsock)
+                min_cid = guest_cid + 1
+            os.close(vsock_fd)
 
         # Add Memory devices
         add_memorys(devices, params)
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -694,3 +694,6 @@ inputs = ""
 # systems that require this extra setting.
 Host_Ubuntu.m18.u10:
     firewalld_dhcp_workaround = yes
+
+# Add vsock device
+# vsocks = vhost_vsock0
--- a/virttest/arch.py
+++ b/virttest/arch.py
@@ -38,6 +38,8 @@ if ARCH in ('ppc64', 'ppc64le'):
     NLMSG_ERROR = 2
     # From linux/socket.h
     AF_PACKET = 17
+    # From linux/vhost.h
+    VHOST_VSOCK_SET_GUEST_CID = 0x8008af60
 else:
     # From include/linux/sockios.h
     SIOCSIFHWADDR = 0x8924
@@ -72,6 +74,8 @@ else:
     NLMSG_ERROR = 2
     # From linux/socket.h
     AF_PACKET = 17
+    # From linux/vhost.h
+    VHOST_VSOCK_SET_GUEST_CID = 0x4008af60
 
 
 def get_kvm_module_list():
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -1839,7 +1839,7 @@ class QPCIEBus(QPCIBus):
         :param device: the QBaseDevice object
         :return: bool value for directly plug or not.
         """
-        pcie_devices = ["virtio-net-pci", "virtio-blk-pci",
+        pcie_devices = ["virtio-net-pci", "virtio-blk-pci", "vhost-vsock-pci"
                         "virtio-scsi-pci", "virtio-balloon-pci",
                         "virtio-serial-pci", "virtio-rng-pci",
                         "e1000e", "virtio-gpu-device", "qemu-xhci"]


virttest/qemu_vm.py
```diff
@@ -1900,6 +1900,15 @@ def sort_key(dev):
                               pci_bus=pci_bus)
                 iov += 1
 
+        # Add vsock device, cid 0-2 are reserved by system
+        default_cid = 3
+        for vsock in params.objects('vsock'):
+            vsock_params = params.object_params(vsock)
+            guest_cid = vsock_params.get("guest_cid", default_cid)
+            devices.insert(qdevices.QDevice(
+                "vhost-vsock-pci", {"id": vsock, "guest-cid": guest_cid}))
```
luckyh: Should be `vhost-vsock-device` for `aarch64:mmio` and `vhost-vsock-ccw` for `s390x`.

virttest/qemu_vm.py
```diff
@@ -1900,6 +1900,15 @@ def sort_key(dev):
                               pci_bus=pci_bus)
                 iov += 1
 
+        # Add vsock device, cid 0-2 are reserved by system
+        default_cid = 3
+        for vsock in params.objects('vsock'):
```
luckyh: Suggest the name `vsocks` if each vm can have more than one vsock.

virttest/qemu_vm.py
```diff
@@ -1900,6 +1900,15 @@ def sort_key(dev):
...
+            vsock_params = params.object_params(vsock)
+            guest_cid = vsock_params.get("guest_cid", default_cid)
+            devices.insert(qdevices.QDevice(
+                "vhost-vsock-pci", {"id": vsock, "guest-cid": guest_cid}))
+            default_cid += default_cid
```
luckyh: Seems libvirt has a more robust approach to allocate cid, here are some snippets, please have a look, thanks.
```c
int
qemuProcessOpenVhostVsock(virDomainVsockDefPtr vsock)
{
    qemuDomainVsockPrivatePtr priv = (qemuDomainVsockPrivatePtr)vsock->privateData;
    const char *vsock_path = "/dev/vhost-vsock";
    int fd;

    if ((fd = open(vsock_path, O_RDWR)) < 0) {
        virReportError(VIR_ERR_CONFIG_UNSUPPORTED,
                       "%s", _("unable to open vhost-vsock device"));
        return -1;
    }

    if (vsock->auto_cid == VIR_TRISTATE_BOOL_YES) {
        if (virVsockAcquireGuestCid(fd, &vsock->guest_cid) < 0)
            goto error;
    } else {
        if (virVsockSetGuestCid(fd, vsock->guest_cid) < 0)
            goto error;
    }

    priv->vhostfd = fd;
    return 0;

 error:
    VIR_FORCE_CLOSE(fd);
    return -1;
}

#if HAVE_DECL_VHOST_VSOCK_SET_GUEST_CID
static int
virVsockSetGuestCidQuiet(int fd,
                         unsigned int guest_cid)
{
    uint64_t val = guest_cid;

    return ioctl(fd, VHOST_VSOCK_SET_GUEST_CID, &val);
}

#else
static int
virVsockSetGuestCidQuiet(int fd ATTRIBUTE_UNUSED,
                         unsigned int guest_cid ATTRIBUTE_UNUSED)
{
    errno = ENOSYS;
    return -1;
}
#endif


/**
 * virVsockSetGuestCid:
 * @fd: file descriptor of a vsock interface
 * @guest_cid: guest CID to be set
 *
 * Wrapper for VHOST_VSOCK_SET_GUEST_CID ioctl.
 * Returns: 0 on success, -1 on error.
 */
int
virVsockSetGuestCid(int fd,
                    unsigned int guest_cid)
{
    if (virVsockSetGuestCidQuiet(fd, guest_cid) < 0) {
        virReportSystemError(errno, "%s",
                             _("failed to set guest cid"));
        return -1;
    }

    return 0;
}

#define VIR_VSOCK_GUEST_CID_MIN 3

/**
 * virVsockAcquireGuestCid:
 * @fd: file descriptor of a vsock interface
 * @guest_cid: where to store the guest CID
 *
 * Iterates over usable CIDs until a free one is found.
 * Returns: 0 on success, with the acquired CID stored in guest_cid
 *         -1 on error.
 */
int
virVsockAcquireGuestCid(int fd,
                        unsigned int *guest_cid)
{
    unsigned int cid = VIR_VSOCK_GUEST_CID_MIN;

    for (; virVsockSetGuestCidQuiet(fd, cid) < 0; cid++) {
        if (errno != EADDRINUSE) {
            virReportSystemError(errno, "%s",
                                 _("failed to acquire guest cid"));
            return -1;
        }
    }
    *guest_cid = cid;

    return 0;
}
```

virttest/qemu_vm.py
```diff
@@ -1900,6 +1900,15 @@ def sort_key(dev):
                               pci_bus=pci_bus)
                 iov += 1
 
+        # Add vsock device, cid 0-2 are reserved by system
+        default_cid = 3
+        for vsock in params.objects('vsock'):
+            vsock_params = params.object_params(vsock)
+            guest_cid = vsock_params.get("guest_cid", default_cid)
```
luckyh: Suggest using unified type for it (either int or str, I prefer int)

virttest/qemu_vm.py
vivianQizhu: Yes, will update.

virttest/qemu_vm.py
vivianQizhu: Sure, that's what I meant.

virttest/qemu_vm.py
vivianQizhu: Okay, I will try to translate it into python.

virttest/qemu_vm.py
vivianQizhu: Yep.

virttest/qemu_vm.py
vivianQizhu: @luckyh Before I go further, since the definition of `VHOST_VSOCK_SET_GUEST_CID` is `#define VHOST_VSOCK_SET_GUEST_CID  _IOW(VHOST_VIRTIO, 0x60, __u64)`, there is the linux macro`_IOW` to calculate the request of  `VHOST_VIRTIO`. So probably we need introduce something like https://python-ioctl.readthedocs.io/en/stable/linux.html#ioctl.linux.IOW to achieve it . Do we want that? Or any other suggestion? Thanks.

luckyh: Simply, we can directly record the target values since they are (usually) stable, see `virttest/arch.py`. And yes, the improvement is like something you mentioned, and I guess it will take more time..

vivianQizhu: @luckyh I don't see a way to directly get the value without _IOW support, it is not different from those values defined in virttest/arch.py, which are all directly defined values. And also there is no such macro defined like sockios.h, to allow use get and set something like if properties. It needs to increase the value each time user call it, so read and write action together(like 1++), so _IOW definitely required.

vivianQizhu: @luckyh Honestly I don't think it worth that for this feature, so I will just keep current implementation it require IOW support, although it needs user define the cid if there is multiple vms use vsock, but I guess we will never have such scenario, since we don't have this device by default and will only have few cases test with it.

luckyh: You can use the following C program to get the value.
```c
#include <stdio.h>
#include <linux/vhost.h>

int main(void) {
    printf("0x%x\n", VHOST_VSOCK_SET_GUEST_CID);
    return 0;
}
```
I have run on few platforms, just take a reference:
* x86_64 (kernel 3.10 & 4.20):  0x4008af60
* ppc64le (kernel 4.19): 0x8008af60

Anyway, it's okay for me if you do think it is not worth for now, but please leave a FIXME comment for that.

Thanks.

vivianQizhu: @luckyh That's not what we want to get, what you get is the value of `VHOST_VSOCK_SET_GUEST_CID`,  it's like the address of the function to set the CID, while what we want is the CID set by it, in libvirt it's `ioctl(fd, VHOST_VSOCK_SET_GUEST_CID, &val)`, so I think that's the `*__u64` from `_IOW(VHOST_VIRTIO, 0x60, __u64)`, which should supposed to be a int(or unsigned long long) we passed to it(>=3), and that's how libvirt works.
And also, that's should be increased, a stationary value would help nothing for multiple vsock CID, still need ++ them in the loop.
The correct way to get it in C is just copy the whole libvirt code. I don't think we should do that, if we do want it, it should be in python, so introduce IOW in python.

luckyh: I'm sorry I don't get your point, doesn't the libvirt's logic be:
```
Open vsock (qemuProcessOpenVhostVsock()):

1. check whether the vsock feature is supported (open(vsock_path, O_RDWR)) < 0)

2. try to occupy cid
2a. auto_cid scenario (if (vsock->auto_cid == VIR_TRISTATE_BOOL_YES))
    2a.1. pick a candidate cid, started by 3 (VIR_VSOCK_GUEST_CID_MIN)

    2a.2. test if the interest cid can be used (ioctl(fd, VHOST_VSOCK_SET_GUEST_CID, &val))
    2a.2a. cid is valid (virVsockSetGuestCidQuiet(fd, cid) >= 0)
        2a.2a.1. end the loop and use that one

    2a.2b. cid is invalid but not in-used (if (errno != EADDRINUSE))
        2a.2b.1. give up to find a valid cid and report error

    2a.2c. cid is invalid due to in-used
        2a.2c.1. try next one for lucks (cid++)

    2a.3. use that verified cid (*guest_cid = cid)

2b. user-defined cid scenario
    2b.1. set cid or report error (virVsockSetGuestCid())

3. assign cid to domain
```

So except `VHOST_VSOCK_SET_GUEST_CID` there is nothing requires `_IOW`, you can change the `ioctl` statement to `ioctl(fd, 0x4008af60, &val)` on x86_64 and it would work, the `&val` in this statement is the cid.

The differences between your current approach and libvirt's are:

1. the libvirt one would do cid verification for both auto_cid and user-defined scenario
2. for auto_cid scenario, the libvirt one would skip those candidates which are in-used (and your approach not)

vivianQizhu: @luckyh My fault, I misunderstood what you mean by 'target values' previously, and also I lead myself to the wrong direction. Certainly we can get the value by ioctl when we have the function address. I will update it. Thank you for bearing with me.

virttest/qemu_vm.py
```diff
@@ -1897,6 +1898,46 @@ def sort_key(dev):
...
+        vsock_path = "/dev/vhost-vsock"
+        try:
+            vsock_fd = open(vsock_path, 'r+')
+        except IOError as e:
+            if e.errno == 2:  # [Errno 2] No such file or directory
+                logging.info("vsock module was not loaded, loading it...")
+                process.system("modprobe vsock && modprobe vhost_vsock")
+                vsock_fd = open(vsock_path, 'r+')
+            else:
+                raise e
```
luckyh: Suggest the following approach:
```python
from avocado.utils import linux_module

min_cid = 3
vsock_path = "/dev/vhost-vsock"
vsock_fd = None
for vsock in params.objects('vsocks'):
    if vsock_fd is None:
        if not os.path.exists(vsock_path):
            if not linux_module.load_module('vhost_vsock'):
                test.error("...")
        vsock_fd = os.open(vsock_path, ...)
    ...
if vsock_fd is not None:
    os.close(vsock_fd)
```

vivianQizhu: @luckyh Yes, but I would like to load the module and get vsock_fd out of the for loop, it only needs to be executed once.

luckyh: Well, if user does not use vsock devices, `/dev/vhost-vsock` does not have to be opened, hence a module missing issue would not be raised in unsupported platforms.

vivianQizhu: @luckyh You are right. Updating.

virttest/qemu_vm.py
```diff
@@ -1897,6 +1900,48 @@ def sort_key(dev):
...
+                except IOError as e:
+                    if e.errno == errno.EADDRINUSE:
+                        cid += 1
+                        continue
+                    else:
+                        raise e
+                else:
+                    return cid
+
+        # Add vsock device, cid 0-2 are reserved by system
```
hereischen: What does `by system` mean?  by host OS or avcoado-vt?

vivianQizhu: @hereischen Operating System. I won't call avocado-vt a 'system'.

virttest/qemu_vm.py
```diff
@@ -1897,6 +1900,48 @@ def sort_key(dev):
...
+        vsocks = params.objects('vsocks')
+        if vsocks:
+            vsock_path = "/dev/vhost-vsock"
+            if not os.path.exists(vsock_path):
+                logging.info("vsock module was not loaded, loading it...")
+                loaded = linux_modules.load_module('vsock')
+                loaded &= linux_modules.load_module('vhost_vsock')
```
yanan-fu: The module `vhost_vosck` depends on three modules: 
```
depends:        vmw_vsock_virtio_transport_common,vhost,vsock
```
Since modprobe will load all the dependent modules, we can remove `load_module('vsock')` here.

virttest/qemu_vm.py
```diff
@@ -1897,6 +1900,48 @@ def sort_key(dev):
...
 
+        def get_cid(cid):
+            """ Get an unused guest cid from system """
+            while cid:
+                cid_c = struct.pack('L', cid)
+                try:
+                    fcntl.ioctl(
+                        vsock_fd, arch.VHOST_VSOCK_SET_GUEST_CID, cid_c)
+                except IOError as e:
+                    if e.errno == errno.EADDRINUSE:
```
yanan-fu: I prefer to add a comment here, to explain the given cid already in use. Not must, it is up to you.

virttest/qemu_vm.py
```diff
@@ -1897,6 +1900,48 @@ def sort_key(dev):
...
+                loaded &= linux_modules.load_module('vhost_vsock')
+                if not loaded:
+                    raise exceptions.TestError(
+                        "Failed on loading module vsock or vhost_vsock.")
+            vsock_fd = os.open(vsock_path, os.O_RDWR)
+            min_cid = 3
+            for vsock in vsocks:
+                guest_cid = get_cid(min_cid)
+                vsock_params = {"id": vsock, "guest-cid": guest_cid}
+                if '-mmio:' in params.get('machine_type'):
```
yanan-fu: For both `arm64-mmio` and `riscv64-mmio` ?

virttest/qemu_vm.py
vivianQizhu: @yanan-fu I think this process should be silence instead of verbose, since this is generated automatically, not set by user, and user actually do not need care about the cid.

virttest/qemu_vm.py
vivianQizhu: @yanan-fu I think so, all mmio arch use `-device` instead of `-pci` suffix, that's all same with other virtio device handling. e.g. `virtio-net-device`

virttest/qemu_vm.py
vivianQizhu: @yanan-fu Updated, please help review, thanks.

yanan-fu:
LGTM

vivianQizhu:
Thanks all.

PR #1910, opened 2019-01-15
Update vm name first when create_serial_console after migration

fangge1212:
When create_serial_console after migration, vm name should be updated
first in case --dname is specified for migration.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2033,6 +2033,16 @@ class VM(virt_vm.BaseVM):
         # Since dest_uri could be None, checking it is necessary.
         if result.exit_status == 0 and dest_uri:
             self.connect_uri = dest_uri
+
+        # Set vm name in case --dname is specified.
+        migrate_options = ""
+        if option:
+            migrate_options = str(option)
+        if extra:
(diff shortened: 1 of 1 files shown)


fangge1212:
@balamuruhans Could you please help me review the pr? Thanks.

virttest/libvirt_vm.py
```diff
@@ -2033,6 +2033,16 @@ def migrate(self, dest_uri="", option="--live --timeout 60", extra="",
         # Since dest_uri could be None, checking it is necessary.
         if result.exit_status == 0 and dest_uri:
             self.connect_uri = dest_uri
+
+        # Set vm name in case --dname is specified.
+        migrate_options = ""
+        if option:
+            migrate_options = str(option)
+        if extra:
+            migrate_options += " %s" % str(extra)
```
balamuruhans: `str` is not necessary in this line.

balamuruhans: `%s` does the typecast, so we don't have to do twice.

fangge1212: Updated. Thanks.

balamuruhans:
Thanks @fangge1212, let's merge it.

PR #1912, opened 2019-01-15
Allow for absolute paths to additional virtual tests

pevogam:
Some test suites would prefer to place their tests outside of and
not relative to the default folder with virtual tests.

### DIFF
--- a/avocado_vt/test.py
+++ b/avocado_vt/test.py
@@ -330,7 +330,8 @@ class VirtTest(test.Test):
 
         other_subtests_dirs = params.get("other_tests_dirs", "")
         for d in other_subtests_dirs.split():
-            d = os.path.join(*d.split("/"))
+            # If d starts with a "/" an absolute path will be assumed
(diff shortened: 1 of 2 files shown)


pevogam:
This PR is performing a fairly small change but I think reviews might be taking longer due to some better approach. Does anyone have a better suggestion on how to discover additional tests which are not placed in a location relative to the default virt test folder?

pevogam:
As this PR is more than 8 months old right now I will be really thankful if somebody helps @clebergnu and gives it a review too.

pevogam:
Hi @ldoktor, are you willing to give your review/approval to this PR? Thanks a lot for being helpful with reviews. We already have one approval here and the change is very minimal but gives a lot of freedom on our side.

pevogam:
Hi @ldoktor, I just pushed the same change applied to the legacy `virt.py` file as well. I hope this addresses your issue.

pevogam:
Thanks a lot both for the quick feedback!

PR #1914, opened 2019-01-15
Add various parameter manipulation functionality to the param utils

pevogam:
This includes custom key checking, list and dictionary parameter
retrieval, advanced object parameter manipulation, etc.

The only additional change other than pure copying from our downstream
code was done in order to adapt to object method access rather than module
function one:

index 40b21f63..b61e5864 100644
@@ -139,21 +139,21 @@ class Params(IterableUserDict):
         :returns: empty dict if if the key is not in the parameters, a dict
                   or an ordered dictionary if the item order is important
         :rtype: {str: str}

         This uses :py:meth:`get_list` to convert the list entries to dict.
         """
         if need_order:
             result = OrderedDict()
         else:
             result = dict()
-        for entry in get_list(self, key, default, delimiter):
+        for entry in self.get_list(key, default, delimiter):
             index = entry.find('=')
             if index == -1:
                 raise ValueError('failed to find "=" in "{0}" (value for {1})'
                                  .format(entry, key))
             result[entry[:index]] = entry[index+1:]
         return result

     def get_multi_param(self, name):
         """
         Get multiple parameters by appending a number

### DIFF
--- a/selftests/unit/test_utils_params.py
+++ b/selftests/unit/test_utils_params.py
@@ -3,6 +3,7 @@
 import unittest
 import os
 import sys
+from collections import OrderedDict
 
 # simple magic for using scripts within a source tree
 basedir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
@@ -85,6 +86,64 @@ class TestParams(unittest.TestCase):
 
     def testGetItem(self):
         self.assertEqual(self.params['image_size'], "10G")
+
+    def testGetBoolean(self):
+        self.params["foo1"] = "yes"
+        self.params["foo2"] = "no"
+        self.params["foo3"] = "on"
+        self.params["foo4"] = "off"
+        self.params["foo5"] = "true"
+        self.params["foo6"] = "false"
+        self.assertEqual(True, self.params.get_boolean('foo1'))
+        self.assertEqual(False, self.params.get_boolean('foo2'))
+        self.assertEqual(True, self.params.get_boolean('foo3'))
+        self.assertEqual(False, self.params.get_boolean('foo4'))
+        self.assertEqual(True, self.params.get_boolean('foo5'))
+        self.assertEqual(False, self.params.get_boolean('foo6'))
+        self.assertEqual(False, self.params.get_boolean('notgiven'))
+
+    def testGetNumeric(self):
+        self.params["something"] = "7"
+        self.params["foobar"] = 11
+        self.params["barsome"] = 13.0
+        self.assertEqual(7, self.params.get_numeric('something'))
+        self.assertEqual(11, self.params.get_numeric('foobar'))
+        self.assertEqual(11.0, self.params.get_numeric('foobar'), float)
+        self.assertEqual(13, self.params.get_numeric('barsome'))
+        self.assertEqual(13.0, self.params.get_numeric('barsome'), float)
+        self.assertEqual(17, self.params.get_numeric('joke', 17))
+        self.assertEqual(17.0, self.params.get_numeric('joke', 17), float)
+
+    def testGetList(self):
+        self.params["primes"] = "7 11 13 17"
+        self.params["dashed"] = "7-11-13"
+        self.assertEqual(["7", "11", "13", "17"], self.params.get_list('primes'))
+        self.assertEqual([7, 11, 13, 17], self.params.get_list('primes', "1 2 3", " ", int))
+        self.assertEqual([1, 2, 3], self.params.get_list('missing', "1 2 3", " ", int))
+        self.assertEqual([7, 11, 13], self.params.get_list('dashed', "1 2 3", "-", int))
+
+    def testGetDict(self):
+        self.params["dummy"] = "name1=value1 name2=value2"
+        self.assertEqual({"name1": "value1", "name2": "value2"}, self.params.get_dict('dummy'))
+        result_dict = self.params.get_dict('dummy', need_order=True)
+        right_dict, wrong_dict = OrderedDict(), OrderedDict()
+        right_dict["name1"] = "value1"
+        right_dict["name2"] = "value2"
+        wrong_dict["name2"] = "value2"
+        wrong_dict["name1"] = "value1"
+        self.assertEqual(right_dict, result_dict)
+        self.assertNotEqual(wrong_dict, result_dict)
+
+    def dropDictInternals(self):
+        self.params["a"] = "7"
+        self.params["b"] = "11"
+        self.params["_b"] = "13"
+        pruned = self.params.drop_dict_internals()
+        self.assertIn("a", pruned.keys())
+        self.assertEqual(pruned["a"], self.params["a"])
+        self.assertIn("b", pruned.keys())
+        self.assertEqual(pruned["b"], self.params["b"])
+        self.assertNotIn("_b", pruned.keys())
 
 
 if __name__ == "__main__":
--- a/virttest/utils_params.py
+++ b/virttest/utils_params.py
@@ -3,6 +3,7 @@ try:
     from collections import UserDict as IterableUserDict
 except ImportError:
     from UserDict import IterableUserDict
+from collections import OrderedDict
 
 from avocado.core import exceptions
 
@@ -104,3 +105,91 @@ class Params(IterableUserDict):
             if self.get(key):
                 new_dict[key] = self.get(key)
         return new_dict
+
+    def get_boolean(self, key, default=False):
+        """
+        Check if a config option is set to a default affirmation or not.
+
+        :param key: config option key
+        :type key: str
+        :param bool default: whether to assume "yes" or "no" if param
+                             does not exist (defaults to False/"no").
+        :return whether option is set to 'yes'
+        :rtype: bool
+        """
+        value = self.get(key, "yes" if default else "no")
+        if value in ("yes", "on", "true"):
+            return True
+        if value in ("no", "off", "false"):
+            return False
+        raise ValueError("Cannot get boolean parameter value for %s: %s", key, value)
+
+    def get_numeric(self, key, default=0, target_type=int):
+        """
+        Get numeric value converting to integer if necessary.
+
+        :param str key: config option key
+        :param int default: default numeric value
+        :param type target_type: numeric type to return like int or float
+        :return numerical type `target_type` converted parameter value
+        :rtype: int or float
+        """
+        return target_type(self.get(key, default))
+
+    def get_list(self, key, default="", delimiter=' ', target_type=str):
+        """
+        Get a parameter value that is a character delimited list.
+
+        :param str key: parameter whose value is list
+        :param str default: default list value
+        :param str delimiter: character to split list items
+        :param type target_type: type of each item, default is string
+        :returns: empty list if if the key is not in the parameters
+        :rtype: [str]
+
+        .. note:: This an extension to the :py:func:`Params.objects` method as
+          it allows for delimiters other than white space.
+        .. seealso:: :py:func:`param_dict`
+        """
+        param_string = self.get(key, default)
+        if not param_string:
+            return []
+        else:
+            return [target_type(entry) for entry in param_string.split(delimiter)]
+
+    def get_dict(self, key, default="", delimiter=' ', need_order=False):
+        """
+        Get a param value that has the form 'name1=value1 name2=value2 ...'.
+
+        :param str key: parameter whose value is dict
+        :param str default: default dict value
+        :param str delimiter: character to split list items
+        :param bool need_order: whether to return an OrderedDict instead of
+                                a regular dict
+        :returns: empty dict if if the key is not in the parameters, a dict
+                  or an ordered dictionary if the item order is important
+        :rtype: {str: str}
+
+        This uses :py:meth:`get_list` to convert the list entries to dict.
+        """
+        if need_order:
+            result = OrderedDict()
+        else:
+            result = dict()
+        for entry in self.get_list(key, default, delimiter):
+            index = entry.find('=')
+            if index == -1:
+                raise ValueError('failed to find "=" in "{0}" (value for {1})'
+                                 .format(entry, key))
+            result[entry[:index]] = entry[index+1:]
+        return result
+
+    def drop_dict_internals(self):
+        """
+        Drop internal keys which are not of our concern and
+        return the modified params.
+
+        :returns: parameters without the internal keys
+        :rtype: {str, str}
+        """
+        return Params({key: value for key, value in self.items() if not key.startswith("_")})


pevogam:
@luckyh I will need a few days before I can provide you with sufficiently detailed comments on this PR since it will take time to extract proper use cases. Thanks!

pevogam:
The five functions towards the end of the module are helping me achieve a final mix of parameters where we do one parsing of guest variants for each of multiple vm objects (including "only" and "no" restrictions) and then mix the final dictionaries from each vm object together in one test. In this way we can perform tests involving multiple vms with different guest OS, emulated hardware, and other variants. We choose one "default" or "base" vm whose parameters are also allowed to be general (no `_vm` suffix) but contain the configuration for each vm in the final parameters. Is there a better way you could suggest in doing this?

I can share more details about the above methods and how they play together later on and drop them for now in order for the rest of the commits to be properly reviewed if you would prefer so. This could become more complicated topic which is why we could also do this.

luckyh:
> The five functions towards the end of the module are helping me achieve a final mix of parameters where we do one parsing of guest variants for each of multiple vm objects (including "only" and "no" restrictions) and then mix the final dictionaries from each vm object together in one test. In this way we can perform tests involving multiple vms with different guest OS, emulated hardware, and other variants. We choose one "default" or "base" vm whose parameters are also allowed to be general (no `_vm` suffix) but contain the configuration for each vm in the final parameters. Is there a better way you could suggest in doing this?

ok, got the point. Currently the VT framework does make the trouble for developing the tests running with heterogeneous VMs, because some of the variants (e.g. guest's distribution, devices..) are pinned for every single VM, one cannot define VMs in different settings by using cfg files. We can only achieve that by handling the definition and the creation of VM in our test scripts, hence, the five functions are introduced to assist us.

Well, I think we have no better choice than this unless we have another schema to define VMs in cfg files. Hi @clebergnu @ldoktor, any good ideas?

> I can share more details about the above methods and how they play together later on and drop them for now in order for the rest of the commits to be properly reviewed if you would prefer so. This could become more complicated topic which is why we could also do this.

ok, let's separate them into two PRs.

Thanks.

pevogam:
The variant separation by objects is already possible in our test suite and we run tests with arbitrary guest vms (in terms of any guest os or driver varaint) using the above methods and a few more. The above methods could look a bit like trickery to bend the virt-test architecture into allowing this but it works and tests where multiple networked vms with heterogeneous variants can validate multiple components on each machine. A few examples (from our own use cases) I can give include an android device connecting to a linux server through VPN, firewall or proxy tests where a lighter linux machine (RIP Linux) is connected as a client behind a server machine which is where most of the real life testing happens, etc. We do have additional plugin that we want to introduce to avocado later on but this is getting out of the scope of this review and we can open a separate topic about it later on.

Should I split the pull request myself? The one condition I would like to keep is having first commit for each pull request preserving the code before the review so that we preserve the integrity of the code when moving between repos. Any later commit can perform one or all changes from the review.

ldoktor:
Hello guys, I don't really have time to work properly look at this, but I know @Andrei-Stepanov and @rdkdd are running heterogenous systems in spice tests, maybe they would have some ideas to share.

pevogam:
Since as I expected this discussion will grow and take much longer I suggest to split this PR into the part that is adding more basic extensions of the `Params` object and the part containing the last few functions to be considered further and possibly added later.

selftests/unit/test_utils_params.py
```diff
@@ -86,6 +86,96 @@ def testGetItemMissing(self):
...
+        self.assertEqual(0, len(self.params.get_multi_param("something")))
+
+    def testMultiParamOnlyOne(self):
+        self.params['something'] = 'foobar'
+
+        found = self.params.get_multi_param("something")
+        self.assertEqual(1, len(found))
+        self.assertEqual('foobar', found[0])
+
+    def testMultiParamMany(self):
```
ldoktor: I remember you mentioned that you want to remove this method as `object_params` should suffice. If you want to keep it, please consider the name `get_range` or `get_range_value` as it'd be shorter and IMO fits well (we get range of named items).

Also if you intend this for serialization purposes only, I'd perhaps suggest to implement `iter_range` instead of `get` as for serialization we might not need the list. (and getting list from generator is simple)

selftests/unit/test_utils_params.py
```diff
@@ -86,6 +86,96 @@ def testGetItemMissing(self):
...
+        self.params['something'] = 'foobar'
+        self.params['something2'] = 'foobar2'
+        self.params['something3'] = 'foobar3'
+
+        found = self.params.get_multi_param("something")
+        self.assertEqual(3, len(found))
+        self.assertEqual('foobar', found[0])
+        self.assertEqual('foobar2', found[1])
+        self.assertEqual('foobar3', found[2])
```
ldoktor: note you can compare `self.assertEqual(["foobar", "foobar2", "foobar3"], found)`

selftests/unit/test_utils_params.py
```diff
@@ -86,6 +86,96 @@ def testGetItemMissing(self):
     def testGetItem(self):
         self.assertEqual(self.params['image_size'], "10G")
 
+    def testCfgEnabled(self):
```
ldoktor: I like this method

selftests/unit/test_utils_params.py
```diff
@@ -86,6 +86,96 @@ def testGetItemMissing(self):
...
+        self.params['something2'] = 'foobar2'
+        self.params['something3'] = 'foobar3'
+
+        found = self.params.get_multi_param("something")
+        self.assertEqual(3, len(found))
+        self.assertEqual('foobar', found[0])
+        self.assertEqual('foobar2', found[1])
+        self.assertEqual('foobar3', found[2])
+
+    def testIsObjectSpecific(self):
```
ldoktor: Could be useful.

selftests/unit/test_utils_params.py
```diff
@@ -86,6 +86,96 @@ def testGetItemMissing(self):
...
+        self.assertTrue(is_object_specific("god_vm2", ["vm1", "vm2"]))
+        self.assertFalse(is_object_specific("god_vm2", ["nic1", "nic2"]))
+        self.assertFalse(is_object_specific("genie", ["vm1", "vm2"]))
+        self.assertFalse(is_object_specific("god", ["vm1", "vm2"]))
+        self.assertFalse(is_object_specific("wizard", ["vm1", "vm2"]))
+        self.assertFalse(is_object_specific("wizard", ["wizard"]))
+
+    def testObjectParams(self):
+        vm_params = utils_params.Params({"name_vm1": "josh", "name_vm2": "jean", "name": "jarjar", "surname": "jura"})
+        params = object_params(vm_params, "vm1", ["vm1", "vm2"])
```
ldoktor: on many places you use `utils_params` methods without using `utils_params` module resulting in failures... (I mean object_params`, `is_object_specific` and such are not directly imported, but imported with `utils_params`...)

selftests/unit/test_utils_params.py
```diff
@@ -86,6 +86,96 @@ def testGetItemMissing(self):
...
+        for key in params.keys():
+            self.assertFalse(is_object_specific(key, ["vm1", "vm2"]))
```
ldoktor: So what is the purpose, to consume less memory? Is it worth the effort? (I mean you spent a lot of cycles to remove the entries that should not affect the current object.

selftests/unit/test_utils_params.py
```diff
@@ -86,6 +86,96 @@ def testGetItemMissing(self):
...
+
+    def testObjectifyParams(self):
+        params = utils_params.Params({"name_vm1": "josh", "name_vm2": "jean", "name": "jarjar", "surname": "jura"})
+        vm_params = objectify_params(params, "vm1", ["vm1", "vm2"])
+        self.assertTrue("name_vm1" in vm_params, "Parameters already specific to the object must be preserved")
+        self.assertEqual(vm_params["name_vm1"], params["name_vm1"], "Parameters already specific to the object must be preserved")
+        self.assertFalse("name_vm2" in vm_params, "Parameters specific to a different object must be pruned")
+        self.assertFalse("name" in vm_params, "Parameters not specific to any object must be pruned if there is specific alternative")
+        self.assertTrue("surname_vm1" in vm_params, "Parameters not specific to any object must become specific to the object")
+        self.assertFalse("surname" in vm_params, "Parameters not specific to any object must become specific to the object")
```
ldoktor: btw it's useful to list the expected and actual values (if not everything) instead just the message. Usually it's better to simply add a comment above the call, leave the default reporting and use the right methods like `assertIn`.

ldoktor:
I don't really have time to review this, but my recommendation would be to post at least simplified way you intend to use this. There are things I'd have to think about more (eg. the environment values, don't like that part much), but on first glance I don't see anything that should break Avocado-vt (which I was afraid of, that's why I found the time to briefly look at it).

Therefor, no nack from me but please find some reviewers (I already emailed @rdkdd and @Andrei-Stepanov who have experience with combining params of different vms).

ldoktor:
[avocado-vt issue/pr](https://trello.com/c/QK2esyNY/1490-avocado-vt-issue-pr)

selftests/unit/test_utils_params.py
pevogam: No, as @luckyh  pointed out, it can be replaced very easily with already existing functionality from `objects`.  I will also resolve this since the change where you commented is outdated.

selftests/unit/test_utils_params.py
pevogam: Good point, still less compatible with one-line versions that are easier to copy and paste like this should the number of unit tests grow. Also outdated though, please check the most recent unit tests from a few days ago.

selftests/unit/test_utils_params.py
pevogam: It was renamed to `get_boolean` after @luckyh's suggestion. Also updated unit tests.

selftests/unit/test_utils_params.py
pevogam: Hopefully, not modified in the push from some days ago and the same.

selftests/unit/test_utils_params.py
pevogam: Yes, old version again - however, you did get one or two instances more - thanks! I wish I could detect them but the unit tests are failing before reaching this stage.

selftests/unit/test_utils_params.py
pevogam: It is part of the magic we do in order to obtain a final set of parameters containing `_vmX` suffixes for each of the heterogeneous variant vms (thus making them coexist in the same set of parameters per test). This memory consumption will only affect code calling the method. It is also a subject of optimization and refactoring but this is a good start and we can even discuss the heterogeneous variants in more depth later on, gradually adding these.

selftests/unit/test_utils_params.py
pevogam: Great suggestion, added.

pevogam:
> I don't really have time to review this, but my recommendation would be to post at least simplified way you intend to use this. There are things I'd have to think about more (eg. the environment values, don't like that part much), but on first glance I don't see anything that should break Avocado-vt (which I was afraid of, that's why I found the time to briefly look at it).
> 
> Therefor, no nack from me but please find some reviewers (I already emailed @rdkdd and @Andrei-Stepanov who have experience with combining params of different vms).

I would still suggest breaking the commit into two because I am pretty sure additional discussion of this depth will take longer time than the more basic extensions this PR is adding (like boolean, numeric, and list types). As I mentioned above, these functions are still work quite much a form of magic and can thus definitely be improved in the direction of simplification (as long there are better suggestions, some years ago when I wrote it I tried to minimize complexity but couldn't go any simpler than this with the Cartesian configuration offered by autotest).

rdkdd:
Hi @pevogam, @ldoktor, @luckyh 
as @Andrei-Stepanov - develper of our multi-VM system - is not involved in https://github.com/spiceqa/tp-spice any more, I will take brief explanation:
AFAIK we can utilize two VMs - one serves as client, the other as guest. Using https://github.com/spiceqa/tp-spice/blob/master/spice/cfg/join.cfg config file we are able to have multiple options in `params` one for client and one for guest (with `_client` or `_guest` suffix). `join.cfg` file is included at the end of testcase configuration if desired (see https://github.com/spiceqa/tp-spice/blob/master/spice/cfg/tests-variants.cfg). If both config options are the same, suffix is dropped (e.g. `vga_client = qxl` and `vga_guest = qxl` becomes `vga = qxl`). You can test behaviour for example with `cartesian-config.py -(s)c tests.cfg`. See also corresponding commits for `suffix` and `join` keywords: https://github.com/avocado-framework/avocado-vt/commit/26aceb19147ce7db8c459ba1e6157d35781cfbbb and https://github.com/avocado-framework/avocado-vt/commit/105010594f2c9ab41dd6667445076478e41080d6
I see both keywords `join` and `suffix` are not documented. I think they should be mentioned in https://avocado-vt.readthedocs.io/en/latest/CartesianConfig.html

virttest/utils_params.py
```diff
@@ -104,3 +107,252 @@ def copy_from_keys(self, keys):
...
+
+        :param key: config option key
+        :type key: str
+        :param bool enabled_by_default: whether to assume "yes" or "no" if param
+                                        does not exist (defaults to False/"no").
+        :return whether option is set to 'yes'
+        :rtype: bool
+        """
+        if enabled_by_default:
+            return self.get(key, "yes") == "yes"
```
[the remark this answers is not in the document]
pevogam: This last statement is richer in supported value pairs but modifies the logic of the original function. As explained in the docstring, `enabled_by_default` is only meant to consider default "yes" or "no" value. A simple call like `params.enabled_by_default("test")` with `params["test"]="yes"` will result in the above code returning `False` which is not intended originally. I would suggest

```python
value = self.get(key, "yes" if enabled_by_default else "no")
if value in ("yes", "on", "true"):
    return True
if value in ("no", "off", "false"):
    return False
raise ValueError("Cannot get boolean parameter value for %s", key)
```
which preserves the original intent of the parameters.

pevogam:
Hi @rdkdd, these are all great hints. In fact, we faced similar issue and implemented some of the methods above, one of which does exactly that - it drops repeating parameters with different suffices. It is nice to see such keywords were added directly to the Cartesian config in the past so we will try to adapt our own use to these (since they are already upstreamed) and as a result document them (no documentation so far indeed) as well as contribute our own extensions as functionality we need but still could not find.

virttest/utils_params.py
```diff
@@ -104,3 +107,252 @@ def copy_from_keys(self, keys):
...
+
+            i = i+1
+            next_name = '{0}{1}'.format(name, i)
+
+        return items
+
+    def drop_dict_internals(self):
+        """
+        Drop internal keys which are not of our concern and
+        return the modified params.
```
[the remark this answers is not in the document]
pevogam: Removed - will be added in a separate PR after further integration with the methods already provided by the Cartesian config.

virttest/utils_params.py
```diff
@@ -104,3 +107,252 @@ def copy_from_keys(self, keys):
...
+    :type params: {str, str}
+    :param param_objects: the parameter objects to multiply with
+    :type param_objects: [str]
+    :returns: multiplied object-specific parameters
+    :rtype: Params object
+
+    .. note:: If a `PREFIX` environment variable is set, the multiplied
+        paramers will also be prefixed with its value. This is useful for
+        performing multiple test runs in parallel.
+    .. note:: Currently only implemented for vm objects.
```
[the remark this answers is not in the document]
pevogam: Removed - will be added in a separate PR after further integration with the methods already provided by the Cartesian config.

pevogam:
> Since as I expected this discussion will grow and take much longer I suggest to split this PR into the part that is adding more basic extensions of the `Params` object and the part containing the last few functions to be considered further and possibly added later.



> > I don't really have time to review this, but my recommendation would be to post at least simplified way you intend to use this. There are things I'd have to think about more (eg. the environment values, don't like that part much), but on first glance I don't see anything that should break Avocado-vt (which I was afraid of, that's why I found the time to briefly look at it).
> > Therefor, no nack from me but please find some reviewers (I already emailed @rdkdd and @Andrei-Stepanov who have experience with combining params of different vms).
> 
> I would still suggest breaking the commit into two because I am pretty sure additional discussion of this depth will take longer time than the more basic extensions this PR is adding (like boolean, numeric, and list types). As I mentioned above, these functions are still work quite much a form of magic and can thus definitely be improved in the direction of simplification (as long there are better suggestions, some years ago when I wrote it I tried to minimize complexity but couldn't go any simpler than this with the Cartesian configuration offered by autotest).

I pushed a commit removing all last 5 methods that sparked the above discussion. This leaves the PR *solely* with already reviewed methods extending the `Params` class. I am not squashing commits since I would like to preserve the commit history of the pre-merge branch (only few commits anyway). The 5 functions related to heterogeneous will be added in a separate PR once they are better integrated with the functionality already existing in the Cartesian config as @rdkdd pointed out.

pevogam:
I did a final push to make all checks pass. @luckyh All of your requested changes so far have been committed as well as all of @ldoktor's that I recall. The PR now contains only the few type extensions on the parameter class so I hope it is good to go but in any case I am ready to implement any other requested changes you might have.

virttest/utils_params.py
```diff
@@ -104,3 +105,94 @@ def copy_from_keys(self, keys):
...
+        """
+        Get numeric value converting to integer if necessary.
+
+        :param key: config option key
+        :type key: str
+        :param int default: default numeric value
+        :return integer converted numeric value
+        :rtype: int
+        """
+        return int(self.get(key, default))
```
luckyh: Better to let user be able to specify the type instead of always using int.

virttest/utils_params.py
```diff
@@ -104,3 +105,94 @@ def copy_from_keys(self, keys):
...
+    def drop_dict_internals(self):
+        """
+        Drop internal keys which are not of our concern and
+        return the modified params.
+
+        :returns: parameters without the internal keys
+        :rtype: {str, str}
+        """
+        internal_keys = [key for key in self.keys() if key.startswith("_")]
+        for key in internal_keys:
```
luckyh: Seems that we don't have to do the iteration twice time.

virttest/utils_params.py
```diff
@@ -104,3 +105,94 @@ def copy_from_keys(self, keys):
...
+            del self[key]
+        return self
```
luckyh: I think it would make more sense if here would return a copy of the original params dict, otherwise, we can just alter `self` and return nothing.

virttest/utils_params.py
pevogam: Good point, resolved in the next push.

virttest/utils_params.py
pevogam: Good point, resolved in the next push.

virttest/utils_params.py
pevogam: Yes, I combined this with the double loop refactoring above.

pevogam:
Hi @luckyh, all your recommendations have now been implemented. Also, I have improved the coverage of the unit tests for extra validation.

luckyh:
@pevogam thank you! so far the code looks good to me, but it is better to rearrange the commits (to me the five can be merged into one), this can be done by `git rebase -i HEAD~5`, thanks.

pevogam:
> @pevogam thank you! so far the code looks good to me, but it is better to rearrange the commits (to me the five can be merged into one), this can be done by `git rebase -i HEAD~5`, thanks.

@luckyh Is there a way to keep the current granularity and all the logical steps within? I am intentionally not squashing these commits since this loses resolution of the development process. If you think this is really necessary and have convincing reasons I wouldn't mind to hear them of course. The commits also contain migrated previously developed code (the first commit) where I want to be able to see all changes applied afterwards (the remaining four commits).

luckyh:
> @luckyh Is there a way to keep the current granularity and all the logical steps within? I am intentionally not squashing these commits since this loses resolution of the development process.

@pevogam well, in short, your development process should be a blackbox to the upstream project, you still can keep it in downstream, but for upstream, since the entire PR provides only one new feature it is not necessary to have so many commits, one is enough.

BTW, by rebasing you can restore the heterogeneous related code into another commit and then sent it later, as you've mentioned, so that won't be a burden.

Thanks.

pevogam:
Hi @luckyh, thanks for sharing your reasoning which is also generally important for my PR workflow. Migrating code from other repositories is not the typical PR so preserving the code's integrity in-between repo is important for me but I already have some downstream patches to remember the change. Not ideal but good enough for one-time migration.

I have now squashed all commits into one as requested.

pevogam:
> @pevogam ok, thank you! then LGTM, and let's wait for another ACK.

@ldoktor As you already left a review (are one of the 2 reviewers), do you mind taking a quick look as well? This PR is half a year old now and I would really like to get it merged, hopefully before Avocado VT 70.0 rolls out.

pevogam:
That's great news, we already have 2 approvals so I hope this PR is now eligible for merging.

selftests/unit/test_utils_params.py
```diff
@@ -86,6 +87,64 @@ def testGetItemMissing(self):
...
+        self.params["something"] = "7"
+        self.params["foobar"] = 11
+        self.params["barsome"] = 13.0
+        self.assertEqual(7, self.params.get_numeric('something'))
+        self.assertEqual(11, self.params.get_numeric('foobar'))
+        self.assertEqual(11.0, self.params.get_numeric('foobar'), float)
+        self.assertEqual(13, self.params.get_numeric('barsome'))
+        self.assertEqual(13.0, self.params.get_numeric('barsome'), float)
+        self.assertEqual(17, self.params.get_numeric('joke', 17))
+        self.assertEqual(17.0, self.params.get_numeric('joke', 17), float)
```
ldoktor: Well 17.0 equals 17, to be sure we'd have to add float test (eg. 13.1) or check for returned type. Anyway it's not critical...

virttest/utils_params.py
```diff
@@ -104,3 +105,91 @@ def copy_from_keys(self, keys):
...
+
+        .. note:: This an extension to the :py:func:`Params.objects` method as
+          it allows for delimiters other than white space.
+        .. seealso:: :py:func:`param_dict`
+        """
+        param_string = self.get(key, default)
+        if not param_string:
+            return []
+        else:
+            return [target_type(entry) for entry in param_string.split(delimiter)]
```
ldoktor: This might produce unwanted empty entries (eg. `"foo  bar baz" => ['foo', '', 'bar', 'baz']`). Is it intended?

virttest/utils_params.py
```diff
@@ -104,3 +105,91 @@ def copy_from_keys(self, keys):
...
+
+        :param str key: config option key
+        :param int default: default numeric value
+        :param type target_type: numeric type to return like int or float
+        :return numerical type `target_type` converted parameter value
+        :rtype: int or float
+        """
+        return target_type(self.get(key, default))
+
+    def get_list(self, key, default="", delimiter=' ', target_type=str):
```
ldoktor: `None` would be IMO better delimiter as it is default for `split` (and removes the unwanted empty spaces issue...)

virttest/utils_params.py
```diff
@@ -104,3 +105,91 @@ def copy_from_keys(self, keys):
...
+        if need_order:
+            result = OrderedDict()
+        else:
+            result = dict()
+        for entry in self.get_list(key, default, delimiter):
+            index = entry.find('=')
+            if index == -1:
+                raise ValueError('failed to find "=" in "{0}" (value for {1})'
+                                 .format(entry, key))
+            result[entry[:index]] = entry[index+1:]
```
ldoktor: This might be tricky with human-entered values as extra spaces might be present. Do you think `strip` would be useful here? (I don't think it's needed, just asking whether you considered this)

pevogam:
Hi @ldoktor, thanks for taking the time! Will take a look and add a PR with your comments soon as well.

selftests/unit/test_utils_params.py
pevogam: Does this look better?:

```python
    def testGetNumeric(self):
        self.params["something"] = "7"
        self.params["foobar"] = 11
        self.params["barsome"] = 13.17
        self.assertEqual(7, self.params.get_numeric('something'))
        self.assertEqual(7, self.params.get_numeric('something'), int)
        self.assertEqual(7.0, self.params.get_numeric('something'), float)
        self.assertEqual(11, self.params.get_numeric('foobar'))
        self.assertEqual(11, self.params.get_numeric('something'), int)
        self.assertEqual(11.0, self.params.get_numeric('foobar'), float)
        self.assertEqual(13, self.params.get_numeric('barsome'))
        self.assertEqual(13, self.params.get_numeric('barsome'), int)
        self.assertEqual(13.17, self.params.get_numeric('barsome'), float)
        self.assertEqual(17, self.params.get_numeric('joke', 17))
        self.assertEqual(17.13, self.params.get_numeric('joke', 17.13), float)
```

virttest/utils_params.py
pevogam: Good point, resolved in a future PR.

virttest/utils_params.py
pevogam: We could allow more flexibility for the user even though it might result in a worse format configuration.

virttest/utils_params.py
pevogam: Not really considered since this method follows a similar approach like the list method and is stricter when it comes to config formatting. However, we can also make it more flexible to use the same approach for both methods again.

-> Will be added in another smaller PR.

selftests/unit/test_utils_params.py
ldoktor: yep

pevogam: Great, will push the PR in about 2 hours after some minimal testing gives me green light.

PR #1737, opened 2018-08-31
Support virtio input devices for qemu_vm

luckyh:
Example:
```
  <configuration>
  inputs = mouse1 keyboard1
  input_dev_type_mouse1 = mouse
  input_dev_type_keyboard1 = keyboard
  input_dev_bus_type = virtio

  <qemu command line>
  -device virtio-mouse-pci,id=input_mouse1,bus=pci.0,addr=0x6 \
  -device virtio-keyboard-pci,id=input_keyboard1,bus=pci.0,addr=0x7
```
Signed-off-by: Xu Han <xuhan@redhat.com>

ID: 1456174

PR #1916, opened 2019-01-15
Add qemu command line option for nondefault (ps/2) mouse

pevogam:
In particular, add the usbdevice-mouse and usbdevice-tablet.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -867,6 +867,9 @@ class VM(virt_vm.BaseVM):
             else:
                 logging.warn("option '-%s' not supportted" % name)
 
+        def add_usbdevice(devices, mouse_type):
+            return " -usbdevice %s" % mouse_type
+
         def add_pcidevice(devices, host, params, device_driver="pci-assign",
                           pci_bus='pci.0'):
             if devices.has_device(device_driver):
@@ -1507,6 +1510,9 @@ class VM(virt_vm.BaseVM):
         devices.insert(StrDev('-S', cmdline="-S"))
         # Add the VM's name
         devices.insert(StrDev('vmname', cmdline=add_name(name)))
+        mouse_type = params.get('mouse_type', 'default')
+        if mouse_type != "default":
+            devices.insert(StrDev('mouse', cmdline=add_usbdevice(devices, mouse_type)))
 
         qemu_sandbox = params.get("qemu_sandbox")
         if qemu_sandbox == "on":


virttest/qemu_vm.py
```diff
@@ -1507,6 +1510,9 @@ def sort_key(dev):
         devices.insert(StrDev('-S', cmdline="-S"))
         # Add the VM's name
         devices.insert(StrDev('vmname', cmdline=add_name(name)))
+        mouse_type = params.get('mouse_type', 'default')
+        if mouse_type != "default":
+            devices.insert(StrDev('mouse', cmdline=add_usbdevice(devices, mouse_type)))
```
luckyh: Usually, we add usb input devices by setting the usb related configuration, for instance, we have the following snippet:
https://github.com/avocado-framework/avocado-vt/blob/3d2b0feb51a42d29c6dcd679f43c716303581ea6/shared/cfg/base.cfg#L98-L102
it will be converted to:
```
```
so supposes that you want usb mouse to be the default pointer device for all vt tests, the patch would be:
```patch
diff --git a/shared/cfg/base.cfg b/shared/cfg/base.cfg
index 87578f40..c8a4e617 100644
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -96,10 +96,10 @@ usb_type_usb1 = ich9-usb-ehci1
 usb_max_port = 6
 
 # USB device object names (whitespace separated)
-usb_devices = tablet1
+usb_devices = mouse1
 # USB device type, run following command to see device list on your host.
 # `qemu-kvm -device \? 2>&1 | grep "bus USB"`
-usb_type_tablet1 = usb-tablet
+usb_type_mouse1 = usb-mouse
 # USB Controller type which device uses.
 usb_controller = ehci
 usb_bus = usb1.0
```

Also, we have another schema to add input devices (see https://github.com/avocado-framework/avocado-vt/pull/1737 for details):
https://github.com/avocado-framework/avocado-vt/blob/16f745b3d02188964a22d191b2d6ab572f8de172/shared/cfg/base.cfg#L683-L688
and it will support types other than `virtio` in future.

pevogam: Sounds perfect, our `base.cfg` is slightly outdated from the time before this option was added. I will add it and run some testing, then probably resolve this conversation and this PR can be closed.

pevogam: Suggestion was perfect, thanks!

pevogam:
Closed as the patches here are outdated.

PR #1923, opened 2019-01-17
Remove custom item retrieval in the params object

pevogam:
Not only does this hide a proper stack trace of where the missing
parameter was detected but it also makes it impossible to use
other useful methods like pop() with a default and so on, all
because of the unnecessary custom ParamNotFound error which is
never caught in an exception clause at the abc collection on
a higher level.


pevogam:
Some more comments:
- This pull request is a suggestion that I am open about discussing. There might be better approaches than the one I had to do in order to make my own debugging easier and restore the functionality that was previously available.
- AFAIK such problems can be reproduced on python3 and not python2 due to changes in the collections modules.

PR #1927, opened 2019-01-18
unattended_install: To create image_dir before being used

luckyh:
Make image_dir always be created before being used.

ID: 1667362

### DIFF
--- a/virttest/tests/unattended_install.py
+++ b/virttest/tests/unattended_install.py
@@ -247,6 +247,11 @@ class UnattendedInstallConfig(object):
                 os.makedirs(os.path.dirname(self.floppy))
 
         self.image_path = os.path.dirname(self.kernel)
+        if not os.path.isdir(self.image_path):
+            # To prevent errors
+            if os.path.isfile(self.image_path):
+                os.unlink(self.image_path)
+            os.makedirs(self.image_path)
 
         # Content server params
         # lookup host ip address for first nic by interface name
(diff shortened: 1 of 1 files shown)


virttest/tests/unattended_install.py
```diff
@@ -1012,6 +1012,9 @@ def setup_nfs(self):
                  (self.nfs_server, self.nfs_dir, self.nfs_mount))
         process.run(m_cmd, verbose=DEBUG)
 
+        if not os.path.exists(self.image_path):
+            os.mkdir(self.image_path)
```
balamuruhans: we can have it in `__init__()` that avoids the redundant code in `setup_cdrom()`, `setup_url()` and here

luckyh: I was worry about making regressions so I didn't do that. But you're right, for code quality we should do so, thank you.

luckyh:
@balamuruhans modified per your suggestion, thanks.

CongLi:
Ack.

luckyh:
ok, thanks all.

virttest/tests/unattended_install.py
luckyh: Unfortunately, it does make a regression (see https://github.com/avocado-framework/avocado-vt/pull/1940), and I guess it was not the only one, so let's revert this approach and use the first one I sent, thanks.

PR #1929, opened 2019-01-21
qemu_vm: add addr option for balloon device

huangyum:
ID: 1645981

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -2305,6 +2305,9 @@ class VM(virt_vm.BaseVM):
             options["deflate-on-oom"] = deflate_on_oom
             guest_polling = balloon_params.get("balloon_opt_guest_polling")
             options["guest-stats-polling-interval"] = guest_polling
+            balloon_addr = balloon_params.get("balloon_addr")
+            if balloon_addr:
+                options["addr"] = balloon_addr
             add_balloon(devices, devid=balloon_devid, bus=balloon_bus,
                         use_old_format=use_ofmt, options=options)
 


virttest/qemu_vm.py
```diff
@@ -2305,6 +2305,9 @@ def sort_key(dev):
             options["deflate-on-oom"] = deflate_on_oom
             guest_polling = balloon_params.get("balloon_opt_guest_polling")
             options["guest-stats-polling-interval"] = guest_polling
+            balloon_addr = balloon_params.get("balloon_addr")
+            if balloon_addr:
+                options["addr"] = balloon_addr
```
luckyh: Looks it only works for PCI buses, however, wo should also consider about other bus types (for aarch64, s390x..)

vivianQizhu: That's really a bad workaround, but since we can not expect migration support been fixed recently, I will accept it, by requiring a comment to declare this is only a workaround for migration, do not recommend to use it other way.

vivianQizhu: @luckyh I guess other arch wont need this option being specified, address pattern should be always automatically generated, only for migration + hotplug it needs to be specified by user. Or better we define only x86 and ppc for such balloon case.

huangyum: @vivianQizhu If you were saying the referenced tp-qemu patch, I wouldn't argue for it.  But for this one, it implements the addr option for balloon device, just like other devices. I won't say it is a workaround.  Thanks.

yanan-fu: Agree with `balloon_pci_addr` is better, and no need to add comment here :)

vivianQizhu: In case of wrong usage, maybe filter the bus type like`if isinstance(balloon_bus, qdevices.QPCIBus) and balloon_addr:`, what do you think? @luckyh

luckyh: I am still not clear about the primary use case for introducing this feature. If it is used to workaround the address mismatch issue of migration, then I would give **NACK** for this, because obviously, we have a better solution: let VT allocate the address automatically and mark it up in qemu command line.

If it is supposed to let user be able to customize balloons' bus address, I would accept, but I do request it works for all the supported platforms.

Thanks.

luckyh: Btw, it's the time to stress the necessity of writing a reasonable commit message.

PaulYuuu: Actually,  on ppc platform, **hotplug nic device + migration** also hit this issue. It is necessary to specify pci bus and address when we hot-plug a device, and it needs to match the divece bus info of the dst guest if we want to migrate it.
Agree with @luckyh that let VT allocate the address automatically

huangyum: As to the supporting for other platforms,  I need do some research first, thanks for your reminder.

huangyum: @vivianQizhu @yanan-fu Thanks for your review.  I guess there might be some negative tests in the future which need setting addr manually.  It's not a must to set it manually, so I think it won't do any harm. As for the variant name, I will update when we have a final conclusion about this patch. Thanks.

vivianQizhu: @PaulYuuu That does match the test scenario. We have hotplug before migration scenario for almost all device test.

vivianQizhu:
@luckyh No matter how we solve the hotplug-before-migration issue, I guess these lines are needed. The user of it can be VT itself(might in migration part), ideally we have to get address of all devices of source vm, and then allocate the same address to destination vm. Cause the sequence of the devices can be different(introduced by hotplug or possibly other case), so generate them automatically(which rely on the sequence of the devices) with certainly lead to address mismatch issue. Any other suggestion?

vivianQizhu:
@luckyh @PaulYuuu And FYI, seems x86 works well with currently implementation, only ppc hit the address issue, not sure if it because x86 allow address mismatch condition or what.

virttest/qemu_vm.py
luckyh: @huangyum so it turns out that there are bugs in `vm.clone()`: it does not reserve the addresses correctly for initial devices, as @PaulYuuu  mentioned, the issue is not only against balloon tests, but also nics, blocks, .. so for this kind of issue we need a common solution.

huangyum:
@vivianQizhu For x86, seems the address happens to be the same, not allow address mismatch.

PaulYuuu:
> @luckyh @PaulYuuu And FYI, seems x86 works well with currently implementation, only ppc hit the address issue, not sure if it because x86 allow address mismatch condition or what.

I checked the log, ppc will uses "00.0" first to hotplug a device
--- info qtree ---
```
bus: pci.0
  type PCI
  dev: virtio-net-pci, id "hotadded"
    ioeventfd = true
    vectors = 4 (0x4)
    virtio-pci-bus-master-bug-migration = false
    disable-legacy = "off"
    disable-modern = false
    migrate-extra = true
    modern-pio-notify = false
    x-disable-pcie = false
    page-per-vq = false
    x-ignore-backend-features = false
    ats = false
    x-pcie-deverr-init = true
    x-pcie-lnkctl-init = true
    x-pcie-pm-init = true
    addr = 00.0
```
Maybe X86 won't? Need your confirmation @huangyum , many thanks!

virttest/qemu_vm.py
huangyum: @luckyh In balloon cases, it does not only not reserve the address but also the balloon device itself. I see some workaround lines in balloon_hotplug.py.
'''
 84         # temporary workaround for migration
 85         vm.params["balloon"] = "balloon%d" % idx
 86         vm.params["balloon_dev_devid"] = "balloon%d" % idx
 87         vm.params["balloon_dev_add_bus"] = "yes"
 88         devs = vm.devices.get_by_params({"id": 'balloon%d' % idx})
 89         vm.params["balloon_pci_bus"] = devs[0]["bus"]
'''
Without these lines, dst qemu won't boot with a balloon device at all.  I'm not sure about other devices.
Besides, whether we have a common solution or not, this patch should be a must to support specifying address for balloon device. Thanks.

huangyum:
@PaulYuuu The address is not fixed, it's related to devices assigned to guest.  Eg, I boot guest with different numbers of devices,  then hotplug a balloon device, it gets different addresses.

huangyum:
@PaulYuuu Get your point. Tried several times on x86 host, never get 00.0.   
Could you please take a look at https://github.com/avocado-framework/avocado-vt/pull/1929#discussion_r250065450,  did nic device need such workaround? Thanks.

huangyum:
Tested on ppc host, it turns out the hotplugged device's addr is 0x0 on ppc while devs[0]["addr"] is 0x6(https://github.com/autotest/tp-qemu/pull/1642). Tried to assign 0x0 directly, and guest failed to boot due to address conflict. Don't have a good solution to fix it currently. Close this one.

PR #1930, opened 2019-01-21
Ignore numa node for non-numa systems

harihare:
Hotplug xml will have node tag by default, due to which the test
fails to hotplug on non numa systems. Patch will not add node tag
for non numa systems.

Signed-off-by: Hariharan T.S <hari@linux.vnet.ibm.com>

### DIFF
--- a/virttest/utils_hotplug.py
+++ b/virttest/utils_hotplug.py
@@ -514,7 +514,8 @@ def create_mem_xml(tg_size, pg_size=None, mem_addr=None, tg_sizeunit="KiB",
         tg_xml = memory.Memory.Target()
         tg_xml.size = int(tg_size)
         tg_xml.size_unit = tg_sizeunit
-        tg_xml.node = int(tg_node)
+        if tg_node != "":
+            tg_xml.node = int(tg_node)
         mem_xml.target = tg_xml
     if pg_size:
         src_xml = memory.Memory.Source()


virttest/utils_hotplug.py
```diff
@@ -514,7 +514,8 @@ def create_mem_xml(tg_size, pg_size=None, mem_addr=None, tg_sizeunit="KiB",
         tg_xml = memory.Memory.Target()
         tg_xml.size = int(tg_size)
         tg_xml.size_unit = tg_sizeunit
-        tg_xml.node = int(tg_node)
+        if tg_node != 'None':
```
balamuruhans:
```python
if tg_node != None:
```

balamuruhans: let's have default value as `tg_node=None`

harihare:
RESULTS    : PASS 36 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 7065.24 s
JOB HTML   : /home/hariharan/avocado-fvt-wrapper/results/job-2019-01-23T05.55-83c2595/results.html
07:52:56 INFO    : 
07:52:56 INFO    : Summary of test results can be found below:
TestSuite                                                                   TestRun    Summary             
 
guest_mem                                                                   Run        Successfully executed

balamuruhans:
Thanks @harihare, let's merge it.

PR #1940, opened 2019-01-23
unattended_install.nfs: To create image_dir before being used

luckyh:
An issue was introduced by https://github.com/avocado-framework/avocado-vt/pull/1927 which cause windows guests failed to be installed successfully.
```
ERROR: [Errno 2] No such file or directory: ''
```
That because windows guests do not have a valid `kernel` set, so let's revert the previous approach and use the first one I sent.

Signed-off-by: Xu Han xuhan@redhat.com

ID: 1668237

### DIFF
--- a/virttest/tests/unattended_install.py
+++ b/virttest/tests/unattended_install.py
@@ -247,11 +247,6 @@ class UnattendedInstallConfig(object):
                 os.makedirs(os.path.dirname(self.floppy))
 
         self.image_path = os.path.dirname(self.kernel)
-        if not os.path.isdir(self.image_path):
-            # To prevent errors
-            if os.path.isfile(self.image_path):
-                os.unlink(self.image_path)
-            os.makedirs(self.image_path)
 
         # Content server params
         # lookup host ip address for first nic by interface name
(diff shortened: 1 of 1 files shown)


peixiu:
(1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.unattended_install.cdrom.extra_cdrom_ks.default_install.aio_threads.q35: PASS (1839.66 s)

Acked-by: Peixiu Hou phou@redhat.com

luckyh:
Thanks all.

balamuruhans:
@luckyh, if it's the case we could have add check for windows guest or check for kernel in `__init__()` right ? Is there specific reason you thought for going with old approach. I still inclined to have it in __init__(),  anyway thank you for letting me know.

luckyh:
> @luckyh, if it's the case we could have add check for windows guest or check for kernel in `__init__()` right ? Is there specific reason you thought for going with old approach. I still inclined to have it in **init**(), anyway thank you for letting me know.

Hi @balamuruhans , I think the point is the code would look weird that `image_path` is depended on `kernel`, I can understand it is "clever" approach, but it is unfriendly to guests like windows which do not have a single kernel file to be extracted. However, windows guests does require the same directory to keep unattended configuration, and strangely, not be created by referencing `image_path`. I think the script needs a further optimization but I have no enough time to do it.

Another reason is I saw that there is another variant `setup_import()`, and it doesn't seem to use the `image_path` directory, that means creating the directory is an optional step, and that's why I reverted my previous patch.

Thanks.

balamuruhans:
@luckyh got it, thanks for sharing your thoughts.

luckyh:
@balamuruhans ok, no problem.

PR #1943, opened 2019-01-28
Update function:check_if_vm_vcpu_match to allow str input

chloerh:
Allow vcpu_desire to be str


harihare:
LGTM

PR #1947, opened 2019-01-30
env_process: add nested virtualization support to run tests in nested VMs

balamuruhans:
This patch enables existing tests to be run on nested VMs to
multilevels, is supports N * N * N VMs where multiple guests
can be brought up and nested with multilevel and multiple VMs
on every level. The params can be set to the granularity of any
particular VM on any level.

The point here is to dynamically calculate and distribute memory to
higher level VMs based on the available memory in current level
VM/host. It is not neccesary but it is required for VMs to run
test seamlessly on positive scenarios, still negative scenarios
can be tested by overriding the param to perform this distribution.

We handle the test across all the VMs at all the Levels in linear
way as parallelism from here will not have control on the test
flow but configuring different testcases on different VMs at any
level is possible. Parallel test scenarios can be added from testcase
indeed.

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -72,6 +72,8 @@ _post_hugepages_surp = 0
 #: information from the output produced by `qemu -version`
 QEMU_VERSION_RE = r"QEMU (?:PC )?emulator version\s([0-9]+\.[0-9]+\.[0-9]+)\s?\((.*?)\)"
 
+THREAD_ERROR = False
+
 
 def preprocess_image(test, params, image_name, vm_process_status=None):
     """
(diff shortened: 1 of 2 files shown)


balamuruhans:
Tested it with 
**Host:**
```

**Comment by balamuruhans** [comment]
@sathnaga @luckyh @vivianQizhu would you guys help for reviewing

File: virttest/env_process.py [hunk 1]
```diff
@@ -1123,16 +1146,33 @@ def preprocess(test, params, env):
                 testlist = [new_params.get("avocado_guest_vt_test",
                                            "boot")]
                 avocadotestargs = new_params.get("avocado_guest_add_args", "")
+                # add nested-hv feature attribute to VM's libvirt xml
+                if new_params.get("vm_type") == "libvirt":
```
sathnaga: we can choose to use --virt-install extra-params aswell, if create_vm_libvirt param is set, so that we can avoid creating a vm at all to change.

balamuruhans: Thanks @sathnaga, I have taken care of the suggestion.

balamuruhans:
The warning is unrelated to the PR.

virttest/utils_test/__init__.py
```diff
@@ -1087,8 +1087,10 @@ def runtest(self):
                 avocado_cmd += " --vt-no-filter \"%s\"" % self.vt_no_filter
             if self.vt_arch:
                 avocado_cmd += " --vt-arch %s" % self.vt_arch
+            if self.guest_image:
+                avocado_cmd += " --vt-guest-os %s" % self.guest_image
             if self.vt_extra_params:
-                avocado_cmd += " --vt-extra-params \"%s\"" % self.vt_extra_params
+                avocado_cmd += " --vt-extra-params %s" % self.vt_extra_params
```
sathnaga: any reason for this change, this was there to handle any space/spl chars to be wrapped right, what changed now?

balamuruhans: for extra params, we really don't need to wrap, and framework errors out if we try to wrap it. where we give `key='value'`

PR #1951, opened 2019-02-04
Add numa distance support for virt-install

sathnaga:
Add numa distance support for virt-install
so that numa distance can be configured duing
installtion/starting of guest.


sathnaga:
virt-install support: https://github.com/avocado-framework/avocado-vt/pull/1951

PR #1961, opened 2019-02-14
Add a param to force use serial session in case of failure

sathnaga:
Currently, we do have a logic in wait_for_login to
return serial session incase of failure if provided
by the user through function param, this patch extends
that functionality by adding a param which can be use set,
so that serial session is returned in case of failure,
this will be helpful as we most of the testcases using
wait_for_login with default params and we do not need
to modify the testcases, this can be useful during certain
env issues if we want to force use serial logins for the same test.


vivianQizhu:
Sometimes we need to be ware of the network broken, so I suppose we better let users decide use serial or not(if they care about the network issue).

sathnaga:
@vivianQizhu True, bydefault the behaviour is same as previous, just introducing additional user param, to decide to override consciously in some situations, to do that, this patch would help in general for all tests without any changes..

sathnaga:
@vivianQizhu Agree, this will be a useful workaround by not touching the old testcases.
and it is safe through a user param aswell.

PR #1967, opened 2019-02-20
fix vcpupin_output format got from vcpupin

brucepatricklee:
in affinity_from_vcpupin(), the value of vcpupin_output is got from
command "virsh vcpupin". The right vcpupin result such as " 0-1,3-7 "
will be thought to be a not a supported format without being stripped
as follows:

L0231 DEBUG| Cpus_string= 0-1,3-7 is not a supported format for cpu_list.


brucepatricklee:
> @brucepatricklee I hope this PR #1796 should address your issue, let me know if it is helpful.
the feedback was added in PR https://github.com/avocado-framework/avocado-vt/pull/1796

PR #1970, opened 2019-02-25
qemu_vm: Correct the check in `VM.mig_finished()`

luckyh:
Now we have many migration statuses other than `*active`, so we
could not make the judgment of a finished migration by verifying
the migration status is not `*active`. Hence, we should let it
return `True` only if the current status is in certain ones which
are `none`, `completed`, `failed` and `cancelled`.

ID: 1679860


luckyh:
@CongLi @maxujun request for your reviews, thanks.

maxujun:
@luckyh tested the case "migrate.tcp.with_filter_off.with_post_copy" 1000 times .hit the status "setup" and  didn't raise exception.the migration is successful finally.LGTM
Acked by:xuma@redhat.com

CongLi:
Ack.

luckyh:
Thanks all.

PR #1974, opened 2019-02-27
Enable dmesg white list

chunfuwen:
In some case, some dmesg from Vm is expected due to specific hardware,
but it still throw exception and fail the case. By adding one white list,
it will ignore exception throw if white list is enabled

### DIFF
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -506,6 +506,10 @@ guest_dmesg_ignore = False
 # Whether to dump guest dmesg output to console
 guest_dmesg_dump_console = no
 
+# Guest dmesg white list that need to be ignored in case, multiple item can be seperated by ';'
+# For examples: hardware has not undergone testing by Red Hat and might not be certified;xx" 
+guest_dmesg_white_list = "this hardware has not undergone testing by Red Hat and might not be certified"
+
 # Screendump thread params
 convert_ppm_files_to_png = no
 keep_ppm_files = no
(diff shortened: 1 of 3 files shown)


shared/cfg/base.cfg
```diff
@@ -471,6 +471,10 @@ guest_dmesg_ignore = False
 # Whether to dump guest dmesg output to console
 guest_dmesg_dump_console = no
 
+# Guest dmesg white list that need to be ignored in case, multiple item can be seperated by ';'
+# For examples: hardware has not undergone testing by Red Hat and might not be certified;xx" 
+guest_dmesg_white_list = "this hardware has not undergone testing by Red Hat and might not be certified"
```
sathnaga: let's pass either this as a list, so that we can iterate or atleast a comma(any other spl char) separated string and
split it while get it in `guest_dmesg_white_list = self.params.get("guest_dmesg_white_list").split(",")` something like this..otherwise this does work properly ,`for white_list_item in dmesg_white_list:`

chunfuwen: updated in the virt_vm.py, and get parameter as one list with spl char ";"

virttest/virt_vm.py
```diff
@@ -985,6 +985,7 @@ def verify_dmesg(self, dmesg_log_file=None, connect_uri=None):
         """
         level = self.params.get("guest_dmesg_level", 3)
         ignore_result = self.params.get("guest_dmesg_ignore", "no") == "yes"
+        guest_dmesg_white_list = self.params.get("guest_dmesg_white_list").split(';')
```
Yingshun: If there is no "guest_dmesg_white_list" in params, this line will get error like -- AttributeError:  'NoneType' object has no attribute 'split'.

pevogam:
Sorry why was this PR closed? I think it has a good idea and some of us have also had such issues with dmesg barking on any content it encounters. It might be good to be able to disable certain content there.

PR #1978, opened 2019-02-28
env_process: backup image in postprocess only if test pass

yanan-fu:
It is meaningless to backup image when test fail, so restrict it

id: 1683557

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -360,7 +360,8 @@ def postprocess_image(test, params, image_name, vm_process_status=None):
         # with a new backup. i.e. assume pre-existing image/backup
         # would not be usable after this test succeeds. The best
         # example for this is when 'unattended_install' is run.
-        if params.get("backup_image", "no") == "yes":
+        if (params.get("backup_image_after_testing_passed", "no") == "yes" and
+                params.get("test_passed") == "True"):
             image.backup_image(params, base_dir, "backup", True)
 
     if (not restored and params.get("restore_image", "no") == "yes"):


yanan-fu:
@luckyh could you help review, thanks!

vivianQizhu:
@hereischen

virttest/env_process.py
```diff
@@ -363,6 +363,10 @@ def postprocess_image(test, params, image_name, vm_process_status=None):
         if params.get("backup_image", "no") == "yes":
             image.backup_image(params, base_dir, "backup", True)
 
+        if (params.get("backup_image_after_testing_pass", "no") == "yes" and
```
luckyh: `backup_image_after_testing_passed`?

yanan-fu: both of them are ok for me, updated, please help review, thanks!

luckyh:
@ldoktor @balamuruhans @kylazhang @dzhengfy request for your review, thanks!

virttest/env_process.py
```diff
@@ -360,7 +360,8 @@ def postprocess_image(test, params, image_name, vm_process_status=None):
         # with a new backup. i.e. assume pre-existing image/backup
         # would not be usable after this test succeeds. The best
         # example for this is when 'unattended_install' is run.
-        if params.get("backup_image", "no") == "yes":
+        if (params.get("backup_image_after_testing_passed", "no") == "yes" and
```
sathnaga: we can still use the same param`backup_image` and additional check of `test_passed` is good.

hereischen: `backup_image` implies both test passed or failed.
If you want to save an image when test for debug, you can use `save_image`.

hereischen:

virttest/env_process.py
sathnaga: exactly, so lets default `backup_image` when `test_passed` = True, why add new param, and unness change in all testcases too.

balamuruhans:
I have difference of opinion, please correct me if I am wrong,

1. This backing up image would be helpful in unattended_install testcase, but we do also just import existing installed image using same unattended_install where we do not need to backup even if test is passed.

2. In some of the testcases that performs disk related tests such as image snapshot tests, we might need to backup for failed testcase to debug what went wrong.

so does it really matter we have to check whether test is pass or not ?
we set `backup_image = "no"` by default and only for unattended_install (that performs complete installation) and any other test that needs backup can set the same param as "yes".

virttest/env_process.py
yanan-fu: @sathnaga  That is a old version of this patch, i update it to the current version as i think it is better to add the usage scenario to the param name, same with `backup_image_before_testing`, `restore_image_after_testing`.  To make it more clearly when use this param. Thanks!

yanan-fu:
> I have difference of opinion, please correct me if I am wrong,
> 
>     1. This backing up image would be helpful in unattended_install testcase, but we do also just import existing installed image using same unattended_install where we do not need to backup even if test is passed.

Thanks  balamuruhans, if i understand you correctly,  I think it is same for the old design (back up no matter test pass or fail) too.  If we not need to backup even if test is passed,  set it to `no` is ok.

> 
>     2. In some of the testcases that performs disk related tests such as image snapshot tests, we might need to backup for failed testcase to debug what went wrong.

For this scenario, `save_image` can do it better,  It will save the image with a new name, which is different with the default image name. refer:
https://github.com/avocado-framework/avocado-vt/blob/a6ebf61e7b1de35113562505606ede0b6140a96c/virttest/env_process.py#L320

In this way, it will not effect other cases in a loop if some case use `restore_image_after_testing`.

> 
> so does it really matter we have to check whether test is pass or not ?
> we set `backup_image = "no"` by default and only for unattended_install (that performs complete installation) and any other test that needs backup can set the same param as "yes".

balamuruhans:
> Thanks  balamuruhans, if i understand you correctly,  I think it is same for the old design (back up no matter test pass or fail) too.  If we not need to backup even if test is passed,  set it to `no` is ok.

what do you think is better ?

> For this scenario, `save_image` can do it better,  It will save the image with a new name, which is different with the default image name. refer:
> https://github.com/avocado-framework/avocado-vt/blob/a6ebf61e7b1de35113562505606ede0b6140a96c/virttest/env_process.py#L323
> 
> In this way, it will not effect other cases in a loop if some case use `restore_image_after_testing`.

@yanan-fu cool, thanks for letting me know.

virttest/env_process.py
sathnaga: make sense, pls make sure take care of  all instances this param usage.

yanan-fu: @sathnaga thanks for reminder,  do not worry, i checked tp-qemu. tp-libvirt and also tp-spice, will cover all ^^

yanan-fu:
> > Thanks  balamuruhans, if i understand you correctly,  I think it is same for the old design (back up no matter test pass or fail) too.  If we not need to backup even if test is passed,  set it to `no` is ok.
> 
> what do you think is better ?
> 
@balamuruhans  this is a param not only for `unattended_install..import`, we can use it as needed.  I think it is better to add the usage scenario to the param name, make it more clearly when use it, Thanks!

balamuruhans:
@yanan-fu thanks! can we squash the commits ?

yanan-fu:
@balamuruhans updated, could you help check again, thanks a lot!

balamuruhans:
@yanan-fu still I am seeing 2 commits in this PR.

yanan-fu:
@balamuruhans  merge into one,  please help check again, many thanks!

PR #1159, opened 2017-09-14
qcontainer: fix duplicated device hotplug in simple_hotplug

sitoliu:
Device 'virtio_serial_pci' will be hotplugged two times in simple_hotplug function. This error affects the devices which can be catched by 'info qtree'.

id: 1491577
Signed-off-by: Sitong Liu <siliu@redhat.com>

PR #1982, opened 2019-03-04
Automatically specify bus and address when hotplugging a pci device

PaulYuuu:
`simple_hotplug()` insert the device at first before hotplugging it, will remove it if it  doesn't support hot plugging.

ID: 1645981

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -625,24 +625,24 @@ class DevContainer(object):
                             break
             if bus is not None:
                 bus.prepare_hotplug(device)
-        out = device.hotplug(monitor)
-        ver_out = device.verify_hotplug(out, monitor)
-
-        if ver_out is False:
-            self.set_clean()
-            return out, ver_out
 
         try:
+            # Insert the device first to assign slot
             qdev_out = self.insert(device)
             if not isinstance(qdev_out, list) or len(qdev_out) != 1:
                 raise NotImplementedError("This device %s require to hotplug "
-                                          "multiple devices %s, which is not "
-                                          "supported." % (device, out))
-            if ver_out is True:
-                self.set_clean()
+                                          "multiple devices, which is not "
+                                          "supported." % device)
         except DeviceError as exc:
             raise DeviceHotplugError(device, 'According to qemu_device: %s'
-                                     % exc, self, ver_out)
+                                     % exc, self)
+        else:
+            out = device.hotplug(monitor)
+            ver_out = device.verify_hotplug(out, monitor)
+            if ver_out is False:
+                self.remove(device)
+            self.set_clean()
+
         return out, ver_out
 
     def simple_unplug(self, device, monitor, timeout=30):


PaulYuuu:
(1/2) Host_RHEL.m8.u0.raw.spapr_vscsi.up.virtio_net.Guest.RHEL.8.0.ppc64le.io-github-autotest-qemu.nic_hotplug.migration.nic_virtio: PASS (351.28 s)
(2/2) Host_RHEL.m8.u0.raw.spapr_vscsi.up.virtio_net.Guest.RHEL.8.0.ppc64le.io-github-autotest-qemu.balloon_hotplug.with_migrate.migrate_after_hotplug: PASS (310.36 s)

PaulYuuu:
@huangyum @vivianQizhu Please help to review.

PaulYuuu:
@huangyum vt reports like follows:
```
20:32:24 DEBUG| (monitor avocado-vt-vm1.qmp1) Sending command 'device_add'
20:32:24 DEBUG| Send command: {'execute': 'device_add', 'arguments': OrderedDict([('id', 'balloon0'), ('driver', 'virtio-balloon-pci'), ('bus', 'pci.0')]), 'id': '138fXdJ7'}
20:32:24 DEBUG| (monitor avocado-vt-vm1.qmp1) Sending command 'device_del'
20:32:24 DEBUG| Send command: {'execute': 'device_del', 'arguments': {'id': 'balloon0'}, 'id': 'qxwY7Rzv'}
20:32:25 DEBUG| Waiting for unplug q'balloon0' to finish (1.001334 secs)
20:32:25 DEBUG| (monitor avocado-vt-vm1.qmp1) Sending command 'device_add'
20:32:25 DEBUG| Send command: {'execute': 'device_add', 'arguments': OrderedDict([('id', 'balloon0'), ('driver', 'virtio-balloon-pci'), ('bus', 'pci.0'), ('addr', '0x5')]), 'id': 'odvPR34T'}
```
For `q'balloon0'` , it is because you have not define the `aid` of the device.
you can use `new_dev.set_aid(new_dev.get_param('id'))` or `new_dev.set_aid(new_dev.get_qid())` after `new_dev = qdevices.QDevice()` from `balloon_hotplug.py`. If you think it looks better.

virttest/qemu_devices/qcontainer.py
```diff
@@ -616,12 +616,11 @@ def simple_hotplug(self, device, monitor, bus=None):
                 if self.is_pci_device(device['driver']):
                     bus = self.get_buses({'aobject': 'pci.0'})[0]
                 if not isinstance(device.parent_bus, (list, tuple)):
-                    device.parent_bus = [device.parent_bus]
-                for parent_bus in device.parent_bus:
-                    for _bus in self.get_buses(parent_bus):
-                        if _bus.bus_item == 'bus':
-                            bus = _bus
-                            break
+                    for parent_bus in [device.parent_bus]:
```
vivianQizhu: Why move this? If the bus is already a list or tuple, it wont get correct bus this way.

PaulYuuu: According to `qcontainer.py`, line 620. 
```python
for parent_bus in device.parent_bus:
```
and we can see other devices info define by framework
```
device
  aid = idJo6csb
  aobject = None
  parent_bus = ({'aobject': 'pci.0'},)
  child_bus = []
  params:
    driver = virtio-net-pci
    mac = 9a:9e:9f:a0:a1:a2
    id = idJo6csb
    vectors = 4
    netdev = id9yY2co
    bus = pci.0
    addr = 0x4
```
If set `device.parent_bus = [device.parent_bus]`, the device info will be `parent_bus = [{'aobject': 'pci.0'}]` instead of `parent_bus = ({'aobject': 'pci.0'},)`
So use `for parent_bus in [device.parent_bus]` will be better.

PaulYuuu:
> 'hotplug -> unplug -> hotplug' does not equal to a single 'hotplug' from testing perspective, I suppose we should find another better solution to get the correct address.

@vivianQizhu What is your opinion about the wrong fixed #1159 
`simple_hotplug()` not mean single hotplug, the first `out = device.hotplug(monitor)` is order to check if verification (ver_out) is True, if the result is True, we need to insert the device into bus first, and then return the second `out = device.hotplug(monitor)` in the finish.

On the other hand, I can move `qdev_out = self.insert(device)` to make it in front of the first `out = device.hotplug(monitor)`, but it seems that `ver_out` will become meaningless.

@ldoktor Could you please help confirm the meaning of this functional design?

huangyum:
I agree with @vivianQizhu,  when we invoke simple_hotplug, all we want is to hotplug the device. And I don't really understand how hotplugging twice solves the address problem since you referred to bz1645981.

PaulYuuu:
@huangyum `self.insert()` will update device params or set if none.
Before insert, `device.str_long()` shows:
```
device
  aid = None
  aobject = None
  parent_bus = [{'aobject': 'pci.0'}]
  child_bus = []
  params:
    id = balloon0
    driver = virtio-balloon-pci
    bus = pci.0
```
After insert, it shows:
```
device
  aid = None
  aobject = None
  parent_bus = [{'aobject': 'pci.0'}]
  child_bus = []
  params:
    id = balloon0
    driver = virtio-balloon-pci
    bus = pci.0
    addr = 0x5
```
So if `ver_out` is `False`, this function will hotplug a device without `addr`, otherwise it will insert to its `parent_bus` and reserve that slot address, re-hotplug it.

virttest/qemu_devices/qcontainer.py
vivianQizhu: @PaulYuuu My point is, `if not isinstance(device.parent_bus, (list, tuple))` should not be a condition for `bus = _bus`. Both `[{'aobject': 'pci.0'}]` and `({'aobject': 'pci.0'},)` is acceptable format for `device.parent_bus`. Similar usage here: https://github.com/avocado-framework/avocado-vt/blob/master/virttest/qemu_devices/qcontainer.py#L494

vivianQizhu:
> > 'hotplug -> unplug -> hotplug' does not equal to a single 'hotplug' from testing perspective, I suppose we should find another better solution to get the correct address.
> 
> @vivianQizhu What is your opinion about the wrong fixed #1159

I wont say it is a wrong fix, hotplug twice does not fit for our test requirement, and why remove the second `hotplug` instead of first one, or why `hotplug` before `insert`? Because if `hotplug` failed then it should not be `inserted`(not allocate the address of the bus).

> `simple_hotplug()` not refer to single hotplug, the first `out = device.hotplug(monitor)` is order to check if verification (ver_out) is True, if the result is True, we need to insert the device into bus first, and then return the second `out = device.hotplug(monitor)` in the finish.

That should be a single hotplug, otherwise it will be out of most tester's expectation. As explained above, instead of `insert` before `hotplug`, if we do want an address before hotplug, maybe get a free slot and assign it to the device as a parameter.

virttest/qemu_devices/qcontainer.py
vivianQizhu: @PaulYuuu And what you get in the output, might get from here: https://github.com/avocado-framework/avocado-vt/blob/master/virttest/qemu_devices/qcontainer.py#L535 So we might modified same way if you want it be consistent. Thanks.

PaulYuuu:
> That should be a single hotplug, otherwise it will be out of most tester's expectation. As explained above, instead of `insert` before `hotplug`, if we do want an address before hotplug, maybe get a free slot and assign it to the device as a parameter.

If only a single hotplug is needed, it seems that `device.hotplug()` more matches than `simple_hotplug()`.
"get a free slot and assign it to the device as a parameter" can resolve it when `prepare_hotplug()`, but it conflicts with `insert()` that `insert()` will ignore its pci address.

vivianQizhu:
> > That should be a single hotplug, otherwise it will be out of most tester's expectation. As explained above, instead of `insert` before `hotplug`, if we do want an address before hotplug, maybe get a free slot and assign it to the device as a parameter.
> 
> If only a single hotplug is needed, it seems that `device.hotplug()` more matches than `simple_hotplug()`.

Actually previously nobody really use simple_hotplug, most test cases use device.hotplug() directly, where they verify the hotplug, but not `insert` it. The issue of this method is, the hotplugged device will not be managed by DevContainer, and the address will be allocated in qemu, but for the bus object in DevContainer, the address are still free. What `insert` does, besides to assign `addr` property to device, also is to manage it in DevContainer and allocate the bus address: https://github.com/avocado-framework/avocado-vt/blob/master/virttest/qemu_devices/qdevices.py#L1503

While actually most hotplug test cases do not care if the device is managed by DevContainer. Only they need is to send out the hotplug command and check the device exists in qemu monitor.

Recently a new issue is pcie, so we modified simple_hotplug to let it adopt pcie bus hotplug, and replaced most device.hotplug to simple_hotplug. Now if the user use device.hotplug, it requires more implementation even if they don't care about DevContainer.

> "get a free slot and assign it to the device as a parameter" can resolve it when `prepare_hotplug()`, but it conflicts with `insert()` that `insert()` will ignore its pci address.

I am not sure, but it should not be conflict, set `addr` property before `insert` should works, you get a free slot but not really allocate it(no `self.bus[addr] = device` executed), even `insert` scan a free slot again, they will be the same one.

PaulYuuu:
> I am not sure, but it should not be conflict, set `addr` property before `insert` should works, you get a free slot but not really allocate it(no `self.bus[addr] = device` executed), even `insert` scan a free slot again, they will be the same one.

Yes, I tried. they are the same one but the `addr` is an invalid parameter when `insert()` this device, it will be set again.
I agree with you, those devices need manage by DevContainer, So if the verification is False, the hotplugged device will be managed by qemu and return directly, otherwise it will be managed by DevContainer and re-hotplug order to use the pci address assigned by DevContainer, but not what we set .

vivianQizhu:
> > I am not sure, but it should not be conflict, set `addr` property before `insert` should works, you get a free slot but not really allocate it(no `self.bus[addr] = device` executed), even `insert` scan a free slot again, they will be the same one.

It might be invalid for `insert`, but it is enough for hotplug command generation(add the `addr` option). So no need for the re-hotplug action.

> 
> Yes, I tried. they are the same one but the `addr` is an invalid parameter when `insert()` this device, it will be set again.
> I agree with you, those devices need to be managed by DevContainer. So if the verification is False, the hotplugged device will be managed by qemu and return directly, otherwise it will be managed by DevContainer and re-hotplug order to use the pci address assigned by DevContainer, but not what we set .

Actually this way what we set is also get from DevContainer, that might not be a perfect approach, but I have no better solution with current framework, while the re-hotplug approach breaks business logic correction so even worse...

PaulYuuu:
@vivianQizhu I modified the code, but I think if we specify a different address, we still need to match the framework.
@luckyh Do you think we need to always keep the address of the device equal to that managed by DevContainer?

vivianQizhu:
Hi @PaulYuuu After reconsidering about this issue, current code looks okay to me, just this way we should not need `prepare_hotplug` anymore, cause `insert` would set correct bus for device, and also we'd better test all cases use simple_hotplug to make sure no regression introduced. Would you help check it again? Thanks.

PaulYuuu:
> Hi @PaulYuuu After reconsidering about this issue, current code looks okay to me, just this way we should not need `prepare_hotplug` anymore, cause `insert` would set correct bus for device, and also we'd better test all cases use simple_hotplug to make sure no regression introduced. Would you help check it again? Thanks.

@vivianQizhu Will check and test during this weekend, Thanks for review again.

PaulYuuu:
> Hi @PaulYuuu After reconsidering about this issue, current code looks okay to me, just this way we should not need `prepare_hotplug` 

@vivianQizhu For PCI device, we can remove `prepare_hotplug`, but PCIE device needs it. Otherwise, `insert` it directly will cause this issue:
```
  params:
    id = balloon0
    driver = virtio-balloon-pci
    bus = pcie.0-root-port-6
    addr = 0x0
```

vivianQizhu:
@PaulYuuu Alright, let's keep it then. Thanks for confirming it.

nanliu-r:
Test command looks good in my side, but I'm a little confused about what you are discussing.

testcase=irtio_port_hotplug.unplug_chardev

Command:
23:22:32 DEBUG| Send command: {'execute': 'device_add', 'arguments': OrderedDict([('id', 'vc1'), ('driver', 'virtserialport'), ('name', 'vc1'), ('chardev', 'chardev_vc1'), ('bus', 'virtio_serial_pci0.0'), ('nr', 1)]), 'id': 'p5SuvE8A'}
23:22:35 DEBUG| Send command: {'execute': 'device_add', 'arguments': OrderedDict([('id', 'vc2'), ('driver', 'virtconsole'), ('name', 'vc2'), ('chardev', 'chardev_vc2'), ('bus', 'virtio_serial_pci1.0'), ('nr', 1)]), 'id': 'fSLRYzQv'}

PaulYuuu:
> Test command looks good in my side, but I'm a little confused about what you are discussing.
> 
> testcase=irtio_port_hotplug.unplug_chardev
> 
> Command:
> 23:22:32 DEBUG| Send command: {'execute': 'device_add', 'arguments': OrderedDict([('id', 'vc1'), ('driver', 'virtserialport'), ('name', 'vc1'), ('chardev', 'chardev_vc1'), ('bus', 'virtio_serial_pci0.0'), ('nr', 1)]), 'id': 'p5SuvE8A'}
> 23:22:35 DEBUG| Send command: {'execute': 'device_add', 'arguments': OrderedDict([('id', 'vc2'), ('driver', 'virtconsole'), ('name', 'vc2'), ('chardev', 'chardev_vc2'), ('bus', 'virtio_serial_pci1.0'), ('nr', 1)]), 'id': 'fSLRYzQv'}

@nanliu-r Command returns consistent with expectations, this test case failed due to:
```python
            if ver_out is False:
                self.remove(device)
```
Test step did not check it. However,
https://github.com/avocado-framework/avocado-vt/blob/89299589305683859e51f2d4204e761680cc9c55/virttest/qemu_devices/qdevices.py#L1386-L1402
These two methods always `return False`. And why? Because type returned by `monitor.query("chardev")` is a list like `[{}, {}, {}]` but not string `str`,

So I suggest:
1. Check ver_out in test case
2. fix `verify_hotplug()` and `verify_unplug()` in `class CharDevice`

@vivianQizhu I am not sure how chardev works, so please to help fix them.

vivianQizhu:
@PaulYuuu Sure, thanks for the analysis and notification. @nanliu-r If `'bus' is 'virtio_serial_pci0.0'` instead of 'virtio_serial_pci0` then the patch works right, I will fix the rest part.

vivianQizhu:
Some negative test might be affected, when test expect fail on hotplug command execution(like redundant id something), with this commit it will fail on `insert()`, the hotplug command won't be executed, so the case will not achieve its purpose. Anyway need test more to check if there really exists such cases.

PaulYuuu:
> Some negative test might be affected, when test expect fail on hotplug command execution(like redundant id something), with this commit it will fail on `insert()`, the hotplug command won't be executed, so the case will not achieve its purpose. Anyway need test more to check if there really exists such cases.

Negative params should not be affected by this, but it will affect the invalid bus/addr.
For negative test, is it more suitable to use `device.hotplug()`? Seems that they should not be managed.

vivianQizhu:
> > Some negative test might be affected, when test expect fail on hotplug command execution(like redundant id something), with this commit it will fail on `insert()`, the hotplug command won't be executed, so the case will not achieve its purpose. Anyway need test more to check if there really exists such cases.
> 
> Negative params should not be affected by this, but it will affect the invalid bus/addr.
> For negative test, is it more suitable to use `device.hotplug()`? Seems that they should not be managed.

Yes I agree, I found it when my new scenario can't directly use simple_hotplug here: https://github.com/autotest/tp-qemu/pull/1890/files#diff-209195933168f97ba23db8ddfb9a0100R54, it will hit error on `insert` instead of `hotplug`. That's reasonable to let such case utilize `device.hotplug` directly, I just worried that there is existed similar test case would be affected, but from my test to most hotplug test case, I don't find there is no such usage till now, let me test more and I will update my conclusion later. Thanks.

PaulYuuu:
Found 2 negative cases, they are also unaffected by this. Will test more later.
 (1/2) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.ppc64le.io-github-autotest-qemu.block_hotplug_negative: PASS (80.82 s)
 (2/2) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.ppc64le.io-github-autotest-qemu.block_hotplug_negative.data_plane: PASS (73.42 s)

vivianQizhu:
Test passed on x86, with --testcase=qemu_guest_agent.virtio_serial.check_hotplug_frozen,block_hotplug.default..fmt_qcow2.default.with_plug.default.multi_pci,virtio_port_hotplug.hotplug_port_pci,balloon_hotplug.with_reboot,block_hotplug_negative.default,pci_bridge.hotplug_unplug.multi_parallel_bridges,vsock_test.hotplug,block_hotplug_in_pause.default.with_plug.multi_pci and --machines=i440fx,q35:
 (01/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.pci_bridge.hotplug_unplug.multi_parallel_bridges: PASS (66.68 s)
 (02/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.virtio_port_hotplug.hotplug_port_pci: PASS (51.70 s)
 (03/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.virtio_port_hotplug.hotplug_port_pci.q35: PASS (72.63 s)
 (04/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.block_hotplug.block_virtio.fmt_qcow2.with_plug.multi_pci: PASS (58.64 s)
 (05/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.block_hotplug.block_virtio.fmt_qcow2.with_plug.multi_pci.q35: PASS (74.35 s)
 (06/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.block_hotplug.block_scsi.fmt_qcow2.with_plug.multi_pci: PASS (58.46 s)
 (07/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.block_hotplug.block_scsi.fmt_qcow2.with_plug.multi_pci.q35: PASS (54.15 s)
 (08/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_hotplug_frozen: PASS (128.46 s)
 (09/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_hotplug_frozen.q35: ERROR: local variable 'mnt_point' referenced before assignment (126.94 s)
 (10/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.balloon_hotplug.with_reboot: PASS (180.63 s)
 (11/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.balloon_hotplug.with_reboot.q35: FAIL: Balloon memory fail with error message: QMP command 'balloon' failed    (arguments: {'value': 7516192768},    error message: {'class': 'DeviceNotActive', 'desc': 'No balloon device has been activated'}) (107.85 s)
 (12/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.block_hotplug_in_pause.with_plug.multi_pci: /PASS (630.87 s)
 (13/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.block_hotplug_in_pause.with_plug.multi_pci.q35: PASS (632.45 s)
 (14/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.block_hotplug_negative: PASS (37.64 s)
 (15/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.block_hotplug_negative.q35: PASS (39.26 s)
 (16/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.vsock_test.hotplug: ERROR: Timeout expired while waiting for shell command to complete: 'rm -f /tmp/vsock_file_uOalSx'    (output: '') (251.11 s)
 (17/17) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.vsock_test.hotplug.q35: FAIL: lspci failed, no device "Communication controller: Red Hat.* Device 1012" (49.82 s)
All the failure and error are not related to this commit.

@PaulYuuu Please help test them on ppc host, thanks.

PaulYuuu:
(1/9) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.ppc64le.io-github-autotest-qemu.balloon_hotplug.with_reboot: PASS (259.16 s)
 (2/9) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.ppc64le.io-github-autotest-qemu.block_hotplug_in_pause.with_plug.multi_pci: PASS (1126.40 s)
 (3/9) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.ppc64le.io-github-autotest-qemu.block_hotplug_negative: PASS (70.13 s)
 (4/9) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.ppc64le.io-github-autotest-qemu.block_hotplug.block_virtio.fmt_qcow2.with_plug.multi_pci: PASS (101.85 s)
 (5/9) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.ppc64le.io-github-autotest-qemu.block_hotplug.block_scsi.fmt_qcow2.with_plug.multi_pci: PASS (86.33 s)
 (6/9) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.ppc64le.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_hotplug_frozen: ERROR: local variable 'mnt_point' referenced before assignment (158.61 s)
 (7/9) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.ppc64le.io-github-autotest-qemu.virtio_port_hotplug.hotplug_port_pci: PASS (104.57 s)
 (8/9) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.ppc64le.io-github-autotest-qemu.vsock_test.hotplug: PASS (197.74 s)
 (9/9) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.ppc64le.io-github-autotest-qemu.pci_bridge.hotplug_unplug.multi_parallel_bridges: FAIL: Could not find the usb device serial:[TARGET_DISK2] (435.86 s)

The last case failed due to product issue.

vivianQizhu:
Thanks @PaulYuuu, let's merge it.

PR #1989, opened 2019-03-07
utils_time: remove "re.A" which is not supported by python2

yanan-fu:
1. python2 don't support "re.A"
2. only for python3, "re.A" is used for ASCII-only matching instead of full Unicode matching. We need to match non-ASCII letters here if it have (such as Chinese character)

id: 1677189

### DIFF
--- a/virttest/utils_time.py
+++ b/virttest/utils_time.py
@@ -81,7 +81,7 @@ def verify_timezone_win(session):
         match_pattern = "(?:\(UTC([+|-]\d{2}:\d{2})?)(?:.*\n)(\w+.*(?:\s\w+)*)"
         timezone_sets = []
         for para in re.split("(?:\r?\n){2,}", timezone_list.strip()):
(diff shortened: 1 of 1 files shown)


yanan-fu:
Test with python3:
 (1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.timedrift.shared_ntp_date.with_reboot: PASS (213.09 s)

Test with python2:
 (1/1) Host_RHEL.m7.u7.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.timedrift.shared_ntp_date.with_reboot: PASS (443.81 s)

yanan-fu:
@sitoliu Could you please help review this PR?
I don't know the reason to add "re.A" before, but after some checking, i think we can remove it.
If i miss considering some situation, please correct me, thanks!

PR #1995, opened 2019-03-11
Give a default value for get_qemu_version

lento-sun:
If not required, a empty dict could be passed. So set the default
value to be a empty dict for easier usage.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2298,7 +2298,7 @@ def get_qemu_best_cpu_model(params):
     return params.get("default_cpu_model", None)
 
 
-def get_qemu_version(params):
+def get_qemu_version(params=None):
     """
     Get the qemu-kvm(-rhev) version info.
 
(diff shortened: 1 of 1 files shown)


virttest/utils_misc.py
```diff
@@ -2298,7 +2298,7 @@ def get_qemu_best_cpu_model(params):
     return params.get("default_cpu_model", None)
 
 
-def get_qemu_version(params):
+def get_qemu_version(params={}):
```
lolyu: Please do not use mutable objects as function params, please refer to https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html.

lento-sun: thx for the tips, updated

virttest/utils_misc.py
```diff
@@ -2298,7 +2298,7 @@ def get_qemu_best_cpu_model(params):
     return params.get("default_cpu_model", None)
 
 
-def get_qemu_version(params):
+def get_qemu_version(params=None):
```
chunfuwen: set params={ } as default

lento-sun: pls refer to https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html as @lolyu mentioned above

PR #1997, opened 2019-03-12
utils_vsock: Abstract vsock related functions

vivianQizhu:
This is for same steps when hot plugging vsock devices, instead of
limiting it inside qemu command generation phase

id: 1671211

### DIFF
--- a/virttest/utils_vsock.py
+++ b/virttest/utils_vsock.py
@@ -1,0 +1,34 @@
+import os
+import fcntl
+import struct
+import errno
+
+from virttest import arch
+
+VSOCK_PATH = '/dev/vhost-vsock'
+
+
+def get_guest_cid(guest_cid):
+    """
+    Get an unused guest cid from system
+
+    :param guest_cid: Requested guest cid
+    :return Available guest cid
+    """
+    vsock_fd = os.open(VSOCK_PATH, os.O_RDWR)
+    try:
+        while guest_cid:
+            cid_c = struct.pack('L', guest_cid)
+            try:
+                fcntl.ioctl(
+                    vsock_fd, arch.VHOST_VSOCK_SET_GUEST_CID, cid_c)
+            except IOError as e:
+                if e.errno == errno.EADDRINUSE:
+                    guest_cid += 1
+                    continue
+                else:
+                    raise e
+            else:
+                return guest_cid
+    finally:
+        os.close(vsock_fd)
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -9,10 +9,8 @@ import time
 import os
 import logging
 import fcntl
-import struct
 import re
 import random
-import errno
 
 from functools import partial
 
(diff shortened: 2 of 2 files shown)


vivianQizhu:
@yanan-fu @luckyh Please help review, thanks.

virttest/utils_vsock.py
```diff
@@ -0,0 +1,46 @@
...
+                vsock_fd, arch.VHOST_VSOCK_SET_GUEST_CID, cid_c)
+        except IOError as e:
+            if e.errno == errno.EADDRINUSE:
+                guest_cid += 1
+                continue
+            else:
+                os.close(vsock_fd)
+                raise e
+        else:
+            os.close(vsock_fd)
```
yanan-fu: Prefer move it to finally.

vivianQizhu: @yanan-fu I think no, there is one condition that should not be closed, at L#40.

yanan-fu: @vivianQizhu  yes,  right,  need continue to get cid.

yanan-fu:
LGTM

virttest/utils_vsock.py
luckyh: It can be done by doing like below:
```python
vsock_fd = os.open(VSOCK_PATH, os.O_RDWR)
try:
    while guest_cid:
        ...
finally:
    os.close(vsock_fd)
```

virttest/utils_vsock.py
```diff
@@ -0,0 +1,46 @@
...
+VSOCK_PATH = '/dev/vhost-vsock'
+
+
+def load_vsock_module():
+    """ Load vhost_vsock module, to enable vsock function """
+    if not os.path.exists(VSOCK_PATH):
+        logging.info('vsock module was not loaded, loading it...')
+        if not linux_modules.load_module('vhost_vsock'):
+            raise exceptions.TestError(
+                'Failed on loading module vhost_vsock.')
```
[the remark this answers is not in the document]
vivianQizhu: @luckyh Would you help explain why? Cause I will still need to raise an error outside if get False here.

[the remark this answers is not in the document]
vivianQizhu: @luckyh Just found that linux_modules.load_module('vhost_vsock') will return False when the module is already loaded, so should not raise error anywhere, will update, thanks.

vivianQizhu:
@luckyh Updated, please help review, thanks.

vivianQizhu:
Well, `load_module()` will check if the module is loaded or not, so the `if` condition for `VSOCK_PATH` existence is not needed, I am removing the load module function and using `load_module('vhost_vsock')` directly. @yanan-fu Please help double review too, thanks.

virttest/utils_vsock.py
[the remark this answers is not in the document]
yanan-fu: Nice catch

yanan-fu:
Ack

vivianQizhu:
Thanks all, merging it.

PR #2002, opened 2019-03-13
Fixup: utils_stress

sathnaga:
_*Commit 1:*_
**Fixup: validate stress events param before split**

stress events param string is split to convert as list and
when stress events are given as empty string, it does create
list entry with empty string and leads to thread exception, like
below

 [stderr]   File "..../virttest/utils_stress.py", line 165, in vm_stress_events
 [stderr]   File "..../virttest/utils_stress.py", line 165, in vm_stress_events
 [stderr]   File "..../virttest/utils_stress.py", line 190, in host_stress_event
 [stderr]     raise NotImplementedError
 [stderr]     raise NotImplementedError
 [stderr]     raise NotImplementedError
 [stderr] NotImplementedError
 [stderr] NotImplementedError
 [stderr] NotImplementedError

and let's fix it by validating the stress events param before split.

_*Commit 2:*_
**Introduce sleep between each stress events iteration**

Let's introduce a user defined sleep between each
guest and host stress events, as some events like vm suspend
does not allow to guest stress to run as most of the time vm is
suspended during iterations.

### DIFF
--- a/virttest/utils_stress.py
+++ b/virttest/utils_stress.py
@@ -23,6 +23,7 @@ class VMStressEvents():
         self.iterations = int(params.get("stress_itrs", 1))
         self.host_iterations = int(params.get("host_event_itrs", 10))
         self.event_sleep_time = int(params.get("event_sleep_time", 10))
+        self.itr_sleep_time = int(params.get("itr_sleep_time", 10))
         self.current_vcpu = params.get("smp", 32)
         self.max_vcpu = params.get("vcpu_maxcpus", 32)
         self.ignore_status = params.get("ignore_status", "no") == "yes"
@@ -171,6 +172,7 @@ class VMStressEvents():
                         libvirt.delete_local_disk(self.disk_type, disk_name)
             else:
                 raise NotImplementedError
+            time.sleep(self.itr_sleep_time)
 
     def host_stress_event(self, event):
         """
@@ -196,3 +198,4 @@ class VMStressEvents():
                 cpu.online(processor)
             else:
                 raise NotImplementedError
+            time.sleep(self.itr_sleep_time)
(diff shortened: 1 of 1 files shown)


sathnaga:
Though this error does not lead to test failure, it will result in considerable increase of test time...

**without patch:**
```
JOB LOG    : /home/sath/avocado-fvt-wrapper/results/job-2019-03-13T07.09-0b67ca0/job.log
 (1/1) type_specific.powerkvm-libvirt.multivm_cpustress.without_hoststress.stress_reboot: PASS (354.32 s)
JOB TIME   : 356.72 s
JOB HTML   : /home/sath/avocado-fvt-wrapper/results/job-2019-03-13T07.09-0b67ca0/results.html
```

**with patch:**
```
JOB LOG    : /home/sath/avocado-fvt-wrapper/results/job-2019-03-13T07.55-79c0de3/job.log
 (1/1) type_specific.powerkvm-libvirt.multivm_cpustress.without_hoststress.stress_reboot: PASS (70.26 s)
JOB TIME   : 72.65 s
```

virttest/utils_stress.py
```diff
@@ -27,8 +27,16 @@ def __init__(self, params, env):
         self.max_vcpu = params.get("vcpu_maxcpus", 32)
         self.ignore_status = params.get("ignore_status", "no") == "yes"
         self.vms = env.get_all_vms()
-        self.events = params.get("stress_events", "reboot").split(',')
-        self.host_events = params.get("host_stress_events", "").split(',')
+        self.events = params.get("stress_events", "reboot")
```
balamuruhans: we could possibly hit this issue only with `host_stress_events` as we default it to `""` this fix would not be necessary if we default to `reboot` in case of `stress_events`

sathnaga: user can assign `stress_events=""` explicitly and it would lead such situation..

balamuruhans: ah, okay got it.

balamuruhans:
Thanks @sathnaga

PR #2004, opened 2019-03-15
utils_net: add bridge name as a param for list_iface func

xiywang:
id: 1688577

In some cases, we need to get all the interfaces in a dedicated
bridge. Add a param 'br' to list_iface.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -1026,10 +1026,14 @@ class Bridge(object):
     def list_br(self):
         return list(self.get_structure().keys())
 
-    def list_iface(self):
+    def list_iface(self, br=None):
         """
         Return all interfaces used by bridge.
+
+        :param br: Name of bridge
         """
+        if br:
+            return self.get_structure()[br]['iface']
         interface_list = []
         for br in self.list_br():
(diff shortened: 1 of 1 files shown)


xiywang:
@pezhang @PaulYuuu @vivianQizhu Please hlep review, thanks.

virttest/utils_net.py
```diff
@@ -1026,10 +1026,14 @@ def get_structure(self):
     def list_br(self):
         return list(self.get_structure().keys())
 
-    def list_iface(self):
+    def list_iface(self, br=None):
         """
         Return all interfaces used by bridge.
```
PaulYuuu: Please update it since we can specify the bridge we want.

xiywang: You mean the doc string? I think "all the interfaces used by bridge" fits well. bridge could be None or specified. The function returns the interfaces used by bridge(None or brname). You think?

PaulYuuu: Okay. Since `br` alias `bridge`, we can ignore this doc.
Acked-by: Yihuang Yu <yihyu@redhat.com>

Issue #2005, opened 2019-03-18
[Bug report] the virsh.py maybe cause RuntimeError in Python 3.7


hsmj1412:
Hi all,

There is a bug I meet but could not find the root cause(it should be a python bug), and I try to reproduce this bug without avocado but failed. So I mark it here.

I will provide a easy way to reproduce it with avocado.
1. Prepare a large python file like following(This file is simplified virttest/virsh.py):
```
class Testtest(object):
    def __init__(self):
        a = list(globals().items())

def func0(test1=1, test2=1, test3=1, test4=1, test5=1, test6=1, test7=1, test8=1, test9=1, test10=1, **dargs):
    return 0

...
Skipped the middle functions
...

def func999(test1=1, test2=1, test3=1, test4=1, test5=1, test6=1, test7=1, test8=1, test9=1, test10=1, **dargs):
    return 0
```

2. Use this file in any case like following:

```
from virttest import testtest

def run(test, params, env):
    a = testtest.Testtest()
```

3. Run the case and the first time it maybe pass, but from the second time, it will report a RuntimeError:

Traceback (most recent call last):

File "/var/lib/libvirt_gating/code/avocado/avocado/core/test.py", line 938, in _run_avocado
    raise test_exception

File "/var/lib/libvirt_gating/code/avocado/avocado/core/test.py", line 839, in _run_avocado
    testMethod()

File "/var/lib/libvirt_gating/code/avocado-vt/avocado_vt/test.py", line 312, in runTest
    raise self.__status  # pylint: disable=E0702

File "/var/lib/libvirt_gating/code/avocado-vt/avocado_vt/test.py", line 263, in setUp
    self._runTest()

File "/var/lib/libvirt_gating/code/avocado-vt/avocado_vt/test.py", line 438, in _runTest
    run_func(self, params, env)

File "/var/lib/avocado/data/avocado-vt/test-providers.d/downloads/io-github-autotest-libvirt/libvirt/tests/src/test.py", line 85, in run
    tttt = testttest.Testtest()

File "/var/lib/libvirt_gating/code/avocado-vt/virttest/testtest.py", line 16, in __init__
    a = list(globals().items())

RuntimeError: dictionary changed size during iteration

Additional info:
In real test env, the Virsh class in virttest/virsh.py will use the globals().items(), so it will report the same error.

hsmj1412:
When the iterator is running, the globals() will be added a new item.
key:  __warningregistry__
value:  {'version': 28}

PR #2006, opened 2019-03-19
Minorfix: Make a copy for globals to avoid RuntimeError

hsmj1412:


hsmj1412:
Due to the issue #2005 , try to avoid the error in avocado-vt

PR #2007, opened 2019-03-19
Minorfix: Fixup the polkit user error

hsmj1412:

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -1870,22 +1870,23 @@ class LibvirtPolkitConfig(object):
         self.polkit_rules_path = "/etc/polkit-1/rules.d/"
         self.polkit_rules_path += "500-libvirt-acl-virttest.rules"
         self.polkit_name = "polkit"
+        self.params = params
         distro_obj = distro.detect()
         # For ubuntu polkitd have to be used
         if distro_obj.name.lower().strip() == 'ubuntu':
             self.polkit_name = "polkitd"
         self.polkitd = service.Factory.create_service(self.polkit_name)
 
-        if params.get("action_id"):
-            self.action_id = params.get("action_id").split()
+        if self.params.get("action_id"):
+            self.action_id = self.params.get("action_id").split()
         else:
             self.action_id = []
-        self.user = params.get("unprivileged_user")
-        if params.get("action_lookup"):
+        self.user = self.params.get("unprivileged_user")
+        if self.params.get("action_lookup"):
             # The action_lookup string should be separated by space and
             # each separated string should have ':' which represent key:value
             # for later use.
-            self.attr = params.get("action_lookup").split()
+            self.attr = self.params.get("action_lookup").split()
         else:
             self.attr = []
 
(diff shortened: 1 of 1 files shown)


virttest/test_setup.py
```diff
@@ -2005,26 +2005,28 @@ def _set_polkit_conf(self):
         except Exception as e:
             raise PolkitRulesSetupError("Set polkit rules file failed: %s", e)
 
-    def setup(self):
+    def setup(self, params):
```
balamuruhans: we would pass the `params` during object creation, so it is not right to again pass it during method invocation of the same class. let's initialize `self.params = params` in `__init__()` and use it.

hsmj1412:
@balamuruhans Updated. Please help review

virttest/test_setup.py
```diff
@@ -2034,9 +2037,12 @@ def cleanup(self):
             if os.path.exists(self.libvirtd_backup_path):
                 os.rename(self.libvirtd_backup_path, self.libvirtd_path)
             if self.user.count('EXAMPLE'):
-                logging.debug("Delete the created user 'testacl'.")
```
balamuruhans: @hsmj1412 so you face issue during clean up ? if so would this change could help instead of introducing `add_polkit_user` and without params?
```python
if self.user and not self.user.count('EXAMPLE'):
    logging.debug("Delete the created user 'testacl'.")
    cmd = "userdel -r testacl"
```

hsmj1412: @balamuruhans no, the issue is when the cases give the 'testacl' instead EXAMPLE, but the testacl is not a existed user.

virttest/test_setup.py
```diff
@@ -2034,9 +2037,12 @@ def cleanup(self):
...
-                cmd = "userdel -r testacl"
+                self.user = 'testacl'
+            if self.params.get('add_polkit_user'):
```
balamuruhans: still we would not require `add_polkit_user`, we can do a similar check that we perform before creating the user,
```python
if self.user and if process.system("id %s" % self.user, ignore_status=True) == 0:
    # delete the user
```

hsmj1412: @balamuruhans Thanks, but I don't want to delete the exited user in the test env, it looks like need a flag to save the user is created by scripts or not

balamuruhans: @hsmj1412, even if anyother existing user are there in test env, self.user will be set only by the script right ? then how does it affect other user ? IMHO `add_polkit_user` is not necessary.

hsmj1412: @balamuruhans I mean some people run this script in their workstation, and they have created and used this user(testacl), so we should not delete it.

virttest/test_setup.py
```diff
@@ -2034,9 +2037,12 @@ def cleanup(self):
             if os.path.exists(self.libvirtd_backup_path):
                 os.rename(self.libvirtd_backup_path, self.libvirtd_path)
             if self.user.count('EXAMPLE'):
-                logging.debug("Delete the created user 'testacl'.")
-                cmd = "userdel -r testacl"
+                self.user = 'testacl'
```
balamuruhans: Then in that case we should not ideally do this, because if we run in workstation testacl should not be hardcoded user. always it should be taking from param `unprivileged_user`

hsmj1412: sorry, not sure you mean, we need set a unprivileged_user in case cfg file?

balamuruhans: Right, I will try to rephrase my opinion,
self.user is initialized in `__init__` with param `unprivileged_user` that we set in case cfg file
so during cleanup why do we hardcode self.user ?

hsmj1412: Not sure why due to from this class doc
"""
Create a non-privileged user 'testacl' for test if given
    'unprivileged_user' contains 'EXAMPLE', and delete the user at cleanup.
"""
I just want to create a user due to the old code will not create the user if self.user is not EXAMPLE, and dont delete the user if the use is not created by script.

PR #2010, opened 2019-03-21
env_process: add param of enable_guest_iommu to guest kernel line

xiywang:
id: 1691263

### DIFF
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -78,6 +78,11 @@ netdst = virbr0
 # present in VM, default is 'no'
 flexible_nic_index = no
 
+# Enable guest iommu (intel_iommu=on) in guest kernel cmd line
+# This parameter only support intel platform so the default
+# value is set to None.
+# enable_guest_iommu = no
+
 # List of block device object names (whitespace separated)
 images = image1
 # List of block device object names with order (whitespace separated)
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1038,6 +1038,13 @@ def preprocess(test, params, env):
             params["ker_remove_similar_pci"] = "yes"
         else:
             kernel_extra_params_remove += " pci=nomsi"
+    if (params.get("enable_guest_iommu") and
+            utils_misc.get_cpu_vendor(verbose=False) == 'GenuineIntel'):
+        enable_guest_iommu = params.get("enable_guest_iommu")
+        if enable_guest_iommu == "yes":
+            kernel_extra_params_add += " intel_iommu=on"
+        else:
+            kernel_extra_params_remove += " intel_iommu=on"
 
     # Clone master image from vms.
     base_dir = data_dir.get_data_dir()


vivianQizhu:
@yanan-fu @jingzhao84 Please help review, I think this will help fix 1662504 too. Thanks.

vivianQizhu:
@jingzhao84 Specifically, do we always need `iommu=pt` together with `intel_iommu=on`?

yanan-fu:
@vivianQizhu  yes,  this patch can fix bz 1662504 too.
@xiywang  code LGTM, but my concern is,  do we need to add it in avocado-vt.  How many cases need to add this arg in guest kernel line ?

From bz,  it is used to test `iommu_platform`,  as my understanding, test step is:
1.  boot  with `iommu_platform = on` for virtio-net-pci, (no  intel_iommu=on in guest kernel), some check ?
2.  add `intel_iommu=on`, reboot, some check ?
If we use `kernel_extra_params_add` to control it, this arg will be added during preprocess (reboot to make it take effect in this stage too). I am not sure if it can match your request.
Other wise we can use `utils_test.update_boot_option` directly in case level.

vivianQizhu:
@yanan-fu I suppose from project perspective it is needed here, actually not only virtio-net require this test, but also other devices. Considering we might need cover it in test loop(to cover all devices), here we need it in env_process anyway.

virttest/env_process.py
```diff
@@ -1038,6 +1038,12 @@ def preprocess(test, params, env):
             params["ker_remove_similar_pci"] = "yes"
         else:
             kernel_extra_params_remove += " pci=nomsi"
+    if params.get("enable_intel_iommu"):
```
luckyh: Not a good design, please also consider if it makes sense for other platforms (e.g. amd ..).

vivianQizhu: @luckyh What I can call to mind is to only enable it on supported platform, or restrict unsupported platform here. Any other suggestion? Thanks.

luckyh: Sounds good. The main point is that it is designed for controling the iommu behavior of guest, so we should point out that it is for **guest**, not intel platform, and I suggest the name `enable_guest_iommu`.

Just forgot one thing, please also add the documentation for it, thanks.

xiywang: @vivianQizhu @luckyh Updated. Please help review again. Thanks.

shared/cfg/base.cfg
```diff
@@ -78,6 +78,11 @@ netdst = virbr0
 # present in VM, default is 'no'
 flexible_nic_index = no
 
+# Enable guest iommu (intel_iommu=on) in guest kernel cmd line
+# This parameter only support intel platform so the default
+# value is set to None.
+#enable_guest_iommu = no
```
vivianQizhu: Space after `#`

xiywang: Updated.

jingzhao84:
> @jingzhao84 Specifically, do we always need `iommu=pt` together with `intel_iommu=on`?

@xiywang @vivianQizhu  yes, we always need iommu=pt` together with `intel_iommu=on` in guest kernel line.

Thanks
Jing

xiywang:
@jingzhao84 Is there any other kinds of parameters of iommu=? except 'pt'.

@vivianQizhu How about I only fix intel_iommu=on issue in this patch since the test case is included from a bug, in which the original cmd line is provided by developer with only intel_iommu=on.

vivianQizhu:
@xiywang, that's okay for me, we might add the other options later.

PR #2012, opened 2019-03-22
Update the cpu check function

hsmj1412:
The cpu will not auto to online in fedora

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -1385,12 +1385,12 @@ class BaseVM(object):
             else:
                 self.send_key(char)
 
-    def get_cpu_count(self):
+    def get_cpu_count(self, check_cmd='cpu_chk_cmd'):
         """
         Get the cpu count of the VM.
         """
         session = self.wait_for_login()
-        cmd = self.params.get("cpu_chk_cmd")
+        cmd = self.params.get(check_cmd)
         try:
             out = session.cmd_output_safe(cmd)
             return int(re.search("\d+", out, re.M).group())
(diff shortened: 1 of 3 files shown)


vivianQizhu:
@huangyum

virttest/virt_vm.py
```diff
@@ -1397,6 +1397,30 @@ def get_cpu_count(self):
         finally:
             session.close()
 
+    def get_online_cpu_count(self):
```
balamuruhans: this method is exact duplicate of `get_cpu_count()`.

balamuruhans: instead we can have fix in existing method with `check_cmd` as extra argument that could default to "cpu_chk_cmd",
```python
def get_cpu_count(self, check_cmd="cpu_chk_cmd"):
    cmd = self.params.get(check_cmd)
    try:
        out = session.cmd_output_safe(cmd)
        return int(re.search("\d+", out, re.M).group())
    finally:
        session.close()
```

hsmj1412: Thanks, a better way.

hsmj1412:
@balamuruhans updated. Please help review

shared/cfg/guest-os/Windows.cfg
```diff
@@ -27,6 +27,7 @@
     cdrom_cd1 = isos/windows/winutils.iso
 
     cpu_chk_cmd = echo %NUMBER_OF_PROCESSORS%
+    cpu_chk_all_cmd = echo %NUMBER_OF_PROCESSORS%
```
balamuruhans: how does this is different from `cpu_chk_cmd` for windows ?

hsmj1412: just avoid get cpu_chek_all_cmd in windows will return None

balamuruhans: in windows, it will never call `vm.get_cpu_count("cpu_chk_all_cmd")` as the check in L2364 will never succeed.

hsmj1412: Updated

hsmj1412:
@vivianQizhu Hi, any update? Or the PR will be merged.

vivianQizhu:
@hsmj1412 No comments, please go ahead.

PR #2015, opened 2019-03-25
test_setup: support shell command for param "physical_nic"

quanwenli:
In different kernel, the network interface will be unpredictable
changed, wen can use shell command to get the network interface
name

id:1692233

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -800,6 +800,14 @@ class PrivateBridgeConfig(object):
                 ports.append(u_port)
             self.iptables_rules = self._assemble_iptables_rules(ports)
             self.physical_nic = params.get("physical_nic")
+            if self.physical_nic:
+                if self.physical_nic.split(':', 1)[0] == "shell":
+                    self.physical_nic = decode_to_text(process.system_output(
+                        self.physical_nic.split(':', 1)[1], shell=True))
+                if self.physical_nic not in utils_net.get_host_iface():
+                    raise exceptions.TestSetupFail("Physical network '%s'"
+                                                   "does not exist" %
+                                                   self.physical_nic)
             self.force_create = False
             if params.get("bridge_force_create", "no") == "yes":
                 self.force_create = True


virttest/test_setup.py
```diff
@@ -799,7 +799,13 @@ def __init__(self, params=None):
...
                 ports.append(u_port)
             self.iptables_rules = self._assemble_iptables_rules(ports)
-            self.physical_nic = params.get("physical_nic")
+            if process.system(params.get("physical_nic"),
+                              shell=True, ignore_status=True) == 0:
+                self.physical_nic = decode_to_text(
+                        process.system_output(params.get("physical_nic"),
+                                              shell=True))
+            else:
+                self.physical_nic = params.get("physical_nic")
```
vivianQizhu: You might exectue `params.get("physical_nic")` only once and also execute process.system_output() only once.

vivianQizhu:
@xiywang @PaulYuuu Please help review, thanks.

virttest/test_setup.py
quanwenli: yes, it looks repetitive．I have update the code. please help a review. thanks

virttest/test_setup.py
```diff
@@ -799,7 +799,13 @@ def __init__(self, params=None):
...
                 ports.append(u_port)
             self.iptables_rules = self._assemble_iptables_rules(ports)
-            self.physical_nic = params.get("physical_nic")
+            physical_nic = params.get("physical_nic")
+            cmd_result = process.run(physical_nic, shell=True,
+                                     ignore_status=True)
+            if cmd_result.exit_status == 0:
+                self.physical_nic = cmd_result.stdout_text
+            else:
+                self.physical_nic = physical_nic
```
vivianQizhu: Code can be optimized to:
```
self.physical_nic = params.get("physical_nic")
cmd_result = process.run(self.physical_nic, shell=True, ignore_status=True)
if cmd_result.exit_status == 0:
    self.physical_nic = cmd_result.stdout_text
```
But I don't think that's a good idea to mix up the definition of "physical_nic" this way, to make it can be either a fixed variant or an command.
`physical_nic = enp5s0f1`
`physical_nic = ip link | sed -n '/90:e2:ba:7d:33:3d/{g;1!p;};h'|awk -F': ' '{print $2}'|head -1`
Even for the command, to get it from a hard code mac address in the command is also not a good idea.

quanwenli: Thanks, qizhu, I have update the codes.

it's very usefully in multiple ethernet machine. like 3 or more ethernet, it's not good to re-find the network name in different system(like rhel7. rhel8), if the network name could be find with mac address, we don't care about it and it provides more automated．

@luckyh, do you have any commands?

luckyh: Well, it makes sense to me that you want a dynamic way to get a physical nic, but I'd like to request you to explicitly point out that the value string is a shell command, e.g.
```
physical_nic = shell:ip link | sed ...
```
thus, you can exactly know whether you should call process.run
```python
if physical_nic.split(':', 1) == "shell":
    self.physical_nic = ...
```

PaulYuuu:
1. we need to check if `physical_nic` exists in `/sys/class/net/` (`utils_net.get_host_iface()`)
2. If you want to call host to get the interface, make `physical_nic` empty(optional, refer to 1) and use `physical_nic_cmd` to let the code clearer.
3. I want to confirm what scenarios we need to use the MAC address to get the interface name. mac address could be found in `/sys/class/net/$iface_name/address`, and then we can return this $iface_name in utils_net.get_host_iface()

virttest/test_setup.py
quanwenli: @luckyh , I have update the code. please help review. thanks

quanwenli:
>     1. we need to check if `physical_nic` exists in `/sys/class/net/` (`utils_net.get_host_iface()`)
> 
yes, we need to check it.  I have update the code, please help review.

>     2. If you want to call host to get the interface, make `physical_nic` empty(optional, refer to 1) and use `physical_nic_cmd` to let the code clearer.

I accept @luckyh 's suggestion and thanks for your comment.
> 
>     3. I want to confirm what scenarios we need to use the MAC address to get the interface name. mac address could be found in `/sys/class/net/$iface_name/address`, and then we can return this $iface_name in utils_net.get_host_iface()
In network performance testing, there are several nics to be tested in one loop.

vivianQizhu:
@quanwenli If I understand correctly, when you test different nics in your loop, you still need to change the mac address in the command line for each nic, I don't see significant improvement in this scenario comparing to change their device name.

quanwenli:
@vivianQizhu, I will send another test_loops patch in kar to add variants 82599/xl710, in each variants, it uses unique mac address. so do not need to change it for rhel7, rhel8 or other system.

PR #2016, opened 2019-03-26
Refactor screen inactivity detection

lolyu:
Issues with previous implementation:
1. It will floods test.background_error with redundant inactivity
errors.
2. Consider the following screendump sequence:
Ti is the checking point sequence
Sj is the screendump produced at checking point Ti
------------------------------------------------
T0->T1->T2->T3->T4->T5->T6->T7->T8->T9
S0->S1->S2->S3->S4->S5->S6->S0->S1->S2
------------------------------------------------
if vm is rebooted at T7, and same booting screen sequence is replayed,
if T7 - T0 > inactivity_threshold, an inactivity error will be reported,
which is a false alarm.

In this patch, generator `screen_inactivity_timer` is introduced to
check the inactivity. And each vm will be allocated with a timer,
which is stored globally in `virt_vm.timers`. So vm could have access to
reset the time in case like vm pause.

After this patch:
1. only one inactivity error will be reported for one inactivity period.
2. inactivity is detected only if the screendumps in consecutive
checking points are the same.

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -6,6 +6,7 @@ import os
 import re
 import socket
 import traceback
+import threading
 
 from aexpect.exceptions import ShellError
 from aexpect.exceptions import ExpectError
@@ -13,6 +14,7 @@ from avocado.core import exceptions
 
 import six
 from six.moves import xrange
+from collections import OrderedDict
 
 from virttest import utils_misc
 from virttest import utils_net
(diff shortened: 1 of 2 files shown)


virttest/virt_vm.py
```diff
@@ -464,6 +472,72 @@ def __init__(self, model=None, vendor=None, flags=None, family=None,
...
+
+    :param vm: vm instance
+    :param threshold: inactivity_treshold
+    :param delay: screendump_delay
+    :param watcher: inactivity_watcher
+    """
+    cache = OrderedDict()
+    ret = None
+    while True:
+        op, image_hash = yield ret
```
hereischen: What does this line mean?

PR #2017, opened 2019-03-26
test_setup: add several functions to handle hugepages

hsmj1412:
Add 2 functions to achieve following:
1.set kernel hugepages
2.get the number of certain hugepages totally allocated by kernel
Update 1 function:
1. get the number of free hugepages for given NUMA node

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -24,6 +24,7 @@ from avocado.utils import distro
 from avocado.core import exceptions
 
 from virttest import data_dir
+from virttest import kernel_interface
 from virttest import error_context
 from virttest import utils_misc
 from virttest import versionable_class
@@ -577,21 +578,51 @@ class HugePageConfig(object):
             raise ValueError("Root hugepage control sysfs directory %s did not"
                              " exist" % self.pool_path)
 
-    def get_node_num_huge_pages(self, node, pagesize):
+    def get_kernel_hugepages(self, pagesize):
+        """
+        Get specific hugepages number allocated by kernel at runtime
+        read page number from
+        /sys/kernel/mm/hugepages/hugepages-${pagesize}kB/nr_hugepages
+
+        :param pagesize: string or int, page size in kB
+        :return: page number
+        """
+        pgfile = "%s/hugepages-%skB/nr_hugepages" % (self.pool_path, pagesize)
+
+        obj = kernel_interface.SysFS(pgfile)
+        return obj.sys_fs_value
+
+    def set_kernel_hugepages(self, pagesize, pagenum):
+        """
+        Let kernel allocate some specific hugepages at runtime
+        write page number to
+        /sys/kernel/mm/hugepages/hugepages-${pagesize}kB/nr_hugepages
+
+        :param pagesize: string or int, page size in kB
+        :param pagenum: page number
+        """
+        pgfile = "%s/hugepages-%skB/nr_hugepages" % (self.pool_path, pagesize)
+
+        obj = kernel_interface.SysFS(pgfile)
+        obj.sys_fs_value = pagenum
+
+    def get_node_num_huge_pages(self, node, pagesize, type="total"):
         """
         Get number of pages of certain page size under given numa node.
 
         :param node: string or int, node number
         :param pagesize: string or int, page size in kB
+        :param type: total pages or free pages
         :return: int, node huge pages number of given page size
         """
+        if type == 'free':
+            ptype = "free_hugepages"
+        else:
+            ptype = "nr_hugepages"
         node_page_path = "%s/node%s" % (self.sys_node_path, node)
-        node_page_path += "/hugepages/hugepages-%skB/nr_hugepages" % pagesize
-        if not os.path.isfile(node_page_path):
-            raise ValueError("%s page size nr_hugepages file of node %s did "
-                             "not exist" % (pagesize, node))
-        out = decode_to_text(process.system_output("cat %s" % node_page_path))
-        return int(out)
+        node_page_path += "/hugepages/hugepages-%skB/%s" % (pagesize, ptype)
+        obj = kernel_interface.SysFS(node_page_path)
+        return obj.sys_fs_value
 
     def set_node_num_huge_pages(self, num, node, pagesize):
         """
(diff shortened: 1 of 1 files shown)


hsmj1412:
Update from #1289 
REF:
1. avocado.utils.memory and virttest.staging.utils_memory only handle default hugepages size.

below are relevant commands used by similar fuctions in avocado.utils.memory or virttest.staging.utils_memory:
grep Hugepagesize /proc/meminfo --> get default hugepage size
/sbin/sysctl vm.nr_hugepages --> get number of default hugepage size

we can get that those fuctions is only to handle default hugepages size, such as 2048K, so it can't be used to set/get for other huge pages, such as 1G.

So provide more common function to handle each hugepages size.

virttest/test_setup.py
```diff
@@ -577,16 +577,58 @@ def get_multi_supported_hugepage_size(self):
             raise ValueError("Root hugepage control sysfs directory %s did not"
                              " exist" % self.pool_path)
 
-    def get_node_num_huge_pages(self, node, pagesize):
+    def get_kernel_hugepages(self, pagesize):
```
balamuruhans: please avoid redundant code by abstracting methods `get_node_num_huge_pages()` and `set_node_num_huge_pages()` with your requirement of different sysfs path.

hsmj1412: @balamuruhans you mean merge the get_node_num_huge_pages and set_node_num_huge_pages? Due to the code have been used before, it maybe affect some existed cases in tp-libvirt and tp-qemu

balamuruhans: no we should always maintain backward compatibility, I meant to reuse `kernel_interface.py` APIs
```python
def func(self, node_path, value=None):
    hp = kernel_interface.SysFS(node_path)
    if value:
        hp.sys_fs_value = value
        return hp.sys_fs_value == str(value)
     return hp.sys_fs_value
def existing_func(self): # to maintain compatibility
     node_path = "/sys/devices/system/node//hugepages/hugepages-%skB/nr_hugepages"
     return func(node_path)
def new_func(self):
     mem_path = "/sys/kernel/mm/hugepages/hugepages-${pagesize}kB/nr_hugepages"
     return func(mem_path)
```

hsmj1412: @balamuruhans Updated, please help review again

virttest/test_setup.py
```diff
@@ -577,21 +578,49 @@ def get_multi_supported_hugepage_size(self):
...
+        obj = kernel_interface.SysFS(pagefile)
+        return obj.sys_fs_value
+
+    def set_kernel_hugepages(self, pagesize, pagenum):
+        """
+        Let kernel allocate some specific hugepages at runtime
+         write page number to
+        /sys/kernel/mm/hugepages/hugepages-${pagesize}kB/nr_hugepages
+         :param pagesize: hugepages size
+        :param pagenum: page number
```

virttest/test_setup.py
```diff
@@ -577,21 +578,49 @@ def get_multi_supported_hugepage_size(self):
...
-        if not os.path.isfile(node_page_path):
-            raise ValueError("%s page size nr_hugepages file of node %s did "
-                             "not exist" % (pagesize, node))
-        out = decode_to_text(process.system_output("cat %s" % node_page_path))
-        return int(out)
+        node_page_path += "/hugepages/hugepages-%skB/%s" % (pagesize, ptype)
+        obj = kernel_interface.SysFS(node_page_path)
+        return obj.sys_fs_value
 
     def set_node_num_huge_pages(self, num, node, pagesize):
```
balamuruhans: would you mind fixing it for this method ?

balamuruhans:
@huangyum request for your review. Thanks

virttest/test_setup.py
hsmj1412: @balamuruhans Updated

huangyum: @hsmj1412  Forgot to push?

hsmj1412: @huangwei sorry, updated

virttest/test_setup.py
```diff
@@ -577,21 +578,49 @@ def get_multi_supported_hugepage_size(self):
             raise ValueError("Root hugepage control sysfs directory %s did not"
                              " exist" % self.pool_path)
 
-    def get_node_num_huge_pages(self, node, pagesize):
+    def get_kernel_hugepages(self, pagesize):
+        """
+        Get specific hugepages number allocated by kernel at runtime
+        read page number from /sys/kernel/mm/hugepages/hugepages-${pagesize}kB/nr_hugepages
+
+        :param pagesize: hugepages size
```
huangyum: I guess it's better to explicitly tell the unit of pagesize should be KB.

virttest/test_setup.py
```diff
@@ -603,11 +632,9 @@ def set_node_num_huge_pages(self, num, node, pagesize):
...
         node_page_path = "%s/node%s" % (self.sys_node_path, node)
         node_page_path += "/hugepages/hugepages-%skB/nr_hugepages" % pagesize
-        if not os.path.isfile(node_page_path):
-            raise ValueError("%s page size nr_hugepages file of node %s did "
-                             "not exist" % (pagesize, node))
-        process.system("echo %s > %s" % (num, node_page_path), shell=True)
-        if int(num) != self.get_node_num_huge_pages(node, pagesize):
+        obj = kernel_interface.SysFS(node_page_path)
+        obj.sys_fs_value = num
+        if obj.sys_fs_value != int(num):
```
huangyum: `num` should always be int.  Would you please correct the parameter description as well, for both set_ and get_ function, thanks.

virttest/test_setup.py
hsmj1412: @huangyum Updated

virttest/test_setup.py
hsmj1412: Due to some num maybe get a str from the config, and the coder forget to turn into int.

balamuruhans:
@hsmj1412 also please have commit msg much more clear with the changes that are done. Let us get the patch in after the review comments are addressed.

hsmj1412:
@balamuruhans @huangyum Updated

balamuruhans:
Thanks @huangyum for review and testing, can we merge it ?

huangyum:
@balamuruhans Yes, I think so. Thanks.

PR #2018, opened 2019-03-27
allow libvirt.check_result to take a string as expected error message

brucepatricklee:
it's common to take a string as an expected error message. So here
allow this kind of usage

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1585,20 +1585,28 @@ def check_result(result,
     expectation.
 
     :param result: Command result instance.
-    :param expected_fails: list of regex of expected stderr patterns. The check
-                           will pass if any of these patterns matches.
-    :param skip_if: list of regex of expected patterns. The check will raise a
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1613,6 +1613,8 @@ def check_result(result,
 
     if result.exit_status:
         if expected_fails:
+            if isinstance(expected_fails, str):
+                expected_fails = [expected_fails]
```
hsmj1412: try:
    unicode
except NameError:
    unicode = str

if isinstance(expected_fails, (str, unicode)):

brucepatricklee: updated

brucepatricklee: updated

kylazhang:
please also update the explanation for the parameters if str is acceptable

brucepatricklee:
@kylazhang parameter explanation updated

PR #2023, opened 2019-03-28
utils_net: support shell command for param netdst_nic

quanwenli:
In different kernel, the network interface will be unpredictable
changed, wen can use shell command to get the network interface
name for param netdst_nic

id: 1692276

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -2831,6 +2831,12 @@ class ParamsNet(VMNet):
                 except IndexError:
                     existing_value = None
                 nic_dict[propertea] = nic_params.get(propertea, existing_value)
+                if propertea == "netdst" and "shell:" in nic_dict[propertea]:
+                    nic_dict[propertea] = process.getoutput(
+                        nic_dict[propertea].split(':', 1)[1])
+                    if not nic_dict[propertea]:
+                        raise exceptions.TestError(
+                            "netdst is null, please check the shell command")
             result_list.append(nic_dict)
         VMNet.__init__(self, self.container_class, result_list)
 


virttest/utils_net.py
```diff
@@ -2688,7 +2688,14 @@ def __init__(self, params, vm_name):
                     existing_value = None
                 except IndexError:
                     existing_value = None
-                nic_dict[propertea] = nic_params.get(propertea, existing_value)
+                if propertea == "netdst" and existing_value is not None:
+                    if existing_value.split(':', 1)[0] == "shell":
+                        existing_value = decode_to_text(process.system_output(
+                            existing_value.split(':', 1)[1], shell=True))
+                    nic_dict[propertea] = existing_value
```
vivianQizhu: Seems it change the original logic a little, when `propertea == "netdst" and existing_value is not None`, originally if `nic_params[propertea]` exist it will not use `existing_value`, but now it ignore `nic_params[propertea]`, not sure how other utils_net users use it, so I will prefer you keep original logic and only fit your own need when `existing_value.split(':', 1)[0] == "shell"`

vivianQizhu: Also need check if `existing_value` you get is expected, the he code might be:
```
nic_dict[propertea] = nic_params.get(propertea, existing_value)
if propertea == "netdst" and existing_value is not None:
    if existing_value.split(':', 1)[0] == "shell":
       existing_value = process.getoutput(existing_value.split(':', 1)[1], shell=True)
       if existing_value is not None:
           nic_dict[propertea] = existing_value
       else:
           # raise error here since the shell command might not be correct
```

vivianQizhu: Or try this:
```
if propertea == "netdst" and existing_value is not None:
    if existing_value.split(':', 1)[0] == "shell":
       existing_value = process.getoutput(existing_value.split(':', 1)[1], shell=True)
       if existing_value is not None:
           nic_dict[propertea] = existing_value
       else:
           # raise error here since the shell command might not be correct
       continue
nic_dict[propertea] = nic_params.get(propertea, existing_value)
```

quanwenli: @vivianQizhu , could you hep review, I have updated code and test results in bz.

virttest/utils_net.py
```diff
@@ -2831,6 +2831,17 @@ def __init__(self, params, vm_name):
...
+                    if existing_value.split(':', 1)[0] == "shell":
+                        existing_value = decode_to_text(process.system_output(
+                            existing_value.split(':', 1)[1], shell=True))
+                        if existing_value:
+                            nic_dict[propertea] = existing_value
+                        else:
+                            raise exceptions.TestError(
+                                "netdst is null,please check the shell command")
+                    else:
+                        nic_dict[propertea] = existing_value
```
vivianQizhu: This change seems meaningless, it has nothing different with your original code. Previously even `if propertea == "netdst" and existing_value is not None`, `nic_dict[propertea]` should equals to `nic_params[propertea]` when it exists, you still change it.

quanwenli: @vivianQizhu , I reassign the code and will not touch "existing_value ", could you help review , the test results has been updated in bz.

PR #2024, opened 2019-03-28
utils_net: verify_ip_address_ownership() on remote host if VM is migrated

balamuruhans:
`verify_ip_address_ownership()` fails when ip address of VM is queried
after it is migrated to target host and cause the tests to error, handle
it using session to check on the remote host.

Signed-off-by: Balamuruhan S <bala24@linux.vnet.ibm.com>

PR #2025, opened 2019-03-28
utils_test/libvirt: post migration check query vm address from remote host

balamuruhans:
As post migration check, pinging the vm to verify its network status and
vm availability is performed but vm.get_address() would fail to get
IP address by querying verify_ip_address_ownership() as it checks from
source host, so handle it by passing remote host session object to
make it work


balamuruhans:
depends on: https://github.com/avocado-framework/avocado-vt/pull/2024

balamuruhans:
@sathnaga Thanks for reviewing, I have tested this patch.

PR #2026, opened 2019-03-29
libvirt: Add an utility to run remote command

dzhengfy:
This is a function dedicated to run a command on remote host.

### DIFF
--- a/virttest/remote.py
+++ b/virttest/remote.py
@@ -961,6 +961,44 @@ def copy_files_from(address, client, username, password, port, remote_path,
                                    "values are scp and rss" % client)
 
 
+def run_remote_cmd(cmd, params, remote_runner=None, ignore_status=True):
+    """
+    A function to run a command on remote host.
+
+    :param cmd: the command to be executed
+    :param params: the parameter for executing
+    :param remote_runner: a remote runner object on remote host
+    :param ignore_status: True - not raise exception when failed
+                          False - raise exception when failed
+
+    :return: CmdResult object
+    :raise: exceptions.TestFail or exceptions.TestError
+    """
+    try:
+        if not remote_runner:
+            remote_ip = params.get("server_ip", params.get("remote_ip"))
+            remote_pwd = params.get("server_pwd", params.get("remote_pwd"))
+            remote_user = params.get("server_user", params.get("remote_user"))
+            remote_runner = RemoteRunner(host=remote_ip,
+                                         username=remote_user,
+                                         password=remote_pwd)
+
+        cmdresult = remote_runner.run(cmd, ignore_status=ignore_status)
+        logging.debug("Remote runner run result:\n%s", cmdresult)
+        if cmdresult.exit_status and not ignore_status:
+            raise exceptions.TestFail("Failed to run '%s' on remote: %s"
+                                      % (cmd, cmdresult))
+        return cmdresult
+    except (LoginError, LoginTimeoutError,
+            LoginAuthenticationError, LoginProcessTerminatedError) as e:
+        logging.error(e)
+        raise exceptions.TestError(e)
+    except process.CmdError as cmderr:
+        logging.error("Remote runner run failed:\n%s", cmderr)
+        raise exceptions.TestFail("Failed to run '%s' on remote: %s"
+                                  % (cmd, cmderr))
+
+
 class Remote_Package(object):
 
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -3665,6 +3665,33 @@ def check_qemu_cmd_line(content, err_ignore=False):
...
+        remote_user = params.get("server_user", params.get("remote_user"))
+
+        remote_runner = remote.RemoteRunner(host=remote_ip,
+                                            username=remote_user,
+                                            password=remote_pwd)
+    cmdResult = remote_runner.run(cmd, ignore_status=True)
+    logging.debug("Remote runner run result:\n%s", cmdResult)
+    if cmdResult.exit_status:
+        raise exceptions.TestFail("Failed to run '%s' on remote: %s"
+                                  % (cmd, results_stderr_52lts(cmdResult).strip()))
```
kylazhang: How about add ignore_status? Maybe some usage do not want testfail if status error.

virttest/utils_test/libvirt.py
```diff
@@ -3665,6 +3665,33 @@ def check_qemu_cmd_line(content, err_ignore=False):
     return True
 
 
+def run_remote_cmd(cmd, params, remote_runner=None, ignore_status=True):
```
balamuruhans: It is better to place this method in `remote.py`

virttest/utils_test/libvirt.py
```diff
@@ -3665,6 +3665,33 @@ def check_qemu_cmd_line(content, err_ignore=False):
...
+    """
+    if not remote_runner:
+        remote_ip = params.get("server_ip", params.get("remote_ip"))
+        remote_pwd = params.get("server_pwd", params.get("remote_pwd"))
+        remote_user = params.get("server_user", params.get("remote_user"))
+
+        remote_runner = remote.RemoteRunner(host=remote_ip,
+                                            username=remote_user,
+                                            password=remote_pwd)
+    cmdResult = remote_runner.run(cmd, ignore_status=ignore_status)
```
balamuruhans: This doesn't work, we need to handle with `try` block.

virttest/utils_test/libvirt.py
dzhengfy: updated.

virttest/utils_test/libvirt.py
dzhengfy: updated

virttest/utils_test/libvirt.py
dzhengfy: updated

virttest/remote.py
```diff
@@ -961,6 +961,43 @@ def copy_files_from(address, client, username, password, port, remote_path,
                                    "values are scp and rss" % client)
 
 
+def run_remote_cmd(cmd, params, remote_runner=None, ignore_status=True):
+    """
+    A function to run a command on remote host.
+
+    :param cmd: the command to be executed
+    :param params: the parameter for executing
+    :param remote_runner: a remote runner object on remote host
```
kylazhang: missing 1 parameters. Others looks good for me

virttest/remote.py
```diff
@@ -961,6 +961,45 @@ def copy_files_from(address, client, username, password, port, remote_path,
...
+            remote_runner = RemoteRunner(host=remote_ip,
+                                         username=remote_user,
+                                         password=remote_pwd)
+
+        cmdresult = remote_runner.run(cmd, ignore_status=ignore_status)
+        logging.debug("Remote runner run result:\n%s", cmdresult)
+        if cmdresult.exit_status and not ignore_status:
+            raise exceptions.TestFail("Failed to run '%s' on remote: %s"
+                                      % (cmd, cmdresult))
+        return (cmdresult, remote_runner)
```
chunfuwen: I assume return cmdresult is enough, rather than cmdresult, remote_runner

dzhengfy: @chunfuwen , it is useful to return remote_runner because the client code can reuse this remote runner to execute other commands in following steps.

virttest/remote.py
dzhengfy: updated

virttest/remote.py
dzhengfy: @chunfuwen , updated according to discussion

PR #2027, opened 2019-03-29
Define a function in ceph.py to create ceph.conf file

meinaLi:
Define a function in ceph.py to create ceph.conf file, which will need to be used in other several scripts
Signed-off-by: Meina Li <meili@redhat.com>


meinaLi:
This function will be used in https://github.com/autotest/tp-libvirt/pull/2060

PR #2028, opened 2019-03-29
Add support for vhostuser+dpdk testing

kylazhang:

### DIFF
--- a/virttest/openvswitch.py
+++ b/virttest/openvswitch.py
@@ -294,6 +294,9 @@ class OpenVSwitchControlCli_140(OpenVSwitchControl):
     def list_br(self):
         return results_stdout_52lts(self.ovs_vsctl(["list-br"])).splitlines()
 
+    def list_interface(self):
+        return self.ovs_vsctl(["list", "interface"]).stdout_text.strip()
+
     def add_port(self, br_name, port_name):
         self.ovs_vsctl(["add-port", br_name, port_name])
 
@@ -457,7 +460,7 @@ class OpenVSwitch(OpenVSwitchSystem):
         """
         self.tmpdir = "/%s/openvswitch" % (tmpdir)
         try:
-            os.mkdir(self.tmpdir)
+            os.makedirs(self.tmpdir)
         except OSError as e:
             if e.errno != 17:
                 raise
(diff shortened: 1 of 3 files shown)


virttest/openvswitch.py
```diff
@@ -473,30 +476,65 @@ def init_db(self):
...
     def start_ovs_vswitchd(self):
+        """
+        Start ovs vswitch daemon with only 1 socket
+        """
+        self.ovs_vsctl(["--no-wait", "set", "Open_vSwitch",
+                        ".", "other_config:dpdk-init=true"])
+        self.ovs_vsctl(["--no-wait", "set", "Open_vSwitch",
+                        ".", "other_config:dpdk-socket-mem='1024'"])
+        self.ovs_vsctl(["--no-wait", "set", "Open_vSwitch",
+                        ".", "other_config:dpdk-lcore-mask='0x1'"])
```
chunfuwen: Do we need check vswitchd status after those operations?

virttest/openvswitch.py
```diff
@@ -473,30 +476,65 @@ def init_db(self):
...
         process.run('%s %s %s %s' %
                     (path.find_command("ovs-vswitchd"),
                      "--detach",
                      "--pidfile=%s" % self.ovs_pidfile,
                      "unix:%s" % self.db_socket))
 
+    def create_br_port(self, br_name, port_names):
```
chunfuwen: suggest separate creating and port creating into two functions. In this way, each time I want to add port, it need create bridge again.

virttest/utils_net.py
```diff
@@ -1191,6 +1195,68 @@ def wrap_init(*args, **kargs):
...
+
+    :param run_dir: openvswitch run dir
+    :param selinux_mode: permissive or enforcing
+    :param page_size: size for setting hugepage
+    :param clean_ovs: clean ovs service or not
+    """
+    # Clean dir
+    if os.path.exists(run_dir):
+        shutil.rmtree(run_dir)
+    os.mkdir(run_dir)
```
chunfuwen: is any reason to create the folder again since it is cleanup logic?

virttest/openvswitch.py
kylazhang: no need, it run process.run and will report error when failed.

virttest/utils_net.py
kylazhang: As the dir exist before the operation

virttest/openvswitch.py
kylazhang: updated, thanks

kylazhang:
show result here https://github.com/autotest/tp-libvirt/pull/2107#issuecomment-584490499

virttest/openvswitch.py
```diff
@@ -469,30 +472,72 @@ def init_db(self):
...
+        """
+        Add ports into bridge
+
+        :param br_name: name of bridge
+        :param port_names: port names split by space
+        """
+        for port_name in port_names.split():
+            self.ovs_vsctl(["add-port", br_name, port_name, "-- set Interface",
+                            port_name, "type=dpdkvhostuser"])
+            process.run('chown qemu:qemu %s/%s' % (self.pid_files_path, port_name))
```
chunfuwen: add shell=true for process.run()

virttest/utils_net.py
```diff
@@ -30,6 +31,9 @@
 from virttest import propcan
 from virttest import utils_misc
 from virttest import arch
+from virttest import utils_selinux
+from virttest import utils_package
+from virttest.staging import utils_memory
```
chunfuwen: add one blank line to separate them

virttest/openvswitch.py
kylazhang: It can work well and no need to add it.

virttest/utils_net.py
kylazhang: ok

virttest/openvswitch.py
chunfuwen: it may cause some unexpected issue on some other linux distribution, suggest add shell=true

virttest/utils_net.py
dzhengfy: If you are sure the dir exists before the operation, do we need to check its existence in L1244?

kylazhang: It just to make sure that the operation are done valuable

virttest/openvswitch.py
kylazhang: updated

PR #2033, opened 2019-04-03
utils_net.py: add traceview check for windows guests

kimi1978wy:
check queues and msi via traceview

Signed-off-by: Yu Wang <wyu@redhat.com>

bz:1485225

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -29,6 +29,8 @@ from virttest import utils_misc
 from virttest import arch
 from virttest.compat_52lts import results_stdout_52lts, results_stderr_52lts, decode_to_text
 from virttest.versionable_class import factory
+from virttest.utils_windows import virtio_win
+from virttest.utils_windows import system
 
 
 try:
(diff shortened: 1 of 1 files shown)


kimi1978wy:
(1/4) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.Win7.i386.sp1.io-github-autotest-qemu.boot_with_max_queues: |PASS (215.39 s)
 (2/4) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.Win7.x86_64.sp1.io-github-autotest-qemu.boot_with_max_queues: PASS (228.48 s)
 (3/4) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.Win8.i386.io-github-autotest-qemu.boot_with_max_queues: PASS (252.92 s)
 (4/4) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.Win8.x86_64.io-github-autotest-qemu.boot_with_max_queues: PASS (235.03 s)

virttest/utils_net.py
```diff
@@ -3744,3 +3746,190 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
                 return False
     logging.info("All the hostnames and IPs are mapped in %s", hosts_file)
     return True
+
+
+def _remove_file(session, file_path):
```
vivianQizhu: The implementation is windows only but hard to tell from the function name, also it is only used once, so I would recommend removing the function. You might have: `clean_cmd = "del %s"` and L#3897: `session_serial.cmd(clean_cmd % log_path)`, L#3922: `session_serial.cmd(clean_cmd % dump_file)`.

virttest/utils_net.py
```diff
@@ -3744,3 +3746,190 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+
+    :param file_path: file path to remove
+    :param session: a logined session
+    """
+
+    clean_cmd = "del %s" % file_path
+    session.cmd_status(clean_cmd)
+
+
+def _copy_file(session, src_path, dst_folder="c:\\"):
```
vivianQizhu: For same reason suggest removing this function.

virttest/utils_net.py
```diff
@@ -3744,3 +3746,190 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+
+
+def _get_traceview_path(session, params):
+    """
+    :param session: a session to send cmd
+    :param params: the test params
+    :return: the proper traceview path
+    """
+
+    traceview_path_template = "WIN_UTILS:\\traceview\\%s\\%s\\traceview.exe"
```
vivianQizhu: What about keep it configurable in params, and have this path as default value?
```
traceview_path_template = params.get('traceview_path_template', "WIN_UTILS:\\traceview\\%s\\%s\\traceview.exe")
```

virttest/utils_net.py
```diff
@@ -3744,3 +3746,190 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+    # traceview_ver = params["traceview_version"]
```
vivianQizhu: Useless?

virttest/utils_net.py
```diff
@@ -3744,3 +3746,190 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+    traceview_ver = "win10"
+    os_version = system.version(session)
```
vivianQizhu: Is there a module named `system`? Or you mean `sys.version`? I guess you might get `os_variant` from `params`.

virttest/utils_net.py
```diff
@@ -3744,3 +3746,190 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+    main_ver = int(os_version.split('.')[0])
+    if main_ver < 10:
+        traceview_ver = "win8"
+    os_arch = "%PROCESSOR_ARCHITECTURE%"
```
vivianQizhu: Seems this is a fixed string, unnecessary to have a `%s` for it in `traceview_path_template`.

virttest/utils_net.py
```diff
@@ -3744,3 +3746,190 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+    # copy traceview.exe
+    logging.info("Copy traceview.exe to drive C")
+    traceview_path = _get_traceview_path(session, params)
+    _copy_file(session, traceview_path)
+
+    # copy Netkvm.pdb
+    logging.info("Find Netkvm.pdb and copy to drive C")
+    viowin_ltr = virtio_win.drive_letter_iso(session)
+    if not viowin_ltr:
+        logging.error("Could not find virtio-win drive in guest")
```
vivianQizhu: Seems you can't continue the rest steps without a correct `viowin_ltr`, so `logging.error` is not enough here, you might want raise error exception or return False to let user(test case) be aware of the failure.

virttest/utils_net.py
```diff
@@ -3744,3 +3746,190 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+    if not guest_arch:
+        logging.error("Could not get architecture dirname of the vm")
+
+    driver_name = params.get("driver_name", "netkvm")
+    pdb_middle_path = "%s\\%s" % (guest_name, guest_arch)
+    pdb_find_cmd = 'dir /b /s %s\\%s.pdb | findstr "\\%s\\\\"'
+    pdb_find_cmd %= (viowin_ltr, driver_name, pdb_middle_path)
+    pdb_path = session.cmd(pdb_find_cmd, timeout=timeout).strip()
+    logging.info("Found Netkvm.pdb file at '%s'", pdb_path)
+    _copy_file(session, pdb_path)
```
vivianQizhu: Better return the full path of traceview and pdb in this function, instead of hard coding at L#3899-3900.

virttest/utils_net.py
```diff
@@ -3744,3 +3746,190 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+    dump_cmd %= (traceview_local_path, log_path, pdb_local_path, dump_file)
+    status, output = session_serial.cmd_status_output(dump_cmd)
+    if status:
+        logging.error("Cann't dump log file %s: %s" % (log_path, output))
+    time.sleep(100)
+    session_serial.cmd("taskkill /im traceview.exe")
+    time.sleep(10)
+    status, output = session_serial.cmd_status_output("type %s" % dump_file)
+    if status:
+        logging.error("Cann't read dumped file %s: %s" % (dump_file, output))
```
vivianQizhu: Same for these `logging.error`s, we need raise exception or return False/None to interrupt the execution when it can't continue the following steps.

virttest/utils_net.py
kimi1978wy: Yes, there is system module in utils_windows (line 33: from virttest.utils_windows import system).

virttest/utils_net.py
```diff
@@ -3744,3 +3746,166 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+    # checkout traceview output
+    logging.info("Check etl file generated by traceview")
+    dump_file = "c:\\trace.txt"
+    session_serial.cmd(clean_cmd % dump_file)
+    dump_cmd = "%s -process %s -pdb %s -o %s"
+    dump_cmd %= (traceview_local_path, log_path, pdb_local_path, dump_file)
+    status, output = session_serial.cmd_status_output(dump_cmd)
+    if status:
+        logging.error("Cann't dump log file %s: %s" % (log_path, output))
+    time.sleep(100)
```
lolyu: is `time.sleep` a good approach to wait for the file-dumping and process-killing? I believe it is better to use `wait_for`

kimi1978wy:
@vivianQizhu @lolyu pls help to review again, thanks!

virttest/utils_net.py
```diff
@@ -3744,3 +3746,198 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+    Check the dump file size periodically, untill the file size doesn't change,
+    considered the dump process has finished. Then kill the idled progress.
+
+    :param session: a session to send command
+    :param dump_file_path: the dump file to check
+    """
+    def _get_file_size(last_size, file_path):
+        check_file_size_cmd = "for %%I in (%s) do @echo %%~zI" % file_path
+        status, output = session.cmd_status_output(check_file_size_cmd)
+        if status:
```
hereischen:
```python
        if status or not output.isdigit()::
             return False
         file_size = int(output)
         if file_size != last_size[0]:
             last_size[0] = file_size
             return False
         return True
```

virttest/utils_net.py
```diff
@@ -3744,3 +3746,195 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+
+    pdb_middle_path = "%s\\%s" % (guest_name, guest_arch)
+    pdb_find_cmd = 'dir /b /s %s\\%s.pdb | findstr "\\%s\\\\"'
+    pdb_find_cmd %= (viowin_ltr, driver_name, pdb_middle_path)
+    pdb_path = session.cmd(pdb_find_cmd).strip()
+    logging.info("Found %s.pdb file at %s" % (driver_name, pdb_path))
+    return pdb_path
+
+
+def _prepare_traceview_windows(test, params, vm, session, timeout=360):
```
vivianQizhu: Seems the parameter `test` is useless now.

virttest/utils_net.py
```diff
@@ -3744,3 +3746,195 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+
+    last_size = [0]
+    utils_misc.wait_for(lambda: _get_file_size(last_size, dump_file_path),
+                        timeout=timeout,
+                        step=10.0)
+    kill_cmd = "taskkill /im traceview.exe"
+    session.cmd(kill_cmd)
+
+
+def get_MSIs_and_queues_windows(test, params, vm, timeout=360):
```
vivianQizhu: So does here.

kimi1978wy:
@hereischen @vivianQizhu I have updated , pls help to review again, thanks a lot

hereischen:

virttest/utils_net.py
```diff
@@ -3744,3 +3746,195 @@ def map_hostname_ipaddress(hostname_ip_dict, session=None):
...
+        status, output = session.cmd_status_output(check_file_size_cmd)
+        if status or not output.isdigit():
+            return False
+        file_size = int(output)
+        if file_size != last_size[0]:
+            last_size[0] = file_size
+            return False
+        return True
+
+    last_size = [0]
```
vivianQizhu: Why need a list for it? Seems you only need one element.

vivianQizhu:
@kimi1978wy Also please limit each line into 79 characters, thanks.

virttest/utils_net.py
kimi1978wy: The purpose to use list is that we need to record the newest file size each time the 

>` _get_file_size`

 is called

kimi1978wy: The purpose to use list is that we need to record the newest file size each time the `_get_file_size` is called. The use the newest value for the next call to this function. If only a variable is used as parameter, it's only passed as value and can't be changed inside the function. So we use a list to store the new size, as it's passed by reference and its element can be changed and restored. Other suggestion has been solved :), pls help to review again

vivianQizhu: @kimi1978wy I don't really understand, `_get_file_size()` is defined under `_wait_for_traceview_dump_finished()`, it's in the same context with `last_size`, you don't even need to pass `last_size` to the function, you might directly use a 'global' `last_size`.  Also for `file_path`,  the `check_file_size_cmd` does not need to be generated each time calling the function. 
```
def _get_file_size():
        status, output = session.cmd_status_output(check_file_size_cmd)
        if status or not output.isdigit():
            return False
        file_size = int(output)
        if file_size != last_size:
            last_size = file_size
            return False
        return True

check_file_size_cmd = "for %%I in (%s) do @echo %%~zI" % dump_file_path
last_size = 0
utils_misc.wait_for(lambda: _get_file_size()...)
```

vivianQizhu: And the function name `_get_file_size` does not match the implementation very well, maybe something like `_check_file_size_stable` would be better, also better have at least one line docstring for it.

kimi1978wy: @vivianQizhu  Updated, pls help to review  it again, thanks

kimi1978wy:
(1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.Win2016.x86_64.io-github-autotest-qemu.boot_with_max_queues: PASS (396.97 s)

kimi1978wy:
(1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.x86_64.io-github-autotest-qemu.boot_with_max_queues: PASS (60.64 s)

PR #2036, opened 2019-04-08
env_process: Add kernel option iommu=pt for guest

vivianQizhu:
id: 1662504

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1045,6 +1045,9 @@ def preprocess(test, params, env):
             kernel_extra_params_add += " intel_iommu=on"
         else:
             kernel_extra_params_remove += " intel_iommu=on"
+        guest_iommu_option = params.get("guest_iommu_option")
+        if guest_iommu_option:
+            kernel_extra_params_add += " iommu=%s" % guest_iommu_option
 
     # Clone master image from vms.
     base_dir = data_dir.get_data_dir()
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -82,6 +82,8 @@ flexible_nic_index = no
 # This parameter only support intel platform so the default
 # value is set to None.
 # enable_guest_iommu = no
+# Set guest kernel iommu option (iommu=off/force/pt e.g.)
+# guest_iommu_option = pt
 
 # List of block device object names (whitespace separated)
 images = image1


vivianQizhu:
@jingzhao84 @yanan-fu Please help review, thanks.

virttest/env_process.py
```diff
@@ -1045,6 +1045,11 @@ def preprocess(test, params, env):
             kernel_extra_params_add += " intel_iommu=on"
         else:
             kernel_extra_params_remove += " intel_iommu=on"
+        guest_iommu_pt = params.get("guest_iommu_pt")
```
yanan-fu: I don't have very strong feeling about this.  guest `iommu` support not only one option, but `iommu=pt` be used more often in our test.  Prefer this a little more:  define parameter `guest_iommu_option = $option`, if some option were added, add it to `kernel_extra_params_add` with `iommu = $option`.  
refer the info from kernel doc:
```
    iommu=      [x86]                                                               
        off                                                                         
        force                                                                       
        noforce                                                                     
        biomerge                                                                    
        panic                                                                       
        nopanic                                                                     
        merge                                                                       
        nomerge                                                                     
        soft                                                                        
        pt      [x86, IA-64]                                                        
        nobypass    [PPC/POWERNV]                                                   
            Disable IOMMU bypass, using IOMMU for PCI devices. 
```

@jingzhao84 @luckyh call for more opinions.

luckyh: Not all platforms have the option `pt`, so I think @yfu 's suggestion is quite good.

vivianQizhu: @yanan-fu @luckyh Yes, that's better. Just I will need remove the other `iommu=\w+` options when add one(e.g. replace `iommu=off` with `iommu=pt`), this way I might need rewrite `check_kernel_cmdline` for fuzzy matching(not confident, seems also need update all caller, since need a list instead of a string with space for `kernel_extra_params_remove`, since it will mix string and regular expression in the list), or I might get the accurate string from `"cat /proc/cmdline"` first and add it to `kernel_extra_params_remove` directly(easier approach), which one do you prefer? Or any other suggestion?

yanan-fu: @vivianQizhu  Good question.  According to this thinking,  when we use `guest_iommu_pt == "yes"`, we need to check other iommu option first too i think.
1. `guest_iommu_pt = yes`,  automation doesn't provide method to add other iommu option, so check other options are not that must ?
2. `guest_iommu_option = pt`, automation provide method to add other option, so check other exclusive options is must ?

vivianQizhu: @yanan-fu If we don't consider other options I think current solution would be enough, but if we consider all option, we need consider more. I was thinking append an parameter like `update_args` to `update_boot_option`, input `iommu: pt` and in the function replace the option with the desired one if it exists, otherwise add it, but this way I will need update `update_boot_option_ubuntu` too...So I now prefer adding an API in `utils_test` like `get_kernel_option(option_name)` returns option value.

vivianQizhu: @yanan-fu Just found that `kernel_extra_params_add` will finally used as `grubby --update-kernel=`grubby --default-kernel` --args="$kernel_extra_params_add"`, per grubby manual page: `If an argument already exists the new value replaces the old values`, I have tried and it does replace old one. So do no need extra handling in our code for the update issue. Therefore I updated to only append defined option to `kernel_extra_params_add`. Please help review, thanks.

vivianQizhu:
```
01:08:29 INFO | Update guest kernel option.  add args: iommu=pt
01:08:29 DEBUG| Sending command: grubby --update-kernel=`grubby --default-kernel` --args="iommu=pt"
```

virttest/env_process.py
luckyh: IIRC, `grubby` would handle the replacement, that means for the case user wants `iommu=X` it will automatically replace the former value with the new one `X`, so that may not be a problem.

The question is what the param `guest_iommu_*` at here stand for, if it exactly represent the kernel command line option `iommu`, then it should be a enum, not a bool; if it is an abstraction of the user's requirement to set the iommu mode to `pt`, then it can be a bool, and we should provide the usage to users.

Thanks.

vivianQizhu: @luckyh You are right, and I have updated the code, please help review, thanks.

virttest/env_process.py
```diff
@@ -1045,6 +1045,9 @@ def preprocess(test, params, env):
             kernel_extra_params_add += " intel_iommu=on"
         else:
             kernel_extra_params_remove += " intel_iommu=on"
+    guest_iommu_option = params.get("guest_iommu_option")
+    if guest_iommu_option:
+        kernel_extra_params_add += " iommu=%s" % guest_iommu_option
```
luckyh: I think it should work only if `enable_guest_iommu = yes`, do you?

vivianQizhu: @luckyh I think they should be separated, when there is other options like `off`, `force`, `nobypass    [PPC/POWERNV]`, obviously it supports other platform besides Intel.

vivianQizhu: @luckyh https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.1/html/installation_guide/appe-configuring_a_hypervisor_host_for_pci_passthrough :
```
If intel_iommu=on or amd_iommu=on works, you can try replacing them with iommu=pt or amd_iommu=pt. The pt option only enables IOMMU for devices used in passthrough and will provide better host performance. However, the option may not be supported on all hardware. Revert to previous option if the pt option doesn't work for your host.
```
Per this guide, even single `iommu=pt` without `intel_iommu=on`  could be a test scenario.

yanan-fu: @vivianQizhu  The info your provide is correct, `intel_iommu`, `amd_iommu`, `iommu=X` can work with bare mental.   Here is about guest kernel line,  QEMU only support `-device intel-iommu` now.

vivianQizhu: @yanan-fu Yes, so we do not consider `amd_iommu`. My point is, if we consider all `iommu=` options, we are considering it as an independent kernel option, then it should not be bound to `intel_iommu=on`.

yanan-fu: @vivianQizhu Yes, `iommu=*` can work without `intel_iommu=on` (no meaning, but can work) in guest kernel line as an independent kernel option.  My point here is,  QEMU only support `intel_iommu`,  this option only work with `utils_misc.get_cpu_vendor(verbose=False) == 'GenuineIntel'`

vivianQizhu: @yanan-fu So no modification required from you to my current code, right?

yanan-fu: @vivianQizhu No,  let me be more clear. Current code don't check the cpu vendor, but QEMU only support `intel-iommu` device now. We need add this step.

yanan-fu: @vivianQizhu ah...  please ignore my comment above, i only consider x86(only intel, no amd), miss other arch, such as ppc. It is not suitable.

vivianQizhu: @yanan-fu Sorry, I don't agree, qemu command line only support intel has nothing to do with this `iommu` kernel option, that should not be controlled here.

yanan-fu: @vivianQizhu  Without iommu device, it is meaningless to add `iommu=*` option in kernel line. Both for host (hardware iommu) and guest ( - device intel-iommu).

yanan-fu: @vivianQizhu  For `enable_guest_iommu`, we check the vendor, because VM can only have `intel-iommu` device, so we can only add `intel-iommu=on` in guest kernel line. (In fact, if only consider kernel option, both intel_iommu and amd_iommu are support).
For iommu opyion, i think they are same.

vivianQizhu: @yanan-fu I understand your point, but `iommu` itself does not distinguish that, here is not the right place the restrict it, and `intel_iommu` is restrict to intel because the option itself only support intel. Hard to say we don't support other platform in the further. Actually, even qemu guest also works with Amd iommu, we have such product bz like amd should not work with iommu device but devel said it make sense since code are actually same.
If you want to make sure kernel options only added when there is iommu device, then should check if there is such device, not check platform. But still does not make sense to me. Cause the kernel options are independent from the hardware device, they should be configured by user, should not restrict the kernel cmdline when there is no such device.
Anyway, it should be configured in right way, rather than restrict here, which is logically wrong for me.

vivianQizhu:
@yanan-fu @luckyh I have moved `guest_iommu_option` under `enable_guest_iommu`, any more comments? Thanks.

vivianQizhu:
Thanks all, merging.

PR #2042, opened 2019-04-16
utils_test/libvirt: add API to create macvtap xml for VM

balamuruhans:
`create_macvtap_vmxml()` creates xml object for macvtap interface
which is used to perform attach/detach operation to VM.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -73,6 +73,7 @@ from virttest.libvirt_xml.devices import hostdev
 from virttest.libvirt_xml.devices import controller
 from virttest.libvirt_xml.devices import seclabel
 from virttest.libvirt_xml.devices import channel
+from virttest.libvirt_xml.devices import interface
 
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -160,6 +161,25 @@ def cleanup(self):
...
+
+    :param iface: macvtap interface name
+    :param params: Test dict params for macvtap config
+
+    :return: macvtap xml object
+    """
+    mode = params.get('macvtap_mode', 'passthrough')
+    model = params.get('macvtap_model', 'virtio')
+    macvtap_type = params.get('macvtap_type', 'direct')
+    macvtap = interface.Interface('direct')
```
sathnaga: s/`macvtap = interface.Interface('direct')`/`macvtap = interface.Interface(macvtap_type)`

balamuruhans: ah, good catch! sure, I will make the change

PR #2047, opened 2019-04-22
qemu_vm: Delete all guest ports correctly when destroying guest

PaulYuuu:
The current method sometimes cannot delete all ports of guest
accurately. Such as openvswitch backend, ovs takes some time to process
ports to be deleted. So we should not check the port status via 'ip
link'(get_net_if), we need to check if it exists in the bridge to decide
whether to delete it.

ID: 1701815

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -1128,6 +1128,10 @@ class Bridge(object):
         """
         try:
             self._br_ioctl(arch.SIOCBRDELIF, brname, ifname)
+            if utils_misc.wait_for(lambda: ifname in self.list_iface(brname),
+                                   timeout=5, first=0.5):
+                logging.warning("Failed to delete %s from %s" % (ifname,
+                                                                 brname))
         except IOError as details:
             raise BRDelIfError(ifname, brname, details)
 
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -2,6 +2,7 @@ import re
 import os
 import socket
 import fcntl
+import errno
 import struct
 import logging
 import random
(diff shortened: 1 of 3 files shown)


PaulYuuu:
Based on: https://github.com/avocado-framework/avocado-vt/commit/9e75aa30fc071865f2fa63308f34c1c9341edade

PaulYuuu:
@xiywang Please help to review. If this method is correct, we don't need to delete the ports in each testcase.

PaulYuuu:
@pezhang `multi_nics_verify`  This test case can easily reproduce the problem.

PaulYuuu:
@pezhang Updated codes according to your test results.

PaulYuuu:
@luckyh Hello, I did a bit code update, could you help to review? Thanks.

virttest/utils_net.py
```diff
@@ -1045,9 +1046,16 @@ def get_structure(self):
...
+                try:
+                    if "bridge" not in os.listdir(br_iface_path):
+                        continue
+                except OSError as e:
+                    if e.errno == errno.ENOENT:
+                        continue
+                    else:
+                        raise e
+
             result[br_iface] = dict()
```
pevogam: @PaulYuuu: The block after this change will execute if `os.path.isdir(br_iface_path) == False` with your change and will not execute before this change took place. Is this intentional?

We encounter similar errors in the block below possibly due to the same reasons like:

```
20:17:00 ERROR|   File "/usr/lib/python3.7/site-packages/virttest/utils_net.py", line 1100, in port_to_br
20:17:00 ERROR|     if port_name in self.get_structure()[br]['iface']:
20:17:00 ERROR|   File "/usr/lib/python3.7/site-packages/virttest/utils_net.py", line 1065, in get_structure
20:17:00 ERROR|     with open(stp_state_path, "r") as stp_state_file:
20:17:00 ERROR| FileNotFoundError: [Errno 2] No such file or directory: '/sys/class/net/t0-W6kgkf/bridge/stp_state'
```

and sometimes

```

2019-07-09 16:27:01,051 stacktrace       L0045 ERROR|   File "/usr/lib/python3.7/site-packages/virttest/utils_net.py", line 1100, in port_to_br
2019-07-09 16:27:01,051 stacktrace       L0045 ERROR|     if port_name in self.get_structure()[br]['iface']:
2019-07-09 16:27:01,051 stacktrace       L0045 ERROR|   File "/usr/lib/python3.7/site-packages/virttest/utils_net.py", line 1066, in get_structure
2019-07-09 16:27:01,052 stacktrace       L0045 ERROR|     stp_state = int(stp_state_file.read().strip())
2019-07-09 16:27:01,052 stacktrace       L0045 ERROR| OSError: [Errno 9] Bad file descriptor
```

Can we put this entire part in a try block instead?

PaulYuuu: @pevogam Sorry for my fault. Could you take a time to review #2111 ? I think it can solve the above problem.

pevogam: Ok, I will take a look although I don't use Open vSwitch so I can only comment on the parts related to this :)

PR #2055, opened 2019-04-23
Add func set_vm_features to set features of vmxml

chloerh:

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -1570,6 +1570,29 @@ class VMXML(VMXMLBase):
         """
         self.remove_all_device_by_type('disk')
 
+    @staticmethod
+    def set_vm_features(vm_name, **attrs):
+        """
+        Set attrs of vm features xml
+
+        :param vm_name: The name of vm to be set
+        :param attrs: attributes to be set
+        """
+        vmxml = VMXML.new_from_inactive_dumpxml(vm_name)
+        if vmxml.xmltreefile.find('/features'):
+            features_xml = vmxml.features
+        else:
+            features_xml = VMFeaturesXML()
+        try:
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -1570,6 +1570,29 @@ def remove_all_disk(self):
...
+        :param vm_name: The name of vm to be set
+        :param attrs: attributes to be set
+        """
+        vmxml = VMXML.new_from_inactive_dumpxml(vm_name)
+        if vmxml.xmltreefile.find('/features'):
+            features_xml = vmxml.features
+        else:
+            features_xml = VMFeaturesXML()
+        try:
+            for attr_key, value in attrs.items():
```
dzhengfy: Should we use list(attrs.items()) here?

chloerh: I was thinking of that too. But I have tested this piece of code on python3, it works fine. So I guess it doesn't matter...?

kylazhang: it return dict_items so I think it is needed to do transfer

dzhengfy: It is better to have because there is still python 2 test.
With python 3:
>>> a = {'a': 1}
>>> a.items()
dict_items([('a', 1)])

With python 2:
>>> a = {'a': 1}
>>> a.items()
[('a', 1)]

chloerh: Actually when it comes to a loop like 'for', this works fine with both python2 and python3

dzhengfy: Agreed. Here we only need an iterable and attrs.items() can work same with python 3 and python 2. So I approve .

PR #2057, opened 2019-04-24
test_setup and env_process: Track and handle host surplus hugepage

PaulYuuu:
1. Add a function to get total surplus hugepages.
2. Add a param to set number of overcommit hugepages, and we can test it in the future.
3. Handle existing surplus hugepages, make `self.target_hugepages` correct.
4. Tracking existing huge page before and after testing

ID: 1701853

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -62,6 +62,10 @@ _vm_register_thread_termination_event = None
 
 _setup_manager = test_setup.SetupManager()
 
+# default num of surplus hugepage, order to compare the values before and after
+# the test when 'setup_hugepages = yes'
+_pre_hugepages_surp = 0
+_post_hugepages_surp = 0
 
 #: QEMU version regex.  Attempts to extract the simple and extended version
 #: information from the output produced by `qemu -version`
@@ -1040,7 +1044,9 @@ def preprocess(test, params, env):
         params["setup_hugepages"] = "yes"
 
     if params.get("setup_hugepages") == "yes":
+        global _pre_hugepages_surp
         h = test_setup.HugePageConfig(params)
+        _pre_hugepages_surp = h.ext_hugepages_surp
         suggest_mem = h.setup()
         if suggest_mem is not None:
             params['mem'] = suggest_mem
@@ -1340,6 +1346,7 @@ def postprocess(test, params, env):
                 test_setup.switch_smt(state="on", params=params)
 
     if params.get("setup_hugepages") == "yes":
+        global _post_hugepages_surp
         try:
             h = test_setup.HugePageConfig(params)
             h.cleanup()
@@ -1348,6 +1355,8 @@ def postprocess(test, params, env):
         except Exception as details:
             err += "\nHP cleanup: %s" % str(details).replace('\\n', '\n  ')
             logging.error(details)
+        else:
+            _post_hugepages_surp = h.ext_hugepages_surp
 
     if params.get("setup_thp") == "yes":
         try:
@@ -1482,6 +1491,9 @@ def postprocess(test, params, env):
 
     if err:
         raise RuntimeError("Failures occurred while postprocess:\n%s" % err)
+    elif _post_hugepages_surp > _pre_hugepages_surp:
+        leak_num = _post_hugepages_surp - _pre_hugepages_surp
+        raise exceptions.TestFail("%d huge pages leaked!" % leak_num)
 
 
 def postprocess_on_error(test, params, env):
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -410,6 +410,8 @@ class HugePageConfig(object):
             self.kernel_hp_file = '/proc/sys/vm/nr_hugepages'
         else:
             raise exceptions.TestSkipError("System doesn't support hugepages")
+        self.over_commit_hp = "/proc/sys/vm/nr_overcommit_hugepages"
+        self.over_commit = kernel_interface.ProcFS(self.over_commit_hp)
         self.pool_path = "/sys/kernel/mm/hugepages"
         self.sys_node_path = "/sys/devices/system/node"
         # Unit is KB as default for hugepage size.
@@ -434,6 +436,11 @@ class HugePageConfig(object):
         vm_mem_minimum = params.get("vm_mem_minimum", "512M")
         self.lowest_mem_per_vm = float(normalize_data_size(vm_mem_minimum))
         utils_memory.drop_caches()
+        self.ext_hugepages_surp = utils_memory.get_num_huge_pages_surp()
+
+        overcommit_hugepages = int(params.get("overcommit_hugepages", 0))
+        if overcommit_hugepages != self.over_commit.proc_fs_value:
+            self.over_commit.proc_fs_value = overcommit_hugepages
 
         target_hugepages = params.get("target_hugepages")
         if target_hugepages is None:
@@ -441,7 +448,8 @@ class HugePageConfig(object):
         else:
             target_hugepages = int(target_hugepages)
 
-        self.target_hugepages = target_hugepages
+        self.target_hugepages = (target_hugepages + self.ext_hugepages_surp -
+                                 overcommit_hugepages)
         self.target_nodes = params.get("target_nodes")
         if self.target_nodes:
             self.target_node_num = {}
@@ -636,7 +644,8 @@ class HugePageConfig(object):
         node_page_path += "/hugepages/hugepages-%skB/nr_hugepages" % pagesize
         obj = kernel_interface.SysFS(node_page_path)
         obj.sys_fs_value = num
-        if obj.sys_fs_value != int(num):
+        # If node has some used hugepage, result will be larger than expected.
+        if obj.sys_fs_value < int(num):
             raise ValueError("Cannot set %s hugepages on node %s, please check"
                              " if the node has enough memory" % (num, node))
 
@@ -690,6 +699,9 @@ class HugePageConfig(object):
         logging.debug("Amount of memory used by each vm: %s", self.mem)
         logging.debug("System setting for large memory page size: %s",
                       self.hugepage_size)
+        if self.over_commit.proc_fs_value > 0:
+            logging.debug("Number of overcommit large memory pages will be set"
+                          " for this test: %s", self.over_commit.proc_fs_value)
         logging.debug("Number of large memory pages needed for this test: %s",
                       self.target_hugepages)
         # Drop caches to clean some usable memory
@@ -713,6 +725,8 @@ class HugePageConfig(object):
             except process.CmdError:
                 return
             process.system("echo 0 > %s" % self.kernel_hp_file, shell=True)
+            self.over_commit.proc_fs_value = 0
+            self.ext_hugepages_surp = utils_memory.get_num_huge_pages_surp()
             logging.debug("Hugepage memory successfully deallocated")
 
 
--- a/virttest/staging/utils_memory.py
+++ b/virttest/staging/utils_memory.py
@@ -141,6 +141,14 @@ def get_num_huge_pages_rsvd(session=None):
     return read_from_meminfo('HugePages_Rsvd', session=session)
 
 
+def get_num_huge_pages_surp(session=None):
+    """
+    Method to get surplus hugepage pages
+    :param session: ShellSession Object of remote host / guest
+    """
+    return read_from_meminfo('HugePages_Surp', session=session)
+
+
 def get_num_anon_huge_pages(pid=0, session=None):
     """
     Method to get total no of anon hugepages


PaulYuuu:
Host_RHEL.m8.u0.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.8.0.ppc64le.io-github-autotest-qemu.boot_vm_in_hugepage.16M: WARN: 6 huge pages leaked! (505.48 s)

PaulYuuu:
@huangyum @MiriamDeng Please help to review, thanks.

virttest/test_setup.py
```diff
@@ -434,14 +434,15 @@ def __init__(self, params):
...
+        self.hugepages_surp = self.get_hugepages_surp()
 
         target_hugepages = params.get("target_hugepages")
         if target_hugepages is None:
             target_hugepages = self.get_target_hugepages()
         else:
             target_hugepages = int(target_hugepages)
 
-        self.target_hugepages = target_hugepages
+        self.target_hugepages = target_hugepages + self.hugepages_surp
```
balamuruhans: why do we include `hugepage_surp` with target_hugepages ?
if we don't really test conditions for surplus hugepages then ideally we have to set `nr_overcommit_hugepages` to zero so that kernel doesn't try to allocate hugepages from normal page pool.

PaulYuuu:
> @PaulYuuu commit msg misguide stating it to be leak ?

Oh no, it may not be leak, but we defined it, I will update the commit message.
> `surplus_hugepages` is a feature where we can set overcommit limit for the applications that uses hugepages, `test_setup` should ideally clear `nr_overcommit_hugepages` if it really doesn't need 

Yes, clear `nr_overcommit_hugepages` also is okay for me.
> (unless we test surplus hugepage feature). Let's have param for surplus hugepage and default it to zero.

Do you mean we need a func `set_hugepages_surp` to handle surplus hugepage? Looks like we are not using it right now.

virttest/test_setup.py
PaulYuuu: target_hugepages will calculate the huge page it wants, if hugepage_surp is not zero then the guest will not be launched.
Set `nr_overcommit_hugepages` to zero is a better method, I will update with your suggestion.

balamuruhans:
> Do you mean we need a func set_hugepages_surp to handle surplus hugepage? Looks like we are not using it right now.

not a func, let's have param for `hugepage_surp` and default it to 0. let's have a check whether it matches with `nr_overcommit_hugepages` before we start using hugepages and set it accordingly.

virttest/test_setup.py
balamuruhans: Thank you.

balamuruhans: one more thing, if you really have to test hugepages_surp, then we should allocate target_hugepages less than what guest required so that kernel will be forced to allocate hugepages from normal pages pool to meet VM to be launched. I think if we allocate `target_hugepages + self.hugepages` then we really don't exercise `hugepage_surp`. please share your thoughts.

virttest/test_setup.py
```diff
@@ -578,6 +579,19 @@ def get_multi_supported_hugepage_size(self):
...
+        """
+        Host may have unexpected huge page leaks. 'HugePages_Surp' from
+        '/proc/meminfo' is the number of huge pages in the pool above the value
+        in /proc/sys/vm/nr_hugepages.
+        Get the current number of HugePages_Surp
+        """
+        surp_file = os.path.join(self.pool_path,
+                                 "hugepages-%dkB" % self.hugepage_size,
+                                 "surplus_hugepages")
+        with open(surp_file, "r") as surp_info:
```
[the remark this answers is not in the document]
PaulYuuu: Thanks for your reminder, I decide to use `/proc/meminfo` to get total of `surplus_hugepages`

virttest/test_setup.py
PaulYuuu: I just have a test case that has failed because of hugepages_surp and I don't test it now.
If we want to let target_hugepages less than what guest required, we can force set `target_hugepages = num` on our test environment.
`target_hugepages + self.hugepages` really not a good way.

virttest/test_setup.py
```diff
@@ -578,6 +584,17 @@ def get_multi_supported_hugepage_size(self):
             raise ValueError("Root hugepage control sysfs directory %s did not"
                              " exist" % self.pool_path)
 
+    def get_hugepages_surp(self):
```
balamuruhans: we can actually reuse `read_from_meminfo('HugePages_Surp')` in `virttest/staging/utils_memory.py`

virttest/test_setup.py
```diff
@@ -434,6 +434,12 @@ def __init__(self, params):
         vm_mem_minimum = params.get("vm_mem_minimum", "512M")
         self.lowest_mem_per_vm = float(normalize_data_size(vm_mem_minimum))
         utils_memory.drop_caches()
+        hugepages_surp = int(params.get("hugepages_surp", 0))
+        if hugepages_surp != self.get_hugepages_surp():
+            with open("/proc/sys/vm/nr_overcommit_hugepages", 'r+') as over_commit:
```
balamuruhans: I thought of something like this,

**Plain and straight forward**
```python
self.hugepages_surp = int(params.get("hugepages_surp", 0))
if self.hugepage_surp != int(utils_memory.read_from_meminfo('HugePages_Surp')):
    logging.warn(...)
    over_commit = kernel_interface.ProcFS("/proc/sys/vm/nr_overcommit_hugepages")
    over_commit.proc_fs_value = self.hugepages_surp
```

**To consider testing hugepages_surp feature,**
```python
self.hugepages_surp = int(params.get("hugepages_surp", 0))
if self.hugepage_surp != int(utils_memory.read_from_meminfo('HugePages_Surp')):
    logging.warn(...)
    over_commit = kernel_interface.ProcFS("/proc/sys/vm/nr_overcommit_hugepages")
    over_commit.proc_fs_value = self.hugepages_surp
:::
:::
# By default `hugepages_surp` would be 0, but if we need to test it
# set the param `hugepages_surp` so that we can take target hugepages
# subtracting hugepages_surp and kernel is expected to allocate rest of
# them from normal pages pool
self.target_hugepages = target_hugepages - self.hugepages_surp
```

virttest/env_process.py
```diff
@@ -1466,6 +1471,10 @@ def postprocess(test, params, env):
 
     if err:
         raise RuntimeError("Failures occurred while postprocess:\n%s" % err)
+    elif (params.get("setup_hugepages") == "yes" and
+          post_hugepages_surp > pre_hugepages_surp):
+        leak_num = post_hugepages_surp - pre_hugepages_surp
+        raise exceptions.TestWarn("%d huge pages leaked!" % leak_num)
```
balamuruhans: changes in this file is not necessary.

virttest/test_setup.py
balamuruhans: After VM boots, we should be able to see `Hugepages_Surp` in /proc/meminfo being updated by kernel with `self.hugepages_surp`

PaulYuuu: I don't think so. If we want to get `Hugepages_Surp` after guest boots, we can invoke `get_hugepages_surp()` in the test case

virttest/env_process.py
PaulYuuu: Yes, maybe. I just want vt to report `WARN` status instead of `PASS`.

virttest/test_setup.py
PaulYuuu: Good point, will update it with other suggestions

PaulYuuu:
Hi @balamuruhans 
There is another situation. If the hugepage_surp is messed up after the test is completed, yes, we need to clean it up, but before that, we should know that our test caused it to be abnormal.

It's okay for me to remove them from `env_process.py`, but I also think it would be better to let the user know that the value of hugepage_surp is not correct after the test, it will be clearer than reporting a warning in the log file.

PowerPC now has a bug that causes hugepage to leak, this patch is also for tracking it.

balamuruhans:
Hi @PaulYuuu,

> There is another situation. If the hugepage_surp is messed up after the test is completed, yes, we need to clean it up, but before that, we should know that our test caused it to be abnormal.

how do we consider it as abnormal ? if non zero value is configured in
`/proc/sys/vm/nr_overcommit_hugepages` then kernel may use and update
`hugepage_surp` in /proc/meminfo.

1. if test configures non zero value then it is expected to clean it
at the end of the testcase.
2. if test did not configure non zero value and if we see `hugepage_surp`
at the end of test then the test have to setup the environment properly before
starting the test. right ?

> It's okay for me to remove them from env_process.py, but I also think it would be better to let the user know that the value of hugepage_surp is not correct after the test, it will be clearer than reporting a warning in the log file.

Better solution that we can do about it is we can move the code from test_setup.py
to env_process.py during preprocess() to initialize the
`/proc/sys/vm/nr_overcommit_hugepages` and clean it up in postprocess().

In env_process.py,

```python
def preprocess():
    :::
    # initialize hugepage_surp with user set params
    hugepages_surp = int(params.get("hugepages_surp", 0))
    over_commit_hp = "/proc/sys/vm/nr_overcommit_hugepages"
    over_commit = kernel_interface.ProcFS(over_commit_hp)
    if hugepages_surp != over_commit.proc_fs_value:
        over_commit.proc_fs_value = hugepages_surp
    :::
:::
:::
def postprocess():
    :::
    # clean it up so that it would not affect the next test
    over_commit_hp = "/proc/sys/vm/nr_overcommit_hugepages"
    over_commit = kernel_interface.ProcFS(over_commit_hp)
    if over_commit.proc_fs_value != 0:
        over_commit.proc_fs_value = 0
    :::

```

In test_setup.py,
```python
class HugePageConfig(object):
    def __init__():
        :::
        self.target_hugepages = target_hugepages - int(params.get("hugepages_surp", 0))
        :::

```

> PowerPC now has a bug that causes hugepage to leak, this patch is also for tracking it.

sorry, can you share me the Bug ID ? I couldn't understand how it is called
as leak. I would check the Bug for my better understanding.

> Anyway, it's up to you. Many thanks for you.

Thank you :)

PaulYuuu:
> Better solution that we can do about it is we can move the code from test_setup.py
> to env_process.py during preprocess() to initialize the
> `/proc/sys/vm/nr_overcommit_hugepages` and clean it up in postprocess().

I think it is more reasonable to handle hugepage_surp and nr_overcommit_hugepages in `class HugePageConfig()`
> > PowerPC now has a bug that causes hugepage to leak, this patch is also for tracking it.
> 
> sorry, can you share me the Bug ID ? I couldn't understand how it is called
> as leak. I would check the Bug for my better understanding.

ID: 1662864, hope you have permission to view it.

In short: (boot_vm_in_hugepage.16M)
After the first test is completed, some huge pages leak so hugepage_surp is not zero, qemu failed to launch the guest during the second test.

PaulYuuu:
Hi @balamuruhans 
So if we really want to test hugepages_surp, we need to handle the existing `HugePages_Surp`(default is zero).
And the `self.target_hugepages` may look like as follow:
```python
self.hugepage_surp = utils_memory.get_num_huge_pages_surp()
hugepage_surp = int(params.get("hugepages_surp", 0))
over_commit_hp = "/proc/sys/vm/nr_overcommit_hugepages"
over_commit = kernel_interface.ProcFS(over_commit_hp)
over_commit.proc_fs_value = hugepages_surp
self.target_hugepages = target_hugepages + self.hugepage_surp - hugepage_surp
```
Right?

(1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.6.10.ppc64.io-github-autotest-qemu.boot_vm_in_hugepage.16M: ERROR: 12 huge pages leaked! (546.33 s)

balamuruhans:
@huangyum request for your review.

virttest/test_setup.py
```diff
@@ -434,14 +434,19 @@ def __init__(self, params):
...
+            over_commit.proc_fs_value = self.hugepages_surp
 
         target_hugepages = params.get("target_hugepages")
         if target_hugepages is None:
             target_hugepages = self.get_target_hugepages()
         else:
             target_hugepages = int(target_hugepages)
 
-        self.target_hugepages = target_hugepages
+        self.target_hugepages = target_hugepages - self.hugepages_surp
```
huangyum: Is it possible the surplus hugepage is being used by other process,  then it should be `        self.target_hugepages = target_hugepages + self.hugepages_surp` ?

PaulYuuu: > Is it possible the surplus hugepage is being used by other process

Yes, you're right. And @balamuruhans means requesting a new parameter to test surplus hugepage.
So maybe we need 1 new parameter and 2 new variables of it.
In cfg:
```
surp_hugepage = int
```
In py:
```python
self.ext_surp_hugepage = int(utils_memory.get_num_huge_pages_surp())
hugepages_surp = int(params.get("hugepages_surp", 0))
self.target_hugepages = target_hugepages + self.ext_surp_hugepage - hugepages_surp
```

huangyum:
IIUC,  /proc/sys/vm/nr_overcommit_hugepages means the number of hugepages which COULD be used more than /proc/sys/vm/nr_hugepages.  HugePages_Surp means the hugepages being used more than /proc/sys/vm/nr_hugepages.  
For example,  set  nr_hugepages to 1024 and nr_overcommit_hugepages to 2048,  HugePages_Total is 1024 (It's same to nr_hugepages at beginning).
```
# cat /proc/sys/vm/nr_overcommit_hugepages 
1024
# cat /proc/sys/vm/nr_hugepages
1024
#  cat /proc/meminfo  | grep -i huge
AnonHugePages:   2058240 kB
ShmemHugePages:        0 kB
HugePages_Total:    1024
HugePages_Free:     1024
HugePages_Rsvd:        0
HugePages_Surp:        0
Hugepagesize:       2048 kB

```
Then boot guest with 3G hugepages and use -mem-prealloc, HugePages_Total becomes 1536(3G), HugePages_Surp is 512(1G=3G-1024*2M)
```
# cat /proc/meminfo  | grep -i huge
AnonHugePages:   2074624 kB
ShmemHugePages:        0 kB
HugePages_Total:    1536
HugePages_Free:        0
HugePages_Rsvd:        0
HugePages_Surp:      512
Hugepagesize:       2048 kB
# cat /proc/sys/vm/nr_hugepages
1024
# cat /proc/sys/vm/nr_overcommit_hugepages 
2048
```
Then I set nr_hugepages to 0(keep guest running), HugePages_Surp becomes 1536.
```
# cat /proc/sys/vm/nr_hugepages
0
# cat /proc/meminfo  | grep -i huge
AnonHugePages:   2095104 kB
ShmemHugePages:        0 kB
HugePages_Total:    1536
HugePages_Free:        0
HugePages_Rsvd:        0
HugePages_Surp:     1536
```
Finally shut down the guest, HugePages_Total and HugePages_Surp both are 0.
```
#  cat /proc/meminfo  | grep -i huge
AnonHugePages:   2091008 kB
ShmemHugePages:        0 kB
HugePages_Total:       0
HugePages_Free:        0
HugePages_Rsvd:        0
HugePages_Surp:        0
Hugepagesize:       2048 kB
```

huangyum:
My point is, HugePages_Surp is not always equal to nr_overcommit_hugepages , and it's normal.  
To check if hugepage leak, comparing HugePages_Surp before and after test is ok. 
To test hugepage overcommit, I guess we just need to set target hugepages less than needed and set nr_overcommit_hugepages correctly. If guest can boot well, it would be fine. 
@balamuruhans @PaulYuuu What do you think?

balamuruhans:
> My point is, HugePages_Surp is not always equal to nr_overcommit_hugepages , and it's normal.
> To check if hugepage leak, comparing HugePages_Surp before and after test is ok.
> To test hugepage overcommit, I guess we just need to set target hugepages less than needed and set nr_overcommit_hugepages correctly. If guest can boot well, it would be fine.
> @balamuruhans @PaulYuuu What do you think?

@huangyum Agreed, nr_overcommit_hugepages is the limit upto which kernel can allow the process to use hugepages more than allocated in nr_hugepages and Hugepage_Surp can grow till the value configured in nr_overcommit_hugepages and need not necessarily equal always.

PaulYuuu:
> > My point is, HugePages_Surp is not always equal to nr_overcommit_hugepages , and it's normal.
> > To check if hugepage leak, comparing HugePages_Surp before and after test is ok.
> > To test hugepage overcommit, I guess we just need to set target hugepages less than needed and set nr_overcommit_hugepages correctly. If guest can boot well, it would be fine.
> > @balamuruhans @PaulYuuu What do you think?
> 
> @huangyum Agreed, nr_overcommit_hugepages is the limit upto which kernel can allow the process to use hugepages more than allocated in nr_hugepages and Hugepage_Surp can grow till the value configured in nr_overcommit_hugepages and need not necessarily equal always.

That's right. So now there are two points to discuss:
1. In order to check if hugepage leak, check `Hugepage_Surp`(before and after testing) with `nr_overcommit_hugepages = 0` and without mounting `hugetlbfs`. This is my original intention.
2. In order to test `nr_overcommit_hugepages`, add a param `overcommit_hugepages`. If we don't add `-mem-prealloc`, `Hugepage_Surp` may not change immediately after guest be launched, so I decided to check it on test case side.

In summary, in order to deal with these two points, `self.target_hugepages` becomes like this:
```python
self.target_hugepages = (target_hugepages + self.ext_hugepages_surp -
                         overcommit_hugepages)
```
I am not sure that I am completely correct. Thanks guys for your patient explanation.

balamuruhans:
@PaulYuuu , Thanks for explanation. ACK from me, let's get @huangyum's suggestion.

PaulYuuu:
I will attach the log result when I finish testing.

balamuruhans:
@PaulYuuu,

> In order to check if hugepage leak, check Hugepage_Surp(before and after testing) with nr_overcommit_hugepages = 0 and without mounting hugetlbfs. This is my original intention.

we should also consider to check `HugePages_Total` and `HugePages_Free` are same before and after test, because in `Bug 1662864` you had mentioned, it is observed sometimes `Hugepage_Surp` remains to be 0 but there is leak causing HugePages_Free < HugePages_Total after the test.

huangyum:
LGTM, ACK.

PaulYuuu:
> @PaulYuuu,
> 
> > In order to check if hugepage leak, check Hugepage_Surp(before and after testing) with nr_overcommit_hugepages = 0 and without mounting hugetlbfs. This is my original intention.
> 
> we should also consider to check `HugePages_Total` and `HugePages_Free` are same before and after test, because in `Bug 1662864` you had mentioned, it is observed sometimes `Hugepage_Surp` remains to be 0 but there is leak causing HugePages_Free < HugePages_Total after the test.

@balamuruhans Is it like this?
```bash
cat /proc/meminfo | grep -i hugepage
AnonHugePages:         0 kB
ShmemHugePages:        0 kB
HugePages_Total:    1024
HugePages_Free:     1022
HugePages_Rsvd:        0
HugePages_Surp:        0
Hugepagesize:      16384 kB
```
It happens when the hugepage has not been cleaned up.
So I think just need to check `Hugepage_Surp` after cleanup.

logs (with over commit):
```
2019-07-31 04:26:34,903 test_setup       L0698 DEBUG| Number of VMs this test will use: 1
2019-07-31 04:26:34,904 test_setup       L0699 DEBUG| Amount of memory used by each vm: 8192
2019-07-31 04:26:34,904 test_setup       L0701 DEBUG| System setting for large memory page size: 16384
2019-07-31 04:26:34,910 test_setup       L0704 DEBUG| Number of overcommit large memory pages will be set for this test: 20
2019-07-31 04:26:34,911 test_setup       L0706 DEBUG| Number of large memory pages needed for this test: 500
2019-07-31 04:26:34,943 test_setup       L0680 DEBUG| Successfully set 500 large memory pages on host
...
...
    -m 8192  \
    -mem-path /mnt/kvm_hugepage \
    -mem-prealloc \
```
Check hugepage after the guest starts.
```
HugePages_Total:     512
HugePages_Free:        0
HugePages_Rsvd:        0
HugePages_Surp:       12
Hugepagesize:      16384 kB
```
We can see `(500 + 12) * 16384 / 1024 = 8192`
So `HugePages_Surp:       12` is expected.

balamuruhans:
> > @PaulYuuu,
> > > In order to check if hugepage leak, check Hugepage_Surp(before and after testing) with nr_overcommit_hugepages = 0 and without mounting hugetlbfs. This is my original intention.
> > 
> > 
> > we should also consider to check `HugePages_Total` and `HugePages_Free` are same before and after test, because in `Bug 1662864` you had mentioned, it is observed sometimes `Hugepage_Surp` remains to be 0 but there is leak causing HugePages_Free < HugePages_Total after the test.
> 
> @balamuruhans Is it like this?

yes
> 
> ```shell
> cat /proc/meminfo | grep -i hugepage
> AnonHugePages:         0 kB
> ShmemHugePages:        0 kB
> HugePages_Total:    1024
> HugePages_Free:     1022
> HugePages_Rsvd:        0
> HugePages_Surp:        0
> Hugepagesize:      16384 kB
> ```
> 
> It happens when the hugepage has not been cleaned up.

Got it, Thanks for explaining!

> So I think just need to check `Hugepage_Surp` after cleanup.
> 
> logs (with over commit):
> 
> ```
> 2019-07-31 04:26:34,903 test_setup       L0698 DEBUG| Number of VMs this test will use: 1
> 2019-07-31 04:26:34,904 test_setup       L0699 DEBUG| Amount of memory used by each vm: 8192
> 2019-07-31 04:26:34,904 test_setup       L0701 DEBUG| System setting for large memory page size: 16384
> 2019-07-31 04:26:34,910 test_setup       L0704 DEBUG| Number of overcommit large memory pages will be set for this test: 20
> 2019-07-31 04:26:34,911 test_setup       L0706 DEBUG| Number of large memory pages needed for this test: 500
> 2019-07-31 04:26:34,943 test_setup       L0680 DEBUG| Successfully set 500 large memory pages on host
> ...
> ...
>     -m 8192  \
>     -mem-path /mnt/kvm_hugepage \
>     -mem-prealloc \
> ```
> 
> Check hugepage after the guest starts.
> 
> ```
> HugePages_Total:     512
> HugePages_Free:        0
> HugePages_Rsvd:        0
> HugePages_Surp:       12
> Hugepagesize:      16384 kB
> ```
> 
> We can see `(500 + 12) * 16384 / 1024 = 8192`
> So `HugePages_Surp: 12` is expected.

cool :+1:

virttest/env_process.py
```diff
@@ -1479,6 +1488,9 @@ def postprocess(test, params, env):
 
     if err:
         raise RuntimeError("Failures occurred while postprocess:\n%s" % err)
+    elif _post_hugepages_surp > _pre_hugepages_surp:
+        leak_num = _post_hugepages_surp - _pre_hugepages_surp
+        raise exceptions.TestError("%d huge pages leaked!" % leak_num)
```
balamuruhans: It should be considered as TestFail right ?

PaulYuuu: Yes, TestFail looks better than TestError, has been updated.

balamuruhans:
Thank you @PaulYuuu @huangyum.

PaulYuuu:
Update log result (huge page leaks):
- Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.6.10.ppc64.io-github-autotest-qemu.boot_vm_in_hugepage.16M: FAIL: 20 huge pages leaked! (992.32 s)
- Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.6.10.ppc64.io-github-autotest-qemu.boot_vm_in_hugepage.16M.with_overcommit: FAIL: 16 huge pages leaked! (762.68 s)

```
HugePages_Total:      36
HugePages_Free:        0
HugePages_Rsvd:        0
HugePages_Surp:       36
Hugepagesize:      16384 kB
```

PR #2064, opened 2019-04-27
Log warning for unreadable serial console in case of IOError

pevogam:
In some cases the file is open just fine but the serial console
prints no output. Not being able to find kernel panic or other
messages in an empty output should be equivalent to good output
or otherwise this will spawn lots of warnings throughout the run.

For instance, a recent check for these warnings for our runs got:

grep -r "Unable to read serial" * |wc -l
-> 9774

Empty output is not the same as no output.


luckyh:
Thanks all.

PR #2069, opened 2019-05-06
utils_test: Enhance the ssh key setup

dzhengfy:
Sometimes there will be a warning or info for ssh-copy-id prompted
when setting up ssh key. Those info can be ignored.

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -2662,7 +2662,9 @@ class RemoteVMManager(object):
                          % (ssh_options, pub_key, vm_ip))
         while True:
             matched_strs = [r"[Aa]re you sure", r"[Pp]assword:\s*$",
-                            r"lost connection", r"]#"]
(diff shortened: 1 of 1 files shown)


virttest/utils_test/__init__.py
```diff
@@ -2685,6 +2687,9 @@ def setup_ssh_auth(self, vm_ip, vm_pwd, vm_user="root",
             elif index == 3:
                 logging.debug("Logged in now...")
                 break
+            elif index in [4, 5]:
```
balamuruhans: we should be handling carefully for the real error that can be expected in the if conditions and break, IIUC the existing code would anyway ignore the warning and continue to the loop as it doesn't enter any conditions

dzhengfy: @balamuruhans , In my previous test, the problem came to me that the '/usr/bin/ssh-copy-id' and 'warning xxx' can not be bypassed and failed to login the VM.  But I can not reproduce it now. So I would like to close it for now. And if it happens again, I will open it. Thank you.

PR #2072, opened 2019-05-10
utils_netperf: Wait a few seconds for running up netserver

luckyh:
We might miss the catching of netserver processes, that because we only
do a one-shot check immediately after the process being launched. Now
let's make the check not too harsh.

ID: 1296804

### DIFF
--- a/virttest/utils_netperf.py
+++ b/virttest/utils_netperf.py
@@ -5,6 +5,7 @@ import re
 import aexpect
 from avocado.utils import download
 from avocado.utils import aurl
+from avocado.utils import wait
 
 from six.moves import xrange
 
@@ -281,7 +282,7 @@ class NetperfServer(Netperf):
             logging.info("Start netserver with cmd: '%s'" % server_cmd)
             self.session.cmd_output_safe(server_cmd)
 
-        if not self.is_server_running():
+        if not wait.wait_for(self.is_server_running, 5):
             raise NetserverError("Can not start netperf server!")
         logging.info("Netserver start successfully")
 


luckyh:
@vivianQizhu @xiangchunfu @yfu request for your review, thanks.

luckyh:
```
 (1/1) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.netperf_stress_test.UDP_STREAM.guest2guest: PASS (976.02 s)
```

hereischen:

virttest/utils_netperf.py
```diff
@@ -281,7 +282,7 @@ def start(self, restart=False):
             logging.info("Start netserver with cmd: '%s'" % server_cmd)
             self.session.cmd_output_safe(server_cmd)
 
-        if not self.is_server_running():
+        if not wait.wait_for(self.is_server_running, 5):
```
yanan-fu: The logic look good to me.
A bit concern, why do not use `utils_misc.wait_for` that in the same repo ?  Both of them can solve this problem :)
From my checking, `avocado.utils.wait.wait_for` can accept  args for `func`, we do not need it here.

luckyh: Yes, either of them could work, but consider about that `utils_misc` is not a well-defined module and we may restructure it in the future (at least I think so), I prefer use the one in `avocado.utils.wait`, thanks.

balamuruhans: ACK

yanan-fu:
Run test for 50 times,  did not hit the issue again. we can merge it ~

luckyh:
Thanks all.

PR #2073, opened 2019-05-10
change the attribute of class Hostdev from hostdev_type to type

brucepatricklee:
the attribute in class Hostdev should be unified with other device class

### DIFF
--- a/virttest/libvirt_xml/devices/hostdev.py
+++ b/virttest/libvirt_xml/devices/hostdev.py
@@ -9,12 +9,12 @@ from virttest.libvirt_xml import accessors
 
 class Hostdev(base.TypedDeviceBase):
 
-    __slots__ = ('hostdev_type', 'mode', 'managed', 'sgio', 'rawio',
+    __slots__ = ('type', 'mode', 'managed', 'sgio', 'rawio',
                  'source', 'boot_order', 'readonly', 'shareable'
                  )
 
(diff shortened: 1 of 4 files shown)


brucepatricklee:
the change has little impact on the existing tp-libvirt code as follows:
[root@rhel7 tp-libvirt]# grep hostdev_type . -r
./libvirt/tests/src/libvirt_pci_passthrough_hotplug.py:    dev.hostdev_type = 'pci'
./libvirt/tests/src/usb_device.py:                dev.hostdev_type = 'usb'
./libvirt/tests/src/usb_device.py:                if dev.hostdev_type == "usb":
./libvirt/tests/src/usb_device.py:                if dev.hostdev_type == "usb":

kylazhang:
Also need to check avocado-vt related change for hostdev_type. Why you want to change it? As it is have already been used for a long time. And also if you change it, please also give all update change on avocado-vt/tp-libvirt and add pr number here so that we need to merge them all

brucepatricklee:
@kylazhang in the libvirt usb test case, usb device has two sorts of device type,  hostdev and redirdev. the ugly attribute code "hostdev_type" will introduce many duplicated code. I have updated all the related code in avocado-vt and tp-libvirt. And what's more, the type is NOT a key word of python in itself. So it can be used as a attribute of a class

https://github.com/autotest/tp-libvirt/pull/2165

kylazhang:
need to merge this pr and autotest/tp-libvirt#2165 at the same time

PR #2074, opened 2019-05-13
cleanup_iommu: Cleanup iommu in postprocess stage

ybduan:
ID: 1709234

Signed-off-by: ybduan <yduan@redhat.com>

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1134,6 +1134,13 @@ def postprocess(test, params, env):
     # recover the changes done to kernel params in postprocess
     kernel_extra_params_add = params.get("kernel_extra_params_add", "")
     kernel_extra_params_remove = params.get("kernel_extra_params_remove", "")
+
+    if params.get("enable_guest_iommu") == "yes":
+        kernel_extra_params_add += " intel_iommu=on"
+        guest_iommu_option = params.get("guest_iommu_option")
+        if guest_iommu_option:
+            kernel_extra_params_add += " iommu=%s" % guest_iommu_option
+
     if kernel_extra_params_add or kernel_extra_params_remove:
         for vm in env.get_all_vms():
             if vm:


virttest/env_process.py
```diff
@@ -1134,6 +1134,15 @@ def postprocess(test, params, env):
     # recover the changes done to kernel params in postprocess
     kernel_extra_params_add = params.get("kernel_extra_params_add", "")
     kernel_extra_params_remove = params.get("kernel_extra_params_remove", "")
+
+    if params.get("enable_guest_iommu"):
+        enable_guest_iommu = params.get("enable_guest_iommu")
+        if enable_guest_iommu == "yes":
+            kernel_extra_params_add += " intel_iommu=on"
```
vivianQizhu: If you wanna clean up, should be `kernel_extra_params_remove`? And what's the use case of this action?

ybduan: 'kernel_extra_params_add' is the correct usage, please refer to Line 1151.
The use case is described in bug.
Thanks!

vivianQizhu: Okay, the code looks okay to me, just please demonstrate same content in commit log. Thanks.

virttest/env_process.py
```diff
@@ -1134,6 +1134,15 @@ def postprocess(test, params, env):
     # recover the changes done to kernel params in postprocess
     kernel_extra_params_add = params.get("kernel_extra_params_add", "")
     kernel_extra_params_remove = params.get("kernel_extra_params_remove", "")
+
+    if params.get("enable_guest_iommu"):
+        enable_guest_iommu = params.get("enable_guest_iommu")
+        if enable_guest_iommu == "yes":
```
balamuruhans: this line is not necessary,
```python
if params.get("enable_guest_iommu") == "yes" and params.get("guest_iommu_option"):
    kernel_extra_params_add += " iommu=%s" % guest_iommu_option
```

ybduan: You code above only clean up guest_iommu_option, but not enable_guest_iommu option.
I prefer to reset all iommu-related configuration now that we add clean up step.
Thanks!

virttest/env_process.py
ybduan: Updated!

virttest/env_process.py
vivianQizhu: Or it might be:
```
if params.get("enable_guest_iommu") == "yes":
    kernel_extra_params_add += " intel_iommu=on"
```
@balamuruhans What do you think?

balamuruhans: @vivianQizhu right! I gave an example, where we really don't need to assign a variable and again checking it.
as you have mentioned it would be good to have,
```python
if params.get("enable_guest_iommu") == "yes":
    kernel_extra_params_add += " intel_iommu=on"
    if params.get("guest_iommu_option"):
        kernel_extra_params_add += " iommu=%s" % guest_iommu_option
```

ybduan: Updated, Thanks!

YongxueHong:
@ybduan For the line 1141, will raise AttributeError: 'VM' object has no attribute 'start' for qemu_vm since the VM class doesn't have the method start(), only libvirt_vm's VM class has it. right ? CC @vivianQizhu

vivianQizhu:
@YongxueHong Without this patch it's line 1141, and with this patch it's line 1148, where to call `vm.start`, that seems nothing to do with this patch.

YongxueHong:
@ybduan  pls ignore my comments.

PR #2076, opened 2019-05-14
Add device class redirdev

brucepatricklee:
add attributes type, bus, source, protocol, alias and address

### DIFF
--- a/virttest/libvirt_xml/devices/redirdev.py
+++ b/virttest/libvirt_xml/devices/redirdev.py
@@ -4,13 +4,62 @@ redirdev device support class(es)
 http://libvirt.org/formatdomain.html#elementsRedir
 """
 
-import six
+from virttest.libvirt_xml import accessors
+from virttest.libvirt_xml.devices import base, librarian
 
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/devices/redirdev.py
```diff
@@ -4,13 +4,60 @@
...
+        """
+        Return a new Redirdev Source instance and set properties from dargs
+        """
+        new_one = self.Source(virsh_instance=self.virsh)
+        for k, v in dargs.items():
+            if v:
+                setattr(new_one, k, v)
+        return new_one
+
+    class Source(base.base.LibvirtXMLBase):
```
qiankehan: Since redir dev could be used as **unix** type:
```xml
<redirdev bus="usb" type="unix">
      <source mode="connect" path="/tmp/connect.socket" />
      <protocol type="raw" />
    </redirdev>
```
It is more perfect to add **source->code** in avocado-vt although it is seldomly used in  libvirt.

brucepatricklee: updated

PR #2078, opened 2019-05-16
libvirt_xml/vm_xml: `nested-hv` feature xml attribute for pseries VM

balamuruhans:
From libvirt, state of `nested-hv` attribute need to be set 'on' to
launch qemu commandline with `cap-nested-hv=on` for VM (L1) to use kvm_hv
in L0 and to support nested virtualized VM (L2) on it.


balamuruhans:
Thanks @dzhengfy :+1:

PR #2085, opened 2019-05-20
RFC: Fixup: Test hung

sathnaga:
Seeing tp-libvirt test cases runs in RHEL 8 host hung at the
preprocess stage itself, rootcaused the issue seems
to be in the below place

	env.start_ip_sniffing(params)
		self._sniffer.start()
			aexpect.run_tail(...) --> indefinitely hung here

Changing it to aexpect.run_bg(...) seems to be working fine,

 (1/1) powerkvm-qemu.reboot: PASS (132.85 s)

But have looked around change of code, no new change in any of
the places.

Avocado-vt version: 69.0+(6579028d)
Aexpect version: 1.5.1

Though this patch fixes the issue, need to understand better
why suddenly, we started seeing this issue and is this proper
fix.

### DIFF
--- a/virttest/ip_sniffing.py
+++ b/virttest/ip_sniffing.py
@@ -200,7 +200,7 @@ class Sniffer(object):
         if self._remote_opts:
             self._start_remote()
         else:
-            self._process = aexpect.run_tail(
+            self._process = aexpect.run_bg(
                 command=cmd,
                 output_func=self._output_logger_handler)
 


vivianQizhu:
@sathnaga I have hit this issue before, it is mostly caused by aexpect version mismatch(e.g. have 2 different version of aexpect installed), anyway if it hangs then the process does not start as we expected, to make it background this way might make the process continue as some case won't be affect when ip sniffing actually failed, but it might affects those cases do need it being success. In short, we should be aware of the failure.

sathnaga:
> @sathnaga I have hit this issue before, it is mostly caused by aexpect version mismatch(e.g. have 2 different version of aexpect installed), anyway if it hangs then the process does not start as we expected, to make it background this way might make the process continue as some case won't be affect when ip sniffing actually failed, but it might affects those cases do need it being success. In short, we should be aware of the failure.

@vivianQizhu Agree, but, in my env, having only single aexpect version installed, not sure how else to handle, I tried manually invoke the command it runs, it seems to be running fine without hung, 

```
tcpdump -tnpvvvi any 'port 68 or port 546'
tcpdump: listening on any, link-type LINUX_SLL (Linux cooked), capture size 262144 bytes
IP (tos 0x0, ttl 128, id 13011, offset 0, flags [none], proto UDP (17), length 308)
    0.0.0.0.bootpc > 255.255.255.255.bootps: [no cksum] BOOTP/DHCP, Request from e0:91:f5:94:b3:ee, length 280, xid 0x38fcb67b, secs 55473, Flags [Broadcast] (0x8000)
	  Client-Ethernet-Address e0:91:f5:94:b3:ee
	  Vendor-rfc1048 Extensions
	    Magic Cookie 0x63825363
	    DHCP-Message Option 53, length 1: Discover
	    Client-ID Option 61, length 7: ether e0:91:f5:94:b3:ee
	    Lease-Time Option 51, length 4: 4294967295
	    Hostname Option 12, length 7: "GS748TS"
	    Parameter-Request Option 55, length 9: 
	      Subnet-Mask, Default-Gateway, TFTP, BF
	      Server-ID, Client-ID, Option 129, Domain-Name-Server
	      Domain-Name
	    END Option 255, length 0
	    PAD Option 0, length 0
```

If we go by this, `run_bg` change should be fine, as long as it does not introduce any other regression...

vivianQizhu:
@sathnaga Seems the process run success in your env, but when we do hit the situation when it does not run(what I hit), that might bring regression and hard to debug since we are not aware of the background failure.

sathnaga:
> @sathnaga Seems the process run success in your env, but when we do hit the situation when it does not run(what I hit), that might bring regression and hard to debug since we are not aware of the background failure.

ah, got it... then we need to dig more for reasons and fix....

vivianQizhu:
> > @sathnaga Seems the process run success in your env, but when we do hit the situation when it does not run(what I hit), that might bring regression and hard to debug since we are not aware of the background failure.
> 
> ah, got it... then we need to dig more for reasons and fix....

@sathnaga Em...although the issue I hit might not be the same one with you, I suppose there is something wrong with aexpect.client.Spawn() initialization.

sathnaga:
I no longer notice this issue, will close for now.

PR #2087, opened 2019-05-21
Add a function to get disk by serial string

Yingshun:
This function to get the disk by given serial string.
Need to check ID_SERIAL in VM if disk is attached with serial property.

### DIFF
--- a/virttest/utils_disk.py
+++ b/virttest/utils_disk.py
@@ -25,6 +25,7 @@ from avocado.utils.service import SpecificServiceManager
 from virttest import error_context
 from virttest import utils_numeric
 from virttest.compat_52lts import decode_to_text
+from virttest.compat_52lts import results_stdout_52lts
 
 PARTITION_TABLE_TYPE_MBR = "msdos"
 PARTITION_TABLE_TYPE_GPT = "gpt"
@@ -790,6 +791,48 @@ def configure_empty_disk(session, did, size, ostype, start="0M", n_partitions=1,
     return configure_empty_linux_disk(session, did, size, start,
                                       n_partitions, fstype,
                                       labeltype, timeout)
+
+
+def get_parts_list(session=None):
+    """
+    Get all partition lists.
+    """
+    parts_cmd = "cat /proc/partitions"
+    if session:
+        _, parts_out = session.cmd_status_output(parts_cmd)
+    else:
+        parts_out = results_stdout_52lts(process.run(parts_cmd))
+    parts = []
+    if parts_out:
+        for line in parts_out.rsplit("\n"):
+            if line.startswith("major") or line == "":
+                continue
+            parts_line = line.rsplit()
+            if len(parts_line) == 4:
+                parts.append(parts_line[3])
+    logging.debug("Find parts: %s", parts)
+    return parts
+
+
+def get_disk_by_serial(serial_str, session=None):
+    """
+    Get disk by serial in VM or host
+
+    :param serial_str: ID_SERIAL of disk, string value
+    :param session: VM session or None
+    :return: Disk name if find one with serial_str, else None
+    """
+    parts_list = get_parts_list(session=session)
+    for disk in parts_list:
+        cmd = ("udevadm info --query=all --name=/dev/{} | grep ID_SERIAL={}"
+               .format(disk, serial_str))
+        if session:
+            status = session.cmd_status(cmd)
+        else:
+            status = process.run(cmd, shell=True, ignore_status=True).exit_status
+        if not status:
+            logging.debug("Disk %s has serial %s", disk, serial_str)
+            return disk
 
 
 class Disk(object):
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -901,27 +901,6 @@ def mkfs(partition, fs_type, options="", session=None):
         process.run(mkfs_cmd)
 
 
-def get_parts_list(session=None):
-    """
-    Get all partition lists.
-    """
-    parts_cmd = "cat /proc/partitions"
-    if session:
-        _, parts_out = session.cmd_status_output(parts_cmd)
-    else:
-        parts_out = results_stdout_52lts(process.run(parts_cmd))
-    parts = []
-    if parts_out:
-        for line in parts_out.rsplit("\n"):
-            if line.startswith("major") or line == "":
-                continue
-            parts_line = line.rsplit()
-            if len(parts_line) == 4:
-                parts.append(parts_line[3])
-    logging.debug("Find parts: %s" % parts)
-    return parts
-
-
 def yum_install(pkg_list, session=None):
     """
     Try to install packages on system
@@ -3303,7 +3282,7 @@ def new_disk_vol_name(pool_name):
         return None
     disk = poolxml.get_source().device_path[5:]
     part_num = len(list(filter(lambda s: s.startswith(disk),
-                               get_parts_list())))
+                               utils_misc.utils_disk.get_parts_list())))
     return disk + str(part_num)
 
 


virttest/utils_test/libvirt.py
```diff
@@ -922,6 +922,24 @@ def get_parts_list(session=None):
...
+    parts_list = get_parts_list(session)
+    for disk in parts_list:
+        cmd = "udevadm info --query=all --name=/dev/{0} | grep ID_SERIAL=".format(disk)
+        s, o = session.cmd_status_output(cmd)
+        if s == 0:
+            serial_tmp = o.strip().split('=')[1]
+            if serial_tmp == serial_str:
+                logging.debug("Disk %s has serial %s" % (disk, serial_str))
+                return disk
+    return None
```
lolyu: LINE 940 is redundant.

Yingshun: Removed.

Yingshun: Removed.

virttest/utils_test/libvirt.py
```diff
@@ -922,6 +922,23 @@ def get_parts_list(session=None):
     return parts
 
 
+def get_disk_by_serial(session, serial_str):
+    """
+    Get disk by serial in VM
+
+    :param serial_str: ID_SERIAL of disk, string value
+    """
```
dzhengfy: Add :return docstring, please.

Yingshun: @dzhengfy  Updated.

virttest/utils_test/libvirt.py
```diff
@@ -922,6 +922,25 @@ def get_parts_list(session=None):
...
+    Get disk by serial in VM
+
+    :param session: VM session 
+    :param serial_str: ID_SERIAL of disk, string value
+    :return: Disk name if find one with serial_str, else None
+    """
+    parts_list = get_parts_list(session)
+    for disk in parts_list:
+        cmd = "udevadm info --query=all --name=/dev/{0} | grep ID_SERIAL=".format(disk)
```
dzhengfy: Could you consider to use 'grep ID_SERIAL=serial_str'? That will make codes simpler.

virttest/utils_test/libvirt.py
```diff
@@ -922,6 +922,23 @@ def get_parts_list(session=None):
...
+def get_disk_by_serial(session, serial_str):
+    """
+    Get disk by serial in VM
+
+    :param serial_str: ID_SERIAL of disk, string value
+    """
+    parts_list = get_parts_list(session)
+    for disk in parts_list:
+        cmd = "udevadm info --query=all --name=/dev/{0} | grep ID_SERIAL=".format(disk)
+        s, o = session.cmd_status_output(cmd)
```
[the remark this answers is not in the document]
Yingshun: Updated.

virttest/utils_test/libvirt.py
Yingshun: Updated.

virttest/utils_test/libvirt.py
```diff
@@ -922,6 +922,23 @@ def get_parts_list(session=None):
...
+        if s == 0:
```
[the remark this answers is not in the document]
Yingshun: updated.

virttest/utils_test/libvirt.py
```diff
@@ -922,6 +922,23 @@ def get_parts_list(session=None):
     return parts
 
 
+def get_disk_by_serial(session, serial_str):
```
balamuruhans: this method should reside in virttest/utils_disk.py

virttest/utils_test/libvirt.py
```diff
@@ -922,6 +922,23 @@ def get_parts_list(session=None):
...
+
+    :param session: VM session
+    :param serial_str: ID_SERIAL of disk, string value
+    :return: Disk name if find one with serial_str, else None
+    """
+    parts_list = get_parts_list(session)
+    for disk in parts_list:
+        cmd = ("udevadm info --query=all --name=/dev/{} | grep ID_SERIAL={}"
+               .format(disk, serial_str))
+        if not session.cmd_status(cmd):
```
balamuruhans: I would suggest to make this method common for host/VM by having session as default argument.
if user gives session use it, else query in host.

virttest/utils_test/libvirt.py
Yingshun: @balamuruhans ,If move this method to virttest/utils_disk.py, there will be a circular imports.
Do you think we must move it to virttest/utils_disk.py?

balamuruhans: @Yingshun, I think we should keep the methods in the respective files. All the disk related methods should be placed in utils_disk.
Currently utils_misc and utils_libvirt is growing and needs to be reorganised by moving the methods to appropriate files. 

Can you move the disk related methods from libvirt.py to utils_disk.py so that this circular import situation can be fixed. `get_disk_by_serial()` calls `get_parts_list()`and if we place in utils_disk.py, then there wont be circular import issue ?

Yingshun: @balamuruhans  Updated. Please help review again. Thanks very much!

virttest/utils_disk.py
```diff
@@ -792,6 +792,48 @@ def configure_empty_disk(session, did, size, ostype, start="0M", n_partitions=1,
...
+        parts_out = results_stdout_52lts(process.run(parts_cmd))
+    parts = []
+    if parts_out:
+        for line in parts_out.rsplit("\n"):
+            if line.startswith("major") or line == "":
+                continue
+            parts_line = line.rsplit()
+            if len(parts_line) == 4:
+                parts.append(parts_line[3])
+    logging.debug("Find parts: %s" % parts)
```
balamuruhans: `logging.debug("Find parts: %s", parts)`

virttest/utils_disk.py
```diff
@@ -792,6 +792,48 @@ def configure_empty_disk(session, did, size, ostype, start="0M", n_partitions=1,
...
+    parts_list = get_parts_list(session=session)
+    for disk in parts_list:
+        cmd = ("udevadm info --query=all --name=/dev/{} | grep ID_SERIAL={}"
+               .format(disk, serial_str))
+        if session:
+            status = session.cmd_status(cmd)
+        else:
+            status = process.run(cmd, shell=True, ignore_status=True).exit_status
+        if not status:
+            logging.debug("Disk %s has serial %s" % (disk, serial_str))
```
balamuruhans: `logging.debug("Disk %s has serial %s", disk, serial_str)`

virttest/utils_test/libvirt.py
```diff
@@ -3303,7 +3282,7 @@ def new_disk_vol_name(pool_name):
         return None
     disk = poolxml.get_source().device_path[5:]
     part_num = len(list(filter(lambda s: s.startswith(disk),
-                               get_parts_list())))
+                               utils_misc.utils_disk.get_parts_list())))
```
balamuruhans: request to move `new_disk_vol_name()` to utils_disk.py

virttest/utils_disk.py
```diff
@@ -24,7 +24,7 @@
 
 from virttest import error_context
 from virttest import utils_numeric
-from virttest.compat_52lts import decode_to_text
+from virttest.compat_52lts import decode_to_text, results_stdout_52lts
```
balamuruhans: as good coding practise let's import individually,
`from virttest.compat_52lts import results_stdout_52lts`

Yingshun: Updated.

virttest/utils_disk.py
Yingshun: updated.

virttest/utils_disk.py
Yingshun: Done.

virttest/utils_test/libvirt.py
Yingshun: @balamuruhans , I will create other PR for this request.

Yingshun: @balamuruhans  It's going on beyond this Pull request. I think we can make a long term plan to reorganize the codes but not in here.  What do you think?

balamuruhans: Okay @Yingshun, I will take care of that part. I thought it would be easy when we do then and there. Thanks!

PR #2048, opened 2019-04-22
utils_disk: update package python-libguestfs for python3

YanhuiMa:
ID: 1694648
Signed-off-by: yama <yama@redhat.com>

PR #2094, opened 2019-05-28
virttest: Resolve the circular import between utils modules

vivianQizhu:
There is circular import between utils_disk, utils_misc and
utils_package, resolve it by replacing importing modules with
importing functions

id: 1714428


vivianQizhu:
Only when we want to import `utils_package` to `utils_disk` like https://github.com/avocado-framework/avocado-vt/pull/2048 So I am closing this one, should be resolve together with https://github.com/avocado-framework/avocado-vt/pull/2048

PR #2104, opened 2019-06-05
utils_disk: Add a method to check basic disk function in guest

Yingshun:
For virtual_disks tests, there is a function to check if the disk
partition in vm works well in individual .py files. It's a common
function and should be moved to avocado-vt. So add it to utils_disk.

### DIFF
--- a/virttest/utils_disk.py
+++ b/virttest/utils_disk.py
@@ -793,6 +793,27 @@ def configure_empty_disk(session, did, size, ostype, start="0M", n_partitions=1,
                                       labeltype, timeout)
 
 
+def linux_disk_check(session, did):
+    """
+    Check basic functions for linux guest disk on localhost.
+    Create partition on disk, format, mount, create a file and then unmount, clean
+    partition
+
+    :param session: VM session
+    :param did: Disk Kname in VM. eg. vdb
+    """
+
+    mount_point = configure_empty_linux_disk(session, did, "100M")
+    try:
+        if len(mount_point) != 1:
+            raise exceptions.TestError("Incorrect mount point {}".format(mount_point))
+        cmd = "echo teststring >  {}/testfile".format(mount_point[0])
+        if session.cmd_status(cmd):
+            raise exceptions.TestError("Failed to run {}".format(cmd))
+    finally:
+        clean_partition_linux(session, did)
+
+
 def get_parts_list(session=None):
     """
     Get all partition lists.


virttest/utils_disk.py
```diff
@@ -793,6 +793,25 @@ def configure_empty_disk(session, did, size, ostype, start="0M", n_partitions=1,
                                       labeltype, timeout)
 
 
+def check_disk_func_linux(session, did):
+    """
+    Check basic functions for linux guest disk.
```
dzhengfy: It seems the function can only support local guest session. So it is better to add 'local' in L798.

Yingshun: updated.

virttest/utils_disk.py
```diff
@@ -793,6 +793,25 @@ def configure_empty_disk(session, did, size, ostype, start="0M", n_partitions=1,
                                       labeltype, timeout)
 
 
+def check_disk_func_linux(session, did):
```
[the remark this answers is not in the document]
Yingshun: @balamuruhans updated.

virttest/utils_disk.py
```diff
@@ -793,6 +793,25 @@ def configure_empty_disk(session, did, size, ostype, start="0M", n_partitions=1,
...
+    """
+    Check basic functions for linux guest disk on localhost.
+    Create partition on disk, format, mount, create a file and then unmount, clean
+    partition
+
+    :param session: VM session
+    :param did: Disk Kname in VM. eg. vdb
+    """
+
+    mount_point = configure_empty_linux_disk(session, did, "100M")
```
xutian: It looks you should call function  'clean_partition_linux' in a finally statement,  right?

code should be looks like:
`try:
    if len(mount_point) != 1:
        raise exceptions.TestError("Incorrect mount point {}".format(mount_point))
    cmd = "echo teststring >  {}/testfile".format(mount_point[0])
    if session.cmd_status(cmd):
        raise exceptions.TestError("Failed to run {}".format(cmd))
finally:
    clean_partition_linux(session, did)`

Yingshun: @xutian There is a function to recover VM in test run finally block, I don't think we need to add it here.

xutian: @Yingshun  please ensure clean code in utils level.

Yingshun: @xutian  updated.

Yingshun: @xutian  Can you please approve it if changes look good?

PR #2105, opened 2019-06-07
Optional vm persistence

pevogam:
Adding optional vm persistence throughout tests/jobs is an old behavior in virt-test that seems to be missing in avocado-vt but is highly appreciated when it comes to improved performance due to no vm rebooting as well as setup reuse through qcow2 vm snapshots. The only way this can be possible is if the vms are not immediately destroyed and the environment/sockets/address_pool cleaned up, all in turn requiring a persistent temporary location that is not regenerated for each job. This branch is doing exactly this, keeping the cleaner (but less efficient) default behavior while making such new behavior optional.

### DIFF
--- a/avocado_vt/conf.d/vt.conf
+++ b/avocado_vt/conf.d/vt.conf
@@ -4,11 +4,15 @@ backup_image_before_test = True
 # Restore image after testing (if backup present)
 restore_image_after_test = True
 # Keep guest running between tests (faster, but unsafe)
-keep_guest_running=False
+keep_guest_running = False
 
 [vt.common]
 # Data dir path. If none specified, the default virt-test data dir will be used
 data_dir =
+# Make the temporary dir path persistent across jobs if needed.
+# By default the data in the temporary directory will be wiped after each test
+# in some cases and after each job in others.
+tmp_dir =
 # Enable only type specific tests. Shared tests will not be tested
 type_specific_only = False
 # RAM dedicated to the main VM
--- a/avocado_vt/test.py
+++ b/avocado_vt/test.py
@@ -395,17 +395,14 @@ class VirtTest(test.Test):
             test_modules[t_type] = imp.load_module(t_type, f, p, d)
             f.close()
 
-        # TODO: the environment file is deprecated code, and should be removed
-        # in future versions. Right now, it's being created on an Avocado temp
-        # dir that is only persisted during the runtime of one job, which is
-        # different from the original idea of the environment file (which was
-        # persist information accross virt-test/avocado-vt job runs)
+        # Open the environment file
         env_filename = os.path.join(data_dir.get_tmp_dir(),
                                     params.get("env", "env"))
         env = utils_env.Env(env_filename, self.env_version)
-        self.runner_queue.put({"func_at_exit": cleanup_env,
-                               "args": (env_filename, self.env_version),
-                               "once": True})
+        if params.get_boolean("job_env_cleanup", "yes"):
+            self.runner_queue.put({"func_at_exit": cleanup_env,
+                                   "args": (env_filename, self.env_version),
+                                   "once": True})
 
         test_passed = False
         t_type = None
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1011,15 +1011,21 @@ def preprocess(test, params, env):
             test.cancel("Failed to map hostname and ipaddress of target host")
         session.close()
 
-    # Destroy and remove VMs that are no longer needed in the environment
+    # Destroy and remove VMs that are no longer needed in the environment or
+    # leave them untouched if they have to be disregarded only for this test
     requested_vms = params.objects("vms")
+    keep_unrequested_vms = params.get_boolean("keep_env_vms", "no")
     for key in list(env.keys()):
         vm = env[key]
         if not isinstance(vm, virt_vm.BaseVM):
             continue
         if vm.name not in requested_vms:
-            vm.destroy()
-            del env[key]
+            if keep_unrequested_vms:
+                logging.debug("The vm %s is registered in the env and disregarded "
+                              "in the current test", vm.name)
+            else:
+                vm.destroy()
+                del env[key]
 
     if (params.get("auto_cpu_model") == "yes" and
             vm_type == "qemu"):
--- a/avocado_vt/loader.py
+++ b/avocado_vt/loader.py
@@ -126,6 +126,7 @@ class VirtTestLoader(loader.TestLoader):
         _add_if_not_exist('vt_log_level', 'debug')
         _add_if_not_exist('vt_console_level', 'debug')
         _add_if_not_exist('vt_datadir', data_dir.get_data_dir())
+        _add_if_not_exist('vt_tmp_dir', '')
         _add_if_not_exist('vt_config', None)
         _add_if_not_exist('vt_arch', None)
         _add_if_not_exist('vt_machine_type', None)
--- a/avocado_vt/options.py
+++ b/avocado_vt/options.py
@@ -76,6 +76,8 @@ class VirtTestOptionsProcess(object):
         # common section
         self.options.vt_data_dir = settings.get_value(
             'vt.common', 'data_dir', default=None)
+        self.options.vt_tmp_dir = settings.get_value(
+            'vt.common', 'tmp_dir', default='')
         self.options.vt_type_specific = settings.get_value(
             'vt.common', 'type_specific_only', key_type=bool,
             default=False)
--- a/virttest/data_dir.py
+++ b/virttest/data_dir.py
@@ -11,6 +11,7 @@ import shutil
 import stat
 
 from avocado.core import data_dir
+from avocado.core.settings import settings
 from avocado.utils import distro
 from avocado.utils import path as utils_path
 
@@ -208,6 +209,10 @@ def get_tmp_dir(public=True):
 
     :param public: If public for all users' access
     """
+    persistent_dir = settings.get_value('vt.common', 'tmp_dir',
+                                        default="")
+    if persistent_dir != "":
+        return persistent_dir
     tmp_dir = None
     # apparmor deny /tmp/* /var/tmp/* and cause failure across tests
     # it is better to handle here


virttest/qemu_monitor.py
```diff
@@ -113,7 +113,8 @@ def get_monitor_filename(vm, monitor_name):
     :param monitor_name: The monitor name.
     :return: The string of socket file name for qemu monitor.
     """
-    return os.path.join(data_dir.get_tmp_dir(),
+    basedir = vm.params.get("persistent_tmp_path", data_dir.get_tmp_dir())
```
xutian: it should not affect the storage location of test log files.

virttest/qemu_vm.py
```diff
@@ -580,8 +581,8 @@ def add_log_seabios(devices):
                 return ""
 
             default_id = "seabioslog_id_%s" % self.instance
-            filename = os.path.join(data_dir.get_tmp_dir(),
-                                    "seabios-%s" % self.instance)
+            basedir = params.get("persistent_tmp_path", data_dir.get_tmp_dir())
+            filename = os.path.join(basedir, "seabios-%s" % self.instance)
```
xutian: same comments in line#116

virttest/qemu_monitor.py
pevogam: Hi @xutian, can you elaborate on what you mean? If the "persistent_tmp_path" parameter is not set, the behavior should be identical to the one before. In order to be able to keep vms for longer than a test run we have to preserve the monitor filename together with others and thus have to change to a permanent location.

xutian: @pevogam 
If the persistent_tmp_path parameter is configured, the serial and monitor's logs are saved in the defined directory, but other test log files (for example, debug files) are still saved in the default location. This is not a good design.

pevogam: If by other files you mean the job log, test results and and so on, then I tend to disagree. The point here is *precisely* to keep only the files related to the persistence and communication with the vm object permanent (env file, monitor socket and so on) beyond tests and all these other files are test/job specific.

We do not store permanently any log files related to the serial console and monitor at least according to my checks but only the socket files which we need beyond a single test/job.

xutian: @pevogam 
If a VM is reused many times in a round of testing, but in one of the tests there is an abnormal state and we need to analyze the serial port log, but since all serial port logs are in the same file, this will give Analysis of the logs creates difficulties. This is my concern

pevogam: Hi @xutian can you provide me with pointers for the log filenames you are referring to? This is the file structure after a job run with the patch above applied:

```
# ls -1 <persistent_tmp_path>/
address_pool
address_pool.lock
env
monitor-catch_monitor-20190704-112820-jZnnwm3a
monitor-catch_monitor-20190708-051444-wD5AWvfT
monitor-catch_monitor-20190713-021459-puWT6b9d
monitor-catch_monitor-20190713-195338-GSytHpRz
monitor-catch_monitor-20190713-235755-MEg3TO1E
monitor-catch_monitor-20190714-143311-ncY8HKXJ
monitor-hmp1-20190704-112820-jZnnwm3a
monitor-hmp1-20190708-051444-wD5AWvfT
monitor-hmp1-20190713-021459-puWT6b9d
monitor-hmp1-20190713-195338-GSytHpRz
monitor-hmp1-20190713-235755-MEg3TO1E
monitor-hmp1-20190714-143311-ncY8HKXJ
seabios-20190704-112820-jZnnwm3a
seabios-20190708-051444-wD5AWvfT
seabios-20190713-021459-puWT6b9d
seabios-20190713-195338-GSytHpRz
seabios-20190713-235755-MEg3TO1E
seabios-20190714-143311-ncY8HKXJ
serial-serial1-20190704-112820-jZnnwm3a
serial-serial1-20190708-051444-wD5AWvfT
serial-serial1-20190713-021459-puWT6b9d
serial-serial1-20190713-195338-GSytHpRz
serial-serial1-20190713-235755-MEg3TO1E
serial-serial1-20190714-143311-ncY8HKXJ

ls -1 <results>/latest/test-results/<test-variant>/
catch_monitor-vm1.log
catch_monitor-vm2.log
catch_monitor-vm3.log
catch_monitor-vm4.log
data
debug.log
hmp1-vm1.log
hmp1-vm2.log
hmp1-vm3.log
hmp1-vm4.log
ip-sniffer.log
output
remote.log
screendumps_vm1_3206_iter0
screendumps_vm2_9319_iter0
screendumps_vm3_38756_iter0
screendumps_vm4_36179_iter0
stderr
stdout
sysinfo
vm_register_vm1_3206
vm_register_vm2_9319
vm_register_vm3_38756
vm_register_vm4_36179
whiteboard
```

From what I gather you are referring to the monitor-related logs like `catch_monitor` and `hmp1`? If so they are already within the test-specific result folder as you suggested. Regarding the serial log file scenario you described, the same thing should apply for it too:

```python
        log_name = "serial-%s-%s.log" % (tmp_serial, self.name)
        self.serial_console_log = os.path.join(utils_misc.get_log_file_dir(),
                                               log_name)
        self.serial_console = aexpect.ShellSession(
            "nc -U %s" % self.get_serial_console_filename(tmp_serial),
```

In this snippet you can see that the serial log file is set from `utils_misc` and only the socket is modified by the optional permanent path in the `get_serial_console_filename` method call.

pevogam: Hi @xutian, is the above answer satisfactory for your inquiries?

pevogam:
Hi @xutian, can you update me on the status of your review on this branch?

xutian:
@pevogam  Sorry for the late reply, I will retest your patch and update my comments  later.

virttest/qemu_monitor.py
xutian: Looks ok

pevogam:
Thank you so much @xutian for spending the time to review and approve this pull request! This is incredibly helpful for us and issue #2134 since this is the last pull request blocker for revolving it. @luckyh Do you mind giving us a hand on this PR too? It will allow our plugin to work without any patches to the Avocado VT repo which is why I would really greatly appreciate it.

avocado_vt/test.py
```diff
@@ -405,17 +405,18 @@ def _runTest(self):
...
-        # different from the original idea of the environment file (which was
-        # persist information accross virt-test/avocado-vt job runs)
-        env_filename = os.path.join(data_dir.get_tmp_dir(),
-                                    params.get("env", "env"))
+        # Open the environment file
+        custom_env_path = params.get("custom_env_path", "")
+        if custom_env_path:
+            env_path = custom_env_path
+        else:
+            env_path = data_dir.get_tmp_dir()
```
luckyh: IMHO, it's a new feature for VT and that means now `tmp_dir` is configurable, if so, we should implement it in `data_dir.get_tmp_dir()`, otherwise all the callers should implement the same logic and that would be horrible because in some context we don't have `params` been passed.

There could be two approaches:

1. introduce a new environment variable (e.g. `AVOCADO_VT_TMP_DIR`) and reference it in `data_dir.get_tmp_dir()`, then if someone wants to use a specified directory the avocado command line would be:  `AVOCADO_VT_TMP_DIR=/path/to/users/tmp avocado run ...`
2. introduce a new option to the VT configuration (https://github.com/avocado-framework/avocado-vt/blob/master/avocado_vt/conf.d/vt.conf), then users are only required to update the configuration and the avocado command line won't be changed. 

What do you think?

pevogam: Makes sense mostly because we could possibly reuse a persistent tmp path for other purposes later on or in all circumstances when we want to avoid the automated cleanup in between tests. I think a config item for avocado VT is more reasonable than an environment variable in order to keep all configuration in one place and a place that is more in the domain of a python program. I will ping you once I push code adding this change.

pevogam: However, why exactly do we have to add it to `data_dir.get_tmp_dir()`? The people would expect a true temporary directory instead of a persistent one. The directory obtained via `data_dir.get_tmp_dir()` is exactly the one that is wiped by the VT code after each test and this is exactly what we want to avoid by adding a parameter (or config or environmental variables) for a *persistent* tmp path. I guess we could still add the persistent tmp path as a config parameter or do you think then it is better off staying as Cartesian config parameter as we don't expect it to be used beyond the current environment processing scope?

pevogam: Hi @luckyh sorry for multiple edits on this comment - it has been quite many months and I constantly needed to refresh my mind on the changes in order to reply to you better. There are some troubles from your suggestion:

1) Changing the behavior of `data_dir.get_tmp_dir()` would affect way more people and code using it and the current overwriting with an optional temporary location is only persistent with respect to vm data across test runs, nothing else. This means that if for instance I set a config item "persistent_path" and change the `get_tmp_dir()` as:

```
    persistent_tmp_dir = settings.get_value('vt.setup', 'persistent_tmp_dir', default=None)
    if persistent_tmp_dir is None:
        return persistent_tmp_dir
```
This will replace the default temporary location with a custom one for all calling instances *while* this PR is only making sure the vm objects (running vms) are persistent, leaving the default temporary directory for everything else. We could call `get_tmp_dir(persistent=True)` only from the current locations but this brings me to the second point.

2) The changed code locations don't necessarily apply the same logic. While some (more than one) will check for persistent location first and if not will default to the original temporary directory, some have to additionally check if persistent location is set (in a boolean fashion) and this will require at least two lines of importing `settings` like above at each location which is about the same number of lines code we tried to reduce by refactoring into `get_tmp_dir`.

3) We also have other parameters used strictly for persistent vm objects like "vm_action" which are also Cartesian parameters like the "persistent_tmp_path". All of these parameters are Cartesian since this is the normal way to operate in `env_process.py` and `qemu_vm.py` and after some search I couldn't find a single instance of importing `avocado.settings` or `avocado_vt.options` into any module in the entire `virttest` folder.

luckyh: > 1. Changing the behavior of `data_dir.get_tmp_dir()` would affect way more people and code using it and the current overwriting with an optional temporary location is only persistent with respect to vm data across test runs, nothing else. This means that if for instance I set a config item "persistent_tmp_dir" and change the `get_tmp_dir()` as:
> 
> 
> ```
>     persistent_tmp_dir = settings.get_value('vt.setup', 'persistent_tmp_dir', default=None)
>     if persistent_tmp_dir is not None:
>         return persistent_tmp_dir
> ```
> 
> This will replace the default temporary location with a custom one for all calling instances _while_ this PR is only making sure the vm objects (running vms) are persistent, leaving the default temporary directory for everything else. We could call `get_tmp_dir(persistent=True)` only from the current locations but this brings me to the second point.

Ok, let me explain further, to make sure the vm objects are persistent, there must be a singleton for tracing and managing them and, as you know, it is the `env` object, and then we can let vms be present across test instances.

Generally, `env` can be in-memory objects and could be shared with test instances directly, and in this case you don't really care about how it is implemented because it can work properly. However, since test instances are running within different processes, it is not easy to make env still in-memory, and hence, it is stored in disk currently and the `tmp` dir of test instances is where it is placed (although IMO, the best place should be the `run` dir, but avocado doesn't have this concept so far). 

Now let's take a look at the issue you met, since each test instances (within one avocado/avocado-vt job) will recreate the `tmp` dir, it is not possible to make sure `env` is persistent, and such that the vm objects can't be persistent too. To resolve the issue, the key point is to make sure `env` is persistent, and then the dir where `env` is placed should be persistent (no matter what and how).

So, what I suggested is to let `env` be still in the **`tmp`** dir and make sure the `tmp` dir is persistent, therefore, `data_dir.get_tmp_dir` should point to the same path as well. Otherwise, as what the current approach does, you are attempting to introduce some new directory resource and it just happens to be called "tmp", and it is explained that, about the second point, why you thought the changed code locations don't apply the same logic, that's because we do have two kinds of directory then.

Moreover, not only `env` would be placed in the `tmp` dir, some other temporary resources used by the vms (such as sockets, directories, images) may be placed in the same room, if we have two kinds of `tmp` dir, these temporary resources should be placed in the persistent one too, but other developers may confuse about that.

> 
> 2. The changed code locations don't necessarily apply the same logic. While some (more than one) will check for persistent location first and if not will default to the original temporary directory, some have to additionally check if persistent location is set (in a boolean fashion) and this will require at least two lines of importing `settings` like above at each location which is about the same number of lines of code we try to reduce by refactoring into `get_tmp_dir` on the first place.
> 
> 3. We also have other parameters used strictly for persistent vm objects like "vm_action" which are also Cartesian parameters like the "persistent_tmp_path". All of these parameters are Cartesian since this is the normal way to operate in `env_process.py` and `qemu_vm.py` and after some search I couldn't find a single instance of importing `avocado.settings` or `avocado_vt.options` into any module in the entire `virttest` folder.

Sorry, I just forgot to send other comments together in the first one. About the functionality of `vm_action` my question is, is it really necessary? If we would let some "outdated" vms run in the current frame, does it mean these vms are more likely *requested*? wouldn't these vms be defined in configurations?

luckyh:
Btw, I just checked that, for the latest avocado, all the test instances are using the same `tmp` dir, would you please help double confirm that?
```
JOB ID     : 8c5aab0266d9def1ebe21db784351431c59f4bad
JOB LOG    : /home/kar/workspace/job-results/job-2020-01-09T18.28-8c5aab0/job.log
 (1/2) repeat1.Guest.RHEL.8.1.0.x86_64.io-github-autotest-qemu.boot: FAIL: /var/tmp/avocado_fqyy62qt (28.58 s)
 (2/2) repeat2.Guest.RHEL.8.1.0.x86_64.io-github-autotest-qemu.boot: FAIL: /var/tmp/avocado_fqyy62qt (28.82 s)
RESULTS    : PASS 0 | ERROR 0 | FAIL 2 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 59.43 s
JOB HTML   : /home/kar/workspace/job-results/job-2020-01-09T18.28-8c5aab0/results.html
```

And it is interested that whether test instances would have different `tmp` dir depends on how avocado initializes the data resources.

Here is an example:
```python
import os
import time
from multiprocessing import Process, Queue

from virttest import data_dir


garbage = Queue()


def cleanup():
    global garbage
    while not garbage.empty():
        try:
            os.rmdir(garbage.get())
        except:
            pass


def get_tmp_dir():
    global garbage
    tmp_dir = data_dir.get_tmp_dir()
    garbage.put(tmp_dir)
    print(tmp_dir)
    time.sleep(3)


if __name__ == '__main__':
    get_tmp_dir()
    pool = [Process(target=get_tmp_dir) for i in range(2)]
    any(map(lambda p: p.start(), pool))
    any(map(lambda p: p.join(), pool))
    cleanup()
```
if the `tmp` dir is created by the parent process, all the test instances would share the same one.
```
/tmp/avocado_toho019z
/tmp/avocado_toho019z
/tmp/avocado_toho019z
```
if the `tmp` dir is created by test instances, then each instance would have its own one.
```patch
 if __name__ == '__main__':
-    get_tmp_dir()
     pool = [Process(target=get_tmp_dir) for i in range(2)]
     any(map(lambda p: p.start(), pool))
     any(map(lambda p: p.join(), pool))
```
```
/tmp/avocado_5y2w_7_c
/tmp/avocado_14rr9x7o
```
Not sure if avocado has any promise about that.

avocado_vt/test.py
pevogam: > Now let's take a look at the issue you met, since each test instances (within one avocado/avocado-vt job) will recreate the `tmp` dir, it is not possible to make sure `env` is persistent, and such that the vm objects can't be persistent too. To resolve the issue, the key point is to make sure `env` is persistent, and then the dir where `env` is placed should be persistent (no matter what and how).

Yes, I confirm everything so far.

> So, what I suggested is to let `env` be still in the **`tmp`** dir and make sure the `tmp` dir is persistent, therefore, `data_dir.get_tmp_dir` should point to the same path as well. Otherwise, as what the current approach does, you are attempting to introduce some new directory resource and it just happens to be called "tmp", and it is explained that, about the second point, why you thought the changed code locations don't apply the same logic, that's because we do have two kinds of directory then.

Do we know the extend to which this will affect other tests and code however? Are we sure that there are no parts of the code that will actually require a temporary storage that should be cleaned up in between tests/runs? I don't want to break other people's code which is why defining an extra optional and more persistent location seemed less intrusive than this. We can make preserving the temporary location optional, but not separately for each part of code, it is either permanent or nonpermanent per run and user configuration.

> Moreover, not only `env` would be placed in the `tmp` dir, some other temporary resources used by the vms (such as sockets, directories, images) may be placed in the same room, if we have two kinds of `tmp` dir, these temporary resources should be placed in the persistent one too, but other developers may confuse about that.

Yes, I see it might be confusing for other developers. If you think making the temporary directory more persistent with a switch should not cause too much trouble we can go that path. Can you confirm this is what you think?

> > 1. The changed code locations don't necessarily apply the same logic. While some (more than one) will check for persistent location first and if not will default to the original temporary directory, some have to additionally check if persistent location is set (in a boolean fashion) and this will require at least two lines of importing `settings` like above at each location which is about the same number of lines of code we try to reduce by refactoring into `get_tmp_dir` on the first place.
> > 2. We also have other parameters used strictly for persistent vm objects like "vm_action" which are also Cartesian parameters like the "persistent_tmp_path". All of these parameters are Cartesian since this is the normal way to operate in `env_process.py` and `qemu_vm.py` and after some search I couldn't find a single instance of importing `avocado.settings` or `avocado_vt.options` into any module in the entire `virttest` folder.
> 
> Sorry, I just forgot to send other comments together in the first one. About the functionality of `vm_action` my question is, is it really necessary? If we would let some "outdated" vms run in the current frame, does it mean these vms are more likely _requested_? wouldn't these vms be defined in configurations?

Yes, the parameter is necessary since our plugin for instance can perform separate setup tests on different vms, say test A for vm1 and test B for vm3 and then can run a conjoint test C using both vm1 and vm3 after the previous test validation. The default action of removing the vm is reasonable for a temporary storage like the one so far but is needed for cases of reuse of previously used vms. I can show you sample code if you are interested.

pevogam:
> Btw, I just checked that, for the latest avocado, all the test instances are using the same `tmp` dir, would you please help double confirm that?

Yes, I can confirm that. However, this doesn't resolve the persistence across jobs as underlined by the previously removed TODO of what `virt-test` was allowing which is the behavior we are trying to restore:

```diff
-        # TODO: the environment file is deprecated code, and should be removed
-        # in future versions. Right now, it's being created on an Avocado temp
-        # dir that is only persisted during the runtime of one job, which is
-        # different from the original idea of the environment file (which was
-        # persist information accross virt-test/avocado-vt job runs)
-        env_filename = os.path.join(data_dir.get_tmp_dir(),
-                                    params.get("env", "env"))
+        # Open the environment file
```

There are multiple immediate benefits from reusing previous vms across jobs, here a few examples:

- We can boot previously installed vms and do some manual customization on them, even develop a test more easily.

- We can easily log into a vm and debug a test that failed and just completed but left the vm running in its final post-test state.

- We can upload and download files to a vm, save it into a certain state (e.g. qemu snapshot), and reuse this state for a lot of test without redoing it and thus saving a lot of time.

Bottom line is that we can manage different vms in real time with automation provided by Avocado VT even if this automation does not involve full self-preparing self-cleaning tests. The above thus although nice does not solve the problem at hand.

avocado_vt/test.py
pevogam: I have pushed the changes you requested - please see below.

pevogam:
Hi @luckyh, I have pushed the branch with two parts now - one that implements the persistence of temporary storage as a VT setting as you requested, and another one that makes the minimal changes possible to preserve the running vms and their registrations in the env file. Please take a look when you can and tell me if it follows your comments exactly.

avocado_vt/conf.d/vt.conf
```diff
@@ -4,7 +4,11 @@ backup_image_before_test = True
 # Restore image after testing (if backup present)
 restore_image_after_test = True
 # Keep guest running between tests (faster, but unsafe)
-keep_guest_running=False
+keep_guest_running = False
+# Make the temporary dir path persistent across jobs if needed.
+# By default the data in the temporary directory will be wiped oafter each test
+# in some cases and after each job in others.
+persistent_tmp_dir =
```
luckyh: Would you please elaborate why the option is put in `vt.setup` instead of `vt.common`?

luckyh: And frankly, I do prefer the name to be `tmp_dir` instead, because VT doesn't care of whether the given directory is truly persistent or not. From VT's point of view it just makes the path of `tmp_dir` to be user customized, otherwise, nothing changed.

virttest/env_process.py
```diff
@@ -1011,15 +1011,24 @@ def preprocess(test, params, env):
             test.cancel("Failed to map hostname and ipaddress of target host")
         session.close()
 
-    # Destroy and remove VMs that are no longer needed in the environment
+    # Destroy and remove VMs that are no longer needed in the environment or
+    # leave them untouched if they have to be disregarded only for this test
     requested_vms = params.objects("vms")
+    vm_env_action = params.get("vm_env_action", "destroy")
```
luckyh: I'd suggest separating this feature into a new commit.

luckyh: How does the name `vm_disposal` sound like?

luckyh: Maybe `unrequested_vm_disposal` is more accurate.

avocado_vt/conf.d/vt.conf
pevogam: Because this is where most options related to setup between test is, e.g.:
```
# Keep guest running between tests (faster, but unsafe)
keep_guest_running = False
```
or

```
# Backup image before testing (if not already backed up)
backup_image_before_test = True
# Restore image after testing (if backup present)
restore_image_after_test = True
```

but if you mean to rename it as general purpose `tmp_dir` then it would make more sense in the `vt.common` section. I will do this in yet another push.

pevogam: Alright, will do in the next push

avocado_vt/test.py
```diff
@@ -395,17 +395,21 @@ def _runTest(self):
             test_modules[t_type] = imp.load_module(t_type, f, p, d)
             f.close()
 
-        # TODO: the environment file is deprecated code, and should be removed
-        # in future versions. Right now, it's being created on an Avocado temp
-        # dir that is only persisted during the runtime of one job, which is
-        # different from the original idea of the environment file (which was
-        # persist information accross virt-test/avocado-vt job runs)
```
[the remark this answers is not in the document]
pevogam: And as someone that agrees, do you have a suggestion for an alternative to achieve persistence of the vms?

The TODO removed here is entirely deprecated if we are to restore optional vm persistence and the TODO states *exactly* that it has to be removed since it is a remnant of `virt-test`'s persistence option.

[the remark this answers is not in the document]
pevogam: And as I mentioned in other places in your review like https://github.com/avocado-framework/avocado-vt/pull/2105#issuecomment-573187772, this persistence is needed for our own plugin for plenty of reasons. I don't see why we can't make this optional for those of us that need the extra flexibility - whenever the `avocado-vt` maintainers considered only permanence within jobs they might not have had in mind all the possible use cases and reasonable extensions. It is a normal part of the development process.

avocado_vt/test.py
```diff
@@ -395,17 +395,21 @@ def _runTest(self):
...
-                               "args": (env_filename, self.env_version),
-                               "once": True})
+        vm_env_action = params.get("vm_env_action", "destroy")
+        if vm_env_action == "disregard":
+            logging.debug("The the env and its registered vms are disregarded "
+                          "in the current test and will be kept further")
+        elif vm_env_action == "destroy":
+            self.runner_queue.put({"func_at_exit": cleanup_env,
+                                   "args": (env_filename, self.env_version),
+                                   "once": True})
```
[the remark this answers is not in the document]
pevogam: Good point, changed with the next push.

virttest/env_process.py
pevogam: If you suggest a separate commit, then I would argue it makes more sense in the commit about making vms persistent which includes the optional cleanup of the env file and the optional destruction of vm objects. If you mean it as a separate PR, then I think this PR is starting to turn into a PR about making the `tmp_dir` permanent which was never its focal point and while this is a good addition, I prefer the PR to remain focused on the vm permanence. An extra commit for the persistent tmp directory should be enough.

pevogam: Given your point about the `env_cleanup` parameter and the fact that we won't use `vm_env_action` for this purpose too, indeed the name is no longer the most precise and should be changed. Let's choose `unrequested_vms_disposal` as a boolean `yes` and `no` parameter then. Coming with the next push.

pevogam: IMO the previous choices of `vm_env_action` were a bit more future proof for multiple possibilities but this is not that important and maybe it won't be extended any further.

virttest/env_process.py
```diff
@@ -1011,15 +1011,21 @@ def preprocess(test, params, env):
             test.cancel("Failed to map hostname and ipaddress of target host")
         session.close()
 
-    # Destroy and remove VMs that are no longer needed in the environment
+    # Destroy and remove VMs that are no longer needed in the environment or
+    # leave them untouched if they have to be disregarded only for this test
     requested_vms = params.objects("vms")
+    vm_disposal = params.get_boolean("unrequested_vms_disposal", "yes")
```
ldoktor: By the current implementation you can safely skip the whole loop in case `vm_disposal` is True, right?

Anyway do you think it'd be useful to allow per-vm granularity by using `object_params(vm.name)` to get the `unrequested_vms_disposal`? (in such case it'd have to be called `unrequested_vm_disposal`)

Also the param is fairly cryptical, how about `keep_env_vms` or `skip_env_vms_postprocess`?

And one more thing, wouldn't this clash with the https://github.com/avocado-framework/avocado-vt/pull/2322 (I mean it would, but can be solved by the same logic...)

pevogam: > By the current implementation you can safely skip the whole loop in case `vm_disposal` is True, right?

I think you mean "if `vm_disposal` is False" and the answer is not quite because I would still like to (debug) log what vms have been discarded.

> Anyway do you think it'd be useful to allow per-vm granularity by using `object_params(vm.name)` to get the `unrequested_vms_disposal`? (in such case it'd have to be called `unrequested_vm_disposal`)

At least not in out use cases - so far we either keep or we destroy unregistered vms. If this feature proves itself to others (and so far this seems to be a hard task from the reviews) it is not trouble to extend it. But vm granularity is a use case that has to prove itself needed considering people still have to cross the first stage and keep vms overall.

> Also the param is fairly cryptical, how about `keep_env_vms` or `skip_env_vms_postprocess`?

It was proposed by @luckyh earlier today in the comments ending with https://github.com/avocado-framework/avocado-vt/pull/2105#discussion_r366833979 where I agreed with him that using 'env' in the parameter name could be misleading. Maybe we could change to `keep_env_vms` if he agrees and we reach one joint choice.

> And one more thing, wouldn't this clash with the #2322 (I mean it would, but can be solved by the same logic...)

Not sure but this PR is much older (June 2019) and @xutian approved it as well so I guess he is aware of the changes undertaken here.

pevogam:
Hi @ldoktor 

> @pevogam vm persistance is a tricky beast. I spent a lot of time fixing bugs in it. On the other hand there are the benefits and the code used to work pretty well so there is no reason in bringing this feature back. But don't complain in case you see weird things, you had been warned...

Since the added functionality is optional and you can always remain on the safe side and not worry about potential problems. For our reasons to use it see below.

> Can I ask what is your usecase? For me surviving one job is usually enough. Or do you have a use-case where you need the same/similar VMs throughout many jobs?

I have pointed to good use cases right above in https://github.com/avocado-framework/avocado-vt/pull/2105#issuecomment-573187772 and in the comment where you were mentioned today. Bottom line for all of them is that we can manage multiple vms across setup reusable from previous jobs.

virttest/env_process.py
pevogam: I took the liberty to rename the parameter to `keep_env_vms` without further feedback. Hopefully this will prevent some unnecessary additional back and forth.

luckyh: > It was proposed by @luckyh earlier today in the comments ending with [#2105 (comment)](https://github.com/avocado-framework/avocado-vt/pull/2105#discussion_r366833979) where I agreed with him that using 'env' in the parameter name could be misleading. Maybe we could change to `keep_env_vms` if he agrees and we reach one joint choice.

@pevogam @ldoktor ok, let me clarify my point: it would be better to make the new param decouple with both the concept and the implement of 'env', that's because even without them we can also have this feature (will tell more about this later). The thing we should highlight is those relevant VMs are "unrequested" (or in other words "unregistered") so I put it into the name. And about the last word "disposal" I just want it to express that how we deal with the VMs, as I'm not a native speaker, maybe it's not that good.

With this feature, users can let the VMs they don't requested be alive across tests or even jobs, and no matter whether the maintainers agree about the proposal of job-level persistence the feature still makes sense, so that's why I suggested a separate commit.

Anyway, the name `keep_env_vms` looks good to me too, feel free to pick that, thanks.

luckyh:
> And as someone that agrees, do you have a suggestion for an alternative to achieve persistence of the vms?

@pevogam the rough idea is by introducing the concept of VM registration, it's about let users be able to register their VMs to the job. It may sounds like to be another env, but the main difference is it's just about the VMs.

There could be different implementations, for example:

* introducing external daemons and let VT talk to them (similar with the libvirt stuff in VT)
* loading from serialized files (similar with the env solution)
...

Once VT takes over these VMs they can be used as usual, and at the end of the job they can be *"dumped"* to where users specified, and then they be registered to the next round if needed.

As you can imagine, these solutions request more complicated code changes than the current one (by using env), and I think it is less likely to be approved by other maintainers, so I didn't mentioned it in this PR.

avocado_vt/test.py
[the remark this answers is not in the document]
ldoktor: Yep, removing the TODO is fine. We wanted to remove this because keeping VMs running after test is tricky, keeping it after job is even trickier and there always were some issues like leaked fds, duplicate VMs or different VMs pretending to be the same.

With this said we don't have an alternative and as far as I know nobody works on changing this. And if you are brave and know what you want to do (keep VMs after job finishes) the deprecation note needs to be removed.

ldoktor:
> Hi @ldoktor
> 
> > @pevogam vm persistance is a tricky beast. I spent a lot of time fixing bugs in it. On the other hand there are the benefits and the code used to work pretty well so there is no reason in bringing this feature back. But don't complain in case you see weird things, you had been warned...
> 
> Since the added functionality is optional and you can always remain on the safe side and not worry about potential problems. We have used this feature for years now so we are used to its characteristics. For our reasons to use it see below.
> 

Sure, I was just warning you, not nacking. We used to have this functionality in autotest so I'm tempted to see it in Avocado as well. On the other hand I know how many false-negatives and even false-positives it resulted in...

> > Can I ask what is your usecase? For me surviving one job is usually enough. Or do you have a use-case where you need the same/similar VMs throughout many jobs?
> 
> I have pointed to good use cases right above in [#2105 (comment)](https://github.com/avocado-framework/avocado-vt/pull/2105#issuecomment-573187772) and in the comment where you were mentioned today as well as the second commit of this PR. Bottom line for all of them is that we can manage multiple vms across setup reusable from previous jobs.

Sure, I read that. I meant something more specific. The reason I'm asking is that this implements the tmpdir location as config option, which might indicate most of the jobs will share the setting. Is your workflow to keep using the same tmpdir for days/weeks or are you going to have tmpdirs per subset of tests? How are you going to clean them etc.

I'm a big fan of yaml_loader so usually all tests fit into a single job and restarting the machines after the long run is usually not that big issue for me (but rather a useful cleanup). That's why I asked whether you really need multi-job persistence.

All in all I welcome this change, the idea looks fairly simple and it does not affect the standard usage. I don't have time for thorough review, but in general it makes sense to me. I'd only consider whether the tmpdir shouldn't be customizable via cmdline (in case you want to change it regularly) to avoid the need to create multiple config files. But if that's not the case having it in config only is less intrusive...

avocado_vt/test.py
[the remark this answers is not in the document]
pevogam: So far we have been using this feature for 7 years, all the way from `virt-test` to patched `avocado-vt` :) We have seen all of this but in the last years things have been fairly clean if we exclude an occasional glitch in LVM that we use for taking vm snapshots and reusing them for tests. This being said, I will resolve this then and summarize everything left from these reviews below.

virttest/env_process.py
pevogam: I am resolving this for the sake of my summary of anything left from these reviews below.

pevogam:
Hi @luckyh 

> > And as someone that agrees, do you have a suggestion for an alternative to achieve persistence of the vms?
> 
> @pevogam the rough idea is by introducing the concept of VM registration, it's about let users be able to register their VMs to the job. It may sounds like to be another env, but the main difference is it's just about the VMs.

Is this meant to be with regular parameters (configuration only approach) or are users supposed to define this in code?

> There could be different implementations, for example:
> 
>     * introducing external daemons and let VT talk to them (similar with the libvirt stuff in VT)
> 
>     * loading from serialized files (similar with the env solution)
>       ...
> 
> 
> Once VT takes over these VMs they can be used as usual, and at the end of the job they can be _"dumped"_ to where users specified, and then they can be registered to the next round if needed.
> 
> As you can imagine, these solutions request more complicated code changes than the current one (by using env), and I think it is less likely to be approved by other maintainers, so I didn't mentioned it in this PR.

I actually really like the flexibility that the first bullet point solution could offer. It would reduce complexity if the handling is similar to the libvirt case and it allow ultimate freedom on where the vms are taken from and where they are left. In our case we also have some semi-automated scenarios where we would like to perform some repetitive work on a vm followed by some manual work and taking a vm (usually from a given start snapshot) then running such setup and leaving the vm in a desired state is perfectly solved by a solution like this. As @ldoktor mentioned perhaps there isn't much activity in this direction but we could implement anything similar as an optional and purely experimental alternative, then truly deprecate worse methods if we reach a stable stage.

------------------------------------

Hi @ldoktor 

> The reason I'm asking is that this implements the tmpdir location as config option, which might indicate most of the jobs will share the setting. Is your workflow to keep using the same tmpdir for days/weeks or are you going to have tmpdirs per subset of tests? How are you going to clean them etc.

Yes, the use case in this regard is fairly straightforward - simply use a permanent form of a `tmp_dir` for all tests as we need the vms for all tests from multiple runs. When it comes to isolation then, we rely on LXC to contain a specific avocado installation and run multiple sets of identical or differentiated vms in parallel. Usually the sockets are cleaned and we don't accumulate such. Some other critical files like `address_pool` are simply overwritten. So the end result is that we don't accumulate files. We do seem to accumulate `aexpect` folders but this is all wiped once the container is destroyed and we constantly have to create new containers for new jobs and modifications. If the `tmp_dir` resides in RAM memory then the same would also happen on container shutdown.

> I'm a big fan of yaml_loader so usually all tests fit into a single job and restarting the machines after the long run is usually not that big issue for me (but rather a useful cleanup). That's why I asked whether you really need multi-job persistence.

It could be a cleanup indeed, but for example we have a "semi-automated" step (in the terminology above) where we would simply boot all vms of a certain type or simply bring all of them to a desired snapshot. None of this is done manually through the backend tools but through a single command line invocation from our plugin and will not be possible if the vms are rebooted - some operations like this require the vm to be on before the test begins like saving a certain RAM state and so on.

> All in all I welcome this change, the idea looks fairly simple and it does not affect the standard usage. I don't have time for thorough review, but in general it makes sense to me. I'd only consider whether the tmpdir shouldn't be customizable via cmdline (in case you want to change it regularly) to avoid the need to create multiple config files. But if that's not the case having it in config only is less intrusive...

I have added the `tmp_dir` setting everywhere I found it could be relevant, not sure if there is any other place at present that would still make sense. So far we follow the minimal approach as usual and can expand on this later on if there is a true demand for it.

pevogam:
Hi @luckyh and @ldoktor, what is the chance this PR would make it into the 75.0 release? I have executed every request you have had, waited over the weekend, and now just received a sudden notice that the release meeting is today oO I really think we should be telling in advance about these meetings.

luckyh:
On Mon, Jan 20, 2020 at 9:18 PM Plamen Dimitrov <notifications@github.com>
wrote:

> Hi @luckyh <https://github.com/luckyh> and @ldoktor
> <https://github.com/ldoktor>, what is the chance this PR would make it
> into the 75.0 release? I have executed every request you have had, waited
> over the weekend, and now just received a sudden notice that the release
> meeting is today oO I really think we should be telling in advance about
> these meetings.
>

Sorry that I have some urgent work in hands these days and have to finish
them before the Lunar New Year holiday (just three days later). I see that
the 75.0 already came out, so, sorry it missed the chance this time, anyway
I will give a review soon.

Since I don't usually attend the release meetings, please ping @clebergnu
in case that you have any important requests to be handled by next coming
releases, thanks.

avocado_vt/test.py
```diff
@@ -395,17 +395,14 @@ def _runTest(self):
...
-        # different from the original idea of the environment file (which was
-        # persist information accross virt-test/avocado-vt job runs)
+        # Open the environment file
         env_filename = os.path.join(data_dir.get_tmp_dir(),
                                     params.get("env", "env"))
         env = utils_env.Env(env_filename, self.env_version)
-        self.runner_queue.put({"func_at_exit": cleanup_env,
-                               "args": (env_filename, self.env_version),
-                               "once": True})
+        if params.get_boolean("env_cleanup", "no"):
```
luckyh: Sorry I just realized that what we needed at here is actually different with the one the `env_cleanup` currently represents, the `env_cleanup` is a per test config while this is a per job one, so have to introduce a new param which could be `job_env_cleanup`.

luckyh:
> > @pevogam the rough idea is by introducing the concept of VM registration, it's about let users be able to register their VMs to the job. It may sounds like to be another env, but the main difference is it's just about the VMs.
> 
> Is this meant to be with regular parameters (configuration only approach) or are users supposed to define this in code?

@pevogam yes I think so, simply it could be regular parameters and I think it's enough for most cases, but it also could be other forms (plugins, test types, ...) if needed, thanks.

avocado_vt/test.py
pevogam: Ok

avocado_vt/test.py
```diff
@@ -395,17 +395,14 @@ def _runTest(self):
...
-        # different from the original idea of the environment file (which was
-        # persist information accross virt-test/avocado-vt job runs)
+        # Open the environment file
         env_filename = os.path.join(data_dir.get_tmp_dir(),
                                     params.get("env", "env"))
         env = utils_env.Env(env_filename, self.env_version)
-        self.runner_queue.put({"func_at_exit": cleanup_env,
-                               "args": (env_filename, self.env_version),
-                               "once": True})
+        if params.get_boolean("job_env_cleanup", "no"):
```
luckyh: The default value should be `"yes"` as normally we always want to do env_cleanup.

pevogam: I think some of the changes got lost from too many iterations of the PR. Fixed now.

luckyh:
@ldoktor any further comments to this PR? thanks!

luckyh:
Thanks.

PR #2106, opened 2019-06-07
Add vm pre- and post-processing hooks for on/off vm stages

pevogam:
These hooks can then be used by custom code for additional stages
during environment setup and cleanup when the vms are on and/or off.

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -71,6 +71,14 @@ _setup_manager = test_setup.SetupManager()
 _pre_hugepages_surp = 0
 _post_hugepages_surp = 0
 
+#: Hooks to use for own customization stages of the virtual machines with
+#: test, params. and env as supplied arguments
+preprocess_vm_off_hook = None
+preprocess_vm_on_hook = None
+postprocess_vm_on_hook = None
+postprocess_vm_off_hook = None
+
+
 #: QEMU version regex.  Attempts to extract the simple and extended version
 #: information from the output produced by `qemu -version`
 QEMU_VERSION_RE = r"QEMU (?:PC )?emulator version\s([0-9]+\.[0-9]+\.[0-9]+)\s?\((.*?)\)"
@@ -751,6 +759,10 @@ def preprocess(test, params, env):
     """
     error_context.context("preprocessing")
 
+    # Run this hook before any network setup stage and vm creation.
+    if callable(preprocess_vm_off_hook):
+        preprocess_vm_off_hook(test, params, env)  # pylint: disable=E1102
+
     # Check if code coverage for qemu is enabled and
     # if coverage reset is enabled too, reset coverage report
     gcov_qemu = params.get("gcov_qemu", "no") == "yes"
@@ -1255,6 +1267,11 @@ def preprocess(test, params, env):
             if THREAD_ERROR:
                 raise exceptions.TestFail("Test inside nested guest "
                                           "reported failure")
+
+    # Run this hook after any network setup stage and vm creation.
+    if callable(preprocess_vm_on_hook):
+        preprocess_vm_on_hook(test, params, env)  # pylint: disable=E1102
+
     return params
 
 
@@ -1269,6 +1286,16 @@ def postprocess(test, params, env):
     """
     error_context.context("postprocessing")
     err = ""
+
+    # Run this hook before closing the connections to the qemu monitors
+    # and possibly destroying the vms.
+    if callable(postprocess_vm_on_hook):
+        try:
+            postprocess_vm_on_hook(test, params, env)  # pylint: disable=E1102
+        except Exception as details:
+            err += "\nPostprocessing living vm hook: %s" % str(details).replace('\\n', '\n  ')
+            logging.error(details)
+
     migration_setup = params.get("migration_setup", "no") == "yes"
     if params.get("verify_guest_dmesg", "yes") == "yes" and params.get("start_vm", "no") == "yes":
         guest_dmesg_log_file = params.get("guest_dmesg_logfile", "guest_dmesg.log")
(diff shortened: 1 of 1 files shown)


virttest/env_process.py
```diff
@@ -62,6 +62,14 @@
 _setup_manager = test_setup.SetupManager()
 
 
+#: Hooks to use for own customization stages of the virtual machines
+#: with params and env as supplied arguments
+preprocess_vm_off_hook = None
```
balamuruhans: the hooks are initialized here, but how to customize or assign this hooks with ?

pevogam: Here is a real life example from our plugin:

```python
# add env processing hooks to handle online/offline state get/set operations
env_process.preprocess_vm_on_hook = state_setup.get_state
def set_online_state(fn):
    def wrapper(run_params, env):
        run_params["skip_types"] = "offline"
        fn(run_params, env)
    return wrapper
def set_offline_state(fn):
    def wrapper(run_params, env):
        run_params["skip_types"] = "online ramdisk"
        fn(run_params, env)
    return wrapper
env_process.postprocess_vm_on_hook = set_online_state(state_setup.set_state)
env_process.postprocess_vm_off_hook = set_offline_state(state_setup.set_state)
```

I don't think this should really be described as hooks are meant to be replaced by customized functions. Or am I missing something?

pevogam:
> please squash the commits to one.

Squashed as requested.

virttest/env_process.py
```diff
@@ -669,6 +677,10 @@ def preprocess(test, params, env):
     """
     error_context.context("preprocessing")
 
+    # Run this hook before any network setup stage and vm creation.
+    if preprocess_vm_off_hook:
```
luckyh: Suggest:
```python
if callable(preprocess_vm_off_hook):
```

virttest/env_process.py
```diff
@@ -62,6 +62,14 @@
 _setup_manager = test_setup.SetupManager()
 
 
+#: Hooks to use for own customization stages of the virtual machines
+#: with params and env as supplied arguments
```
luckyh: Should add the argument `test` as well, to make it consistent with the callers.

virttest/env_process.py
```diff
@@ -62,6 +62,14 @@
...
+preprocess_vm_off_hook = None
+preprocess_vm_on_hook = None
```
luckyh: Is that possible to have hook chains? Just a rough idea, request for your comments.

virttest/env_process.py
```diff
@@ -669,6 +677,10 @@ def preprocess(test, params, env):
     """
     error_context.context("preprocessing")
 
+    # Run this hook before any network setup stage and vm creation.
+    if preprocess_vm_off_hook:
+        preprocess_vm_off_hook(params, env)  # pylint: disable=E1102
```
luckyh: How should we handle the failures (exceptions)?

virttest/env_process.py
pevogam: Incredibly good idea, thanks a lot!

virttest/env_process.py
pevogam: Added, including a comment "or others of your choosing".

virttest/env_process.py
pevogam: I would say the current hoots are most useful as entry points for your own code, once inside your own custom function you can can just implement a chain of calls. Correct me if I misunderstood what you mean of course.

virttest/env_process.py
pevogam: I think handling like argument selection and implementation are all up to the caller. The VT test is already taking care of the pre- and post-processing errors and at least to my experience (already using hooks and raising LVM and other exceptions from there) these are regularly called as errors during pre- or post-processing.

pevogam:
Hi @luckyh @balamuruhans, I have pushed all of your requested changes and hope I have answered all of your questions. I would appreciate if you help merging this as we are already collecting conflicts with the current `virttest/env_process.py` that were not present at the time the PR was first introduced.

balamuruhans:
@pevogam ACK, could you help rebase the patch ?

virttest/env_process.py
luckyh: Sorry that I haven't check this comment yet since I was busy on some urgent works last few months.

Here the problem is, by the current approach, VT test does catch the pre- and post-processing errors to prevent crashing the process, but it is logically incorrect because any exception raised by hooks would break the entire pre-/post-processing, that means the codes after the hooks won't be executed anymore, which includes some important ones. It doesn't matter for the pre-processing as the test has not been started at that time, but it does affect the post- one.

Let see an example:
```python
# Run this hook after any vms are actually off to ensure data is
# written to disk.
if callable(postprocess_vm_off_hook):
    postprocess_vm_off_hook(test, params, env)  # pylint: disable=E1102

if err:
```
supposes that the variable `err` is set to some valid value before `postprocess_vm_off_hook` being called, it means the test will be failed by that reason, however, what if `postprocess_vm_off_hook` raises another error? How will the test status to be set?

IMO, calling hooks are just optional steps for the two phases and others are mandatory, in any cases we should ensure the phases are finished completely.

pevogam: Thanks for elaborating, I was still thinking from the perspective of the old autotest's virttest where exceptions of any kind are always caught at singe preprocessing and postprocessing clauses. I just force pushed a version that hopefully addresses your concerns.

pevogam:
> @pevogam ACK, could you help rebase the patch ?

Rebased as was your request.

Thankfully the conflicts for this PR are easy to manage but there are others that incorporate complicated and elaborate changes that might not have the same luck with slower reviews.

balamuruhans:
Thanks all. let's merge

PR #2118, opened 2019-06-20
 Update libvirt_vm.py with aarch support to undefine similar to vm_xml.py

conv8888:
When running multivm stress, I notice the VMs were not being undefined which cause the subsequent test failures. Lifted this patch from vm_xml.py. Just wondering if it is the right thing to do?


conv8888:
Or if this is more acceptable?       

     if self.params.get("vir_domain_undefine_nvram") == "yes":
                options = "--nvram"
            else:
                options = None

PR #2120, opened 2019-06-21
[WIP]env_process: Support to update the boot options for qemu vm

YongxueHong:
The class VM in qemu_vm module has no attribute start()
for updating the boot options, so add support for qemu vm.

Signed-off-by: Yongxue Hong <yhong@redhat.com>

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1075,7 +1075,25 @@ def preprocess(test, params, env):
         for vm in env.get_all_vms():
             if vm:
                 if not vm.is_alive():
-                    vm.start()
+                    if params.get("vm_type") == "libvirt":
+                        vm.start()
+                    elif params.get("vm_type") == "qemu":
+                        # Start the VM (or restart it if it's already up)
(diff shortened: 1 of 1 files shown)


virttest/env_process.py
```diff
@@ -1149,7 +1167,25 @@ def postprocess(test, params, env):
         for vm in env.get_all_vms():
             if vm:
                 if not vm.is_alive():
-                    vm.start()
+                    if params.get("vm_type") == "libvirt":
+                        vm.start()
+                    elif params.get("vm_type") == "qemu":
+                        # Start the VM (or restart it if it's already up)
+                        if params.get("reuse_previous_config", "no") == "no":
```
vivianQizhu: From my view this is unnecessary, seems `reuse_previous_config = yes` only used for multiple host, here is post process, so directly pass `params` should be sufficient here.

vivianQizhu: Also the comment `(or restart it if it's already up)` should be remove, we have `if not vm.is_alive()` above, so no such condition here.

PR #2128, opened 2019-06-26
kernel_interface: handle invalid literal for int()

balamuruhans:
Sys and proc interface can return value that may not be
type casted to int type, so return as string on receiving
`ValueError`.


balamuruhans:
@sathnaga made the changes as suggested

PR #2129, opened 2019-06-26
Fixup: Improper initialization of vmcpu_xml

sathnaga:
In some scenarios, we would hit below issue due to vmcpu_xml
is not initialized in proper place, this patch tries to fix it.

UnboundLocalError: local variable 'vmcpu_xml' referenced before assignment

Reported-by: Junxiang Li <junli@redhat.com>

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -790,8 +790,8 @@ class VMXML(VMXMLBase):
                                          'threads': threads}
                 vmxml['cpu'] = vmcpu_xml
             try:
+                vmcpu_xml = vmxml['cpu']
                 if update_numa and vmxml.cpu.numa_cell:
-                    vmcpu_xml = vmxml['cpu']
                     no_numa_cell = len(vmxml.cpu.numa_cell)
                     if vcpus >= no_numa_cell:
                         vcpus_num = vcpus // no_numa_cell


hsmj1412:
@sathnaga The line 819 is still out of the if part, so I dont think it will work.

sathnaga:
> @sathnaga The line 819 is still out of the if part, so I dont think it will work.

So, moved the initialisation out of if part aswell,
```
            try:
                vmcpu_xml = vmxml['cpu']
                if update_numa and vmxml.cpu.numa_cell: 
```

hsmj1412:
@sathnaga it LGTM, but I still have another problem about line 794(if update_numa and vmxml.cpu.numa_cell:).
Will the update_numa is True but the numa_cell is not included(defined) in vmxml.cpu part?

sathnaga:
> @sathnaga it LGTM, but I still have another problem about line 794(if update_numa and vmxml.cpu.numa_cell:).
> Will the update_numa is True but the numa_cell is not included(defined) in vmxml.cpu part?

Does that check break for you?, I guess it should not.

Intention of this code is to, correct/update the numa definition only if present, while we change the vcpu numbers from test case explicitly, if not numa cell is present, we are good without any change.

This is not intended to add_numa()...agree?

hsmj1412:
> > @sathnaga it LGTM, but I still have another problem about line 794(if update_numa and vmxml.cpu.numa_cell:).
> > Will the update_numa is True but the numa_cell is not included(defined) in vmxml.cpu part?
> 
> Does that check break for you?, I guess it should not.
> 
> Intention of this code is to, correct/update the numa definition only if present, while we change the vcpu numbers from test case explicitly, if not numa cell is present, we are good without any change.
> 
> This is not intended to add_numa()...agree?

I mean if you are sure the numa_cell is used, then this check is useless due to numa_cell should not be None, False or other negative type. Otherwise, if the numa_cell is not used, it will report UnboundLocalError.
In summary, to use vmxml.cpu.get('numa_cell') is better?

PR #2131, opened 2019-06-27
env_process: refactor updating kernel params during pre/postprocess

balamuruhans:
Perform updating kernel params inside preprocess_vm() and cleanup
in postprocess_vm() by doing so,

    * vm specific kernel params can be provided for multivm cases
    * `start_vm` check and creating/starting vm is taken care
      for libvirt and qemu test.
    * re-iteration of vm objects and re-checks can be avoided

Reported-by: Yongxue Hong <yhong@redhat.com>

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -207,6 +207,36 @@ def preprocess_vm(test, params, env, name):
                           migration_mode=params.get("migration_mode"),
                           migration_fd=params.get("migration_fd"),
                           migration_exec_cmd=params.get("migration_exec_cmd_dst"))
+
+        # Update kernel param
+        kernel_extra_params_add = params.get("kernel_extra_params_add", "")
+        kernel_extra_params_remove = params.get("kernel_extra_params_remove", "")
+        if params.get("disable_pci_msi"):
+            disable_pci_msi = params.get("disable_pci_msi")
+            if disable_pci_msi == "yes":
+                if "pci=" in kernel_extra_params_add:
+                    kernel_extra_params_add = re.sub("pci=.*?\s+", "pci=nomsi ",
+                                                     kernel_extra_params_add)
+                else:
+                    kernel_extra_params_add += " pci=nomsi"
+                params["ker_remove_similar_pci"] = "yes"
+            else:
+                kernel_extra_params_remove += " pci=nomsi"
+        if (params.get("enable_guest_iommu") and
+                utils_misc.get_cpu_vendor(verbose=False) == 'GenuineIntel'):
+            enable_guest_iommu = params.get("enable_guest_iommu")
+            if enable_guest_iommu == "yes":
+                kernel_extra_params_add += " intel_iommu=on"
+            else:
+                kernel_extra_params_remove += " intel_iommu=on"
+            guest_iommu_option = params.get("guest_iommu_option")
+            if guest_iommu_option:
+                kernel_extra_params_add += " iommu=%s" % guest_iommu_option
+        if kernel_extra_params_add or kernel_extra_params_remove:
+            utils_test.update_boot_option(vm,
+                                          args_added=kernel_extra_params_add,
+                                          args_removed=kernel_extra_params_remove)
+
     elif not vm.is_alive():    # VM is dead and won't be started, update params
         vm.devices = None
         vm.params = params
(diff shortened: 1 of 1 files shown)


virttest/env_process.py
```diff
@@ -1073,11 +1073,11 @@ def preprocess(test, params, env):
...
         for vm in env.get_all_vms():
-            if vm:
-                if not vm.is_alive():
+            if params.get("vm_type") == "libvirt":
+                if vm and not vm.is_alive():
                     vm.start()
-                utils_test.update_boot_option(vm, args_added=kernel_extra_params_add,
-                                              args_removed=kernel_extra_params_remove)
+            utils_test.update_boot_option(vm, args_added=kernel_extra_params_add,
+                                          args_removed=kernel_extra_params_remove)
```
YongxueHong: I thought there doesn't need to start VM after  `process(test, params, env, preprocess_image, preprocess_vm)` that will not disturb the status of start_vm of the case level, should keep the initial status of VM.
Thanks.

virttest/env_process.py
```diff
@@ -1147,11 +1147,11 @@ def postprocess(test, params, env):
...
         for vm in env.get_all_vms():
-            if vm:
-                if not vm.is_alive():
+            if params.get("vm_type") == "libvirt":
+                if vm and not vm.is_alive():
                     vm.start()
-                utils_test.update_boot_option(vm, args_added=kernel_extra_params_remove,
-                                              args_removed=kernel_extra_params_add)
+            utils_test.update_boot_option(vm, args_added=kernel_extra_params_remove,
+                                          args_removed=kernel_extra_params_add)
```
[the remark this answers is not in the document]
balamuruhans: I see couple of testcases in tp-qemu uses this param, so I think it is better to be supported for qemu tests as well. Also I believe it is a good to have this working in both qemu and libvirt that helps us to bringup VM and test it with different kernel commandline options.

We have placed this option before postprocessing of VMs in postprocess. So, if test uses the params and VM is dead in the context then framework should raise the error, which it does now.

virttest/env_process.py
balamuruhans: do you mean even for libvirt test, it is not necessary to start the VM as preprocess_vm takes care ?

YongxueHong: umr~~
yes,  from my point of view.

virttest/env_process.py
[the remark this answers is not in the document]
YongxueHong: You said that raises the error is just for qemu vm, but not for libvirt vm since libvirt vm will start even if the vm is dead in the context from the above codes.
So why libvirt vm could do start vm ?  qemu vm can't.

virttest/env_process.py
balamuruhans: Updating the boot options remains the same. the catch here is to handle managing the VM in generic way.

virttest/env_process.py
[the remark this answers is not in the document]
balamuruhans: do you want to create VM for qemu test if vm is dead in qemu test instead of error out ?
In libvirt test, start VM will work because VM xml is defined, persistent and managed by libvirt, but in qemu test we have to handle. The only option I could think of is to create the VM and proceed incase of qemu test. Please share your thoughts

virttest/env_process.py
YongxueHong: One scenario that if the parameter `start_vm` is set  `start_vm = no` in the case level, then set some parameters for VM by preprocess_vm.But the VM has been started at this time, so you have to shut down it then preprocess_vm again. I think it goes against the purpose of the case.

balamuruhans: if `start_vm = no` is set by the test case then test should take care to update kernel param if at all it is going to use VM, so we can raise `TestSetupError()` if user sets both `start_vm = no` and `kernel_extra_params_*` from env_process. what say ?

balamuruhans:
@YongxueHong, I have handled to check for `start_vm`, if the test case is very specific by setting `start_vm = no`, then env_process skips updating kernel_param and test case should call `update_boot_option()` explicitly if it require to update kernel param.

Please help to review.

virttest/env_process.py
```diff
@@ -1072,10 +1072,13 @@ def preprocess(test, params, env):
 
     # change the kernel params if configured
     if kernel_extra_params_add or kernel_extra_params_remove:
-        for vm in env.get_all_vms():
-            if vm:
-                if not vm.is_alive():
-                    vm.start()
+        if params.get("start_vm", "no") == "yes":
```
sathnaga: Am not sure when will we endup in this situation, as we do start the vm here `process(test, params, env, preprocess_image, preprocess_vm)`  if right params are set in config `create_vm_libvirt` for libvirt and `start_vm` for both qemu/libvirt ?

```
  # Preprocess all VMs and images
    if params.get("not_preprocess", "no") == "no":
        process(test, params, env, preprocess_image, preprocess_vm)

    # change the kernel params if configured
    if kernel_extra_params_add or kernel_extra_params_remove:
        if params.get("start_vm", "no") == "yes":
            for vm in env.get_all_vms():
                if vm and not vm.is_alive():
                    if params.get("vm_type") == "libvirt":
                        vm.start()
                    elif params.get("vm_type") == "qemu":
                        vm.create(params=params)
                utils_test.update_boot_option(vm, args_added=kernel_extra_params_add,
                                              args_removed=kernel_extra_params_remove)```

Comment (sathnaga) [reply virttest/env_process.py]: probably we can set start_vm and other relavant params if we see `kernel_extra_params_add` and `kernel_extra_params_remove` are set just before `process(test, params, env, preprocess_image, preprocess_vm)` and let framework do the magic?

Comment (balamuruhans) [reply virttest/env_process.py]: I have taken care of it. Thanks for the suggesion :+1:

**Comment by balamuruhans** [comment]
@YongxueHong @sathnaga addressed all your comments, please help to review.

File: virttest/env_process.py [hunk 4]
```diff
@@ -415,6 +445,26 @@ def postprocess_vm(test, params, env, name):
     if not vm:
         return
 
+    # recover the changes done to kernel params in postprocess
```
sathnaga: check for `start_vm` and then proceed, as we have set only if `start_vm` is set in pre-process.

balamuruhans: Thank you @sathnaga for reviewing and making code better.

balamuruhans:
@YongxueHong request for your Ack.

balamuruhans:
Thanks @YongxueHong @sathnaga

PR #2135, opened 2019-06-28
run_virtio_serial_file_transfer: make serial data transfter script work

vivianQizhu:
Make it running on py2 and Linux first, will fix py3 and Windows issue later separately

id: 1724987, 1734208, 1734210

### DIFF
--- a/shared/deps/serial/VirtIoChannel_guest_send_receive.py
+++ b/shared/deps/serial/VirtIoChannel_guest_send_receive.py
@@ -50,7 +50,7 @@ class VirtIoChannel:
         self.ack_format = "3s"
         self.ack_msg = b"ACK"
         self.hi_format = "2s"
-        self.hi_msg = "HI"
+        self.hi_msg = b"HI"
         self.recv_msg = b'ALLRECEIVED'
         if self.is_windows:
             vport_name = '\\\\.\\Global\\' + device_name
@@ -86,7 +86,7 @@ class VirtIoChannel:
             self.send(self.hi_msg)
             txt = self.receive(hi_msg_len)
             out = struct.unpack(self.hi_format, txt)[0]
-            if out != "HI":
+            if out != b"HI":
                 raise ShakeHandError("Fail to get HI from host.")
             size_s = struct.pack("q", size)
             self.send(size_s)
@@ -96,7 +96,7 @@ class VirtIoChannel:
                 raise ShakeHandError("Host didn't ACK the file size message.")
             return size
         elif action == "receive":
-            txt = ''
+            txt = b''
             while len(txt) < hi_msg_len:
                 txt += self.receive(hi_msg_len)
             hi_str = struct.unpack(self.hi_format, txt)[0]
--- a/shared/deps/serial/VirtIoChannel_guest_send_receive.py
+++ b/shared/deps/serial/VirtIoChannel_guest_send_receive.py
@@ -51,6 +51,7 @@ class VirtIoChannel:
         self.ack_msg = b"ACK"
         self.hi_format = "2s"
         self.hi_msg = "HI"
+        self.recv_msg = b'ALLRECEIVED'
         if self.is_windows:
             vport_name = '\\\\.\\Global\\' + device_name
             from windows_support import WinBufferedReadFile
@@ -182,6 +183,7 @@ def receive(device, filename, p_size=1024):
             md5_value.update(txt)
             file_no.write(txt)
             recv_size += len(txt)
+        vio.send(vio.recv_msg)
     finally:
         file_no.close()
         if vio:
--- a/shared/deps/serial/VirtIoChannel_guest_send_receive.py
+++ b/shared/deps/serial/VirtIoChannel_guest_send_receive.py
@@ -95,7 +95,9 @@ class VirtIoChannel:
                 raise ShakeHandError("Host didn't ACK the file size message.")
             return size
         elif action == "receive":
-            txt = self.receive(hi_msg_len)
+            txt = ''
+            while len(txt) < hi_msg_len:
+                txt += self.receive(hi_msg_len)
             hi_str = struct.unpack(self.hi_format, txt)[0]
             if hi_str != self.hi_msg:
                 raise ShakeHandError("Fail to get HI from host.")
--- a/shared/deps/serial/VirtIoChannel_guest_send_receive.py
+++ b/shared/deps/serial/VirtIoChannel_guest_send_receive.py
@@ -86,22 +86,21 @@ class VirtIoChannel:
             txt = self.receive(hi_msg_len)
             out = struct.unpack(self.hi_format, txt)[0]
             if out != "HI":
-                raise ShakeHandError("Fail to get HI from guest.")
+                raise ShakeHandError("Fail to get HI from host.")
             size_s = struct.pack("q", size)
             self.send(size_s)
             txt = self.receive(ack_msg_len)
             ack_str = struct.unpack(self.ack_format, txt)[0]
             if ack_str != self.ack_msg:
-                raise ShakeHandError("Guest didn't ACK the file size message.")
+                raise ShakeHandError("Host didn't ACK the file size message.")
             return size
         elif action == "receive":
             txt = self.receive(hi_msg_len)
             hi_str = struct.unpack(self.hi_format, txt)[0]
             if hi_str != self.hi_msg:
-                raise ShakeHandError("Fail to get HI from guest.")
+                raise ShakeHandError("Fail to get HI from host.")
             self.send(txt)
             size = self.receive(8)
-            print("xxxx size = %s" % size)
             if size:
                 size = struct.unpack("q", size)[0]
                 txt = struct.pack(self.ack_format, self.ack_msg)
@@ -190,20 +189,20 @@ def receive(device, filename, p_size=1024):
 
 
 def send(device, filename, p_size=1024):
-    recv_size = 0
+    send_size = 0
     f_size = os.path.getsize(filename)
     vio = VirtIoChannel(device)
     vio.shake_hand(f_size, action="send")
     md5_value = md5_init()
     file_no = open(filename, 'rb')
     try:
-        while recv_size < f_size:
+        while send_size < f_size:
             txt = file_no.read(p_size)
             vio.send(txt)
             md5_value.update(txt)
-            recv_size += len(txt)
+            send_size += len(txt)
     finally:
-        print("received size = %s" % recv_size)
+        print("Sent size = %s" % send_size)
         file_no.close()
         if vio:
             vio.close()
--- a/shared/deps/serial/VirtIoChannel_guest_send_receive.py
+++ b/shared/deps/serial/VirtIoChannel_guest_send_receive.py
@@ -242,6 +242,8 @@ if __name__ == "__main__":
         parser.error("Please set -f parameter.")
     p_size = options.package
     action = options.action
+    if action not in ("receive", "send", "both"):
+        parser.error('Please set -a parameter: "receive", "send", "both"')
 
     if action == "receive":
         md5_sum = receive(device, filename, p_size=p_size)
--- a/shared/deps/serial/VirtIoChannel_guest_send_receive.py
+++ b/shared/deps/serial/VirtIoChannel_guest_send_receive.py
@@ -181,7 +181,7 @@ def receive(device, filename, p_size=1024):
             txt = vio.receive(p_size)
             md5_value.update(txt)
             file_no.write(txt)
-            recv_size += p_size
+            recv_size += len(txt)
     finally:
         file_no.close()
         if vio:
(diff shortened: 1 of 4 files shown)


vivianQizhu:
03:48:02 INFO | PASS 1-Host_RHEL.m7.u7.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.virtio_serial_file_transfer

vivianQizhu:
@luckyh Would you help review? Thanks.

shared/deps/serial/VirtIoChannel_guest_send_receive.py
```diff
@@ -95,7 +95,9 @@ def shake_hand(self, size=0, action="receive"):
                 raise ShakeHandError("Host didn't ACK the file size message.")
             return size
         elif action == "receive":
-            txt = self.receive(hi_msg_len)
+            txt = ''
+            while len(txt) < hi_msg_len:
+                txt += self.receive(hi_msg_len)
```
luckyh: It should be either:
```python
txt = b''
while not txt:
    txt = self.receive(hi_msg_len)
```
or
```python
txt = b''
while len(txt) < hi_msg_len:
    txt += self.receive(hi_msg_len - len(txt))
```

luckyh: And if we have to add this waiting behavior, the same thing should be done for `serial_host_send_receive.py` as there is no guarantee of which script must be launched at first.

virttest/utils_test/__init__.py
```diff
@@ -797,13 +797,12 @@ def transfer_data(session, host_cmd, guest_cmd, n_time, timeout,
...
     dir_name = data_dir.get_tmp_dir()
     transfer_timeout = int(params.get("transfer_timeout", 720))
-    tmp_dir = params.get("tmp_dir", data_dir.get_tmp_dir())
     filesize = int(params.get("filesize", 10))
     count = int(filesize)
 
     host_data_file = os.path.join(dir_name,
                                   "tmp-%s" % utils_misc.generate_random_string(8))
-    guest_data_file = os.path.join(tmp_dir,
+    guest_data_file = os.path.join(guest_path,
```
luckyh: `guest_path` is used for keeping the tool script file rather than the data file, it should use `tmp_dir` yet and just pick another default value to resolve the issue you met.

shared/deps/serial/VirtIoChannel_guest_send_receive.py
vivianQizhu: @luckyh Yes, thanks. Updated to second one, but no `b` there, since it might need further handling on `receive` and `send` functions, I will fix py3 issues together later.

vivianQizhu: @luckyh `serial_host_send_receive.py` has no such issue since it use `socket.recv` instead of `os.read`, it will wait until receive something by itself. I have tested it does wait there until receive handshake message.

virttest/utils_test/__init__.py
vivianQizhu: @luckyh Okay, I will use `/var/tmp` instead, just not sure what do we use for windows guest in this case?

luckyh: Maybe `%TEMP%` I guess, @xiagao would you please help confirm this? thanks.

shared/deps/serial/serial_host_send_receive.py
```diff
@@ -191,6 +191,7 @@ def main():
 
     vport = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
     vport.connect(sock)
+    vport.setsockopt(1, socket.SO_SNDBUF, p_size)
```
luckyh: I haven't seen the possible reason to have to do so, so would please explain more on it? For my point of view, `shutdown()`/`close()` would sufficiently resolve the issue.

vivianQizhu: @luckyh To be honest I don't really understand how it affects. I had same think with you, I suppose `close` should resolve the issue, but from the test result, only set send buffer size can resolve the issue.
With this set size line, no matter have shutdown and close lines or not, the test can pass.
Well without this line, no matter have shutdown and close lines or not, the test would fail.

shared/deps/serial/VirtIoChannel_guest_send_receive.py
luckyh: ok, thanks for the clarification.

shared/deps/serial/serial_host_send_receive.py
luckyh: @vivianQizhu well, that might be a product bug or something else, so better to figure out it.

shared/deps/serial/VirtIoChannel_guest_send_receive.py
```diff
@@ -86,22 +86,23 @@ def shake_hand(self, size=0, action="receive"):
...
             txt = self.receive(ack_msg_len)
             ack_str = struct.unpack(self.ack_format, txt)[0]
             if ack_str != self.ack_msg:
-                raise ShakeHandError("Guest didn't ACK the file size message.")
+                raise ShakeHandError("Host didn't ACK the file size message.")
             return size
         elif action == "receive":
-            txt = self.receive(hi_msg_len)
+            txt = ''
+            while len(txt) < hi_msg_len:
```
xutian: why do this update ，   txt = self.receive(hi_msg_len) looks OK in here？

vivianQizhu: @xutian If start guest script before host one, later line `hi_str = struct.unpack(self.hi_format, txt)[0]`  will raise error since `txt` will be None as nothing received.

xutian: Ok, better to comments here.

vivianQizhu: @xutian Explained it in commit message: https://github.com/avocado-framework/avocado-vt/pull/2135/commits/5eaf9039bc7d6bd410fe025b425c5d1d38c16e6b

vivianQizhu:
@luckyh Updated,
1. Use '/var/tmp/' for `tmp_dir`, windows will be configured in case level.
2. Add an ack message after guest receive all data as we discussed.
Please help review, thanks.

shared/deps/serial/serial_host_send_receive.py
```diff
@@ -156,9 +156,23 @@ def send(connect, filename, p_size=1024):
...
+    exp_str = b'ALLRECEIVED'
+    exp_size = len(exp_str)
+    r_str = b''
+    r_size = 0
+    while r_size < exp_size:
+        txt = connect.recv(exp_size)
+        r_size += len(txt)
+        r_str += txt
+    if struct.unpack("11s", txt)[0] != exp_str:
+        raise ShakeHandError("Not get Received ACK from guest.")
```
luckyh: I think we need that works on dual side, I mean both `host -> guest` and `guest -> host`, right?

vivianQizhu: @luckyh I have thought about that, but finally not add `guest->host` side because 1. Host has receive buffer so lower risk for data loss; 2. I was worried about the process would hang at any while loop if meet exception(like previous hang at guest receive loop), one more while loop more risk for that. So I prefer not adding more while loop if not very necessary. But still I am open to it, let me know your thought please, thanks.

vivianQizhu:
@luckyh Add py3 fix and test passed, please hlep review together, thanks.
(1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.x86_64.io-github-autotest-qemu.virtio_serial_file_transfer: PASS (41.87 s)

shared/deps/serial/serial_host_send_receive.py
luckyh: > @luckyh I have thought about that, but finally not add `guest->host` side because 1. Host has receive buffer so lower risk for data loss;

So that means there is the probability to get an automation failure, right? then @yanan-fu, what do you think about that?

> 2. I was worried about the process would hang at any while loop if meet exception(like previous hang at guest receive loop), one more while loop more risk for that. So I prefer not adding more while loop if not very necessary. But still I am open to it, let me know your thought please, thanks.

Well, it doesn't sound like a problem, actually you will either get md5 mismatch or transmission timeout failures when the exception happened. And, to prevent tests getting stuck we can simply give a reasonable `timeout` value to `session.cmd`/`process.run`.

vivianQizhu: @luckyh I have tried let guest also wait for host ack after receive data, with current version code. But it introduce new issue: Host script will exit after send its 'receive ack', so sometimes guest will loss this 'ack',  and go into infinite waiting loop, finally lead guest session timeout and fail the case. That's not 100% reproducible but have very high chance to hit(2/3 in my test today). I think for current implementation it's not a good idea to let guest wait more. And I think the 'probability' for host to loss data is very very low, we can ignore it. What do you think?

yanan-fu: @luckyh  @vivianQizhu  After checking, I agree with xuhan more about check `guest -> host`,  same with `guest -> host`. what's more, I can't understand well why the 'probability' for host to loss data is low, but not guest.
And for host script exit after send 'receive ack', it is should be a bug which need be fix rather than bypass it.

vivianQizhu: @yanan-fu I believe that's because host use socket so it has receive buffer, while guest use os.read() so no receive buffer. Now let guest wait receive ack from host increases the chance guest loss data(the ack message), It failed a lot in my test.
1. We only hit data loss from guest side currently, by adding the receive ack for `host-> guest` it does not fail anymore.
2. Never hit data loss from host so far(test more than 100 times).
3. Add receive ack for `guest->host`, case will fail some times cause host script exit after send ack message but before guest receive it.
In one word, your request will not fix any issue but only highly increase the fail ratio of this case.

vivianQizhu: And the 'probability' for host to loss data, currently it only exists in theory, never really hit it, so I say it is very rare. While the guest loss data is hit a lot of times during my recent test.

yanan-fu: > @yanan-fu I believe that's because host use socket so it has receive buffer, while guest use os.read() so no receive buffer. Now let guest wait receive ack from host increases the chance guest loss data(the ack message), It failed a lot in my test.
> 
>     1. We only hit data loss from guest side currently, by adding the receive ack for `host-> guest` it does not fail anymore.
> 
>     2. Never hit data loss from host so far(test more than 100 times).
> 
>     3. Add receive ack for `guest->host`, case will fail some times cause host script exit after send ack message but before guest receive it.

My point is, this is a issue need be fixed. we can not miss the checking step, just because the change can cause new issue. Thanks!

>        In one word, your request will not fix any issue but only highly increase the fail ratio of this case.

vivianQizhu: @yanan-fu No, don't be confused, there is no such steps, add receive ack is to fix issue, it's not a checking step. The only thing we need to check is that sender and receiver has same md5 value after file transfter, which has to be accomplished. If there is no guest data loss issue(that's introduced by previous code, cause it uses os.read, and it has to use it), we don't need the receive ack at all. 
Actually my previous solution is to set small send buffer to make host sender slower, to match the rhythm of guest receiver, will can reduce the data loss, just in theory it still has the risk.

vivianQizhu: @yanan-fu To be more clear, the issue we need to fix is: guest will loss data if host scrip exit before guest finish all receive action, cause guest has no receive buffer. So the solution is let host wait for guest tells that it finish all data receive action: Add a receive ack after guest receive all data.
Host has no such issue, and because the guest no buffer issue, let guest wait host's ack will cause the same problem: guest might loss the ack message if host exit first.
We can't make host exit first before guest receive all data, that's why we can't let guest wait for host's ack. Hope I explained it better this way.

vivianQizhu: @yanan-fu Based on the above situation, if you have better solution, that will be appreciated. Thanks.

yanan-fu: @vivianQizhu my opinion remains the same,  since i still think `guest -> host` and `host -> guest` are same in theory, to avoid the risk.  However, you can insist yours if you think the risk is acceptable from feature testing perspective. Thanks

vivianQizhu: @yanan-fu I have to say even in theory they are not same, cause their read/write method are totally not same.

vivianQizhu:
@lijinlijin Add commit https://github.com/avocado-framework/avocado-vt/pull/2135/commits/3339d6437c4b84fcfa81fed12726514824121f20 to restrict action in ('receive', 'send', 'both'), please help review, thanks.

lijinlijin:
I run on win2008-32,win7-32, win8-32, win8.1-32 and win10-32 and repeat 5 times, all passed:
Summary:
Finshed=55, PASS=55

vivianQizhu:
@lijinlijin Thanks, let's move forward.

PR #2141, opened 2019-07-03
Add new mechanism to upgrade system during installation

yanan-fu:
id: 1724985

### DIFF
--- a/virttest/tests/unattended_install.py
+++ b/virttest/tests/unattended_install.py
@@ -394,6 +394,11 @@ class UnattendedInstallConfig(object):
             raise ValueError("Unexpected installation medium %s" % self.url)
         contents = re.sub(dummy_medium_re, content, contents)
 
+        dummy_rh_system_stream_id_re = r'\bRH_SYSTEM_STREAM_ID\b'
+        if re.search(dummy_rh_system_stream_id_re, contents):
+            rh_system_stream_id = self.params.get("rh_system_stream_id", "")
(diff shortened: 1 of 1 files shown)


xutian:
Any update for kickstart file ?

virttest/tests/unattended_install.py
```diff
@@ -394,6 +394,11 @@ def answer_kickstart(self, answer_path):
             raise ValueError("Unexpected installation medium %s" % self.url)
         contents = re.sub(dummy_medium_re, content, contents)
 
+        dummy_rh_system_stream_id_re = r'\bRH_SYSTEM_STREAM_ID\b'
+        if re.search(dummy_rh_system_stream_id_re, contents):
+            rh_system_stream_id = self.params.get("rh_system_stream_id", "")
```
xutian: Please use params["rh_system_stream_id"] instead self.params.get("rh_system_stream_id", ""),  If it not replace, that may cause install fail.

luckyh:
@xuma request for your review, thanks.

virttest/tests/unattended_install.py
yanan-fu: @xutian  some guest do not need to do system upgrade, then no param `rh_system_stream_id` is ok,  then we replace `RH_SYSTEM_STREAM_ID`  will  ''.
In kickstart, we do system upgrade according to the value of stream id.

maxujun:
LGTM

luckyh:
Thanks all.

PR #2150, opened 2019-07-05
Fixup: Re-order cpu library functions

sathnaga:
Let's re-order and group all cpu related library
functions in same place, so that it helps user identify
easily and improves modularity.


balamuruhans:
@sathnaga patch needs rebase

sathnaga:
@balamuruhans rebased

dzhengfy:
@sathnaga, you have moved many cpu functions from utils_misc to cpu module in this PR. That is really good. Do you plan to modify tp-libvirt modules which are referring utils_misc for cpu functions? Because those reference will be definitely fail after this PR is merged.

sathnaga:
@dzhengfy I hope, i have changed all affecting tests through, this https://github.com/autotest/tp-libvirt/pull/2222, and one another minor miss also patched here, https://github.com/autotest/tp-libvirt/pull/2316, do you see issues with these PRs included aswell, if then , pls report.

dzhengfy:
@sathnaga, thanks for your quickly response and point to the tp-libvirt PRs. So PR 2222 did already fix the failure in virsh_migrate.py. But due to merge time, our job did not get the fix :). We will rerun. Thanks.

huangyum:
Hi, this patch affected some tp-qemu cases as well, I've fixed them by https://github.com/autotest/tp-qemu/pull/1912. It would be appreciated if it could get ack from us before it gets merged.  Thanks. @luckyh

luckyh:
> Hi, this patch affected some tp-qemu cases as well, I've fixed them by [autotest/tp-qemu#1912](https://github.com/autotest/tp-qemu/pull/1912). It would be appreciated if it could get ack from us before it gets merged. Thanks. @luckyh

Thanks @huangyum !

@sathnaga so, this patch does affect tp-qemu tests. I am aware that you have sent a patch (https://github.com/autotest/tp-qemu/pull/1816) to tp-qemu to follow up with the change accordingly, so firstly thanks for that. However, there is some case which is not covered, for instance, we were developing new test scripts at that time and they referenced the old APIs, and things are broken after this patch got merged as you didn't notice that.

Usually, we don't remove old APIs while introducing new ones at the same time for compatibility reason, because we don't know how many users would be affected. Once we understood about that, it is the time to clean them all.

I don't want to blame anyone, but the truth is I think we need to communicate more on this kind of issue. And I'm agree with @huangyum that it should get us ACKed before merged because we have to know it too. So please do consider us as well for the next time.

luckyh:
@sathnaga Secondly, I want to talk about this approach, there is one concern that it is not good to put every function had the pattern `.*cpu.*` of its name into one single module named `cpu.py`, why? because they are invoked in different context, for example, `get_cpu_vendor` can be called to get the cpu vendor info in host for any types of VT tests (libvirt, qemu, spice ..), but `check_xmlcount` should only be called when `vt_type == libvirt`. It doesn't make sense to me that, when I am running qemu tests, I have to indirectly import `libvirt_xml` by importing `cpu` just for querying the host cpu info.

IMO, at least there are two parts of things, one is the utility functions of cpu and it can go to, i.e.  `utils_cpu.py`, the other is the APIs wrapper for management tools/hypervisor and it should go, i.e. `libvirt_cpu.py`/`qemu_cpu.py`. The latter can import the former, not otherwise.

sathnaga:
> > Hi, this patch affected some tp-qemu cases as well, I've fixed them by [autotest/tp-qemu#1912](https://github.com/autotest/tp-qemu/pull/1912). It would be appreciated if it could get ack from us before it gets merged. Thanks. @luckyh
> 
> Thanks @huangyum !
> 
> @sathnaga so, this patch does affect tp-qemu tests. I am aware that you have sent a patch ([autotest/tp-qemu#1816](https://github.com/autotest/tp-qemu/pull/1816)) to tp-qemu to follow up with the change accordingly, so firstly thanks for that. However, there is some case which is not covered, for instance, we were developing new test scripts at that time and they referenced the old APIs, and things are broken after this patch got merged as you didn't notice that.
> 
> Usually, we don't remove old APIs while introducing new ones at the same time for compatibility reason, because we don't know how many users would be affected. Once we understood about that, it is the time to clean them all.
> 
> I don't want to blame anyone, but the truth is I think we need to communicate more on this kind of issue. And I'm agree with @huangyum that it should get us ACKed before merged because we have to know it too. So please do consider us as well for the next time.

Sure, apologies for the inconvenience caused, reason for faster merge as it touches many files, rebase by delaying will be a task., agree will take care of taking inputs from as many next time,

The reason behind consolidation is to make life easy for new users and in general we know where to look for all cpu related utils, we have plans in similar way to consolidate other components(memory, network, disk etc) methods too, one such you might have noticed for utils_disk, that PR is still open.

Suggestions welcome!

sathnaga:
> get_cpu_vendor

I understand the concern of certain non relevant(vt-type=libvirt) methods for qemu tests but
problem in that would be again we might duplicate certain methods where it can be wrapped into single, consider a possibility of  some of VT(libvirt/qemu) cases, the output would be same but way we get that output can be different for tp-qemu/tp-libvirt, but that can be handled inside the method itself, so that we have common place of implementation, which would be user, we need not force user to implement method for both vt-type, but it will allow user to see what are all methods available for other vt-type and if applicable they can modify and reuse them.

if you look our previous implementation, cpu methods were spread across multiple libraries, in some places it was not even relevant, now its all in one place and am finding it helpful(hopefully for others aswell) and  I do not see a huge burden of having all at one place, but that's my opinion though.

luckyh:
> > @sathnaga Secondly, I want to talk about this approach, there is one concern that it is not good to put every function had the pattern `.*cpu.*` of its name into one single module named `cpu.py`, why? because they are invoked in different context, for example, `get_cpu_vendor` can be called to get the cpu vendor info in host for any types of VT tests (libvirt, qemu, spice ..), but `check_xmlcount` should only be called when `vt_type == libvirt`. It doesn't make sense to me that, when I am running qemu tests, I have to indirectly import `libvirt_xml` by importing `cpu` just for querying the host cpu info.
> > IMO, at least there are two parts of things, one is the utility functions of cpu and it can go to, i.e. `utils_cpu.py`, the other is the APIs wrapper for management tools/hypervisor and it should go, i.e. `libvirt_cpu.py`/`qemu_cpu.py`. The latter can import the former, not otherwise.
> 
> I understand the concern of certain non relevant(vt-type=libvirt) methods for qemu tests but
> problem in that would be again we might duplicate certain methods where it can be wrapped into single, consider a possibility of some of VT(libvirt/qemu) cases, the output would be same but way we get that output can be different for tp-qemu/tp-libvirt, but that can be handled inside the method itself, so that we have common place of implementation, which would be user, we need not force user to implement method for both vt-type, but it will allow user to see what are all methods available for other vt-type and if applicable they can modify and reuse them.

I don't think so. Let's take the following snippet as an example, it is copied from the current `cpu.py`.
```python
def get_domstats(vm, key):
    """
    Get VM's domstats output value for given keyword
    :param vm: VM object
    :param key: keyword for which value is needed
    :return: value string
    """
    domstats_output = virsh.domstats(vm.name)
    for item in results_stdout_52lts(domstats_output).strip().split():
        if key in item:
            return item.split("=")[1]


def check_vcpu_domstats(vm, exp_vcpu):
    """
    Check the cpu values from domstats output
    :param vm: VM object
    :param exp_vcpu: dict of expected vcpus
    :return: True if exp_vcpu matches the domstats output, False if not
    """
    status = True
    cur_vcpu = int(get_domstats(vm, "vcpu.current"))
    max_vcpu = int(get_domstats(vm, "vcpu.maximum"))
    if vm.is_alive():
        exp_cur_vcpu = exp_vcpu['cur_live']
    else:
        exp_cur_vcpu = exp_vcpu['cur_config']
    if exp_cur_vcpu != cur_vcpu:
        status = False
        logging.error("Mismatch in current vcpu in domstats output, "
                      "Expected: %s Actual: %s", exp_cur_vcpu, cur_vcpu)
    if exp_vcpu['max_config'] != max_vcpu:
        status = False
        logging.error("Mismatch in maximum vcpu in domstats output, Expected:"
                      " %s Actual: %s", exp_vcpu['max_config'], max_vcpu)

    return status
```
The first issue is that `get_domstats` is not related to the cpu things, so we could consider moving it to where it should be. The reason of why it is presented here is, from my opinion, it is called by `check_vcpu_domstats`, but obviously, we can do it in a better way, for example, moving it to the `VM` class of `libvirt_vm.py` or `qemu.py` as a method and then call it at here.

Okay, even if we can let `get_domstats` be a private function to make it stay at here, there is still another issue: it doesn't handle qemu VMs. The docstring didn't point out that where it should, otherwise, how does user know that? I mean, as the module name is `cpu`, why is there something can not be used in the qemu or other context?

To solve that, the qemu part should be implemented, even with a fake one (by giving `NotImplementedError` to users). We can either separate it into two functions (i.e. `_get_domstats_virsh` and `_get_domstats_qemu`) or handle them all within the current one, and apparently you like the latter much because you think that it is more convenient by invoking. However, it also can be done by an OO approach, polymorphism is the idea for this case, as I mentioned in the above, we can implement the method both in `libvirt_vm.VM` and `qemu_vm.VM` and then call `vm.get_domstats()`, and I think it is much better. Not only `get_domstats` can be restructured like this, but also some of the rest cpu functions in the current module.

Also, there is other patterns to let users use the polymorphic functions/methods by only importing one module, whereas there are more than one module together providing the functionalities. But it is not the main point, and I will skip that.

> 
> if you look our previous implementation, cpu methods were spread across multiple libraries, in some places it was not even relevant, now its all in one place and am finding it helpful(hopefully for others aswell) and I do not see a huge burden of having all at one place, but that's my opinion though.

Well, why is single monolithic library (module) not that good? If you look the `utils_mise.py`, it is the "ultimated" one, dozens of functions are put into that huge library, and that is what we complain about, right? We suffer from lots of issues, messy namespace, patch conflicts, circular imports, which makes the code hard to be manageable. But you may say that is because `utils_mise.py` is too large and its definition is not clear, right? I do agree with that, and that is what I want to say to the current implement of `cpu.py`, it is still big! As I mentioned before, IMO, `cpu.py` should be cut into separated parts, some of the library functions can be put into `utils_cpu.py`, and some of them are not even library functions, they are the methods of objects, thus they should go where objects are defined.

sathnaga:
> > > @sathnaga Secondly, I want to talk about this approach, there is one concern that it is not good to put every function had the pattern `.*cpu.*` of its name into one single module named `cpu.py`, why? because they are invoked in different context, for example, `get_cpu_vendor` can be called to get the cpu vendor info in host for any types of VT tests (libvirt, qemu, spice ..), but `check_xmlcount` should only be called when `vt_type == libvirt`. It doesn't make sense to me that, when I am running qemu tests, I have to indirectly import `libvirt_xml` by importing `cpu` just for querying the host cpu info.
> > > IMO, at least there are two parts of things, one is the utility functions of cpu and it can go to, i.e. `utils_cpu.py`, the other is the APIs wrapper for management tools/hypervisor and it should go, i.e. `libvirt_cpu.py`/`qemu_cpu.py`. The latter can import the former, not otherwise.
> > 
> > 
> > I understand the concern of certain non relevant(vt-type=libvirt) methods for qemu tests but
> > problem in that would be again we might duplicate certain methods where it can be wrapped into single, consider a possibility of some of VT(libvirt/qemu) cases, the output would be same but way we get that output can be different for tp-qemu/tp-libvirt, but that can be handled inside the method itself, so that we have common place of implementation, which would be user, we need not force user to implement method for both vt-type, but it will allow user to see what are all methods available for other vt-type and if applicable they can modify and reuse them.
> 
> I don't think so. Let's take the following snippet as an example, it is copied from the current `cpu.py`.
> 
> ```python
> def get_domstats(vm, key):
>     """
>     Get VM's domstats output value for given keyword
>     :param vm: VM object
>     :param key: keyword for which value is needed
>     :return: value string
>     """
>     domstats_output = virsh.domstats(vm.name)
>     for item in results_stdout_52lts(domstats_output).strip().split():
>         if key in item:
>             return item.split("=")[1]
> 
> 
> def check_vcpu_domstats(vm, exp_vcpu):
>     """
>     Check the cpu values from domstats output
>     :param vm: VM object
>     :param exp_vcpu: dict of expected vcpus
>     :return: True if exp_vcpu matches the domstats output, False if not
>     """
>     status = True
>     cur_vcpu = int(get_domstats(vm, "vcpu.current"))
>     max_vcpu = int(get_domstats(vm, "vcpu.maximum"))
>     if vm.is_alive():
>         exp_cur_vcpu = exp_vcpu['cur_live']
>     else:
>         exp_cur_vcpu = exp_vcpu['cur_config']
>     if exp_cur_vcpu != cur_vcpu:
>         status = False
>         logging.error("Mismatch in current vcpu in domstats output, "
>                       "Expected: %s Actual: %s", exp_cur_vcpu, cur_vcpu)
>     if exp_vcpu['max_config'] != max_vcpu:
>         status = False
>         logging.error("Mismatch in maximum vcpu in domstats output, Expected:"
>                       " %s Actual: %s", exp_vcpu['max_config'], max_vcpu)
> 
>     return status
> ```
> 
> The first issue is that `get_domstats` is not related to the cpu things, so we could consider moving it to where it should be. The reason of why it is presented here is, from my opinion, it is called by `check_vcpu_domstats`, but obviously, we can do it in a better way, for example, moving it to the `VM` class of `libvirt_vm.py` or `qemu.py` as a method and then call it at here.
> 
> Okay, even if we can let `get_domstats` be a private function to make it stay at here, there is still another issue: it doesn't handle qemu VMs. The docstring didn't point out that where it should, otherwise, how does user know that? I mean, as the module name is `cpu`, why is there something can not be used in the qemu or other context?
> 
> To solve that, the qemu part should be implemented, even with a fake one (by giving `NotImplementedError` to users). We can either separate it into two functions (i.e. `_get_domstats_virsh` and `_get_domstats_qemu`) or handle them all within the current one, and apparently you like the latter much because you think that it is more convenient by invoking. However, it also can be done by an OO approach, polymorphism is the idea for this case, as I mentioned in the above, we can implement the method both in `libvirt_vm.VM` and `qemu_vm.VM` and then call `vm.get_domstats()`, and I think it is much better. Not only `get_domstats` can be restructured like this, but also some of the rest cpu functions in the current module.
> 
> Also, there is other patterns to let users use the polymorphic functions/methods by only importing one module, whereas there are more than one module together providing the functionalities. But it is not the main point, and I will skip that.
> 
> > if you look our previous implementation, cpu methods were spread across multiple libraries, in some places it was not even relevant, now its all in one place and am finding it helpful(hopefully for others aswell) and I do not see a huge burden of having all at one place, but that's my opinion though.
> 
> Well, why is single monolithic library (module) not that good? If you look the `utils_mise.py`, it is the "ultimated" one, dozens of functions are put into that huge library, and that is what we complain about, right? We suffer from lots of issues, messy namespace, patch conflicts, circular imports, which makes the code hard to be manageable. But you may say that is because `utils_mise.py` is too large and its definition is not clear, right? I do agree with that, and that is what I want to say to the current implement of `cpu.py`, it is still big! As I mentioned before, IMO, `cpu.py` should be cut into separated parts, some of the library functions can be put into `utils_cpu.py`, and some of them are not even library functions, they are the methods of objects, thus they should go where objects are defined.

Sure, Get your point, that's why I had moved all cpu related functions to a single place, but as you mentioned it can be handle better and still some unwanted/unrelevant methods present, let's do one thing, will create an issue and let's discuss on methods which can move across different places and let's agree to it, then we will address it. sounds fine?

https://github.com/avocado-framework/avocado-vt/issues/2305

luckyh:
@sathnaga ok, thanks. Let's discuss the thing at there, I'll add my comments later.

PR #2153, opened 2019-07-10
Enhance check_vm_state() to check reason for the state

Yingshun:
Need to check reason for domain state in some migration cases.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1275,7 +1275,7 @@ class MigrationTest(object):
         vm_state = params.get("virsh_migrated_state", "running")
         ping_count = int(params.get("ping_count", 10))
         for vm in vms:
-            if not self.check_vm_state(vm.name, vm_state, uri):
+            if not self.check_vm_state(vm.name, vm_state, uri=uri):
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1594,21 +1594,27 @@ def cleanup_dest_vm(self, vm, srcuri, desturi):
         # Set connect uri back to local uri
         vm.connect_uri = srcuri
 
-    def check_vm_state(self, vm_name, state='paused', uri=None):
+    def check_vm_state(self, vm_name, state='paused', reason=None, uri=None):
```
balamuruhans: This would break virsh_migate.py and virsh_migration.py,

It is better to always add new arguments at last for better backward compatibility or to change all the users accordingly.

Yingshun: @balamuruhans  only virsh_cmd/domain/virsh_migrate_stress.py is using this method on latest code. Updated in tp-libvirt 2225, please help review.

balamuruhans: Thank you @Yingshun, for taking care of it.

PR #2154, opened 2019-07-11
update argument for check_vm_state in libvirt

Yingshun:
This PR corrects #2153 which missed one update.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1631,7 +1631,7 @@ class MigrationTest(object):
         """
         def check_state():
             try:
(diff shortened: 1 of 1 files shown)


Yingshun:
Updated check_vm_state in avocado-vt.
```
[root@host193 virttest]# grep check_vm_state -r *
utils_test/libvirt.py:            if not self.check_vm_state(vm.name, vm_state, uri=uri):
utils_test/libvirt.py:    def check_vm_state(self, vm_name, state='paused', reason=None, uri=None):
utils_test/libvirt.py:                return self.check_vm_state(dest_vm_name, state, uri=uri)
[root@host193 virttest]# 
```

PR #2155, opened 2019-07-15
set hugepage param in respective config section

sathnaga:
Enable hugepage param in respective hugepages
config section so that it get's honored by filter
while using `only hugepages` for libvirt tests.

### DIFF
--- a/shared/cfg/guest-hw.cfg
+++ b/shared/cfg/guest-hw.cfg
@@ -196,6 +196,7 @@ variants:
 variants:
     - @smallpages:
     - hugepages:
+        hugepage = yes
         setup_hugepages = yes
         # In some special conditions, such as low mem PPC systems, you might
         # want to define how many large memory pages you want to set directly.


shared/cfg/guest-hw.cfg
```diff
@@ -196,6 +196,7 @@ variants:
 variants:
     - @smallpages:
     - hugepages:
+        hugepage = yes
```
luckyh: Hi @sathnaga would you please explain why we need this configuration? seems that the param just sets `setup_hugepages = yes` where we have it at the next line.

https://github.com/avocado-framework/avocado-vt/blob/54a2d9a96850ff9ce490e4f4e89d8e1ecc657a6f/virttest/env_process.py#L986-L987

sathnaga: it is getting used in libvirt_vm and decide to use hugepage during `virt-install` 
https://github.com/avocado-framework/avocado-vt/blob/master/virttest/libvirt_vm.py#L1013

sathnaga: I see both are redundant, will check and address it different PR with validating other instances of usages in tests etc..

luckyh: ok, thanks. Then I have a question to this, I see that, in `libvirt_vm.py`, it was enabled by 25ac7d7d5af84c76adeaa2d87532ab16a69c76c8, do you know why `hugepage` was chosen at that time? (maybe @balamuruhans knows that much). I have no idea about the difference between this two params, but I'm wondering if we can replace `hugepage` to `set_hugepages`.

sathnaga: @luckyh looks like it might be by mistake, just checked, not many instances in tests, so will update this PR itself to change it to common same param, lets choose `setup_hugepages` which has been used in tests.

balamuruhans: @luckyh @sathnaga There are 2 things here,
1. `setup_hugepages` is used to allocate and cleanup hugepages in host
2. `hugepages` is used to create virt-install commandline to include hugepage attribute for VM.

In this line we are allocating hugepages in host because VM is going to be booted with hugepage attribute that is set by `hugepage`
https://github.com/avocado-framework/avocado-vt/blob/54a2d9a96850ff9ce490e4f4e89d8e1ecc657a6f/virttest/env_process.py#L986-L987

It is not redundant.

sathnaga: param `hugepage`  which is to be used in virt-install has the dependency of allocating hugepages in host which needs to set `setup_hugepages` param, instead we can use `setup_hugepages` itself fo virt-install hugepage attribute aswell..what say?

balamuruhans: param usage is completely different, so I would prefer to have separate params, if we make it same then we could introduce unwanted complexity in below ways

- In MultiVM environment, VMs can have combination of with/without hugepages and combination of default/non default hugepage size where it is easy to handle with separate params.

- Negative testing cannot be done.

luckyh: @balamuruhans thanks for your explanation, so we have

* the `setup_hugepages` param to specify whether do hugepages configurations on host
* the `hugepage` param (please let me call it `vm_hugepage` since it is more appropriate) to specify whether the given vm consumes hugepages

this totally makes sense to me, but I still have two questions:

1. @sathnaga would you please tell us what the perspective is of this PR? to set `hugepage` for all vms or just the main vm?
2. @balamuruhans then e369092294d0bf31cc51f9a7933f305ab9242985 seems to be defective because `vm_hugepage` is a per-vm param, it will go wrong if we only set the vm local scope param (e.g. `vm_hugepage_vm1`) without the global scope param `vm_hugepage`, do you plan to fix that?

Thanks.

balamuruhans: > @balamuruhans then e369092 seems to be defective because vm_hugepage is a per-vm param, it will go wrong if we only set the vm local scope param (e.g. vm_hugepage_vm1) without the global scope param vm_hugepage, do you plan to fix that?

@luckyh right, I need work on it to fix. I didn't have considered multivm scenarios during https://github.com/avocado-framework/avocado-vt/commit/e369092294d0bf31cc51f9a7933f305ab9242985 timeframe

sathnaga: @balamuruhans make sense, it's better to keep it separate for multi vm scenarios

@luckyh `hugepage` is already per-vm and can be used with multivm scenarios, but i guess not the allocation part, we might need to handle it,

This PR would help us to achieve the missing support for libvirt tests while using `only hugepages` in test config, it will be for all VMs.

luckyh: @sathnaga got it, thanks.

luckyh: @balamuruhans sure, thanks.

PR #2162, opened 2019-07-18
Increase the flexibility of parameter type converting methods

pevogam:
We allow the user more flexibility when it comes to white spaces
and overall Cartesian configuration formatting.


pevogam:
This PR is a small follow up of #1914.

PR #2164, opened 2019-07-22
Add new function get_controller()

fangge1212:
It's used to return controllers list in domain xml according to controller
type and model.

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -548,6 +548,20 @@ class VMXMLBase(base.LibvirtXMLBase):
             devices_element.append(element.getroot())
         self.xmltreefile.write()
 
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -548,6 +548,20 @@ def set_controller(self, controller_list):
             devices_element.append(element.getroot())
         self.xmltreefile.write()
 
+    def get_controller(self, controller_type=None, model=None):
```
balamuruhans: we can modify existing `get_controllers()` method in the same file, rather than a new method

fangge1212: Oh, I didn't notice that method before. I will check, thanks very much.

fangge1212:
I created a new pr: https://github.com/avocado-framework/avocado-vt/pull/2184

PR #2170, opened 2019-07-25
libvirt_xml/vm_xml: Add --nvram flag for guests with nvram

jallen-amd:
Guests with nvram will fail with "ERROR: Failed to undefine " since
the --nvram flag is only added for the aarch platform. Instead, add the
--nvram for any guest with nvram.

Signed-off-by: John Allen john.allen@amd.com


jallen-amd:
Ah, text editor is not setup to use spaces over tabs. Resending...

PR #2172, opened 2019-07-26
libvirt.py: refactor the "check_vm_state".

santwana:
libvirt.py: refactor the "check_vm_state".

The function "check_vm_state" is generic for non-migration scenario's as well.
Placing it out of the MigrationTest class, would be useful to refer in other tests.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -863,6 +863,29 @@ def check_actived_pool(pool_name):
     return True
 
 
+def check_vm_state(vm_name, state='paused', reason=None, uri=None):
+    """
+    checks whether state of the vm is as expected
+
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -953,6 +953,29 @@ def check_actived_pool(pool_name):
     return True
 
 
+def check_vm_state(self, vm_name, state='paused', reason=None, uri=None):
```
balamuruhans: please remove `self` as it is now not a method associated with class.

santwana: @balamuruhans Addressed the changes, please review.

balamuruhans:
@santwana Thanks for the patch.

PR #2173, opened 2019-07-29
Add libvirt_remote.py to manage libvirt on remote

Yingshun:
This is required by migration cases.
Need to update libvirt related configuration files on remote.

### DIFF
--- a/virttest/libvirt_remote.py
+++ b/virttest/libvirt_remote.py
@@ -1,0 +1,46 @@
+"""
+Shared code for tests that need to control libvirt on remote
+"""
+
+import logging
+
+from avocado.core import exceptions
+
+from virttest import remote
+
+from virttest.utils_test import libvirt
+
+
+def update_remote_file(params, value,
+                       file_path='/etc/libvirt/libvirtd.conf',
+                       restart_libvirt=True):
+    """
+    Update file on remote and restart libvirtd if needed
+
+    :param param: Test run params
+    :param value: The value to update
+    :param file_path: The path of the file
+    :param restart_libvirt: True to restart libvirtd
+    :return: remote.RemoteFile object
+
+    """
+    try:
+        tmp_value = eval(value)
+        logging.debug("Update file {} with: {}".format(file_path, value))
+        remote_ip = params.get("server_ip", params.get("remote_ip"))
+        remote_pwd = params.get("server_pwd", params.get("remote_pwd"))
+        remote_user = params.get("server_user", params.get("remote_user"))
+
+        remote_file_obj = remote.RemoteFile(address=remote_ip, client='scp',
+                                            username=remote_user,
+                                            password=remote_pwd,
+                                            port='22', remote_path=file_path)
+        remote_file_obj.sub_else_add(tmp_value)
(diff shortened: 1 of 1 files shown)


virttest/libvirt_remote.py
```diff
@@ -0,0 +1,42 @@
...
+import logging
+
+from avocado.core import exceptions
+
+from virttest import remote
+
+from virttest.utils_test import libvirt
+
+
+def update_remote_config(params, new_key,
```
dzhengfy: It seems this function can apply to any file on remote host, instead of configuration file only. So could you help update it for common file?

Yingshun: @dzhengfy There is an action to restart service at end of this func, so I suggest only for configuration file for now.
Or there is any other files related to libvirt but not config files? Please let me know if I'm wrong.

dzhengfy: @Yingshun , I just think it may helpful in case some scenarios need update a remote file. For libvirtd restarting, we can add a new parameter to indicate if restarting is needed. What do you think of?

Yingshun: @dzhengfy Okay, will update

virttest/libvirt_remote.py
```diff
@@ -0,0 +1,46 @@
...
+
+from avocado.core import exceptions
+
+from virttest import remote
+
+from virttest.utils_test import libvirt
+
+
+def update_remote_file(params, new_key,
+                       config_path='/etc/libvirt/libvirtd.conf',
```
dzhengfy: Change 'config_path' to other common name.

Yingshun: @dzhengfy modified to 'file_path'

virttest/libvirt_remote.py
```diff
@@ -0,0 +1,46 @@
...
+import logging
+
+from avocado.core import exceptions
+
+from virttest import remote
+
+from virttest.utils_test import libvirt
+
+
+def update_remote_file(params, new_key,
```
chunfuwen: I assume update key with value, so the input is not just new_key, but need value as well

virttest/libvirt_remote.py
```diff
@@ -0,0 +1,46 @@
...
+        remote_file_obj = remote.RemoteFile(address=remote_ip, client='scp',
+                                            username=remote_user,
+                                            password=remote_pwd,
+                                            port='22', remote_path=file_path)
+        remote_file_obj.sub_else_add(tmp_key)
+        if restart_libvirt:
+            libvirt.remotely_control_libvirtd(remote_ip, remote_user,
+                                              remote_pwd, action='restart',
+                                              status_error='no')
+        return remote_file_obj
```
chunfuwen: I assume there is no much meaning to return remote_file_obj

Yingshun: @chunfuwen  When RemoteFile object deleted, the updated file was recovered. So I return it here.

virttest/libvirt_remote.py
Yingshun: updated.

PR #2176, opened 2019-07-30
Add missing parameters for 'blkiotune' virsh cmd

lento-sun:

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -4209,10 +4209,18 @@ def vol_upload(name, dfile, options="", **dargs):
     return command(cmd, **dargs)
 
 
-def blkiotune(name, weight=None, device_weights=None, options=None, **dargs):
+def blkiotune(name, weight=None, device_weights=None, device_read_iops_sec=None,
+              device_write_iops_sec=None, device_read_bytes_sec=None,
+              device_write_bytes_sec=None, options=None, **dargs):
     """
     Set or get a domain's blkio parameters
     :param name: name of domain
+    :param weight: overall blkio weight
+    :param device_weights: blkio weight for specific dev
+    :param device_read_iops_sec: read iops for specific dev
+    :param device_write_iops_sec: write iops for specific dev
+    :param device_read_bytes_sec: read bytes per sec for specific dev
+    :param device_write_bytes_sec: write bytes per sec for specific dev
     :param options: options may be live, config and current
     :param dargs: standardized virsh function API keywords
     :return: CmdResult instance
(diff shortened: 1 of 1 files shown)


lento-sun:
related cases also fixed:
https://github.com/autotest/tp-libvirt/pull/2257

virttest/virsh.py
```diff
@@ -4209,7 +4209,9 @@ def vol_upload(name, dfile, options="", **dargs):
     return command(cmd, **dargs)
 
 
-def blkiotune(name, weight=None, device_weights=None, options=None, **dargs):
+def blkiotune(name, weight=None, device_weights=None, device_read_iops_sec=None,
```
Yingshun: please update docstring for the parameters.

virttest/virsh.py
```diff
@@ -4224,6 +4226,14 @@ def blkiotune(name, weight=None, device_weights=None, options=None, **dargs):
         cmd += " --weight %s" % weight
     if device_weights:
         cmd += " --device-weights %s" % device_weights
+    if device_read_iops_sec:
+        cmd += " --device-read-iops-sec %s" % device_read_iops_sec
+    if device_write_iops_sec:
+        cmd += " --device-write-iops-sec %s" % device_write_iops_sec
+    if device_read_bytes_sec:
+        cmd += " --device-read-bytes-se %s" % device_read_bytes_sec
```
Yingshun: typo. s/device-read-bytes-se/device-read-bytes-sec/

virttest/virsh.py
lento-sun: added, thx

virttest/virsh.py
lento-sun: fixed, thx

virttest/virsh.py
Yingshun: @lento-sun, your updates are about params in 'blkdeviotune' but not in 'blkiotune'.

virttest/virsh.py
```diff
@@ -4236,6 +4246,12 @@ def blkdeviotune(name, device=None, options=None,
     Set or get a domain's blkio parameters
```
Yingshun: BTW, I think the description of blkdeviotune() is inappropriate, would you please update it? Thanks!

virttest/virsh.py
lento-sun: oops~
updated. The blkdeviotune() has much less available params than virsh api does, need to be updated later. not covered here in cgroup related fix.

virttest/virsh.py
lento-sun: the whole blkdeviotune() is not containing enough available virsh api params, will be updated in another pr.

Yingshun: It makes sense. LGTM.

virttest/virsh.py
```diff
@@ -4209,10 +4209,18 @@ def vol_upload(name, dfile, options="", **dargs):
     return command(cmd, **dargs)
 
 
-def blkiotune(name, weight=None, device_weights=None, options=None, **dargs):
+def blkiotune(name, weight=None, device_weights=None, device_read_iops_sec=None,
```
dzhengfy, line 4212: I am not sure if we will have other more parameters need to be added in this function. It seems too long. Maybe a param_dict is more simple. But however, it is OK to me so far.

PR #2188, opened 2019-08-06
fix TypeError when create serial device

xutian:
ID: 1737744

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -2695,8 +2695,9 @@ class VM(virt_vm.BaseVM):
 
             # Create serial ports.
             for serial in params.objects('serials'):
-                if not params.object_params(serial).get(
-                        "serial_type").startswith('virt'):
+                serial_params = params.object_params(serial)
+                serial_type = serial_params["serial_type"]
+                if not serial_type.startswith('virt'):
                     self.serial_ports.append(serial)
 
             if (name is None and params is None and root_dir is None and


virttest/qemu_vm.py
```diff
@@ -1567,7 +1567,7 @@ def sort_key(dev):
             devices.insert(serial_devices)
 
             # Create virtio_ports (virtserialport and virtconsole)
-            serial_type = serial_params['serial_type']
+            serial_type = serial_params.get('serial_type', '')
```
vivianQizhu: From my view `serial_type` must be defined when there is `serials` defined, do you have any use case that allow it is empty?

xutian: if 'serial_type' is mandatory  param , line#2699 should be "if not params.object_params(serial)['serial_type']", base on your origin code, I guess serial_type is not a mandatory param

vivianQizhu: @xutian Em, my fault, it should be mandatory, maybe you could help correct that:), thanks.

yanan-fu:
The issue listed in 1737744 is `'NoneType' object has no attribute 'startwith'`,  i think this patch can not fix it,  with the new change, it should failed as `KeyError`.
Any other patch need work together with it ?

xutian:
@yanan-fu 

"serial_type = None" because default value for  'get()' function of  dict is None. now, update code to serial_params['serial_type'] will raise  missing options exception before create serial device.

Thanks,
Xu

xutian:
@yanan-fu  and @vivianQizhu  any more comments, if no let  me merge this small fix.

PR #2206, opened 2019-08-19
Add funcs for gluster backend disk oprations

Yingshun:
This adds some functions to manage the glusterfs backend disk

1. Add functions to mount/umount/check status of glusterfs
2. Add functions to create/remove/check symlink
3. Move setup_or_cleanup_gluster from libvirt.py to gluster.py
and move get_host_ipv4_addr to utils_net due to cyclic import

This PR is required by migration cases which need to migrate
with gluster backend disk.

### DIFF
--- a/virttest/ppm_utils.py
+++ b/virttest/ppm_utils.py
@@ -26,6 +26,12 @@ except ImportError:
                     'and Windows guest BSOD detection are disabled. '
                     'In order to enable it, please install python-imaging or '
                     'the equivalent for your distro.')
+# Prevent logs pollution
+if Image is not None:
+    for _logger_name in logging.root.manager.loggerDict:
+        if _logger_name.split('.', 1)[0] == 'PIL':
+            _pil_logger = logging.getLogger(_logger_name)
+            _pil_logger.setLevel(logging.CRITICAL)
 
 
 try:


virttest/utils_net.py
```diff
@@ -4145,3 +4145,21 @@ def get_netkvm_param_value(vm, param):
         return value
     finally:
         session.close()
+
+
+def get_host_ipv4_addr():
```
xutian: Host maybe have multi nics,  I suggest to get IP address by  nic's MAC or interface name.  you can return default route device's IP address by default.

e,g. 

`
import netifaces as ni

def get_host_ip(interface, ip_version=ni.AF_INET)
      if interface is None:
          interface = ni.gateways()['default'][ip_version][1]
      iface = ni.ifaddresses(interface)
      return iface[ip_version][0]['addr']
`

PaulYuuu: https://github.com/avocado-framework/avocado-vt/blob/5c83fe5eafe748f771adaa406879c02d24c90faf/virttest/utils_net.py#L3388-L3407

Yingshun: @PaulYuuu  Thanks for your reminder, but get_host_ip_address does not meet our requirement for now. There is a param netdst(virbr0) but no hot_ip_addr, in this case, it will pick up IP address on virbr0.

Yingshun: @xutian， Thanks for your review, updated.

PaulYuuu:
```python
def get_host_ip_address(params=None, ifname=None, ip_ver="ipv4", linklocal=False):
    if params:
        host_ip = params.get('host_ip_addr', None)
        if host_ip:
            logging.debug("Use IP address at config %s=%s", 'host_ip_addr', host_ip)
            return host_ip
        net_dev = params.get("netdst")
    else:
        net_dev = ifname
    if not net_dev:
        net_dev = get_host_default_gateway(iface_name=True)
```
I think this change can help you, and it is compatible with the previous way.

xutian: looks OK to me.

Yingshun: @PaulYuuu  Thanks for the so detailed resolution, updated.

PaulYuuu: It's nothing. `utils_net.py` looks good to me, other code needs someone help to review.

PR #2207, opened 2019-08-19
Changes to support client-server and standalone stress tool runs

bssrikanth:
This commit has few changes[1,2] required to support running
stress workloads in server-client mode as well as standalone mode
in parallel. Change [3] is a minor fix for handling stress unload.
1. Rename param which is used to specify custom server-client pairs.
2. Now users can run stress tools which are client-server based as well
as standalone stress tools together. So place existing check under condition
when user chooses only 'client_server' type of stress load.
3. In case of stopping stress app, in some cases have seen delays to stop
stress app, giving couple of seconds sleep after stop_cmd fixes the issue

new param which can be used in testcases to configure stress run mode
is documented in base.cfg.


bssrikanth:
Test case would be posted to tp-libvirt.
Below are results: 
```
avocado run multivm_loadstress --vt-type libvirt --vt-extra-params vms="vm1 vm2 vm3" main_vm=vm1 master_images_clone="ubuntu-18.04-ppc64le" create_vm_libvirt=yes kill_vm_libvirt=yes os_variant=centos7.0 env_cleanup=yes backup_image_before_testing=no restore_image_after_testing=no mem=8096 close_fds=True vcpu_cores=4 smp=4 --vt-guest-os RHEL.7.6alt.ppc64le

JOB ID     : 3b14e850e902735694e0547d33ae3d55cf3a6c40
JOB LOG    : /home/srikanth/avocado-fvt-wrapper/results/job-2019-08-19T07.23-3b14e85/job.log
 (1/1) type_specific.powerkvm-libvirt.multivm_loadstress: /nohup: redirecting stderr to stdout
PASS (301.55 s)
JOB TIME   : 306.88 s
JOB HTML   : /home/srikanth/avocado-fvt-wrapper/results/job-2019-08-19T07.23-3b14e85/results.html
```
Where vm1 and vm2 are running uperf workload [server and client], vm3 is running stress-ng workload which is standalone.

shared/cfg/base.cfg
```diff
@@ -679,6 +679,10 @@ libvirtd_log_cleanup = "yes"
 #uperf_protocol = "tcp"
 #Client profile to run for uperf tests
 #client_profile_uperf = "shared/deps/uperf/iperf.xml"
+#stress_config can have below possible values:
+# 1> client_server: stress workloads which requires client server setup: eg:uperf
+# 2> standalone: stress workloads which are standalone. eg: stress-ng
+# 3> mix: running both client server workload and standalone stress workload in parallel
```
balamuruhans: :+1:

virttest/utils_test/__init__.py
```diff
@@ -2151,6 +2151,7 @@ def unload_stress(self):
         """
         def _unload_stress():
             self.sendline(self.stop_cmd)
+            time.sleep(2)
```
balamuruhans: why do we need this ?

bssrikanth: As mentioned in PR message `In case of stopping stress app, in some cases have seen delays to stop stress app, giving couple of seconds sleep after stop_cmd fixes the issue`.
Even I have tried passing higher delay for first value in wait_for function below.. strangely it as well does not resolve the issue. The test use to wait for 2 secs after issuing stop command.. then find stress running still.. would attempt to sleep for couple of seconds and try executing func again.. but this time it would find stress not running... and test would fail with this reason.. when I tried sleep after stop_cmd I could see desired result all the time.. do you have any better solution here?

balamuruhans: I got you, but this is unnecessary because we do exactly the same thing with current flow,
in unload_stress() we use `first=2.0` as an argument to `wait_for()` that does this., 
https://github.com/avocado-framework/avocado-vt/blob/8372643276fcc6d7b2bd3fabc16f540377ce21d9/virttest/utils_test/__init__.py#L2161-L2162


with the below change I think it should work fine,

```patch
diff --git a/virttest/utils_test/__init__.py b/virttest/utils_test/__init__.py
index c8f00fa0..eb03a894 100755
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -2153,10 +2153,11 @@ class Stress(object):
         stop stress tool manually
         """
         def _unload_stress():
-            self.sendline(self.stop_cmd)
-            if not self.app_running():
+            try:
+                self.sendline(self.stop_cmd)
+            except (process.CmdError, aexpect.ShellError):
                 return True
-            return False
+            return not self.app_running():
 
```

bssrikanth: i had mentioned about same in my previous reply... i tried same...  strangely it as well does not resolve the issue.

bssrikanth: I see proposed change introduce try catch, flags if cmderror. I am not sure if this proposed change would resolve issue as I had tried increasing `first` to 4 secs while calling wait_for . But would try same and get back

bssrikanth: @balamuruhans my bad. I overlooked and had interchanged first and step delays .. original step delay of 1sec seems not enough for stopping stress app on host. I am bumping it to 2 secs. This should remove the sleep call which I had introduced. Testing this change now.. once done will push same.

Issue #2209, opened 2019-08-20
mac generator issue


bssrikanth:
Looks like newly integrated mac generator has a regression. I hit below scenario today.. noticing two issues:
1. generated mac is only 40 bits/5 octets instead of 6.
2. after discarding invalid mac, we seem to be generated newer mac by adding 1 to last octet of older mac without checking if old mac was wrong/right.

I was looking into this for a while now, thought of opening an issue here to get attention/advice from @luckyh 

```
2019-08-19 23:45:40,181 virt_vm          L0920 DEBUG| Generating random mac address for nic
2019-08-19 23:45:40,194 libvirt_vm       L1978 DEBUG| VM.create activating nic {'netdst': 'virbr0', 'ip': None, 'nic_name': 'nic1', 'mac': '52:54:00:be:e1', 'nettype': 'bridge', 'nic_model': 'virtio', 'g_nic_name': None}
2019-08-19 23:45:40,237 utils_net        L2682 WARNI| Discarded invalid mac "52:54:00:be:e1" for nic "nic1" from input, 10 warnings remaining.
2019-08-19 23:45:40,244 libvirt_vm       L0205 INFO | Libvirt VM 'vm1', driver 'qemu', uri 'qemu:///system'
2019-08-19 23:45:41,376 libvirt_vm       L0978 DEBUG| Machine types supported for hvm/ppc64le: ['pseries-rhel7.6.0', 'pseries', 'pseries-rhel7.6.0-sxxm', 'pseries-rhel7.3.0-sxxm', 'pseries-rhel7.5.0', 'pseries-rhel7.5.0-sxxm', 'pseries-rhel7.2.0', 'pseries-rhel7.4.0-sxxm', 'pseries-rhel7.3.0', 'pseries-rhel7.4.0']
2019-08-19 23:45:41,378 libvirt_vm       L0897 DEBUG| vm.add_memballoon returning:  --memballoon model=virtio
2019-08-19 23:45:41,379 virt_vm          L0920 DEBUG| Generating random mac address for nic
2019-08-19 23:45:41,410 libvirt_vm       L1299 DEBUG| make_create_command() setting up command for nic: {'netdst': 'virbr0', 'ip': None, 'nic_name': 'nic1', 'mac': '52:54:00:be:e2', 'nettype': 'bridge', 'nic_model': 'virtio', 'g_nic_name': None}
[...]

PR #2210, opened 2019-08-20
utils_net: Fix the mac generator issue

luckyh:
Got a wrong mac addr `52:54:00:be:e1` by invoking `BaseVM.add_nic`,
that is because there is `0x00` in the high order octets of the addr
and the `_int_to_int_list` method would give a wrong result to that.
Making this patch to fix the issue.

Fix: https://github.com/avocado-framework/avocado-vt/issues/2209

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -2526,7 +2526,7 @@ class VirtIface(propcan.PropCan, object):
         return mac_bytes
 
     @staticmethod
-    def _int_to_int_list(number):
+    def _int_to_int_list(number, align=0):
         """
         Convert integer to integer list split by byte.
         """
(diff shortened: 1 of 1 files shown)


luckyh:
@bssrikanth request for your review, thanks!

luckyh:
```python
def int_to_int_list(number, align=0):
    out = []
    while number > 0:
        out.insert(0, number & 0xff)
        number >>= 8
    if not out:
        out.append(0)
    blen = len(out)
    if align > blen:
        out = ([0] * (align - blen)) + out
    return out

print(int_to_int_list(0x001234567800))
print(int_to_int_list(0x001234567800, 6))
```
```
$ python3 test.py
[18, 52, 86, 120, 0]
[0, 18, 52, 86, 120, 0]
```

balamuruhans:
Thanks @luckyh @bssrikanth.

PR #2227, opened 2019-08-29
utils_net: Remove unhelpful log message

luckyh:
When running tests with windows guest, the following message would
flood the main log stream.

  Retry verifying IP address (0.000007 secs)

However, it isn't helpful since it only tells user what it will do
rather than give the detail of errors. Let's remove it then.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3313,8 +3313,7 @@ def verify_ip_address_ownership(ip, macs, timeout=60.0, devs=None,
         return bool(utils_misc.wait_for(lambda: verify_func(ip, macs, dev,
                                                             timeout,
                                                             session=session, **dargs),
-                                        timeout,
-                                        text="Retry verifying IP address"))
+                                        timeout))
 
 
 def generate_mac_address_simple():


luckyh:
a log snippet:
```
13:12:16 DEBUG| Retry verifying IP address (0.000007 secs)
13:12:16 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:16 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:17 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:17 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:17 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:17 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:17 DEBUG| Retry verifying IP address (0.000014 secs)
13:12:18 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:18 DEBUG| Retry verifying IP address (0.000014 secs)
13:12:18 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:18 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:18 DEBUG| Retry verifying IP address (0.000016 secs)
13:12:18 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:19 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:19 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:19 DEBUG| Retry verifying IP address (0.000013 secs)
13:12:19 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:19 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:20 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:20 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:20 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:20 DEBUG| Retry verifying IP address (0.000015 secs)
13:12:20 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:20 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:21 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:21 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:21 DEBUG| Retry verifying IP address (0.000014 secs)
13:12:21 DEBUG| STREAM b'IHDR' 16 13
13:12:21 DEBUG| STREAM b'IDAT' 41 1216
13:12:21 DEBUG| Retry verifying IP address (0.000009 secs)
13:12:21 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:22 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:22 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:22 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:22 DEBUG| Retry verifying IP address (0.000015 secs)
13:12:22 DEBUG| Retry verifying IP address (0.000013 secs)
13:12:22 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:23 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:23 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:23 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:23 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:23 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:23 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:24 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:24 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:24 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:24 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:24 DEBUG| Retry verifying IP address (0.000012 secs)
13:12:25 DEBUG| Retry verifying IP address (0.000011 secs)
13:12:25 DEBUG| Retry verifying IP address (0.000012 secs)
```

luckyh:
@balamuruhans request for your review, thanks!

vivianQizhu:
Do we consider also reduce its 'step'? it seems too frequently.

balamuruhans:
@luckyh yes, as per @vivianQizhu's suggestion if we increase the step time it would reduce the flooding of log message, but I feel log message would really help us during debugging.

luckyh:
> Do we consider also reduce its 'step'? it seems too frequently.

It doesn't work as expected since it never executed `time.sleep(step)` in this case.

virttest/utils_net.py
```diff
@@ -3313,8 +3313,7 @@ def __verify_neigh(ip, macs, dev, timeout, session=None, **dargs):
         return bool(utils_misc.wait_for(lambda: verify_func(ip, macs, dev,
                                                             timeout,
                                                             session=session, **dargs),
-                                        timeout,
-                                        text="Retry verifying IP address"))
```
balamuruhans: can we add a single log before calling `wait_for()` ?

vivianQizhu:
> > Do we consider also reduce its 'step'? it seems too frequently.
> 
> @vivianQizhu It doesn't work as expected since it never executed `time.sleep(step)` in this case. 

Right, that explains why it retries so often when the default step is 1.0. Anyway current code change is okay to me, cause we will know it fails at the end, do not need to know when it keeps trying in the process most time.

luckyh:
> @luckyh yes, as per @vivianQizhu's suggestion if we increase the step time it would reduce the flooding of log message, but I feel log message would really help us during debugging.

@balamuruhans @vivianQizhu let me clarify more, by the definition of `utils_misc.wait_for`, the `text` would be logged every time before `func` being executed, so the log message should be `Try verifying ..` instead of `Retry verifying ..`
```python
    while time.time() < end_time:
        if text:
            logging.debug("%s (%f secs)", text, (time.time() - start_time))

        output = func()
        if output:
            return output

        time.sleep(step)

    return None
```
And in my case, the `output` is `True`, so it never executed `time.sleep(step)`. That looks like something needed to be improved, but, sorry, I haven't had a further investigation.

Also, in my opinion we should logged the messages when errors occurred because that is what we are interested in, I'd suggest making it to be:
```python
"Failed to verify IP address '%s' for the given MAC '%s' due to %s, retrying" % (mac, ip_addr, reason)
```
but I have to leave the improvement to the future since I don't have enough time at this time, sorry.

virttest/utils_net.py
luckyh: Do you have an example?

balamuruhans: @luckyh, I am okay with this change. I was mentioning something like below that helps us during debugging,
```patch
diff --git a/virttest/utils_net.py b/virttest/utils_net.py
index a743bdd2..bf7b7120 100644
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3310,11 +3310,11 @@ def verify_ip_address_ownership(ip, macs, timeout=60.0, devs=None,
     verify_func = __verify_neigh if ip_ver == 6 else __arping
     for dev in devs:
         # VM might take some time to respond after migration
+        logging.debug("verify IP address ownership for interface: %s", dev)
         return bool(utils_misc.wait_for(lambda: verify_func(ip, macs, dev,
                                                             timeout,
                                                             session=session, **dargs),
-                                        timeout,
-                                        text="Retry verifying IP address"))
+                                        timeout))
```

vivianQizhu: @luckyh Do you want adopt the above suggestion and remove this log this time before we have better improvement? I am testing Windows a lot recently and the message is truly annoying...

luckyh: @balamuruhans @vivianQizhu honestly, I think it is not necessary (as I already explained below), so I don't intent to add it this time, thanks.

balamuruhans: @luckyh I faced difficulty to debug when I worked for https://github.com/avocado-framework/avocado-vt/commit/e1d50763511aa9848c6714adf66546102b6349bb, it is not harm a single line that won't flood the log. If you still not willing, please feel free to merge it. Thanks!

luckyh: @balamuruhans in the case of calling `verify_ip_address_ownership` in the `virt_vm.get_address` context, there is already a log, `"Dropped the address cache of HWADDR ($MAC)"`, being raised when failed to verify the ownership, otherwise, callers should make the log and explain the reason. To me, as I can't see a significant help for debugging by adding such the above log, I'd like to leave it to the future and add it when we realize that we have to do that. Anyway, thanks for your understanding!

luckyh:
@balamuruhans thank you :)

PR #2233, opened 2019-08-30
Update set_vm_disk and create_controller_xml to make disks work with model

Yingshun:
This PR updates set_vm_disk() and create_controller_xml() in
utils_test/libvirt to set model for disk and controller.

There are 2 ways to make virtio device work with legacy guest drivers.
1. Update 'model virtio-transitional' in disk
2. Update model for controller

This is required by legacy guest migration case for now.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1918,23 +1918,47 @@ def create_hostdev_xml(pci_id, boot_order=None, xmlfile=True,
     return hostdev_xml.xml
 
 
-def create_controller_xml(contr_type='scsi', contr_model='virtio-scsi',
-                          contr_index='0', contr_alias=None):
+def create_controller_xml(contr_dict, oper="get_xml", vm_name=None):
     """
-    Create controller xml
+    Create controller xml or add controller
 
-    :param contr_type: controller type name
-    :param contr_model: controller device model
-    :param contr_index: controller device index
-    :param contr_alias: controller device alias name
-    :return: controller xml file
+    :param contr_dict: The dict params includs controller configurations
+    :param oper: Operation need to do, current we have 2 choice
+                1. "get_xml": controller xml according to dict
+                2. "add_controller": add controller according to dict and
+                updated in vm
+    :param vm_name: name of vm, mandatory for 'add_controller' operation
+    :return: controller xml file if oper is equal to "get_xml";
+             controller object for 'add_controller'.
     """
+    contr_type = contr_dict.get("controller_type", 'scsi')
+    contr_model = contr_dict.get("controller_model", "virtio-scsi")
+    contr_index = contr_dict.get("controller_index")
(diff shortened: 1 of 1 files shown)


Yingshun:
depends on: https://github.com/avocado-framework/avocado-vt/pull/2220

vivianQizhu:
It works well for me, even it has defect like creating duplicate disks, and might fail on vm.start if they share same file, but it could be worked around anyway when utilize it, and I have no better solution now.
So Ack.

virttest/utils_test/libvirt.py
```diff
@@ -1918,23 +1918,43 @@ def create_hostdev_xml(pci_id, boot_order=None, xmlfile=True,
     return hostdev_xml.xml
 
 
-def create_controller_xml(contr_type='scsi', contr_model='virtio-scsi',
-                          contr_index='0', contr_alias=None):
+def create_controller_xml(vm_name, contr_dict, oper="add_controller"):
```
kylazhang: get_xml should be the default as it will not update the existing vm

virttest/utils_test/libvirt.py
```diff
@@ -1918,23 +1918,43 @@ def create_hostdev_xml(pci_id, boot_order=None, xmlfile=True,
     return hostdev_xml.xml
 
 
-def create_controller_xml(contr_type='scsi', contr_model='virtio-scsi',
-                          contr_index='0', contr_alias=None):
+def create_controller_xml(vm_name, contr_dict, oper="get_xml"):
```
chunfuwen: I have two concerning brought by  those change:
1)bind create_controller_xml with VM in earlier stage, suppose that controller may not be bind to specific VM
2)As it names imply , create is 'create' , DO NOT infuse create and get together, otherwise it will confuse users

virttest/utils_test/libvirt.py
```diff
@@ -1918,23 +1918,43 @@ def create_hostdev_xml(pci_id, boot_order=None, xmlfile=True,
...
+    contr_model = contr_dict.get("controller_model", "virtio-scsi")
+    contr_index = contr_dict.get("controller_index")
+    contr_alias = contr_dict.get("contr_alias")
     contr = controller.Controller(contr_type)
     contr.model = contr_model
-    contr.index = contr_index
+    if contr_index:
+        contr.index = contr_index
     if contr_alias:
         contr.alias = dict(name=contr_alias)
```
vivianQizhu: I have new request here, need specify bus for controller, 
```
<controller index="0" model="virtio-transitional" type="scsi">
    <address **bus**="0x08" domain="0x0000" function="0x0" slot="0x01" type="pci" />
</controller>
```
Maybe something like:
```
address = contr.new_controller_address({'bus': controller_bus})
contr.address = address
```

Yingshun: @vivianQizhu updated.

PR #2237, opened 2019-09-02
graphical_console: Add mouse related functions

peixiu:
Include mouse button down/up/press func,
mouse scroll up/down func,
and mouse move func.

id: 1724060, 1724061

### DIFF
--- a/virttest/graphical_console.py
+++ b/virttest/graphical_console.py
@@ -2,11 +2,14 @@
 Graphical Console
 """
 
+from __future__ import division
 import time
 import os
 import csv
+import math
 
 from virttest import data_dir
+from virttest import ppm_utils
 
 
 _STR_DASH = '-'
@@ -31,6 +34,38 @@ def get_keymap(keymap_file):
         return keymap_dict
 
 
+def uniform_linear(duration, rate):
+    """
+    Uniform linear trace for mouse move.
+
+    :param duration: the move duration time.
+    :param rate: frequency of sending move event.
+    :param src: start position.
+    :param dst: destination position.
+
+    :return: the object with a position in this line.
+    :rtype: function for motion.
+    """
+
+    def motion(src, dst):
+        start_x, start_y = src
+        end_x, end_y = dst
+        nr_samples = (duration * rate) if duration != 0 else 1
+        dx = (end_x - start_x) / nr_samples
+        dy = (end_y - start_y) / nr_samples
+        x, y = start_x, start_y
+        interval = (duration / nr_samples) if duration != 0 else 0
+        while True:
+            x += dx
+            y += dy
+            time.sleep(interval)
+            if (abs(x - end_x) <= 1) and (abs(y - end_y) <= 1):
+                break
+            yield (round(x), round(y))
+        yield (end_x, end_y)
+    return motion
+
+
 class UnsupportedKeyError(Exception):
     """
     Unsupported Key Error for Console
@@ -45,8 +80,20 @@ class BaseConsole(object):
     KEY_DOWN = 1
     KEY_UP = 0
 
+    BTN_DOWN = 1
+    BTN_UP = 0
+
+    SCROLL_FORWARD = 1
+    SCROLL_BACKWARD = -1
+
+    _pointer_pos = None
+
     def __init__(self, vm, logfile=None):
         self._vm = vm
+        self._width, self._height = self.screen_size
+        # TODO: screen size is changeable, should ensure they were synchronized before access.
+        if not self._pointer_pos:
+            self._set_pointer_pos((self._width//2, self._height//2))
         # TODO: logfile trace
 
     def key_press(self, keystroke, interval=0):
@@ -96,6 +143,116 @@ class BaseConsole(object):
         """
         time.sleep(hold_time)
 
+    def btn_press(self, btnstroke, interval=0):
+        """
+        Press the button.
+
+        :param btnstroke: button, like left, right, middle.
+        :param interval: the interval between two buttons, default interval=0.
+        """
+
+        for btn in self._get_seq(btnstroke, interval):
+            self._btn_event(self.BTN_DOWN, btn)
+
+    def btn_release(self, btnstroke, interval=0):
+        """
+        Release the button.
+
+        :param btnstroke: button, like left, right, middle.
+        :param interval: the interval between two buttons, default interval=0.
+        """
+
+        for btn in self._get_seq(btnstroke, interval):
+            self._btn_event(self.BTN_UP, btn)
+
+    def btn_click(self, btnstroke, hold_time=0, interval=0):
+        """
+        Click the button.
+
+        :param btnstroke: Mouse button, like left, right, middle.
+        :param hold_time: the time that keep one btn pressed, default hole_time=0.
+        :param interval: the interval between two btns, default interval=0.
+        """
+
+        self.btn_press(btnstroke, interval)
+        self._hold_time(hold_time)
+        btnstroke = self._reverse(btnstroke)
+        self.btn_release(btnstroke, interval)
+
+    def scroll_backward(self, count=1, interval=0):
+        """
+        Scroll the wheel to backward.
+
+        :param count: scroll count, default count=1.
+        :param interval: the interval between two scroll actions, default interval=0.
+        """
+
+        count -= 1
+        for num in range(count):
+            self._scroll_event(self.SCROLL_BACKWARD)
+            if interval:
+                time.sleep(interval)
+        self._scroll_event(self.SCROLL_BACKWARD)
+
+    def scroll_forward(self, count=1, interval=0):
+        """
+        Scroll the wheel to forward.
+
+        :param count: scroll count, default count=1.
+        :param interval: the interval between two scroll actions, default interval=0.
+        """
+
+        count -= 1
+        for num in range(count):
+            self._scroll_event(self.SCROLL_FORWARD)
+            if interval:
+                time.sleep(interval)
+        self._scroll_event(self.SCROLL_FORWARD)
+
+    def pointer_move(self, pos, motion=None, absolute=True):
+        """
+        Move the pointer to a given position.
+
+        :param pos: pointer position as (x, y)
+        :param motion: motion line, like uniform linear, default motion=None.
+        :param absolute: True means motion with absolute coordinates,
+                     False means motion with relatively coordinates.
+        """
+
+        if motion:
+            for _pos in motion(self._pointer_pos, pos):
+                self._motion_event(_pos, absolute)
+        else:
+            self._motion_event(pos, absolute)
+
+    @property
+    def screen_size(self):
+        """
+        Get the console screen size in pixels.
+        Returns a tuple of 2 integers
+        """
+
+        raise NotImplementedError
+
+    @property
+    def pointer_pos(self):
+        """
+        Get the pointer position in console, the unit is pixel.
+        Returns a tuple of 2 integers.
+        """
+
+        return self._pointer_pos
+
+    @classmethod
+    def _set_pointer_pos(cls, pos):
+        """
+        Set the pointer position in console, the unit is pixel.
+
+        :param pos: a tuple of 2 integers.
+        """
+
+        cls._pointer_pos = pos
+
     @staticmethod
     def _get_seq(seq_str, interval):
         """
@@ -144,6 +301,31 @@ class BaseConsole(object):
         """
         raise NotImplementedError()
 
+    def _btn_event(self, event, btn):
+        """
+        Send mouse button event.
+
+        :param event: btn event.
+        :param btn: mouse button.
+        """
+        raise NotImplementedError()
+
+    def _scroll_event(self, event):
+        """
+        Send mouse scroll event.
+
+        :param event: scroll event.
+        """
+        raise NotImplementedError()
+
+    def _motion_event(self, pos, absolute=True):
+        """
+        Send mouse move event.
+
+        :param pos: mouse position as (x,y).
+        """
+        raise NotImplementedError()
+
     def close(self):
         pass
 
@@ -165,6 +347,12 @@ class QMPConsole(BaseConsole):
     KEY_UP = False
 
     KEYMAP = get_keymap("keymap_to_qcode.csv")
+
+    BTN_DOWN = True
+    BTN_UP = False
+
+    SCROLL_FORWARD = "wheel-up"
+    SCROLL_BACKWARD = "wheel-down"
 
     def key_press(self, keystroke, interval=0):
         """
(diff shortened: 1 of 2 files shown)


peixiu:
@luckyh please help to review it when you free, thanks a lot~

peixiu:
@luckyh Hi xuhan, I updated the code, could you help to review it?
Thanks a lot~

peixiu:
Hi @luckyh , I updated the code, please help to review it when you free~
Thanks a lot~

peixiu:
Updated the code, tested on win2019, all passed.

PASS 1-Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.vioinput_mice.device_mouse.q35
PASS 2-Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.vioinput_mice.device_tablet.q35

peixiu:
@luckyh please help to review this patch~
Thanks~

virttest/graphical_console.py
```diff
@@ -31,22 +34,76 @@ def get_keymap(keymap_file):
...
+    BTN_DOWN = 1
+    BTN_UP = 0
+
+    SCROLL_FORWARD = 1
+    SCROLL_BACKWARD = -1
+
     def __init__(self, vm, logfile=None):
         self._vm = vm
+        self.width, self.height = self.screen_size()
+        self.pointer_pos = (int(self.width/2), int(self.height/2))
```
luckyh: Can't assume the pointer always be at the centre of the screen, for example, when you reattach the console with a newly created object, the pointer may be away of the centre.

virttest/graphical_console.py
```diff
@@ -96,6 +153,104 @@ def _hold_time(self, hold_time):
...
+        """
+
+        count -= 1
+        for num in range(count):
+            self._scroll_event("up")
+            if interval:
+                time.sleep(interval)
+        self._scroll_event("up")
+
+    def pointer_move(self, pos, motion=None, mtype='abs'):
```
luckyh: Suggest `mtype='abs'` -> `abs=True`

virttest/graphical_console.py
```diff
@@ -229,6 +415,95 @@ def _key_event(self, down, keys):
...
+                      But qmp only support scroll vertical up/down.
+                      Does not support scroll horizontal left/right currently.
+        """
+
+        if event == "up":
+            self._btn_event(self.BTN_DOWN, self.SCROLL_FORWARD)
+        elif event == "down":
+            self._btn_event(self.BTN_DOWN, self.SCROLL_BACKWARD)
+        else:
+            raise ConsoleNotSupportedError("qmp console doesn't support"
```
luckyh: Seems it won't happen so you can remove the `else` branch.

virttest/graphical_console.py
```diff
@@ -229,6 +415,95 @@ def _key_event(self, down, keys):
...
+            y = pos[1] - self.pointer_pos[1]
+        return (x, y)
+
+    def screen_size(self):
+        """
+        Get the current screen size in pixels.
+        Returns a tuple of 2 integers
+        """
+
+        image_file = "/tmp/screendump"
```
luckyh: Should use the tmp dir provided by VT
```python
from virttest import data_dir

data_dir.get_tmp_dir()
```

luckyh: Have to ensure that the filename has not been used.

virttest/graphical_console.py
```diff
@@ -229,6 +415,95 @@ def _key_event(self, down, keys):
...
+        self._vm.monitor.screendump(image_file)
+        img = Image.open(image_file)
+        return img.size
```
luckyh: Instead of directly manipulating the image, it is better to introduce a function (i.e. `image_size`) to the `ppm_utils` module and call it at here
```python
def image_size(filename):
    img = Image.open(image_file)
    return img.size
```

virttest/graphical_console.py
```diff
@@ -96,6 +153,104 @@ def _hold_time(self, hold_time):
...
+        :param hold_time: the time that keep one btn pressed, default hole_time=0.
+        :param interval: the interval between two btns, default interval=0.
+        """
+
+        self.btn_press(btnstroke, interval)
+        self._hold_time(hold_time)
+        btnstroke = self._reverse(btnstroke)
+        self.btn_release(btnstroke, interval)
+
+    def scroll_down(self, count=1, interval=0):
```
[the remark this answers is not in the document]
peixiu: yes, will update, thanks~

virttest/graphical_console.py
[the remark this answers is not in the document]
peixiu: thanks, will update~

virttest/graphical_console.py
```diff
@@ -96,6 +153,104 @@ def _hold_time(self, hold_time):
...
+        """
+        Scroll the wheel to down.
+
+        :param count: scroll count, default count=1.
+        :param interval: the interval between two scroll actions, default interval=0.
+        """
+
+        count -= 1
+        for num in range(count):
+            self._scroll_event("down")
```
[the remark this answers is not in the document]
peixiu: ok, will update it, thanks~

virttest/graphical_console.py
peixiu: thanks, will updated it~

virttest/graphical_console.py
peixiu: Considering may be need support scroll_left or scroll_right event except scroll_up/down in the future, so add this else branch here, qmp console doesn't support them now.

virttest/graphical_console.py
peixiu: ok, will update, thanks

peixiu: ok, will update it, thanks

virttest/graphical_console.py
peixiu: thanks, will update~

virttest/graphical_console.py
peixiu: I do a pointer position initialization here, if users create a new graphical console object, the default pointer position will set to this console screen central, it will be changed along with pointer move, users will always get the latest pointer position in the same console, but if they create a new object again, the pointer position will also be re-initiated in this new console.

peixiu:
Hi @luckyh I updated the code, for screendump file name, I changed it to screendump + 4 random num+str characters as suffix.
when called screen_size(), all screendump_* file are save in new generated tem dir,  we can cover it if file name is same with before, we catch screen size need to use the latest screen dump file, so I do not add if file is exists codes here. Please help to review it again,  thanks a lot~

Best Regards~
Peixiu

virttest/graphical_console.py
luckyh: Let me explain about that, supposes that you have created an console object and moved the pointer to somewhere, for example the top-right, the screen would look like the following (`^` is presented as the pointer):
```
+-------+
|     ^ |
|       |
|       |
+-------+
```
then if you close the console connection and recreate a new one, by the current approach it would look like:
```
  +-------+
+ | - - + |
  |   ^   |
| |     | |
  +-------+
+ - - - +
```
the box drawn by solid line is the screen seen by the current console, and the one drawn by dash line is the screen seen by the previous, they do not coincide, and that means you have a new coordinate system now, then the question is which one should you choose to get the position of the pointer?

virttest/graphical_console.py
```diff
@@ -96,6 +147,104 @@ def _hold_time(self, hold_time):
...
+        """
+
+        count -= 1
+        for num in range(count):
+            self._scroll_event(self.SCROLL_FORWARD)
+            if interval:
+                time.sleep(interval)
+        self._scroll_event(self.SCROLL_FORWARD)
+
+    def pointer_move(self, pos, motion=None, _abs=True):
```
luckyh: Please use the full spelling `absolute` instead of `_abs`.

virttest/graphical_console.py
```diff
@@ -229,6 +409,90 @@ def _key_event(self, down, keys):
...
+    def _scroll_event(self, event):
+        """
+        Send mouse scroll event.
+
+        :param event: Event as vertical forward/backward or horizontal left/right.
+                      But qmp only support scroll vertical forward/backward.
+                      Does not support scroll horizontal left/right currently.
+        """
+
+        self._btn_event(self.BTN_DOWN, event)
```
luckyh: typo?

virttest/graphical_console.py
```diff
@@ -229,6 +409,90 @@ def _key_event(self, down, keys):
...
+        return (x, y)
+
+    def screen_size(self):
+        """
+        Get the current screen size in pixels.
+        Returns a tuple of 2 integers
+        """
+
+        image = data_dir.get_tmp_dir() + "/screendump_" + \
+            utils_misc.generate_random_string(4)
```
luckyh: Suggest:
```python
tmp_dir = os.path.join(data_dir.get_tmp_dir(), "graphic_console_%s" % self._vm.name)
if not os.path.exists(tmp_dir):
    os.mkdir(tmp_dir)
image = os.path.join(tmp_dir, "screendump")

```

virttest/graphical_console.py
peixiu: ok, will update~

virttest/graphical_console.py
peixiu: For qmp_console, send btn event with qmp command as:
{'execute': 'input-send-event', 'arguments': {'events': [{'type': 'btn', 'data': {'down': True, 'button': 'left'}}]}, 'id': 'W3uMGZHZ'}
send scroll event is same with send btn event, just only has True option, none False option, like:
{'execute': 'input-send-event', 'arguments': {'events': [{'type': 'btn', 'data': {'down': True, 'button': 'wheel-up'}}]}, 'id': 'Qohmxaei'}
{'execute': 'input-send-event', 'arguments': {'events': [{'type': 'btn', 'data': {'down': True, 'button': 'wheel-down'}}]}, 'id': 'VqEnRfs1'}
Due to they have same structure, and I want to separate the _btn_event() and _scroll_event away,  so I call self._btn_event() directly for _scroll_event(),  just I should not use "event" param, it should define scroll action(wheel-up or wheel-down),  "event" is not so clear here, I want change it to "scroll", do you think so that?

virttest/graphical_console.py
peixiu: Thanks a lot, will update it~

virttest/graphical_console.py
peixiu: as we discussed, will use a class attribute to save pointer position, will get pointer position dynamically through this class attribute. Thanks a lot~

virttest/graphical_console.py
luckyh: Ok, thanks for the clarification! so I think we'd better 'leave a comment for that.

Btw, doesn't it need to be followed with an up event?

virttest/graphical_console.py
peixiu: Updated the codes, new added a class method _set_pointer_pos() to change the class attribute POINTER_POS, because instance method cannot change it.
Thanks~

virttest/graphical_console.py
peixiu: updated the code, and yes, qemu doesn't need to do up action for wheel-up and wheel-down, only for btn(left, right, middle,side, extra) need do up action followed down action.

virttest/graphical_console.py
peixiu: updated~

peixiu:
Tested with patch https://github.com/autotest/tp-qemu/pull/1902 together:
 (1/2) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2012.x86_64.r2.io-github-autotest-qemu.vioinput_mice.device_mouse.q35: PASS (41.74 s)
 (2/2) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2012.x86_64.r2.io-github-autotest-qemu.vioinput_mice.device_tablet.q35: PASS (41.84 s)

@luckyh please help to review it again, thanks a lot~

peixiu:
@luckyh please help to review this patch again, thanks a lot~

virttest/graphical_console.py
```diff
@@ -31,6 +34,45 @@ def get_keymap(keymap_file):
...
+        dx_int, dx_float = str((x2 - x1) / num).split('.')
+        dy_int, dy_float = str((y2 - y1) / num).split('.')
+        decimal_x = 0
+        decimal_y = 0
+        for i in range(num):
+            decimal_x += float("0." + dx_float)
+            decimal_y += float("0." + dy_float)
+            if decimal_x >= 1:
+                x += (-1 if "-" in str(dx_int) else 1)
+                decimal_x = float("0." + (str(decimal_x).split(".")[1]))
```
luckyh: Looks too complicated, is there some reason to have to use this handmade signed floor? shall we use `round` (四舍五入) instead? then the code would be more clear.
```python
def motion(src, dst):
    start_x, start_y = src
    end_x, end_y = dst
    nr_samples = (duration * rate) if duration != 0 else 1
    dx = (end_x - start_x) / nr_samples
    dy = (end_y - start_y) / nr_samples
    x, y = start_x, start_y
    interval = (duration / nr_samples) if duration != 0 else 0
    while True:
        x += dx
        y += dy
        time.sleep(interval)
        if (abs(x) >= abs(end_x)) or (abs(y) >= abs(end_y)):
            break
        yield (round(x), round(y))
    yield (end_x, end_y)

File: virttest/graphical_console.py [hunk 12]
```diff
@@ -31,6 +34,45 @@ def get_keymap(keymap_file):
...
+            if decimal_y >= 1:
+                y += (-1 if "-" in str(dy_int) else 1)
+                decimal_y = float("0." + (str(decimal_y).split(".")[1]))
+            x, y = x + int(dx_int), y + int(dy_int)
+            if i == num - 1:
+                x, y = x2, y2
+            yield (x, y)
+            time.sleep(duration / rate)
```
luckyh: not correct, please reference the above comment.

virttest/graphical_console.py
```diff
@@ -45,8 +87,19 @@ class BaseConsole(object):
     KEY_DOWN = 1
     KEY_UP = 0
 
+    BTN_DOWN = 1
+    BTN_UP = 0
+
+    SCROLL_FORWARD = 1
+    SCROLL_BACKWARD = -1
+
+    POINTER_POS = None
```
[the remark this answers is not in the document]
peixiu: that's ok, thanks, will update it~

virttest/graphical_console.py
```diff
@@ -96,6 +149,115 @@ def _hold_time(self, hold_time):
...
+        """
+
+        if motion:
+            #for _pos in motion(self._pointer_pos, pos):
+            for _pos in motion(self.POINTER_POS, pos):
+                self._motion_event(_pos, absolute)
+        else:
+            self._motion_event(pos, absolute)
+
+    def screen_size(self):
```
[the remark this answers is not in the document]
peixiu: ok, thanks, will update~

virttest/graphical_console.py
```diff
@@ -45,8 +87,19 @@ class BaseConsole(object):
...
+    BTN_UP = 0
+
+    SCROLL_FORWARD = 1
+    SCROLL_BACKWARD = -1
+
+    POINTER_POS = None
+
     def __init__(self, vm, logfile=None):
         self._vm = vm
+        self._width, self._height = self.screen_size()
```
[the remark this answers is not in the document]
peixiu: thanks, will update~

virttest/graphical_console.py
```diff
@@ -96,6 +149,115 @@ def _hold_time(self, hold_time):
...
+        Move the pointer to a given position.
+
+        :param pos: pointer position as (x, y)
+        :param motion: motion line, like uniform linear, default motion=None.
+        :param absolute: True means motion with absolute coordinates,
+                     False means motion with relatively coordinates.
+        """
+
+        if motion:
+            #for _pos in motion(self._pointer_pos, pos):
```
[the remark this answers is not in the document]
peixiu: ok, will update~

virttest/graphical_console.py
```diff
@@ -96,6 +149,115 @@ def _hold_time(self, hold_time):
...
+
+    def screen_size(self):
+        """
+        Get the console screen size in pixels.
+        Returns a tuple of 2 integers
+        """
+
+        raise NotImplementedError
+
+    def get_pointer_pos(self):
```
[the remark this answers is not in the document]
peixiu: ok, thanks, will update~

virttest/graphical_console.py
```diff
@@ -96,6 +149,115 @@ def _hold_time(self, hold_time):
...
+        """
+        Get the pointer position in console, the unit is pixel.
+        Returns a tuple of 2 integers.
+        """
+
+        raise NotImplementedError
+
+    @classmethod
+    def _set_pointer_pos(cls, pos):
```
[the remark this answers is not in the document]
peixiu: hi xuhan, thanks for review, I added this class method due to there are 2 situation need to use this method.
1. we need initiate the pointer_pos in center of screen, but screen_size() is a instance method, we cannot change the class attribute value through a instance method, so if use assignment directly, we need define the class attribute under the screen_size() method, it looks not beautiful for me; currently the first place used this class method in line 102. 
2. we also need change the class attribute POINTER_POS when the pointer moved(in line 469) to track the pointer place, also hit the similar situation, we cannot assignment new value to the class attribute value in a instance. method.
So I used a class method to complete it here~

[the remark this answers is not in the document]
luckyh: I mean you can set the value by `QMPConsole.POINTER_POS = (x, y)`, but yes it is not as clear as having a method, for reading. In fact, my true concern is that not all types of console would require `POINTER_POS` had to be an class attribute, then it is not good to implement this as a class method in the base class. However, I didn't figure out yet that how would different console manage their own pointer and display, so let's keep the approach until we find a better one, thanks.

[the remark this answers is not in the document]
peixiu: ok, thanks~

virttest/graphical_console.py
peixiu: use round also is ok, thanks, has updated~

virttest/graphical_console.py
peixiu: updated~

virttest/graphical_console.py
```diff
@@ -229,6 +422,95 @@ def _key_event(self, down, keys):
...
+        self._vm.monitor.screendump(image)
+        return ppm_utils.image_size(image)
+
+    def get_pointer_pos(self):
+        """
+        Get the pointer position in console, the unit is pixel.
+        Returns a tuple of 2 integers.
+        """
+
+        return self.POINTER_POS
```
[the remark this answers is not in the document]
peixiu: updated~

peixiu:
Updated the code and tested with latest code: 
(1/2) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2012.x86_64.r2.io-github-autotest-qemu.vioinput_mice.device_mouse.q35: PASS (50.40 s)
 (2/2) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2012.x86_64.r2.io-github-autotest-qemu.vioinput_mice.device_tablet.q35: PASS (48.70 s)

@luckyh Please help to review it again, thanks a lot~

virttest/graphical_console.py
```diff
@@ -96,6 +143,116 @@ def _hold_time(self, hold_time):
...
+    def screen_size(self):
+        """
+        Get the console screen size in pixels.
+        Returns a tuple of 2 integers
+        """
+
+        raise NotImplementedError
+
+    @property
+    def get_pointer_pos(self):
```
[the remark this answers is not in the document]
peixiu: Updated, thanks~

peixiu:
Tested with latest code:
 (1/2) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2012.x86_64.r2.io-github-autotest-qemu.vioinput_mice.device_mouse.q35: PASS (50.14 s)
 (2/2) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2012.x86_64.r2.io-github-autotest-qemu.vioinput_mice.device_tablet.q35: PASS (50.14 s)

luckyh:
Thanks all.

PR #2248, opened 2019-09-09
utils_conn: update to start socket before libvirtd started

Yingshun:
From libvirt 5.6, need to start below sockets, *before* libvirtd
itself is started to enable tls/tcp connection.
1. systemctl start libvirtd-tls.socket
2. systemctl start libvirtd-tcp.socket

Again, for recovery, stop sockets because they are inactive 
by default.

### DIFF
--- a/virttest/utils_conn.py
+++ b/virttest/utils_conn.py
@@ -14,6 +14,7 @@ from avocado.utils import process
 from virttest import propcan, remote, utils_libvirtd
 from virttest import data_dir
 from virttest import utils_package
+from virttest import libvirt_version
 from virttest.compat_52lts import results_stderr_52lts
 
 
@@ -728,7 +729,12 @@ class TCPConnection(ConnectionBase):
                                             server_user, server_pwd,
                                             r"[\#\$]\s*$")
             libvirtd_service = utils_libvirtd.Libvirtd(session=session)
-            libvirtd_service.restart()
+            if libvirt_version.version_compare(5, 6, 0, session):
+                libvirtd_service.stop()
+                session.cmd("systemctl stop libvirtd-tcp.socket")
+                libvirtd_service.start()
+            else:
+                libvirtd_service.restart()
         except (remote.LoginError, aexpect.ShellError) as detail:
             raise ConnServerRestartError(detail)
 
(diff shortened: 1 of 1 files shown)


virttest/utils_conn.py
```diff
@@ -728,7 +729,12 @@ def conn_recover(self):
                                             server_user, server_pwd,
                                             r"[\#\$]\s*$")
             libvirtd_service = utils_libvirtd.Libvirtd(session=session)
-            libvirtd_service.restart()
+            if libvirt_version.version_compare(5, 6, 0, session):
+                libvirtd_service.stop()
+                session.cmd("systemctl stop libvirtd-tcp.socket")
```
dzhengfy: Does libvirtd-tcp.socket need start before starting libvirtd ?

virttest/utils_conn.py
```diff
@@ -1003,7 +1016,12 @@ def conn_recover(self):
                                             server_user, server_pwd,
                                             r"[\#\$]\s*$")
             libvirtd_service = utils_libvirtd.Libvirtd(session=session)
-            libvirtd_service.restart()
+            if libvirt_version.version_compare(5, 6, 0, session):
+                libvirtd_service.stop()
+                session.cmd("systemctl stop libvirtd-tls.socket")
```
dzhengfy: Why not restart

virttest/utils_conn.py
Yingshun: > Does libvirtd-tcp.socket need start before starting libvirtd ?

It's for recovery, I think libvirtd-tcp.socket should be stopped.

virttest/utils_conn.py
Yingshun: Same reason as above.

virttest/utils_conn.py
dzhengfy: But you said in the description that before starting libvirtd, we need start this socket. And now in L735, you did start libvirtd.

Yingshun: Okay, that confused you. We have to start socket before libvirtd started to enable tls/tcp. 
but to disable them, no need to start these sockets.

dzhengfy: Regarding that these sockets are inactive by default, I can accept this change. But could you help update the description to avoid of the confusion?

Yingshun: comment is updated.

PR #2258, opened 2019-09-11
Add cgroup uitls for libvirt

lento-sun:


virttest/utils_test/libvirt.py
```diff
@@ -3759,3 +3760,278 @@ def check_dumpxml(vm, content, err_ignore=False):
...
+            with open(vm_proc_cgroup_path, 'r') as vm_cg_file:
+                cg_vm_scope = re.findall(r"\S*::(\S*)", vm_cg_file.read())
+            cgroup_path = os.path.join(cg_mount_point[0], cg_vm_scope[0].strip("/"))
+            if "emulator" in cgroup_path:
+                cgroup_path += "/.."
+        else:
+            cgroup_path = utils_cgroup.resolve_task_cgroup_path(int(self.__vm_pid),
+                                                                controller)
+        if not os.path.exists(cgroup_path):
+            logging.error("cgroup path '%s' doesn't exist" % cgroup_path)
```
chunfuwen: I suppose you need return "" or None here if the path doesn't exit

virttest/utils_test/libvirt.py
```diff
@@ -3759,3 +3760,278 @@ def check_dumpxml(vm, content, err_ignore=False):
...
+        return cgroup_path
+
+    def __get_cpu_subdirs(self, controller_path=None, dir_keyword=None):
```
chunfuwen: miss parameter doc

virttest/utils_test/libvirt.py
lento-sun: the cgroup_path set as "" in line3815, and it'll be returned in 3830 as "" if not changed.

virttest/utils_test/libvirt.py
lento-sun: added, thx

virttest/utils_test/libvirt.py
```diff
@@ -3760,3 +3761,354 @@ def check_dumpxml(vm, content, err_ignore=False):
...
+
+    """Class for cgroup related test"""
+
+    __vm_pid = ""
+    __virsh_blkiotune_output_mapping = {"weight": "weight",
+                                        "device_read_iops_sec": "riops",
+                                        "device_write_iops_sec": "wiops",
+                                        "device_read_bytes_sec": "rbps",
+                                        "device_write_bytes_sec": "wbps"}
+    __cgroup_v1_blkio_file_mapping = {"weight": "blkio.bfq.weight",
```
dzhengfy: Please also consider different value, for example, for RHEL 7.x

lento-sun: per we talked about this, cgroup2 can be tested only on rhel8, and rhel7 have existing cases, so we're not adding this job to rhel7 test scope. No need to deal with rhel7 issues for now.

virttest/utils_test/libvirt.py
```diff
@@ -3760,3 +3761,354 @@ def check_dumpxml(vm, content, err_ignore=False):
                 raise exceptions.TestFail("Expected '%s' was not found in "
                                           "%s's xml" % (content, vm.name))
     return True
+
+
+#cgroup related functions
+class CgroupTest(object):
```
chunfuwen: Maybe separate it into one singleton file , such as CgroupUtils.py since there are too many stuff in libvirt.py file

virttest/utils_test/libvirt.py
```diff
@@ -3760,3 +3761,354 @@ def check_dumpxml(vm, content, err_ignore=False):
...
+
+    """Class for cgroup related test"""
+
+    __vm_pid = ""
+    __virsh_blkiotune_output_mapping = {"weight": "weight",
```
chunfuwen: maybe move them ahead to this class, let them as static variables, in this way, other file can easily refer to them if necessary.

virttest/utils_test/libvirt.py
```diff
@@ -3760,3 +3761,354 @@ def check_dumpxml(vm, content, err_ignore=False):
...
+            if dir_keyword in filename:
+                dir_names.append(filename)
+        if not dir_names and "iothread" in dir_keyword:
+            logging.debug("No sub dirs found with keyword: '%s'. "
+                          "Pls check if you've executed virsh cmd "
+                          "'iothreadadd'.", dir_keyword)
+            return None
+        return sorted(dir_names)
+
+    def __get_standardized_cgroup1_info(self, virsh_cmd=None):
```
chunfuwen: maybe enforce virsh_cmd must be given one value, rather than None as default value

lento-sun: No suitable virsh cmd here is a suitable 'default' command. And None will trgger a error notice later in these methods

virttest/utils_test/libvirt.py
lento-sun: done, now it's "virttest/libvirt_cgroup.py", cgroup special utils is virttest/staging/utils_cgroup.py, this file is mainly for libvirt, so I named it libvirt_cgroup.py

virttest/utils_test/libvirt.py
lento-sun: Done, now they're public consts

lento-sun:
And update tp-libvirt pr2324 to import the correct module, all cases passed locally
...
RESULTS    : PASS 168 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 1407.06 s

PR #2279, opened 2019-09-25
virttest.guest_agent: Add function related to memory block

xiagao:
Add guest-get-memory-block-info,guest-get-memory-blocks and
guest-set-memory-blocks functions.
ID: 1754370

### DIFF
--- a/virttest/guest_agent.py
+++ b/virttest/guest_agent.py
@@ -875,3 +875,30 @@ class QemuAgent(Monitor):
         cmd = "guest-get-osinfo"
         self.check_has_command(cmd)
         return self.cmd(cmd)
+
+    def get_memory_block_info(self):
+        """
+        Get information relating to guest memory blocks.
+        """
+        cmd = "guest-get-memory-block-info"
+        self.check_has_command(cmd)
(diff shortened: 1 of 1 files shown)


xiangchunfu:
looks good to me. 
Acked by:Xiangchun Fu xfu@redhat.com

virttest/guest_agent.py
```diff
@@ -875,3 +875,30 @@ def get_osinfo(self):
...
+        Get the list of the guest's memory blocks.
+        """
+        cmd = "guest-get-memory-blocks"
+        self.check_has_command(cmd)
+        return self.cmd(cmd=cmd)
+
+    def set_memory_blocks(self, mem_blocks_list):
+        """
+        Reconfigure (currently: enable/disable) state of guest memory blocks.
+        :param mem_blocks_list: a list of all memory blocks the guest
```
vivianQizhu: From my understanding this should not be 'all' memory blocks, but only the ones you want to configure. Isn't it?

xiagao: You are right, updated.

vivianQizhu:
Ack.

PR #2281, opened 2019-09-25
libvirt_vm: use `nfs_mount_dir` as base_dir for NFS tests

balamuruhans:
framework picks wrong path from `image_base_dir` or `data_dir` path
for nfs tests causing VM to boot/test with wrong disk path. Instead
use `nfs_mount_dir` for VM to use appropriate image for NFS based
tests.

Fixes: commit 0b463201 and 782cc76a

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -1197,12 +1197,14 @@ class VM(virt_vm.BaseVM):
                 virt_install_cmd += add_pcidevice(help_text, pci_id)
 
         for image_name in params.objects("images"):
+            basename = False
             image_params = params.object_params(image_name)
 
             base_dir = image_params.get("images_base_dir",
                                         data_dir.get_data_dir())
-
-            basename = params.get("storage_type") == "nfs"
+            if params.get("storage_type") == "nfs":
+                basename = True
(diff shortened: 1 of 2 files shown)


virttest/libvirt_vm.py
```diff
@@ -1197,12 +1197,14 @@ def add_cputune(vcpu_cputune=""):
...
+            basename = False
             image_params = params.object_params(image_name)
 
             base_dir = image_params.get("images_base_dir",
                                         data_dir.get_data_dir())
-
-            basename = params.get("storage_type") == "nfs"
+            if params.get("storage_type") == "nfs":
+                basename = True
+                base_dir = params["nfs_mount_dir"]
```
kylazhang: How about also give default value for base_dir like other params?

balamuruhans: when user choose "nfs" as storage, the directory path is expected to be configured by user. so it is better to error out early and appropriately, instead of assuming with default values.

balamuruhans:
@sathnaga @kylazhang Thanks for reviewing, It fixes the commits mentioned in commit msg, let's merge it as the migration run with NFS is broken with out this patch.

PR #2291, opened 2019-10-11
Add the new Qemu capabilities key to the qcontainer comparison ignorable ones

pevogam:
The vms in our runs are constantly reset using release 72.0 and the
reason is mismatched env configuration within the same vm. The
mismatched params come from no real configuration change but from
a different `qemu_capabilities.Capabilities` instance which is not
relevant to such use cases.

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -304,8 +304,7 @@ class DevContainer(object):
         qdev2 = qdev2.__dict__
         for key, value in six.iteritems(self.__dict__):
             if key in ("_DevContainer__devices", "_DevContainer__buses",
-                       "_DevContainer__state",
-                       "allow_hotplugged_vm"):
+                       "_DevContainer__state", "caps", "allow_hotplugged_vm"):
                 continue
             if key not in qdev2 or qdev2[key] != value:
                 return False


virttest/qemu_devices/qcontainer.py
```diff
@@ -304,7 +304,7 @@ def __eq__(self, qdev2):
         qdev2 = qdev2.__dict__
         for key, value in six.iteritems(self.__dict__):
             if key in ("_DevContainer__devices", "_DevContainer__buses",
-                       "_DevContainer__state",
+                       "_DevContainer__state", "_DevContainer__caps",
```
luckyh: After https://github.com/avocado-framework/avocado-vt/pull/2274/commits/3f0da17ed7830c00fa20575defdb43aed1d90c2d got merged, it should be `"caps"` I guess.

pevogam: Hi @luckyh, I am not aware of these most recent changes. Do you mean renaming the key `"_DevContainer__caps"` simply to `"caps"`? Unfortunately, we only test official release versions so I won't be able to test this any time soon. We could still proceed with this if you can confirm that this is indeed the correct key now.

luckyh: As capabilities used to be a private attribute, you can use the name `"_DevContainer__caps"` to access it, but it is a public one now, therefore have to change the name to `"caps"`.

pevogam: Alright, I have pushed an adapted version of this.

PR #2302, opened 2019-10-21
Fixup: remove hard coded grep binary path

sathnaga:
In some distribution, grep binary is not placed
under /usr/bin path, rather in /bin path, so let's
find out the grep binary path and use it rather.

### DIFF
--- a/virttest/utils_time.py
+++ b/virttest/utils_time.py
@@ -1,6 +1,7 @@
 import logging
 import re
 
+from avocado.utils import path
 from avocado.utils import process
 from avocado.core import exceptions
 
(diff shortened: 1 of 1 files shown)


virttest/utils_time.py
```diff
@@ -8,15 +8,17 @@
 from virttest import utils_test
 from virttest.compat_52lts import decode_to_text
 
+# command `grep --color` may have alias name `grep` in some systems,
+# so get explicit command 'grep' with path
+grep_binary = decode_to_text(process.system_output("which grep|grep -v alias", shell=True))
```
balamuruhans: This should help here,

```python
from avocado.utils import path
path.find_command("grep")
```

sathnaga: @balamuruhans perfect!, will use it, thanks for the suggestion...

PR #2312, opened 2019-10-28
Update boot loader entry on s390x

smitterl:
On s390x running zipl command is required after updating boot loader entries
(with grubby), see https://bugzilla.redhat.com/show_bug.cgi?id=1764306#c2
In other words, running grubby and then rebooting without zipl won't change
cmdline.

1. update_boot_option:
 a. Added new parameter guest_arch_name: zipl has to be run only on s390x
    guests. Has default value for backwards compatibility.
 b. __run_cmd_handle_error: extracted cmd line handling into private
    method to avoid duplication.

2. virttest/unittests:
 a. Created new package for unit tests because virttest/tests contains
    higher level test cases.

3. test_utils_test__init__.py:
 a. Added unit tests for 1.a./b.
 b. try: import unittest.mock...: In order to run the unit tests with 2.x
    'mock' has to be installed from pip, import will be different.

### DIFF
--- a/virttest/unittests/test_utils_test__init__.py
+++ b/virttest/unittests/test_utils_test__init__.py
@@ -1,0 +1,57 @@
+import unittest
+import logging
+
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
+from virttest.utils_test import update_boot_option
+from virttest import utils_test
+from avocado.core import exceptions
+
+check_kernel_cmdline_mock = mock.MagicMock(return_value=["3", None])
+
+
+@mock.patch('virttest.utils_package.package_install')
+@mock.patch.object(utils_test, 'check_kernel_cmdline', check_kernel_cmdline_mock)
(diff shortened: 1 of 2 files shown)


smitterl:
@ldoktor May I request you review? Thank you :)

smitterl:
@dzhengfy Please, could you review?

smitterl:
@dzhengfy Unit test run:
Testing started at 10:33 AM ...
/root/code/p3-ve/bin/python /opt/pycharm-community-2019.2.3/helpers/pycharm/_jb_unittest_runner.py --target virttest.unittests.test_utils_test__init__.TestUpdateBootOptionZipl -- -v
Launching unittests with arguments python -m unittest -v virttest.unittests.test_utils_test__init__.TestUpdateBootOptionZipl in /root/code/tp-libvirt



Ran 3 tests in 0.007s

OK

PR #2319, opened 2019-11-06
test_setup: Enhance code of create private bridge

PaulYuuu:
1. Check if dnsmasq package exists before _start_dhcp_server
2. Use a better method to check ovs bridge
    - If the bridge name contains digits, the bridge info is similar to the following:
         ```bash
        $ ovs-vsctl show
        65fbb3f9-b98f-4009-8c20-8eb725ba3d2a
            Bridge ovsI
                Port ovsI
                    Interface ovsI
                        type: internal
            Bridge "ovs0"
                Port "ovs0"
                    Interface "ovs0"
                        type: internal
            ovs_version: "2.11.0"
        ```
    - So `re.match` could not match it
https://github.com/avocado-framework/avocado-vt/blob/9acbffa1bd200be205c5cff771c0a1b3e277a80a/virttest/test_setup.py#L1005-L1012
    - This will block the call to the cleanup method
https://github.com/avocado-framework/avocado-vt/blob/9acbffa1bd200be205c5cff771c0a1b3e277a80a/virttest/test_setup.py#L981-L984

ID: 1783859

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -24,7 +24,6 @@ from avocado.utils import distro
 from avocado.core import exceptions
 
 from virttest import data_dir
-from virttest import kernel_interface
 from virttest import error_context
 from virttest import cpu
 from virttest import utils_misc
@@ -34,6 +33,7 @@ from virttest import remote
 from virttest import utils_libvirtd
 from virttest import utils_net
 from virttest import utils_config
+from virttest import utils_package
 from virttest import kernel_interface
 from virttest.staging import service
 from virttest.staging import utils_memory
@@ -797,6 +797,7 @@ class PrivateBridgeConfig(object):
         if params is not None:
             self.brname = params.get("priv_brname", 'atbr0')
             self.subnet = params.get("priv_subnet", '192.168.58')
+            self.netmask = params.get("priv_netmask", '255.255.255.0')
             self.ip_version = params.get("bridge_ip_version", "ipv4")
             self.dhcp_server_pid = None
             ports = params.get("priv_bridge_ports", '53 67').split()
@@ -822,6 +823,7 @@ class PrivateBridgeConfig(object):
             self.force_create = False
             if params.get("bridge_force_create", "no") == "yes":
                 self.force_create = True
+        self.bridge_manager = utils_net.Bridge()
 
     def _assemble_iptables_rules(self, port_list):
         rules = []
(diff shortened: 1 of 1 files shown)


virttest/test_setup.py
```diff
@@ -868,6 +868,8 @@ def _enable_nat(self):
             self._iptables_add(rule)
 
     def _start_dhcp_server(self):
+        if not utils_package.package_install("dnsmasq"):
+            exceptions.TestSkipError("Failed to install dnsmasq package")
         process.system("service dnsmasq stop")
```
xutian: let's stop dnsmasq service with service module in framework  to make code works more stable. eg, 
`
 service.SpecificServiceManager("dnsmasq").stop()`

Other parts looks good to me.

PaulYuuu: > let's stop dnsmasq service with service module in framework to make code works more stable. eg,
> ` service.SpecificServiceManager("dnsmasq").stop()`
> 
Updated.
> Other parts looks good to me.

Also, I replaced some steps with the Bridge Manager method, test pass with `netdst=private` and `testcase=boot`. Please help to review again, thanks.

xutian:
Acked-by: XuTian <xutian@redhat.com>

xutian:
@PaulYuuu   you need to read trigger CI test.

PaulYuuu:
> @PaulYuuu you need to read trigger CI test.

Fixed, thanks for your reminder.

xutian:
Acked-by: XuTian <xutian@redhat.com>

PaulYuuu:
@vivianQizhu Could you help to review? Thanks a lot.

virttest/test_setup.py
```diff
@@ -868,7 +872,9 @@ def _enable_nat(self):
             self._iptables_add(rule)
 
     def _start_dhcp_server(self):
-        process.system("service dnsmasq stop")
+        if not utils_package.package_install("dnsmasq"):
```
xutian: only need to install package if it's not installed.

virttest/test_setup.py
```diff
@@ -975,8 +972,10 @@ def _disable_nat(self):
             self._iptables_del(rule)
 
     def _remove_bridge(self):
-        process.system("ip link delete %s type bridge" % self.brname,
-                       ignore_status=True)
+        try:
+            self.bridge_manager.del_bridge(self.brname)
+        except:
```
xutian: better to print warning here

PaulYuuu: updated

virttest/test_setup.py
PaulYuuu: `package_install` will check it, so don't need to double check.

PaulYuuu:
@yanglei-rh Please help to review, thanks.

PR #2325, opened 2019-11-11
Fix vt-bootstrap

smitterl:
Raising EnvironmentError won't be handled by avocado/core/extension_manager.py anymore since https://github.com/avocado-framework/avocado/commit/c2e30ab980520ab025673a0e60f07de1f32dfa94
Consequence is that if there's no previous avocado installation (/var/lib/avocado)
vt-bootstrap will always fail the first time with
"FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/avocado/data/avocado-vt/test-providers.d/downloads'"
Raising ImportError instead of EnvironmentError to maintain existing
behavior.


smitterl:
@dzhengfy

PR #2329, opened 2019-11-12
Restrict set of seclabels to be removed

smitterl:
1. del_seclabel:
 a. The function originally deleted all seclabels.
 b. Added new parameter maintaining legacy behavior.
 c. New parameter let's restrict the set of seclables
    to be removed (see docstring).

2. virttest/unittests:
 a. Created new package for unit tests and subpackage to have
    consistent paths in unittests.

3. test_vm_xml.py:
 a. Added test cases to cover previous/default and new behavior.

### DIFF
--- a/virttest/unittests/libvirt_xml/test_vm_xml.py
+++ b/virttest/unittests/libvirt_xml/test_vm_xml.py
@@ -1,0 +1,48 @@
+import unittest
+
+from virttest.libvirt_xml import vm_xml, xcepts
+
+XML = """
+<domain type='kvm'>
+  <seclabel type='dynamic' model='selinux' relabel='yes'/>
+  <seclabel type='dynamic' model='dac' relabel='yes'/>
+</domain>
+"""
+
+
+def get_vmxml():
+    vmxml = vm_xml.VMXML()
+    vmxml['xml'] = XML.strip()
+
+    return vmxml
+
+
+class TestVMXMLDelSeclabel(unittest.TestCase):
+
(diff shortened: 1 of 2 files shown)


smitterl:
@dzhengfy

virttest/unittests/libvirt_xml/test_vm_xml.py
```diff
@@ -0,0 +1,135 @@
+import unittest
+
+from virttest.libvirt_xml import vm_xml, xcepts
+
+XML = """
```
chunfuwen: how about put those VM config into one separate xml under the same folder

chunfuwen: How about put those config into separate xml file ? you can refer to https://github.com/autotest/tp-libvirt/pull/2216, for how access this file.

smitterl: @chunfuwen I'd like to not load the file from disk as it's currently not needed in other cases. Instead, I reduced the xml snippet to only contain the xml relevant for the test. What do you think?

```bash
# python -W ignore -m unittest test_vm_xml.py -v
test_del_seclabel_default (test_vm_xml.TestVMXMLDelSeclabel) ... ok
test_del_seclabel_with_conditions (test_vm_xml.TestVMXMLDelSeclabel) ... ok
test_del_seclabel_with_partial_match (test_vm_xml.TestVMXMLDelSeclabel) ... ok

Ran 3 tests in 0.083s

OK
```

PR #2330, opened 2019-11-13
Fix vmx filename in '-i vmx' for v2v

xiaodwan:
The name of a guest in esx, the nfs path of esx guest files and the
name of vmx file may be different

For '-i vmx', It always needs to mount the nfs storage to local host,
so we can get the real vmx file name. But for '-i vmx -it ssh', NFS
storage don't have to be mounted, it's not worth mounting it and only
for getting the correct vmx file name. Luckily we can infer more than
95% guest's name by it's nfs disk path. And if the inference is wrong,
V2V will report an error. If v2v runs successfully, it can be sure
the name is correct.

If some corner guests cannot run successfully, you can set the correct
nfs source as in '-i vmx' for those corver scenarios.

### DIFF
--- a/virttest/utils_v2v.py
+++ b/virttest/utils_v2v.py
@@ -7,6 +7,7 @@ Virt-v2v test utility functions.
 import os
 import re
 import time
+import glob
 import logging
 import random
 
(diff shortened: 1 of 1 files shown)


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.convert_from_file.positive_test.windows.input_mode.vmx.vmx_default.win2012r2.output_mode.libvirt: PASS (453.00 s)

xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.convert_from_file.positive_test.linux.input_mode.vmx.vmx_ssh.latest6.output_mode.rhev.rhv_upload: PASS (1351.33 s)

xiaodwan:
@chloerh @chunfuwen Please help to review this PR, thanks.

PR #2340, opened 2019-11-19
utils_config.py: Add missing opts for libvirtd.conf and qemu.conf

qiankehan:


qiankehan:
ping

PR #2351, opened 2019-11-25
Update pygtk requirement to GTK3 compatibility layer

pevogam:
This is the minimum needed for Python 3 compatibility and update
of the GTK API from 2 to 3 is the minimal needed for the GObject
introspection compatibility layer.

### DIFF
--- a/virttest/step_editor.py
+++ b/virttest/step_editor.py
@@ -12,17 +12,22 @@ import shutil
 import sys
 import logging
 
-import pygtk
+try:
+    from gi import pygtkcompat as pygtk
+except ImportError:
+    pygtk = None
+if pygtk is not None:
+    pygtk.enable()
+    pygtk.enable_gtk(version='3.0')
 import gtk
 
 from virttest import ppm_utils
 
-pygtk.require('2.0')
-
 
 # General utilities
 
-def corner_and_size_clipped(startpoint, endpoint, limits):
+def corner_and_size_clipped(startpoint, endpoint, width, height):
+    limits = width, height
     c0 = startpoint[:]
     c1 = endpoint[:]
     if c0[0] < 0:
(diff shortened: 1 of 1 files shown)


pevogam:
This PR is identical to https://github.com/autotest/tp-qemu/pull/1461 so it must be easier to merge based on these reviews.

pevogam:
Hi @luckyh, sorry for requesting from you again - this PR is identical to autotest/tp-qemu#1461 which you helped reviewing but for a duplicate file located in the virttest folder. Since you already did the other review and the code change is the same, I hope this should be a breeze.

luckyh:
@pevogam LGTM, but I have no ideas about how to test this since I've never used the script before. Would you please provide the test result? thanks.

pevogam:
@luckyh This can be tested if you run a `stepmaker` instead of `steps` test (have to modify the `type` parameter) and simply use the various functionality provided by the stepmaker tool. In any case, we keep using the stepmaker test/tool (i.e. this code) with the above patch now and haven't encountered any problems. The previous functionality is mostly deprecated by now anyway.

pevogam:
> @luckyh This can be tested if you run a `stepmaker` instead of `steps` test (have to modify the `type` parameter) and simply use the various functionality provided by the stepmaker tool. In any case, we keep using the stepmaker test/tool (i.e. this code) with the above patch now and haven't encountered any problems. The previous functionality is mostly deprecated by now anyway.

I meant to say that we have been testing it by using it for some months and now and this is the testing coverage I could provide at present. Also, as I mentioned above it is rarely used and we have done a similar update at another place that you also reviewed in the past. TL;DR: It must be good to go.

luckyh:
@pevogam sorry, just forgot this one, thanks for the reminder.

Let's move forward.

PR #2352, opened 2019-11-25
Handle preseed installations properly

pevogam:
The previous code considers only extra ISO with kickstart file,
bootable kickstart image, or HTTP server providing the kickstart
file, none of which are valid for Debian derivatives where the
initrd is remastered with the kickstart (here preseed) file.

### DIFF
--- a/virttest/tests/unattended_install.py
+++ b/virttest/tests/unattended_install.py
@@ -1080,7 +1080,7 @@ class UnattendedInstallConfig(object):
             raise ValueError("Unexpected installation method %s" %
                              self.medium)
 
-        if self.unattended_file:
+        if self.unattended_file and not self.unattended_file.endswith('.preseed'):
             if self.floppy or self.cdrom_unattended:
                 self.setup_boot_disk()
                 if self.params.get("store_boot_disk") == "yes":


pevogam:
@luckyh Do you know of any Ubuntu users among us that could verify this PR?

luckyh:
> @luckyh Do you know of any Ubuntu users among us that could verify this PR?

@pevogam seems that @balamuruhans and @sathnaga may have the tests running with Ubuntu.

pevogam:
> > @luckyh Do you know of any Ubuntu users among us that could verify this PR?
> 
> @pevogam seems that @balamuruhans and @sathnaga may have the tests running with Ubuntu.

Thanks a lot for your suggestion! It would be better to coordinate with other ubuntu testers so this is good to know.

pevogam:
Are there any other Ubuntu users we can think of? If there are none, than I guess this PR will hang forever :/

pevogam:
As @balamuruhans and @sathnaga are probably not using Ubuntu, is there a way for this PR to make it before the 82.0 LTS release? I am ready to pledge any hotfixes in case someone suddenly uses Ubuntu and complains of trouble in the last minute.

sathnaga (commented):
LGTM and adding a run log summary would help.

pevogam:
Hi @sathnaga 

> LGTM and adding a run log summary would help.

What do you mean by run log summary? A `job.log` of the run within the PR?

sathnaga:
@pevogam something like this would help, https://github.com/avocado-framework/avocado-vt/pull/2751#issuecomment-689594851

pevogam:
Hi @sathnaga, it seems that for the months from Nov 25, 2019 until now something else happened within the repo that already fixed the problem. While this PR doesn't seem to harm the execution in any way, it is thus rendered obsolete and adds some more complexity that is no longer required. I would therefore close it in favor of the fix that somebody else provided and got into the code base before this PR which took more than half a year at best.

PR #2364, opened 2019-12-04
Unattended install params

pevogam:
Provide vm specific parameters before performing any unattended vm setup

This change makes sure we are running vm-restricted configuration during unattended installation and should have no effect on any tests avoiding regex matching through ``?=`` or ``join`` operators in the Cartesian configs. It is however needed if some Cartesian parameters end up with a vm suffix like ``_vm1`` due to any preinstallation preprocessing.

### DIFF
--- a/virttest/tests/unattended_install.py
+++ b/virttest/tests/unattended_install.py
@@ -1259,6 +1259,9 @@ def run(test, params, env):
             ver_from = utils_test.get_image_version(qemu_image)
             utils_test.update_qcow2_image_version(qemu_image, ver_from, ver_to)
 
+    vm = env.get_vm(params["main_vm"])
+    # at this stage we need vm specific parameters
+    params = params.object_params(vm.name)
     src = params.get('images_good')
     vt_data_dir = data_dir.get_data_dir()
     base_dir = params.get("images_base_dir", vt_data_dir)
@@ -1275,7 +1278,6 @@ def run(test, params, env):
         funcatexit.register(env, params.get("type"), copy_file_from_nfs, src,
                             dst, mount_point, image_name)
 
-    vm = env.get_vm(params["main_vm"])
     local_dir = params.get("local_dir", os.path.abspath(vt_data_dir))
     local_dir = utils_misc.get_path(vt_data_dir, local_dir)
     for media in params.get("copy_to_local", "").split():


pevogam:
@luckyh: Since you are more aware of the join variants (not sure who else is) I hope you don't mind taking a look at this single-line change. Thanks!

virttest/tests/unattended_install.py
```diff
@@ -1296,6 +1296,8 @@ def copy_images():
         error_context.context(msg, logging.info)
         download.get_file(nfs_link, local_link)
         params[media] = local_link
+    # at this stage we need vm specific parameters
+    params = params.object_params(vm.name)
```
luckyh: Looks it is needed in the earlier stage, before line 1262, since almost all the params are per VM ones.

pevogam: Great, I only put it at a line where I was sure it will be needed. If it is needed earlier than I will move it right up to line 1262.

pevogam: I had to also move the code `vm = env.get_vm(params["main_vm"])` from line 1278 since it is needed in order to restrict the parameters.

luckyh: Yes, I think so, or you can have something like `params.object_params(params["main_vm"])` at there, thanks.

luckyh: And btw, it is handled well by the wrapper in tp-qemu for multiple VMs installation cases, but unfortunately the single VM one didn't take the benefit at that time.
https://github.com/autotest/tp-qemu/blob/master/generic/tests/unattended_install.py#L38

pevogam: > Yes, I think so, or you can have something like `params.object_params(params["main_vm"])` at there, thanks.

But I think what I suggested with the last push is a bit more natural. It is part of the usual workflow of obtaining a vm object and its parameters.

pevogam: > And btw, it is handled well by the wrapper in tp-qemu for multiple VMs installation cases, but unfortunately the single VM one didn't take the benefit at that time.
> https://github.com/autotest/tp-qemu/blob/master/generic/tests/unattended_install.py#L38

Indeed, but I guess the current change is a minimal one to achieve the same effect here :)

luckyh: > > Yes, I think so, or you can have something like `params.object_params(params["main_vm"])` at there, thanks.
> 
> But I think what I suggested with the last push is a bit more natural. It is part of the usual workflow of obtaining a vm object and its parameters.

ok, but looks you forgot to bring line 1278 to there, would you please check it?

pevogam: turns out the hunk was staged but not committed ^^ done now

pevogam:
@luckyh Do you know of somebody else who is aware of this area of code enough to recommend for reviewing?

luckyh:
> @luckyh Do you know of somebody else who is aware of this area of code enough to recommend for reviewing?

@pevogam maybe you can request @balamuruhans and @sathnaga for a view, thanks.

pevogam:
> > @luckyh Do you know of somebody else who is aware of this area of code enough to recommend for reviewing?
> 
> @pevogam maybe you can request @balamuruhans and @sathnaga for a view, thanks.

ok thanks for your recommendation!

pevogam:
> @pevogam ACK, can you add the details mentioned in description to commit msg ?
> Thanks for the patch!

Hi @balamuruhans, sorry for the slight delay - I have pushed with the PR details added to the commit if this is what you originally meant.

pevogam:
@luckyh @balamuruhans since we have two approvals I think we can merge, what do you think?

pevogam:
Thank you all!

PR #2371, opened 2019-12-10
utils_qemu: set allow_output_check='combined' in has_device_category

PaulYuuu:
The output type of some qemu help information may be different in
different distributions, some are `stdout`, some are `stderr`. Add
allow_output_check='combined' to check them together.

Bugzilla ID: 1781036


PaulYuuu:
@luckyh I forget that the output type of device help info might be stdout/stderr, so `re.findall` does not match correctly, fix it in this patch, please help to review it, thanks.

PR #2376, opened 2019-12-18
virtio_serial: Add virtio_serial extra params handle

nanliu-r:
ID: 1755682

Signed-off-by: nanliu <nanliu@redhat.com>

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -2018,7 +2018,8 @@ class DevContainer(object):
 
     def serials_define_by_variables(self, serial_id, serial_type, chardev_id,
                                     bus_type=None, serial_name=None,
-                                    bus=None, nr=None, reg=None):
+                                    bus=None, nr=None, reg=None,
+                                    bus_extra_params=None):
         """
         Creates related devices by variables
 
@@ -2030,10 +2031,14 @@ class DevContainer(object):
         :param bus: the busid of parent bus(optional, virtio only)
         :param nr: the nr(port) of the parent bus(optional, virtio only)
         :param reg: reg option of isa-serial(optional)
+        :param bus_extra_params: extra virtio_serial_pci params
         :return: the device list that construct the serial device
         """
 
         devices = []
+        if bus_extra_params:
+            bus_params = (dict([_.split('=')
+                                for _ in bus_extra_params.split(',') if _]))
         # For virtio devices, generate controller and create the port device
         if serial_type.startswith('virt'):
             if not bus:
@@ -2050,9 +2055,10 @@ class DevContainer(object):
                     bus = self.list_missing_named_buses(
                         _hba, 'SERIAL', bus + 1)[-1]
                     logging.debug("list missing named bus: %s", bus)
+                    bus_params["id"] = bus
                     devices.append(
                         qdevices.QDevice(bus_type,
-                                         {"id": bus},
+                                         bus_params,
                                          bus,
                                          pci_bus,
                                          qdevices.QSerialBus(
@@ -2109,7 +2115,8 @@ class DevContainer(object):
         serial_devices = self.serials_define_by_variables(
             serial_id, serial_type, chardev_id, bus_type=bus_type,
             serial_name=params.get("serial_name"), bus=params.get("serial_bus"),
-            nr=params.get("serial_nr"), reg=params.get("serial_reg"))
+            nr=params.get("serial_nr"), reg=params.get("serial_reg"),
+            bus_extra_params=params.get("virtio_serial_extra_params"))
 
         return [chardev_device] + serial_devices
 


virttest/qemu_devices/qcontainer.py
```diff
@@ -2018,7 +2018,8 @@ def images_define_by_params(self, name, image_params, media=None,
 
     def serials_define_by_variables(self, serial_id, serial_type, chardev_id,
                                     bus_type=None, serial_name=None,
-                                    bus=None, nr=None, reg=None):
+                                    bus=None, nr=None, reg=None,
+                                    pci_extra_params=None):
```
xutian: It looks this var should be bus_extra_params.

nanliu-r: Thanks and update.

yanan-fu:
LGTM

Let's wait for the tp-qemu one, and test them toghther.

virttest/qemu_devices/qcontainer.py
```diff
@@ -2030,10 +2031,17 @@ def serials_define_by_variables(self, serial_id, serial_type, chardev_id,
...
         :param nr: the nr(port) of the parent bus(optional, virtio only)
         :param reg: reg option of isa-serial(optional)
+        :param bus_extra_params: extra virtio_serial_pci params
         :return: the device list that construct the serial device
         """
 
         devices = []
+        bus_params = {}
+        if bus_extra_params:
+            bus_extra_params_ = (_.split('=', 1) for _ in
```
xutian: above code equal to below, let's update it to make it more readable.

bus_params = dict([_.split('=') for _ in bus_extra_params.split(',') if _])

virttest/qemu_vm.py
```diff
@@ -1777,6 +1777,11 @@ def sort_key(dev):
             serial_name = prefix + str(len(self.virtio_ports))\
                 if prefix else serial
             serial_params['serial_name'] = serial_name
+            if params.get("virtio_serial_extra_params") != '':
```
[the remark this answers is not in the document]
yanan-fu: @xutian   seems this is not the latest code,  the latest do not have this line.

virttest/qemu_devices/qcontainer.py
nanliu-r: thanks and update.

virttest/qemu_devices/qcontainer.py
```diff
@@ -2030,10 +2031,14 @@ def serials_define_by_variables(self, serial_id, serial_type, chardev_id,
...
         :param nr: the nr(port) of the parent bus(optional, virtio only)
         :param reg: reg option of isa-serial(optional)
+        :param bus_extra_params: extra virtio_serial_pci params
         :return: the device list that construct the serial device
         """
 
         devices = []
+        if bus_extra_params:
+            bus_params = (dict([_.split('=')
+                                for _ in bus_extra_params.split(',') if _]))
```
yanan-fu: An extra brackets although the output is same. It is up to you

yanan-fu:
@xutian  Do you have any new comments ?

virttest/qemu_devices/qcontainer.py
nanliu-r: To avoid the newline character "\" I think we should keep this.

xutian:
Merged, thanks.

luckyh:
@nanliu-r @yanan-fu @xutian hello everyone, please be noted that a regression is introduced by this patch: an exception would be raised due to `"local variable 'bus_params' referenced before assignment"` while using `serial_type = virt*` without setting `virtio_serial_extra_params`. Please help fix that, thanks.

And it is also interesting that the CI check didn't catch that, so we should be careful about this kind of issue when reviewing other PRs.

yanan-fu:
@luckyh `bus_params["id"]` is must, so it do not have the problem you mentioned as my understanding!

luckyh:
> @luckyh `bus_params["id"]` is must, so it do not have the problem you mentioned as my understanding!

You can reproduce it with the following change:
```patch
diff --git a/shared/cfg/base.cfg b/shared/cfg/base.cfg
index f946318d..96ff2fde 100644
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -160,7 +160,8 @@ q35, arm64-pci, arm64-mmio:
 # Then the serial_name of "vs1" will be "com.redhat.spice.0"
 
 serials = "serial0"
-serial_type = "isa-serial"
+#serial_type = "isa-serial"
+serial_type = "virtio-serial"
 
 # Emulated machine type, run following command to see supported machine type.
 # qemu-kvm -M ?
```

```
Original Traceback (most recent call last):
  File "/home/kar/workspace/avocado-vt/virttest/qemu_vm.py", line 2850, in create
  File "/home/kar/workspace/avocado-vt/virttest/qemu_vm.py", line 1778, in make_create_command
    serial, serial_params, serial_filename)
  File "/home/kar/workspace/avocado-vt/virttest/qemu_devices/qcontainer.py", line 2263, in serials_define_by_params
    bus_extra_params=params.get("virtio_serial_extra_params"))
  File "/home/kar/workspace/avocado-vt/virttest/qemu_devices/qcontainer.py", line 2202, in serials_define_by_variables
    bus_params["id"] = bus
UnboundLocalError: local variable 'bus_params' referenced before assignment
```

And that's why we should test more, thanks.

yanan-fu:
@luckyh  yes, you are right, will fix in with new patch.

xutian:
@yanan-fu  and @luckyh  sent a hotfix, https://github.com/avocado-framework/avocado-vt/pull/2421

would you help to review it.

PR #2396, opened 2020-01-10
libvirt_vm.py: New methods for device attach and detach

qiankehan:
Add methods attach_device() and detach_device() for libvirt_vm object.
The attribute 'ensure_boot' could be used to make sure hotplug or
hotunplug happens when vm has booted.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2137,6 +2137,46 @@ class VM(virt_vm.BaseVM):
                                       ignore_status=ignore_status,
                                       debug=debug)
 
+    def attach_device(self, xml_file, options="", ensure_boot=False, debug=False):
+        """
+        Attach a device to VM.
+
+        :param xml_file: The device xml for attach.
+        :param options: The options for device attach, which could be '--live',
+                '--config', '--current', '--persistent' or any combinations.
+                Default value is "" for live hotplug.
+        :param ensure_boot: Make sure the device attach happens after vm has
+                booted. That is necessary when checking the function of
+                hotplugged device.
+        :param debug: Open or close debug information.
+        """
+        if (ensure_boot):
+            if (not self.is_alive()):
+                self.start()
+            self.wait_for_login()
+        return virsh.attach_device(self.name, filearg=xml_file,
+                                   flagstr=options, debug=debug)
+
+    def detach_device(self, xml_file, options="", ensure_boot=False, debug=True):
+        """
+        Detach a device from VM.
+
+        :param: xml_file: The device xml for detach.
+        :param: options: The options for device detach, which could be '--live'
+                '--config', '--current', '--persistent' or any combinations.
+                Default value is "" for live hotunplug.
+        :param ensure_boot: Make sure the device detach happens after vm has
+                booted. That is necessary when checking the function of
+                hotunplugged device.
+        :param debug: Open or close debug information.
+        """
+        if (ensure_boot):
+            if (not self.is_alive()):
+                self.start()
+            self.wait_for_login()
+        return virsh.detach_device(self.name, filearg=xml_file,
+                                   flagstr=options, debug=debug)
+
     def destroy(self, gracefully=True, free_mac_addresses=True):
         """
         Destroy the VM.


virttest/libvirt_vm.py
```diff
@@ -2137,6 +2137,46 @@ def detach_interface(self, option="", ignore_status=False,
...
+        :param xml_file: The device xml for attach.
+        :param options: The options for device attach, which could be '--live',
+                '--config', '--current', '--persistent' or any combinations.
+                Default value is "" for live hotplug.
+        :param ensure_boot: Make sure the device attach happens after vm has
+                booted. That is necessary when checking the function of
+                hotplugged device.
+        :param debug: Open or close debug information.
+        """
+        if (ensure_boot):
```
smitterl: You should extract this part in order to avoid code duplication, something like
```python
def _ensure_booted(ensure_booted):
    if ensure_booted:
        ...
    self.wait_for_login()
```
```python
def attach_device(...):
    ...
    self._ensure_booted(ensure_booted)
    return virsh.attach_device(...)

def detach_device(...):
    ...
    self._ensure_booted(ensure_booted)
    return virsh.detach_device(...)
```

virttest/libvirt_vm.py
```diff
@@ -2137,6 +2137,46 @@ def detach_interface(self, option="", ignore_status=False,
                                       ignore_status=ignore_status,
                                       debug=debug)
 
+    def attach_device(self, xml_file, options="", ensure_boot=False, debug=False):
```
smitterl: I'm wondering if it'd be better to use `ensure_guest_booted` (same for `detach_device`).

virttest/libvirt_vm.py
```diff
@@ -2137,6 +2137,46 @@ def detach_interface(self, option="", ignore_status=False,
...
+                                   flagstr=options, debug=debug)
+
+    def detach_device(self, xml_file, options="", ensure_boot=False, debug=True):
+        """
+        Detach a device from VM.
+
+        :param: xml_file: The device xml for detach.
+        :param: options: The options for device detach, which could be '--live'
+                '--config', '--current', '--persistent' or any combinations.
+                Default value is "" for live hotunplug.
```
smitterl: s. above

virttest/libvirt_vm.py
smitterl: We don't need the parenthesis in python.

virttest/libvirt_vm.py
```diff
@@ -2137,6 +2137,46 @@ def detach_interface(self, option="", ignore_status=False,
...
+        :param ensure_boot: Make sure the device detach happens after vm has
+                booted. That is necessary when checking the function of
+                hotunplugged device.
+        :param debug: Open or close debug information.
+        """
+        if (ensure_boot):
```
smitterl: s. above

qiankehan:
> Thank you for your contribution @qiankehan
> 
> IMHO, it's useful to generalize even more by supporting several vm states, like `shutoff`, `running`, `paused` which we find often, e.g. https://github.com/autotest/tp-libvirt/blob/97a5f76530ee206e92a555633cfbd6693d0cdc31/libvirt/tests/src/virsh_cmd/domain/virsh_migrate_setmaxdowntime.py#L203 and we should not add new methods for the very specific case here.
> 
> What do you think if there was a method `set_vm_state(state)` that would help set preconditions on vm anywhere?

Sorry for late reply. I agree that the method like set_vm_state(state) would be more useful than what is in this PR.
I am sure if any enum type defined for vm state in avocado-vt. Should I define that myself?
As I know the vm state enumation has been defined in **libvirt-python**:
>     52     VIR_DOMAIN_NOSTATE = 0,     /* no state */                                  
    53     VIR_DOMAIN_RUNNING = 1,     /* the domain is running */                     
    54     VIR_DOMAIN_BLOCKED = 2,     /* the domain is blocked on resource */         
    55     VIR_DOMAIN_PAUSED  = 3,     /* the domain is paused by user */              
    56     VIR_DOMAIN_SHUTDOWN= 4,     /* the domain is being shut down */             
    57     VIR_DOMAIN_SHUTOFF = 5,     /* the domain is shut off */                    
    58     VIR_DOMAIN_CRASHED = 6,     /* the domain is crashed */                     
    59     VIR_DOMAIN_PMSUSPENDED = 7, /* the domain is suspended by guest             
    60                                    power management */      

However, the **libvirt-python** is not integrated in avocado-vt...

> What I mean is instead of consuming this code
> 
> ```python
> def run(...):
>   ...
>   vm.attach_device(xml_file, options, ensure_boot=True, debug)
> ```
> 
> you'd write
> 
> ```python
> def run(...):
>   ...
>   vm.set_state(state.BOOTED)
>   vm.attach_device(xml_file, options, debug)
> ```

smitterl:
...
> Sorry for late reply. I agree that the method like set_vm_state(state) would be more useful than what is in this PR.
> I am sure if any enum type defined for vm state in avocado-vt. Should I define that myself?
> As I know the vm state enumation has been defined in **libvirt-python**:
> 
> > ```
> > 52     VIR_DOMAIN_NOSTATE = 0,     /* no state */                                  
...
> ```
> 
> However, the **libvirt-python** is not integrated in avocado-vt...

I agree with you that we should not add dependencies but instead you could create your own constants.

I would want to consider the following

1. constant values: you could use the string representations (e.g. "running") instead of actual enum.Enum; by doing so, they would be compatible with the `VM.state()` and related interfaces in `virttest/libvirt_vm.py`; this seems to be used throughout the code and can then also be added to log messages more easily than numbers; then you augment those values by your own like e.g. `BOOTED = "booted"`
2. location: you could add the state constants and the new method directly to the `libvirt_vm.py`, i.e. something like
```python
# libvirt_vm.py
class states():
  RUNNING = "running"
  BOOTED = "booted"
...

class VM(virt_vm.BaseVM):
...
  def set_state(state=None):
    if state == states.RUNNING:
      if self.is_dead():
        ..
    elif state == states.SHUTDOWN:
      ..
```
```python
# some_test.py
from virttest.libvirt_vm import states
...
def run(...):
  ...
  vm = env.get_vm(vm_name)
  vm.set_state(states.SHUTDOWN)
```
or even
```python
# some_test.py
from virttest.libvirt_vm import states
...
def run(...):
  ...
  vm = env.get_vm(vm_name)
  pre_vm_state = params.get("pre_vm_state", states.SHUTDOWN)
  vm.set_state(pre_vm_state)
```

qiankehan:
During the discussion I find what I really need it not only a arguments to make sure the the VM's OS is booted before device live attach/detach, but also a function to switch the VM states to reuse the code in libvirt.
So I send another PR to realize that: https://github.com/avocado-framework/avocado-vt/pull/2815
@smitterl Thanks for your review. And sorry for late update.

PR #2405, opened 2020-01-16
Enhance lvm support

lolyu:
Aims to provide LVM support for qemu-img related tests.
Usage:
```
pv_name = /dev/sdb
vg_name = autotest
images = image1 stg1
image_name_stg1 = stage1
image_format_stg1 = qcow2
image_size_stg1 = 20G
storage_type = lvm
enable_lvm_stg1 = yes
```

limit:
LVM initialization is limited to images listed in param `images`.

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -32,6 +32,7 @@ from virttest import ppm_utils
 from virttest import test_setup
 from virttest import virt_vm
 from virttest import utils_misc
+from virttest import lvm
 from virttest import cpu
 from virttest import storage
 from virttest import utils_libguestfs
@@ -908,10 +909,17 @@ def preprocess(test, params, env):
         params["image_raw_device"] = "yes"
 
     if params.get("storage_type") == "lvm":
-        lvmdev = qemu_storage.LVMdev(params, base_dir, "lvm")
-        params["image_name"] = lvmdev.setup()
-        params["image_raw_device"] = "yes"
-        env.register_lvmdev("lvm_%s" % params["main_vm"], lvmdev)
+        for vm in params.objects("vms"):
+            vm_params = params.object_params(vm)
+            for image in vm_params.objects("images"):
+                image_params = params.object_params(image)
+                if image_params.get("enable_lvm", "no") == "no":
+                    continue
+                image_params["lv_name"] = lvm.get_lv_name(
+                    image_params["image_name"], image_params["image_format"])
+                lvmdev = qemu_storage.LVMdev(image_params, base_dir, image)
+                lvmdev.setup()
+                env.register_lvmdev("lvm_%s" % image, lvmdev)
 
     if params.get("storage_type") == "nfs":
         selinux_local = params.get('set_sebool_local', 'yes') == "yes"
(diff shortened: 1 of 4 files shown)


lolyu:
@luckyh please have a look.

lolyu:
```
 (1/3) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.unattended_install.cdrom.extra_cdrom_ks.default_install.aio_threads: PASS (1579.88 s)
```

virttest/env_process.py
```diff
@@ -908,10 +911,16 @@ def preprocess(test, params, env):
         params["image_raw_device"] = "yes"
 
     if params.get("storage_type") == "lvm":
-        lvmdev = qemu_storage.LVMdev(params, base_dir, "lvm")
-        params["image_name"] = lvmdev.setup()
-        params["image_raw_device"] = "yes"
-        env.register_lvmdev("lvm_%s" % params["main_vm"], lvmdev)
+        for image in params.objects("images"):
```
luckyh: Since `images` is a per VM param, here it should iterate each VM as well in order to prepare every lvm volumes requested by this test.

virttest/env_process.py
```diff
@@ -908,10 +911,16 @@ def preprocess(test, params, env):
...
+            image_params = params.object_params(image)
+            if image_params.get("enable_lvm", "no") == "no":
+                continue
+            image_params["lv_name"] = lvm.get_lv_name(
+                image_params["image_name"], image_params["image_format"])
+            lvmdev = qemu_storage.LVMdev(image_params, base_dir, image)
```
luckyh: If it represents the entity of lvm pools, it should be initialized outside the loop, otherwise pools might be configured twice and that would result in errors.

The correct logic would be something like this:
```python
if params.get("storage_type") == "lvm":
    lvm_pools_initialize()

    ...

    for vm in vms:
        for image in vm.images:
            image_params = params.object_params(image)
            if image_params.get("enable_lvm", "no") == "no":
                continue
            # request volume
            vol = vol_lookup_by_image_params(image_params)
            ...
```

virttest/env_process.py
```diff
@@ -1636,13 +1645,16 @@ def postprocess(test, params, env):
...
 
     if params.get("storage_type") == "lvm":
-        try:
-            lvmdev = env.get_lvmdev("lvm_%s" % params["main_vm"])
-            lvmdev.cleanup()
-        except Exception as details:
-            err += "\nLVM cleanup: %s" % str(details).replace('\\n', '\n  ')
-            logging.error(details)
-        env.unregister_lvmdev("lvm_%s" % params["main_vm"])
+        for image in params.objects("lvm"):
```
luckyh: Is it a typo, or a newly defined param?

lolyu:
@luckyh, updated and test the scenario that one `vg` is created with multiple `pv`s.
```
 (1/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.unattended_install.cdrom.extra_cdrom_ks.default_install.aio_threads: PASS (1534.78 s)
 (2/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.boot: PASS (36.49 s)
```

lolyu:
@luckyh, please help to review, thanks.

virttest/env_process.py
```diff
@@ -1642,13 +1650,18 @@ def postprocess(test, params, env):
...
-        try:
-            lvmdev = env.get_lvmdev("lvm_%s" % params["main_vm"])
-            lvmdev.cleanup()
-        except Exception as details:
-            err += "\nLVM cleanup: %s" % str(details).replace('\\n', '\n  ')
-            logging.error(details)
-        env.unregister_lvmdev("lvm_%s" % params["main_vm"])
+        for vm in params.objects("vms"):
+            vm_params = params.object_params(vm)
+            for image in vm_params.objects("images"):
```
luckyh: Without the abstraction of the pool, it is hard to make the clean-up procedure right, let's consider the following situation:

* there are more than one image (e.g. stg1 stg2)
* vg is dynamically created (requested by stg1)
* one of the image would not be removed after test (`remove_image = no`)

with the current implementation, it would remove the vg at here and then get failure.

lolyu: if `remove_image = no`, this cleanup process won't remove the `vg` created since the real parameter to control this behavior is `force_remove_image`, which is default `no`.

luckyh: Sorry, the example may not be correct (I mean the configuration part in the above comment), but the situation does exist, it is possible to construct a configuration which can trigger the failure I mentioned.

luckyh: From my perspective, supporting host devices is the most important point for the entire requirement: we have to ensure that images can be created, removed, backed up, restored ... with using host's blocks. Then the management layer of LVM could be had as an additional functionality.

lolyu: I've drafted out the `supporting host lvm device` functionality in this [path] (https://github.com/avocado-framework/avocado-vt/pull/2448) without lvm management, please take a look.

PR #2409, opened 2020-01-17
put path for tty chardev

michiboo:
ID: 1791585
To create path automatically for tty backend for chardev

Signed-off-by: Micky Yun Chan <michan@redhat.com>

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -2243,6 +2243,7 @@ class DevContainer(object):
         _backend = 'socket' if 'socket' in backend else backend
         # Generate -chardev device
         chardev_param = Params({'backend': _backend, 'id': chardev_id})
+        file_name = '/dev/ttyS0' if backend == 'tty' else file_name
         if backend in ['unix_socket', 'file', 'pipe', 'serial',
                        'tty', 'parallel', 'parport']:
             chardev_param.update({'path': file_name})


yanan-fu:
@michiboo  wrong "id", please add the correct one.

virttest/qemu_devices/qcontainer.py
```diff
@@ -2243,6 +2243,7 @@ def chardev_define_by_params(self, chardev_id, params, file_name=None):
         _backend = 'socket' if 'socket' in backend else backend
         # Generate -chardev device
         chardev_param = Params({'backend': _backend, 'id': chardev_id})
+        file_name = '/dev/ttyS0' if backend == 'tty' else file_name
```
yanan-fu: `ttyS0` is only for x86 platform. @nanliu-r please confirm if this case need cover other platform, thanks!

nanliu-r: > `ttyS0` is only for x86 platform. @nanliu-r please confirm if this case need cover other platform, thanks!

yes, this case is "RHEL and x86 only"

yanan-fu: I just checked with ppc platform, it support tty too although we do not test it:
```
-chardev tty,id=id,path=path[,mux=on|off][,logfile=PATH][,logappend=on|off]
```

yanan-fu:
`file_name` as a  parameter of `chardev_define_by_params` ,it can be defined when call `chardev_define_by_params`.  So i think it is better to update:
https://github.com/avocado-framework/avocado-vt/blob/ea2eebf6764dfd03e4a239d94c149608521b757a/virttest/qemu_vm.py#L1759

PR #2413, opened 2020-01-28
Allow value replacement in joint variants

pevogam:
Joint variants need flattening and dropping of indices in order
to detect values that should be substituted if a previous parameter
of the given key exists.

### DIFF
--- a/virttest/cartesian_config.py
+++ b/virttest/cartesian_config.py
@@ -140,7 +140,7 @@ import re
 import sys
 
 _reserved_keys = set(("name", "shortname", "dep", "_short_name_map_file", "_name_map_file"))
-
+options = None
 num_failed_cases = 5
 
 
@@ -544,6 +544,13 @@ def _drop_suffixes(d):
 
     This step returns a copy of a suffix flattened dictionary.
     """
+    if options is not None:
+        # This file was invoked through cmdline
+        skipdups = options.skipdups
+    else:
+        # This file was invoked as Python module
+        skipdups = True
+
     # dictionary `d_flat' is going to be the mutated copy of `d`
     d_flat = d.copy()
     for key in d:
@@ -552,14 +559,6 @@ def _drop_suffixes(d):
 
         if not isinstance(key, tuple):
             continue
-
-        try:
-            # This file was invoked through cmdline
-            options.skipdups
-            skipdups = options.skipdups
-        except NameError:
-            # This file was invoked as Python module
-            skipdups = True
 
         if skipdups:
             # Drop vars with suffixes matches general var val
--- a/virttest/cartesian_config.py
+++ b/virttest/cartesian_config.py
@@ -538,9 +538,63 @@ class Node(object):
 match_substitute = re.compile("\$\{(.+?)\}")
 
 
+def _drop_suffixes(d):
+    """
+    Merge suffixes for same var, or drop off unnecessary suffixes
+
+    This step returns a copy of a suffix flattened dictionary.
+    """
+    # dictionary `d_flat' is going to be the mutated copy of `d`
+    d_flat = d.copy()
+    for key in d:
+        if key in _reserved_keys:
+            continue
+
+        if not isinstance(key, tuple):
+            continue
+
+        try:
+            # This file was invoked through cmdline
+            options.skipdups
+            skipdups = options.skipdups
+        except NameError:
+            # This file was invoked as Python module
+            skipdups = True
+
+        if skipdups:
+            # Drop vars with suffixes matches general var val
+            # Example: if a_x == 1, and a == 1. Drop: a_x, leave a
+            gen_var_name = key[0]
+            if gen_var_name in d and d[gen_var_name] == d[key]:
+                # Drop gen_var_name, use general key with same value
+                d_flat.pop(key)
+                continue
+
+            can_drop_all_suffixes_for_this_key = True
+            for k in d:
+                gen_name = k[0] if isinstance(k, tuple) else k
+                if gen_var_name == gen_name:
+                    if d[key] != d[k]:
+                        can_drop_all_suffixes_for_this_key = False
+                        break
+
+        if skipdups and can_drop_all_suffixes_for_this_key:
+            new_key = key[0]
+        else:
+            # merge suffixes, preserve reverse order of suffixes
+            new_key = key[:1] + key[1:][::-1]
+            new_key = ''.join((map(str, new_key)))
+        d_flat[new_key] = d_flat.pop(key)
+
+    return d_flat
+
+
 def _substitution(value, d):
     """
     Only optimization string Template substitute is quite expensive operation.
+
+    .. todo:: The current substitution is limited to simple cases when it comes
+        to using join and suffix operators.
 
     :param value: String where could be $string for substitution.
     :param d: Dictionary from which should be value substituted to value.
@@ -553,7 +607,8 @@ def _substitution(value, d):
         try:
             match = match_substitute.search(value, start)
             while match:
-                val = eval(match.group(1), None, d)
+                d_flat = _drop_suffixes(d)
+                val = eval(match.group(1), None, d_flat)
                 st += value[start:match.start()] + str(val)
                 start = match.end()
                 match = match_substitute.search(value, start)
@@ -1856,55 +1911,6 @@ class Parser(object):
                                          lexer.line))
             raise
 
-    def drop_suffixes(self, d):
-        """
-        Merge suffixes for same var, or drop off unnecessary suffixes
-
-        This step can be done safely only by the top level generator, before
-        outputting the dictionary to outside world.
-        """
-        # dictionary `d' is going to change, keep its original copy
-        d_orig = d.copy()
-        for key in d_orig:
-            if key in _reserved_keys:
-                continue
-
-            if not isinstance(key, tuple):
-                continue
-
-            try:
-                # This file was invoked through cmdline
-                options.skipdups
-                skipdups = options.skipdups
-            except NameError:
-                # This file was invoked as Python module
-                skipdups = True
-
-            if skipdups:
-                # Drop vars with suffixes matches general var val
-                # Example: if a_x == 1, and a == 1. Drop: a_x, leave a
-                gen_var_name = key[0]
-                if gen_var_name in d_orig and d_orig[gen_var_name] == d_orig[key]:
-                    # Drop gen_var_name, use general key with same value
-                    d.pop(key)
-                    continue
-
-                can_drop_all_suffixes_for_this_key = True
-                for k in d_orig:
-                    gen_name = k[0] if isinstance(k, tuple) else k
-                    if gen_var_name == gen_name:
-                        if d_orig[key] != d_orig[k]:
-                            can_drop_all_suffixes_for_this_key = False
-                            break
-
-            if skipdups and can_drop_all_suffixes_for_this_key:
-                new_key = key[0]
-            else:
-                # merge suffixes, preserve reverse order of suffixes
-                new_key = key[:1] + key[1:][::-1]
-                new_key = ''.join((map(str, new_key)))
-            d[new_key] = d.pop(key)
-
     def get_dicts(self, node=None, ctx=[], content=[], shortname=[], dep=[]):
         """
         Process 'join' entry, unpack join filter for node.
@@ -1957,9 +1963,7 @@ class Parser(object):
         if not joins:
             # Return generator
             for d in self.get_dicts_plain(node, ctx, content, shortname, dep):
-                if parent:
-                    self.drop_suffixes(d)
-                yield d
+                yield _drop_suffixes(d) if parent else d
         else:
             # Rewrite all separate joins in one node as many `only'
             onlys = []
@@ -1972,9 +1976,7 @@ class Parser(object):
             old_content = node.content[:]
             node.content = new_content
             for d in self.multiply_join(onlys, node, ctx, content, shortname, dep):
-                if parent:
-                    self.drop_suffixes(d)
-                yield d
+                yield _drop_suffixes(d) if parent else d
             node.content = old_content[:]
 
     def mk_name(self, n1, n2):
--- a/selftests/unit/test_cartesian_config.py
+++ b/selftests/unit/test_cartesian_config.py
@@ -826,6 +826,121 @@ class CartesianConfigTest(unittest.TestCase):
                              cartesian_config.Label("aaa")]]],
                           "Failed to parse filter.")
 
+    def testJoinSubstitution(self):
+        self._checkStringDump("""
+            key0 = "Baz"
+            variants:
+                - one:
+                    key1 = "Hello"
+                    key2 = "Foo"
+
+                    test01 = "${key1}"
+                    # the following substitutions are still not supported
+                    #test02 = "${key1_v1}"
+                    #test03 = "${key1_v2}"
+
+                    suffix _v1
+                - two:
+                    key1 = "Bye"
+                    key3 = "Bar"
+
+                    test04 = "${key1}"
+                    # the following substitutions are still not supported
+                    #test05 = "${key1_v1}"
+                    #test06 = "${key1_v2}"
+
+                    suffix _v2
+            variants:
+                - alpha:
+                    # the following substitutions are still not supported
+                    #test07 = "${key1}"
+                    #test08 = "${key1_v1}"
+                    #test09 = "${key1_v2}"
+                    #test10 = "${key2}"
+                    #test11 = "${key3}"
+
+                    key1 = "Alpha"
+                    test12 = "${key1}"
+
+                    join one two
+                - beta:
+                    # the following substitutions are still not supported
+                    #test13 = "${key1}"
+                    #test14 = "${key1_v1}"
+                    #test15 = "${key1_v2}"
+                    #test16 = "${key2}"
+                    #test17 = "${key3}"
+
+                    join one two
+
+            test100 = "${key0}"
+            # the following substitutions are still not supported
+            #test18 = "${key1}"
+            #test19 = "${key1_v1}"
+            #test20 = "${key1_v2}"
+            #test21 = "${key2}"
+            #test22 = "${key3}"
+            """,
+                              [
+                                  {'_name_map_file': {'<string>': 'alpha.two'},
+                                   '_short_name_map_file': {'<string>': 'alpha.two'},
+                                   'dep': [],
+                                   'key0': 'Baz',
+                                   'key1': 'Alpha',
+                                   'key1_v1': 'Hello',
+                                   'key1_v2': 'Bye',
+                                   'key2': 'Foo',
+                                   'key3': 'Bar',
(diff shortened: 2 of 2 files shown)


pevogam:
Hi @luckyh, this PR was staying in draft state for some time so that I could test its effects. I think it is ready for review now and since you are one of the few people reviewing Cartesian config internals I will be happy to hear your opinion.

Before this PR, if one uses the `join` operator they will no longer be able to substitute parameter values like e.g. `key1="${prefix}val` which is quite a severe limitation. This PR both decouples a method which does not use any instance data and performs the necessary steps to be able to match and substitute the values of joined suffixed parameters. There could only be a slight performance loss in the case of join operator using configs and it is minor for making value substitution possible again for these cases.

Thanks!

luckyh:
> Before this PR, if one uses the `join` operator they will no longer be able to substitute parameter values like e.g. `key1="${prefix}val"` which is quite a severe limitation. This PR both decouples a method which does not use any instance data and performs the necessary steps to be able to match and substitute the values of joined suffixed parameters. There could only be a slight performance loss in the case of join operator using configs and it is minor for making value substitution possible again for these cases.

Hi @pevogam, I tried to understand this proposal with using the following example:
```
variants:
    - one:
        key1 = "Hello"
	key2 = "Foo"

	test01 = "${key1}"
	test02 = "${key1_v1}"
	test03 = "${key1_v2}"

        suffix _v1
    - two:
        key1 = "Bye"
	key3 = "Bar"

	test04 = "${key1}"
	test05 = "${key1_v1}"
	test06 = "${key1_v2}"

        suffix _v2
variants:
    - alpha:
        test07 = "${key1}"
	test08 = "${key1_v1}"
	test09 = "${key1_v2}"
	test10 = "${key2}"
	test11 = "${key3}"

	key1 = "Alpha"

	test12 = "${key1}"

	join one two
    - beta:
	test13 = "${key1}"
	test14 = "${key1_v1}"
	test15 = "${key1_v2}"
	test16 = "${key2}"
	test17 = "${key3}"

	join one two

test18 = "${key1}"
test19 = "${key1_v1}"
test20 = "${key1_v2}"
test21 = "${key2}"
test22 = "${key3}"
```

with this patch the outcome is:
```
dict    1:  alpha.one.two
    _name_map_file = {'e.cfg': 'alpha.two'}
    _short_name_map_file = {'e.cfg': 'alpha.two'}
    dep = []
    key1 = Alpha
    key1_v1 = Hello
    key1_v2 = Bye
    key2 = Foo
    key3 = Bar
    name = alpha.one.two
    shortname = alpha.one.two
    test01 = Hello
    test02 = ${key1_v1}
    test03 = ${key1_v2}
    test04 = Bye
    test05 = ${key1_v1}
    test06 = ${key1_v2}
    test07 = Bye
    test08 = ${key1_v1}
    test09 = ${key1_v2}
    test10 = ${key2}
    test11 = Bar
    test12 = Alpha
    test18 = Alpha
    test19 = ${key1_v1}
    test20 = Bye
    test21 = ${key2}
    test22 = Bar
dict    2:  beta.one.two
    _name_map_file = {'e.cfg': 'beta.two'}
    _short_name_map_file = {'e.cfg': 'beta.two'}
    dep = []
    key1_v1 = Hello
    key1_v2 = Bye
    key2 = Foo
    key3 = Bar
    name = beta.one.two
    shortname = beta.one.two
    test01 = Hello
    test02 = ${key1_v1}
    test03 = ${key1_v2}
    test04 = Bye
    test05 = ${key1_v1}
    test06 = ${key1_v2}
    test13 = Bye
    test14 = ${key1_v1}
    test15 = ${key1_v2}
    test16 = ${key2}
    test17 = Bar
    test18 = Bye
    test19 = ${key1_v1}
    test20 = ${key1_v2}
    test21 = ${key2}
    test22 = Bar
```
As you can see, all the tests of substitutions to `key2` are failed, so it looks like to be an issue, right?

And, except the `test1`, `test12` and the `test18` the rest tests of substitutions to `key1` performed the replacement to the value `Bye` which is the one assigned by the variant `two`. I have checked and found that it was affected by the `join` syntax, the outcome could be changed by reordering `one` and `two`. I'm not sure whether it makes sense, but it's better to have the doc accordingly.

Btw, would you like to build the unit tests?

pevogam:
Hi @luckyh, if I understand the origin of your comparison correctly, the difference between the produced outputs of the master branch and this PR is:
```diff
--- before.cfg	2020-09-09 18:53:26.109006623 +0300
+++ after.cfg	2020-09-09 18:56:32.176935761 +0300
@@ -15,17 +15,17 @@
     test04 = Bye
     test05 = ${key1_v1}
     test06 = ${key1_v2}
-    test07 = ${key1}
+    test07 = Bye
     test08 = ${key1_v1}
     test09 = ${key1_v2}
     test10 = ${key2}
-    test11 = ${key3}
+    test11 = Bar
     test12 = Alpha
     test18 = Alpha
     test19 = ${key1_v1}
-    test20 = ${key1_v2}
+    test20 = Bye
     test21 = ${key2}
-    test22 = ${key3}
+    test22 = Bar
 dict    2:  beta.one.two
     _name_map_file = {'t.py': 'beta.two'}
     _short_name_map_file = {'t.py': 'beta.two'}
@@ -42,13 +42,13 @@
     test04 = Bye
     test05 = ${key1_v1}
     test06 = ${key1_v2}
-    test13 = ${key1}
+    test13 = Bye
     test14 = ${key1_v1}
     test15 = ${key1_v2}
     test16 = ${key2}
-    test17 = ${key3}
-    test18 = ${key1}
+    test17 = Bar
+    test18 = Bye
     test19 = ${key1_v1}
     test20 = ${key1_v2}
     test21 = ${key2}
-    test22 = ${key3}
+    test22 = Bar
```
Thus I guess what you mean to say is that even though some values were replaced, this is by far not all of them.

The truth is that this solution is pretty limited and currently replacements like `${key_suffix}` are not at all supported. The same is true for values defined during the join variants where as you can see the order of parsing of pre-dicts can shadow some initial values and expose others.

The use case that I have in mind is far simpler and still solved by this PR and could rather be represented by the config:
```
key0 = "Baz"

variants:
    - one:
        key1 = "Hello"
	key2 = "Foo"
        suffix _v1
    - two:
        key1 = "Bye"
	key3 = "Bar"
        suffix _v2
variants:
    - alpha:
	key1 = "Alpha"
	join one two
    - beta:
	join one two

test = "${key0}"
```
where the output diff between the master and this PR is
```diff
--- before.cfg	2020-09-09 19:28:47.890078000 +0300
+++ after.cfg	2020-09-09 19:28:21.630078000 +0300
@@ -10,7 +10,7 @@
     key3 = Bar
     name = alpha.one.two
     shortname = alpha.one.two
-    test = ${key0}
+    test = Baz
 dict    2:  beta.one.two
     _name_map_file = {'s.cfg': 'beta.two'}
     _short_name_map_file = {'s.cfg': 'beta.two'}
@@ -22,4 +22,4 @@
     key3 = Bar
     name = beta.one.two
     shortname = beta.one.two
-    test = ${key0}
+    test = Baz
```
As you can see, even if the user relies on a simple base configuration that they would like to use substitutions with, this is currently not possible on master.

So I see two possibilities: I could add this as the unit test to accompany the PR (1) or I could add your unit tests with my addition and lots of TODO comments about the additional cases you tried. In any case, fixing more elaborate cases like the ones you showed in your excellent observations could be something to note for future work and I currently only have the time resources for the current proposal which still improves on the baseline case.

pevogam:
Hi @luckyh, what do you think is the chance for this to be merged before the 82.0 LTS release? I could add unit tests for an hour or two if you are willing to propose the way to go from the two options above. I am asking because to some extend the release of our plugin depends on this patch and our 82.0 LTS release will directly depend on the Avocado VT 82.0 LTS release and cannot require patches on top of it.

luckyh:
> So I see two possibilities: I could add this as the unit test to accompany the PR (1) or ...

@pevogam It's fine to me to have the **limited** support like this, but better to highlight that in either the commit log or the [doc](https://avocado-vt.readthedocs.io/en/latest/CartesianConfig.html#value-substitutions), and the unit test would be helpful, thanks.

virttest/cartesian_config.py
```diff
@@ -1856,55 +1908,6 @@ def _parse(self, lexer, node=None, prev_indent=-1):
                                          lexer.line))
             raise
 
-    def drop_suffixes(self, d):
```
luckyh, line 1859: Not sure if this public(?) method is no longer used, maybe you could confirm with the author @Andrei-Stepanov

luckyh:
> Hi @luckyh, what do you think is the chance for this to be merged before the 82.0 LTS release?

@pevogam may be possible if can get the second ACK in time, anyway, it's better to let @clebergnu be aware of your willing :)

virttest/cartesian_config.py
pevogam: Let's see what he has to say but considering the very low usage of this code and the fact that getting a feedback for a simple thing might take months, I would consider hour long silence on his side as acknowledgement.

pevogam:
> > So I see two possibilities: I could add this as the unit test to accompany the PR (1) or ...
> 
> @pevogam It's fine to me to have the **limited** support like this, but better to highlight that in either the commit log or the [doc](https://avocado-vt.readthedocs.io/en/latest/CartesianConfig.html#value-substitutions), and the unit test would be helpful, thanks.

Both executed in the latest push.

> > Hi @luckyh, what do you think is the chance for this to be merged before the 82.0 LTS release?
> 
> @pevogam may be possible if can get the second ACK in time, anyway, it's better to let @clebergnu be aware of your willing :)

Awaiting second ACK on the Cartesian config with join operators is not going to happen any time soon, probably even taking an infinite time as the maintainers using this code are very likely <2. But fine by me, let's hear Cleber's opinion. I happen to use components of this code base that rarely anybody uses, thus putting myself at an unlucky spot. The most activity within this project I have seen is in the Qemu backend where it is easier to merge if more eyes are set on the same components.

pevogam:
Hi @clebergnu, now that the major updates are out of the way do you mind helping me out too by giving advise about the more general issue here? In particular, there are some very weakly covered areas of Avocado VT in terms of usage where there are <2 people that actually use the functionality. The requirement of at least 2 reviews could then never be met. And I am not saying this just for this PR, I recently closed a PR that waited more than 6 months for a second person to set their eyes on it (incl. of people I pinged as @luckyh suggested they might be using the functionality it concerned). Is there a way to remedy this larger problem of the Avocado VT project? I have been plagued by this problem for years now as very often it turned out we use functionality that was last used in Autotest and is now still present but not touched by anyone.

What are you thoughts about this? I am curious to hear your and @luckyh's opinion since this PR will just be another one to hang ready for merging forever because it has <2 people relying on it. For instance, AFAIK currently we (me and my colleagues) are the only ones I know of that use join operator functionality. Not sure if the SPICE devs are still relying on it at all.

pevogam:
Hi @sathnaga, do you think you can check this fairly niche change? I stated some questions to Cleber before but they are too global and beyond the scope of the current PR anyway.

virttest/cartesian_config.py
```diff
@@ -538,10 +538,64 @@ def dump(self, indent, recurse=False):
...
+            # This file was invoked through cmdline
+            options.skipdups
+            skipdups = options.skipdups
+        except NameError:
+            # This file was invoked as Python module
+            skipdups = True
+
+        if skipdups:
+            # Drop vars with suffixes matches general var val
+            # Example: if a_x == 1, and a == 1. Drop: a_x, leave a
```
sathnaga, line 566: @pevogam 
 Yet to completely understand the usecase of this, here is what I understood, correct me if wrong:
patch has two parts:
1. flaten the keys 
        Ex:
```
        newvcpucores = 12
         vms="vm1 vm2 vm3 vm4"
         cores = 4
         cores_vm1 = 4
         cores_vm2 = 8
         test1:
              cores= ${newvcpucores}
```
        without this patch:
                    so core values of vm1 will 4, vm2 will be 8 and vm3,vm4 will be 12
         with this patch:
                   cores_vm1 will be dropped so 
                   so core values of vm1 will also be 12???, vm2 will be 8 and vm3,vm4 will be 12, but that's not what we want right?
          
           Or am I totally getting it wrong?

2. then attempt substitution [earlier it would directly attempt to substitute]

somehow commit message is  not explain to me

pevogam: Yes, the main idea is that substitution won't work on suffixed/joined variants which need flattening before being able to substitute.  Otherwise no substitution is possible for users using these variants like us.

>     without this patch:
>                 so core values of vm1 will 4, vm2 will be 8 and vm3,vm4 will be 12
>      with this patch:
>                cores_vm1 will be dropped so 
>                so core values of vm1 will also be 12???, vm2 will be 8 and vm3,vm4 will be 12, but that's not what we want right?
>       
>        Or am I totally getting it wrong?

Not sure what you mean here, if I run your example on the current branch the output will be:
```
> cat t1.cfg
newvcpucores = 12
vms="vm1 vm2 vm3 vm4"
cores = 4
cores_vm1 = 4
cores_vm2 = 8
test1:
     cores= ${newvcpucores}
> python3 virttest/cartesian_config.py -c t1.cfg
dict    1:  
    cores = 4
    cores_vm1 = 4
    cores_vm2 = 8
    dep = []
    name = 
    newvcpucores = 12
    shortname = 
    vms = vm1 vm2 vm3 vm4
```
If I run it slightly corrected:
```
> cat t1.cfg
newvcpucores = 12
vms="vm1 vm2 vm3 vm4"
cores = 4
cores_vm1 = 4
cores_vm2 = 8
test1:
     cores= ${newvcpucores}
> python3 virttest/cartesian_config.py -c t1.cfg
dict    1:  
    cores = 4
    cores_vm1 = 4
    cores_vm2 = 8
    dep = []
    name = 
    newvcpucores = 12
    shortname = 
    vms = vm1 vm2 vm3 vm4

> cat t1.cfg 
newvcpucores = 12
vms="vm1 vm2 vm3 vm4"
cores = 4
cores_vm1 = 4
cores_vm2 = 8
#test1:
cores= ${newvcpucores}
> python3 virttest/cartesian_config.py -c t1.cfg
dict    1:  
    cores = 12
    cores_vm1 = 4
    cores_vm2 = 8
    dep = []
    name = 
    newvcpucores = 12
    shortname = 
    vms = vm1 vm2 vm3 vm4    
```
I obtain your "without this patch" version but with the patch.

pevogam: Regarding the commit message, this is exactly what I wrote with
```
Joint variants need flattening and dropping of indices in order
to detect values that should be substituted if a previous parameter
of the given key exists.
```
but if I miss something let me know and I will correct it.

pevogam: Ping @sathnaga, sorry if I bother you.

pevogam:
Hi @luckyh, as we are discussing something slightly related in https://github.com/avocado-framework/avocado-vt/pull/2922, perhaps one advantage you could point out in the `tags` discussion is the ability to use parameter substitution and thus define dynamic tags.

Also it is great to get your ACK here if it is still available.

luckyh:
> Hi @luckyh, as we are discussing something slightly related in #2922, perhaps one advantage you could point out in the `tags` discussion is the ability to use parameter substitution and thus define dynamic tags.

@pevogam well, it just happen to be a side effect of the implementation by params, where I hold a neutral view about that. It would be harmless if using correctly, thanks.

luckyh (approved):
LGTM

pevogam:
Rebased the rather oldish branch to tag 85.0 to reduce the pre-merge diff.

samiraguiar:
As you probably know from past issues, I am a colleague of @pevogam also working at Intra2net almost fully with avocado tests. He asked me to independently take a look at some of his pending PRs as another pair of eyes, so I will add a few comments with my thoughts.

virttest/cartesian_config.py
```diff
@@ -538,10 +538,64 @@ def dump(self, indent, recurse=False):
...
+    # dictionary `d_flat' is going to be the mutated copy of `d`
+    d_flat = d.copy()
+    for key in d:
+        if key in _reserved_keys:
+            continue
+
+        if not isinstance(key, tuple):
+            continue
+
+        try:
```
samiraguiar, line 556: It seems that this `try-except` block (and thus the initialization of `skipdups`) can be moved out of the `for` loop as the value of this variable is only used in this function (and doesn't change).

virttest/cartesian_config.py
```diff
@@ -538,10 +538,64 @@ def dump(self, indent, recurse=False):
...
+            # This file was invoked through cmdline
+            options.skipdups
```
samiraguiar, line 558: Are these two lines needed? It seems the second line `skipdups = options.skipdups` would be enough to trigger the exception.

virttest/cartesian_config.py
```diff
@@ -538,10 +538,64 @@ def dump(self, indent, recurse=False):
...
+            skipdups = options.skipdups
+        except NameError:
```
samiraguiar, line 560: Is `NameError` the exception thrown? In Python 3 at least it seems to throw AttributeError:

```python
>>> import optparse
>>> parser = optparse.OptionParser()
>>> options, args = parser.parse_args()
>>> options.foobar
AttributeError: 'Values' object has no attribute 'foobar'
```

virttest/cartesian_config.py
```diff
@@ -538,10 +538,64 @@ def dump(self, indent, recurse=False):
...
+            # Drop vars with suffixes matches general var val
+            # Example: if a_x == 1, and a == 1. Drop: a_x, leave a
+            gen_var_name = key[0]
+            if gen_var_name in d and d[gen_var_name] == d[key]:
+                # Drop gen_var_name, use general key with same value
+                d_flat.pop(key)
+                continue
+
+            can_drop_all_suffixes_for_this_key = True
+            for k in d:
```
samiraguiar, line 574: Maybe in the future it is possible to refactor this function to iterate over the dict twice in separate `for` loops (e.g. by grouping tuples with the same prefix in a second `dict` in the first pass and handling then separately in a second iteration). Since this function is called multiple times, it would be interesting to reduce its time complexity.

virttest/cartesian_config.py
pevogam: Sure will move it out since this change seems to be harmless and will definitely lead to some performance improvement.

virttest/cartesian_config.py
pevogam: Good point, I merely copied this code snippet out from the previous author but will add the performance and this improvement in a separate commit.

virttest/cartesian_config.py
pevogam: The code was probably developed during the Python 2 era and never migrated. Will fix this too.

virttest/cartesian_config.py
pevogam: Very good idea, however indeed a bit too risky and out of scope for the current PR. I will note it down though.

samiraguiar (approved):

samiraguiar:
LGTM

pevogam:
@luckyh This PR from January 2020 is finally ready for merge.

PR #2415, opened 2020-01-30
Add set_kernel_console support for s390x

smitterl:
In order for the kernel console parameter to be set correctly on s390x
the guest architecture must be passed on. Details s. https://github.com/avocado-framework/avocado-vt/pull/2312

Signed-off-by: Sebastian Mitterle smitterl@redhat.com

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -1443,22 +1443,26 @@ class VM(virt_vm.BaseVM):
         logging.debug("Set root login for %s failed.", device)
         return False
 
-    def set_kernel_console(self, device, speed=None, remove=False):
+    def set_kernel_console(self, device, speed=None, remove=False,
+                           guest_arch_name='x86_64'):
         """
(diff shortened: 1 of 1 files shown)


smitterl:
Thank you @balamuruhans I updated commit and description accordingly.

virttest/libvirt_vm.py
```diff
@@ -1443,22 +1443,25 @@ def set_root_serial_console(self, device, remove=False):
         logging.debug("Set root login for %s failed.", device)
         return False
 
-    def set_kernel_console(self, device, speed=None, remove=False):
+    def set_kernel_console(self, device, speed=None, remove=False, guest_arch_name='x86_64'):
```
balamuruhans: Let's have 80 column length,

```python
def set_kernel_console(self, device, speed=None, remove=False,
                       guest_arch_name='x86_64'):
```

smitterl: You're right, thank you :)

PR #2425, opened 2020-02-10
Fix closing local virsh connection problems in V2V

xiaodwan:
1) For local virsh connection, it may does' have session_id attribute,
   we need to check that.
2) Add '--no-pager' option for journal command, if not, the command may
   be timeout because of wating for user interaction.
3) Add 'lshw' for checking pci devices, because not every guest has 'lspci'
   command.
4) store_vm_info is defined in base.cfg, we need to reset its value after
   base.cfg.
5) Move subtests.cfg to the line after 'include convert_destination.cfg'
   in tests-shared.cfg.

### DIFF
--- a/backends/v2v/cfg/tests-shared.cfg
+++ b/backends/v2v/cfg/tests-shared.cfg
@@ -8,19 +8,19 @@ vm_type = v2v
 # The hypervisor uri (default, qemu://hostname/system, etc.)
 # where default or unset means derive from installed system
 connect_uri = default
-# Don't store vm info for v2v
-store_vm_info = no
 
 # Include the base config files.
 include base.cfg
-include subtests.cfg
 include machines.cfg
 include guest-os.cfg
 include guest-hw.cfg
 # Include configuration for v2v tests
 include convert_source.cfg
 include convert_destination.cfg
+include subtests.cfg
 
+# Don't store vm info for v2v
+store_vm_info = no
 # Modify/comment the following lines if you wish to modify the paths of the
 # image files, ISO files or qemu binaries.
 #
--- a/virttest/utils_v2v.py
+++ b/virttest/utils_v2v.py
@@ -577,8 +577,11 @@ class LinuxVMCheck(VMCheck):
         """
         Get vm pci list.
         """
-        cmd = "lspci"
-        return self.run_cmd(cmd)[1]
+        cmd_list = ['lspci', 'lshw']
+        for cmd in cmd_list:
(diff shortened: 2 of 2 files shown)


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.specific_kvm.positive_test.linux.floppy.kvm.output_mode.rhev.rhv.it_default: PASS (637.20 s)
(1/1) type_specific.io-github-autotest-libvirt.specific_kvm.positive_test.linux.network.multi_netcards.esx.esx_67.output_mode.rhev.rhv.it_vddk: PASS (1220.80 s)
(1/1) type_specific.io-github-autotest-libvirt.function_test_xen.positive_test.libvirt.pool_uuid.output_mode.libvirt: PASS (539.87 s)

xiaodwan:
@chunfuwen @chloerh Please help to review this PR, thanks.

backends/v2v/cfg/tests-shared.cfg
```diff
@@ -8,19 +8,19 @@ vm_type = v2v
...
 # Include the base config files.
 include base.cfg
-include subtests.cfg
 include machines.cfg
 include guest-os.cfg
 include guest-hw.cfg
 # Include configuration for v2v tests
 include convert_source.cfg
 include convert_destination.cfg
+include subtests.cfg
```
chunfuwen: why change the order 'include subtests.cfg' ? For v2v , I remember there is special requirement for include order

virttest/utils_v2v.py
```diff
@@ -577,8 +577,11 @@ def get_vm_pci_list(self):
         """
         Get vm pci list.
         """
-        cmd = "lspci"
-        return self.run_cmd(cmd)[1]
+        cmd_list = ['lspci', 'lshw']
+        for cmd in cmd_list:
+            status, output = self.run_cmd(cmd)
```
chunfuwen: Do you need handle possible exception thrown by executed by 'lspci' ? If exception throw, lshw will never be executed

backends/v2v/cfg/tests-shared.cfg
xiaodwan: The old implementation was by patch 'v2v-subtests.patch' in libvirt-ci. It's better to merge it.

virttest/utils_v2v.py
xiaodwan: I think it's not neccessary, lspci never throw out an exception. once it do happens, lshw will fail too because of same exception.

PR #2436, opened 2020-02-20
Update Getting Started

smitterl:
1. Avocado doc has been restructured: fix broken links
2. Give more priority to pip installation:
 a. It's the primary installation method for Avocado (s. doc
    https://avocado-framework.readthedocs.io/en/latest/guides/user/chapters/installing.html
    "Avocado is primarily written in Python, so a standard Python installation is possible
    and often preferable. You can also install from your distro repository, if available."
    etc.)
 b. Pip installation is available on more environments.
 c. Whoever wants to develop/debug test cases is better off with pip installation.
3. Recommend to use the same installation method for both plugin and main program:
 a. Mixing both methods can lead to issues due to version incompatibility and command
    availability.
4. Add python-devel as dependency: I got installation error from dependency netifaces
   (Python.h: No such file or directory)
5. Add a note that test backend related progrmas should be available on target system
   in order to list test cases.
6. As consequence of GetStartedGuide: Add virsh to recommended programs for qemu test backend
   because Networking is usually done using libvirtd's virbr0 - s. https://avocado-vt.readthedocs.io/en/latest/Networking.html
   A user following the GetStartedGuide and gets to the point where they randomly chose
   any test, e.g.  `lvm.lvm_create` for execution from `avocado list --vt-type qemu`
   will just see "ERROR: Can't add interface ... to bridge virbr0: there is no bridge in the system."
   It's not enough for them to just add a bridge of that name, so it can become quite tedious.
   By having the virsh already as a recommended program the user can at least try install it
   which would then automatically set up the default network with correct bridge.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -11,7 +11,7 @@ Gettings started with Avocado
 First of all, make sure you have Avocado itself installed. You can check
 the Avocado online documentation at:
 
-http://avocado-framework.readthedocs.org/en/latest/GetStartedGuide.html#installing-Avocado
+https://avocado-framework.readthedocs.io/en/latest/guides/user/chapters/installing.html
 
 Getting started with Avocado-VT
 ===============================
--- a/docs/source/GetStartedGuide.rst
+++ b/docs/source/GetStartedGuide.rst
@@ -6,24 +6,22 @@ Getting Started
 
 The first step towards using Avocado-VT is, quite obviously, installing it.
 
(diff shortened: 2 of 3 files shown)


ngu-niny:
@luckyh Xu, could you help to review the patch? It's submitted by our s390x colleague Sebas.

Thanks in advance.

PR #2443, opened 2020-02-24
Add scsi hostdev helper functions

smitterl:
1. `utils_disk.get_scsi_info`: new function to get required information
   for `//hostdev/source/address`
2. `utils_test.libvirt.create_hostdev_xml`:
 a. handle usb/pci as before
 b. add logic for scsi
 c. don't modify name of parameter `pci_id`: it has already been overloaded
    to be used with usb, so explicitly mention which information parameter
    can hold but don't change name for backwards compatibility in case
    function called with named parameters

### DIFF
--- a/virttest/utils_disk.py
+++ b/virttest/utils_disk.py
@@ -621,6 +621,22 @@ def _get_mpoint_fstype_linux(session, partition):
     return mount_info.groups()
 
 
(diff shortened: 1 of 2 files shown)


smitterl:
This change was tested during related PR https://github.com/autotest/tp-libvirt/pull/2566

PR #2445, opened 2020-02-25
The cpulist has to be an integer

tfauck:
In some cases CPU being tested is an empty string resulting in that message. Skipping that case.

Signed-off-by: thierry <tfauck@free.fr>


sathnaga:
@tfauck could you pls add supporting test results, failing vs passing, tnx!

PR #2446, opened 2020-02-25
nodecpustat always send a blank line

tfauck:
Even if command is invalid a blanc line is returned by the command.
Currently it is supposed as an error, so testing that case to prevent a
false error.

Signed-off-by: Thierry <tfauck@free.fr>

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -2585,7 +2585,11 @@ def nodecpustats(option='', **dargs):
     """
 
     cmd_nodecpustat = "nodecpustats %s" % option
-    return command(cmd_nodecpustat, **dargs)
+    result = command(cmd_nodecpustat, **dargs)
+    if result.splitline() == "":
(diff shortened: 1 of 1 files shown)


virttest/virsh.py
```diff
@@ -2585,7 +2585,11 @@ def nodecpustats(option='', **dargs):
     """
 
     cmd_nodecpustat = "nodecpustats %s" % option
-    return command(cmd_nodecpustat, **dargs)
+    result = command(cmd_nodecpustat, **dargs)
```
sathnaga, line 2588: this is library call, check of the api behaviour has to be handled in individual testcase, in https://github.com/autotest/tp-libvirt/blob/master/libvirt/tests/src/virsh_cmd/host/virsh_nodecpustats.py

sathnaga:
Closing as this change is not relevant here.

PR #2451, opened 2020-02-27
Fix keep printing error logs when Client process terminated unexpectedly

xiaodwan:
As you can see below, There are millions of error logs were printed in job.log. so the job.log
is very large which makes jenkins out-of-memory.

2020-02-23 23:28:41,427 virt_vm          L1188 DEBUG| Attempting to log into 'esx6.7-rhel8.1-x86_64WY0' (timeout 480s)
2020-02-23 23:29:21,840 ip_sniffing      L0053 DEBUG| Updated HWADDR (00:50:56:ac:06:41)<->(192.168.122.150) IP pair into address cache
2020-02-23 23:29:22,067 virt_vm          L0919 DEBUG| Found/Verified IP 192.168.122.150 for VM esx6.7-rhel8.1-x86_64WY0 NIC 0
2020-02-23 23:30:43,083 ip_sniffing      L0068 DEBUG| Dropped the address cache of HWADDR (00:50:56:ac:06:41)
2020-02-23 23:30:43,086 virt_vm          L1373 DEBUG| Attempting to log into 'esx6.7-rhel8.1-x86_64WY0' via serial console (timeout 10s)
2020-02-23 23:30:43,087 utils_net        L3648 WARNI| Error occur when update VM address cache: No console available.
2020-02-23 23:30:43,088 virt_vm          L1373 DEBUG| Attempting to log into 'esx6.7-rhel8.1-x86_64WY0' via serial console (timeout 10s)
2020-02-23 23:30:43,088 utils_net        L3648 WARNI| Error occur when update VM address cache: No console available.
2020-02-23 23:30:43,089 virt_vm          L1373 DEBUG| Attempting to log into 'esx6.7-rhel8.1-x86_64WY0' via serial console (timeout 10s)
2020-02-23 23:30:43,089 utils_net        L3648 WARNI| Error occur when update VM address cache: No console available.
2020-02-23 23:30:43,090 virt_vm          L1373 DEBUG| Attempting to log into 'esx6.7-rhel8.1-x86_64WY0' via serial console (timeout 10s)
2020-02-23 23:30:43,090 utils_net        L3648 WARNI| Error occur when update VM address cache: No console available.
2020-02-23 23:30:43,090 virt_vm          L1373 DEBUG| Attempting to log into 'esx6.7-rhel8.1-x86_64WY0' via serial console (timeout 10s)
2020-02-23 23:30:43,091 utils_net        L3648 WARNI| Error occur when update VM address cache: No console available.
2020-02-23 23:30:43,091 virt_vm          L1373 DEBUG| Attempting to log into 'esx6.7-rhel8.1-x86_64WY0' via serial console (timeout 10s)

Let's see how it happens by this simple chart.

wait_for_login ->
    while
        self.login ->
            remote_login ->
                handle_prompts

First wait_for_login was called, in the function, wait_for_get_address was called to get
the ip of the guest. By the log, we can see the ip (192.168.122.150) was acquired successfully.
Then the code goes into the while loop which calls function login. In the login function, it
calls remote_login, in remote_login it calls handle_promts. This process generates an ssh command
to connect to the guest. But for some unknow reason, the ssh command timeout, and handle_prompts
raise a timeout exception, and in the while loop of wait_for_login, the exception errors was only
saved silently and then try self.login again.

In the new self.login call, becase of the same unknow reason, the login timeout again.

After two times failure, the guest's state becomes to shutoff suddently, the shutoff reason is 'unknown'. It
might be casued by a crash, but this is another problem. so right now the Client process was terminated and
a exception 'remote.LoginProcessTerminatedError' was raised and be ignored in the while loop as same as the
timeout exception.

From now on, the new self.login will keep trying to connect to a dead guest, which cause the code to call
update_mac_ip_address to update the address by serial console, then reports 'No console available'. and
the failure log prints untill while loop timeout.

That's how the problem happens.

In fact, when the client process is terminated, it means something unrecoverable happens, the login process
should be stoped. so the fix is just capturing the LoginProcessTerminatedError exception and break the
while loop.

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -1216,6 +1216,8 @@ class BaseVM(object):
                 if serial:
                     break
                 raise
+            except remote.LoginProcessTerminatedError:
+                raise
             except Exception as err:
                 error = err
             not_tried = False


xiaodwan:
The  error was printed as expected.

 (1/1) type_specific.io-github-autotest-libvirt.convert_from_file.positive_test.linux.input_mode.vmx.vmx_ssh.latest8.output_mode.libvirt: ERROR: Login timeout expired    (output: "exceeded 480 s timeout, last failure: Client process terminated    (status: 255,    output: 'ssh: connect to host 192.168.122.150 port 22: No route to host\\n')") (828.56 s)

xiaodwan:
@kylazhang @chunfuwen @chloerh Please help to review this PR, thanks.

xiaodwan:
The ssh timeout is caused by the virbr0 interface in the guest. it has same ip address '192.168.122.1' as the host by default.
Althought we can fix it by removing '/etc/libvirt/qemu/autostart/default.xml', this fix is still needed to ensure the problem will not occur.

xiaodwan:
> I'm not sure about this. But I have met this before and maybe we could insert wait time between every 'raise' to reduce the amount of errors.
> I guess this works too.
I don't think It's a good idea to add a sleep between every call to reduce the logs. 
It's meaningless to still connect to a dead guest. And I can't think of any other scenarios to
connect to a process when LoginProcessTerminatedError was raised.

And please see my another comment about the resaon of ssh timeout. You can fix it in the guest.

virttest/virt_vm.py
```diff
@@ -1216,6 +1216,9 @@ def print_guest_network_info():
                 if serial:
                     break
                 raise
+            except remote.LoginProcessTerminatedError as err:
+                error = err
```
chunfuwen: I notice the output log comes from serial console, so suggest add conditional here:
if serial:
     break

Moreover, have you any chance to test the change ? please upload some test results if you have.

virttest/virt_vm.py
```diff
@@ -1216,6 +1216,9 @@ def print_guest_network_info():
                 if serial:
                     break
                 raise
+            except remote.LoginProcessTerminatedError as err:
```
chunfuwen: From your paste logs, most of them comes from virt_vm L1373 ,which is wait_for_serial_login(). 
This is not right place to suppress the log

xiaodwan: When remote.LoginProcessTerminatedError(or ExpectProcessTerminatedError) is raised, it means the aexpect process is dead, why the loop is still running and keeps trying to connect a dead process? 
We should look into why it happens, but not where it happens. I'm not suppressing the log. This patch tries to fix the problem I'm asking above. If anybody gives some reasons why we should not break the loop, then we should fix it in other places by checking the serial condition(This should modify several functions and may break some script, so it has more risks than my current fix).  

In order to show this problem clearly, I'll give more explain.
After LoginProcessTerminatedError happens, the next code runs as below:
self.login->self.get_address(L1089)->self._get_address(L778)->self.address_cache.drop(L829)->raise VMAddressVerificationError(L835)->utils_net.update_mac_ip_address(L1091)->vm.wait_for_serial_login(utils_net.py:3640)->logging.warn("Error occur when update VM )(utils_net.py:3648)->then a new loop again(keep printing above two line logs).

chunfuwen: LoginProcessTerminatedError inherit from LoginError, which extends generic exception (virttest/remote.py). So LoginProcessTerminatedError throwing DOES NOT mean  'aexpect process is dead' since LoginProcessTerminatedError doesn't implement any additional logic there , but just  override __str__(). 
I think that is why it use generic Exception at last in line 1222.

xiaodwan: Then we can check the status and output, the make sure it is aexpect.ExpectProcessTerminatedError.  ILoginProcessTerminatedError is raised in handle_prompt when ExpectProcessTerminatedError happens.

xiaodwan: I checked other places useing LoginProcessTerminatedError, an exceptions.TestError will be raised. I can say LoginProcessTerminatedError has same meaning as aexpect.ExpectProcessTerminatedError. The current code doesn't need to be changed.

xiaodwan:
@kylazhang @dzhengfy Could you help to review this PR? thanks.

virttest/virt_vm.py
```diff
@@ -1216,6 +1216,9 @@ def print_guest_network_info():
...
+                error = err
+                break
```
dzhengfy: Is it better to use 'raise' instead of break? Because if 'break', then  print_guest_network_info() will be executed, but I do not think that is necessary. This exception is enough to be raised just like above LoginAuthenticationError. What do you think of? @xiaodwan

virttest/virt_vm.py
```diff
@@ -1216,6 +1216,9 @@ def print_guest_network_info():
...
             except Exception as err:
                 error = err
```
dzhengfy: In addition, in order to decrease the times we try to login when an unknown exception is happened, shall we add some sleep here? @xiaodwan

virttest/virt_vm.py
xiaodwan: Yes, 'raise' is better than 'break' here, thanks.

virttest/virt_vm.py
xiaodwan: I don't think 'sleep' is necessary,  because unknown exception is suppressed by the common Exception clause. If there are other scenarios are not considered,  a sleep will not help too much, and it may hidden some real script bugs.

virttest/virt_vm.py
kylazhang: I find the aexpect code to check the raise condition
```
        # Check if the child has terminated
        if utils_wait.wait_for(lambda: not self.is_alive(), 5, 0, 0.1):
            raise ExpectProcessTerminatedError(patterns, self.get_status(),
                                               output)
```
So that should means only when process is not alive, it will raise this kind of error
So I think it should work @chunfuwen

virttest/virt_vm.py
dzhengfy: I do not agree this will hid some script bugs. This suggestion is for your statement below:
```
 But for some unknow reason, the ssh command timeout, and handle_prompts
raise a timeout exception, and in the while loop of wait_for_login, the exception errors was only
saved silently and then try self.login again.
```
In your test log,  there are many times to log in 'Attempting to log into'  at 2020-02-23 23:30:43 , so many logs are printed. I think we can add sleep to avoid of printing so many logs in case there is no LoginProcessTerminatedError happened, ssh timeout exception will always happen, then the  'Attempting to log into'  log will print too many.

xiaodwan: The remote_login has already had a timeout function by default, this can protect the script will not print too much 'Attempting to log into' and it is how it works before. I don't see any 'Attempting to log into' logs in my experience. If add another sleep function, it will delay the script running normally. 
And for other uncaught exceptions, I can't think of such an exception that will block login operation.

virttest/virt_vm.py
chunfuwen: @kylazhang 
@xiaodwan 
aexpect as client ,which essentially call  underlying subprocess.Popen(cmd), and it is hard to know status of cmd accurately.  Suggest combine LoginProcessTerminatedError with previous LoginAuthenticationError , and do some basic test since wait_for_login() is heavily used function.
            except (remote.LoginAuthenticationError,
                    remote.LoginBadClientError, LoginProcessTerminatedError ):
                if serial:
                    break
                raise

xiaodwan: if serial is true, in wait_for_serial_login, lots of logs will still be printed out. 
I don't think it's a good idea to put together with the two login exceptions.

virttest/virt_vm.py
dzhengfy: If that is the case, I can agree so far.

virttest/virt_vm.py
```diff
@@ -1216,6 +1216,8 @@ def print_guest_network_info():
                 if serial:
                     break
                 raise
+            except remote.LoginProcessTerminatedError:
```
chunfuwen: Can this additional exception:remote.LoginProcessTerminatedError  combined with previous  remote.LoginAuthenticationError ?  In previous LoginAuthenticationError: if serial : break, otherwise  raise

xiaodwan: if serial is true, in the while loop of wait_for_serial_login, lots of logs will still be printed out. And when the process is terminated, even serial is true, the code still can do nothing to make a success result.
I don't think it's a good idea to put together with the LoginAuthenticationError exceptions.
@chunfuwen

chunfuwen: this is heavily common used function, please paste your test result once ready

xiaodwan:
The exception can be raised as expected.

2020-03-10 22:05:15,649 stacktrace       L0044 ERROR| Reproduced traceback from: /root/v2v_auto/avocado-vt/avocado_vt/test.py:436
2020-03-10 22:05:15,656 stacktrace       L0047 ERROR| Traceback (most recent call last):
2020-03-10 22:05:15,656 stacktrace       L0047 ERROR|   File "/root/v2v_auto/avocado-vt/virttest/remote.py", line 161, in handle_prompts
2020-03-10 22:05:15,656 stacktrace       L0047 ERROR|     timeout=timeout, internal_timeout=0.5)
2020-03-10 22:05:15,656 stacktrace       L0047 ERROR|   File "/usr/local/lib/python3.6/site-packages/aexpect-1.5.1-py3.6.egg/aexpect/client.py", line 925, in read_until_last_line_matches
2020-03-10 22:05:15,657 stacktrace       L0047 ERROR|     print_func)
2020-03-10 22:05:15,657 stacktrace       L0047 ERROR|   File "/usr/local/lib/python3.6/site-packages/aexpect-1.5.1-py3.6.egg/aexpect/client.py", line 861, in read_until_output_matches
2020-03-10 22:05:15,657 stacktrace       L0047 ERROR|     raise ExpectProcessTerminatedError(patterns, self.get_status(), o)
2020-03-10 22:05:15,657 stacktrace       L0047 ERROR| aexpect.exceptions.ExpectProcessTerminatedError: Process terminated while looking for patterns ['[Aa]re you sure', '[Pp]assword:\\s*', '\\(or (press|type) Control-D to continue\\):\\s*$', '[Gg]ive.*[Ll]ogin:\\s*$', '(?<![Ll]ast )[Ll]ogin:\\s*$', '[Cc]onnection.*closed', '[Cc]onnection.*refused', '[Pp]lease wait', '[Ww]arning', '[Ee]nter.*username', '[Ee]nter.*password', '[Cc]onnection timed out', '[\\#\\$]\\s*$', 'Escape character is.*', 'Command>']    (status: 255,    output: 'ssh: connect to host 192.168.122.58 port 22: No route to host\n')

xiaodwan:
For a right configured guest, the script can run successfully.
 (1/1) type_specific.io-github-autotest-libvirt.convert_from_file.positive_test.linux.input_mode.vmx.vmx_ssh.latest8.output_mode.libvirt: PASS (572.69 s)

smitterl:
> As you can see below, There are millions of error logs were printed in job.log. so the job.log
> is very large which makes jenkins out-of-memory.

Alternative proposal https://github.com/avocado-framework/avocado-vt/pull/2488

PR #2475, opened 2020-03-17
Add debug info after installation of stress tool fails

chloerh:
To debug what causes the failure of installation, we need to check
the content of config.log.

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -2276,6 +2276,8 @@ class Stress(object):
         status, output = self.cmd_status_output(self.make_cmds,
                                                 timeout=self.stress_shell_timeout)
         if status != 0:
+            config_log = self.cmd_output_safe('cd %s;cat config.log' % install_path)
+            logging.debug(config_log)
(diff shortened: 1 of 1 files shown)


virttest/utils_test/__init__.py
```diff
@@ -2276,6 +2276,8 @@ def install(self):
         status, output = self.cmd_status_output(self.make_cmds,
                                                 timeout=self.stress_shell_timeout)
         if status != 0:
+            config_log = self.cmd_output_safe('cd %s;cat config.log' % install_path)
```
dzhengfy: It is not good to hard code the  log file name 'config.log'  here.
Suggestion:
1. Check the file's existence before reading it, Or 
2. 'cat *.log' if you want all the logs.

chloerh: This is the default log file, I don't think it needs check, especially when build fails.

sathnaga: I recommend test of file existence as there can be different stress tools configured by user and I am not sure config.log would be present for all.

PR #2479, opened 2020-03-19
libvirt: split to controller create and add func

dzhengfy:
This is to enable the capabilities of configuring more parameters when creating a controller.
And it is also to support to only modify an exsiting vm xml object instead of a guest's xml.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1438,47 +1438,54 @@ def create_hostdev_xml(pci_id, boot_order=None, xmlfile=True,
     return hostdev_xml.xml
 
 
-def create_controller_xml(contr_dict, oper="get_xml", vm_name=None):
+def add_controller(vm_name, contr):
     """
-    Create controller xml or add controller
+    Add a specified controller to the vm xml
+    :param vm_name: The vm name to be added with a controller
+    :param contr: The controller object to be added
+    :raise: exceptions.TestFail if the controller can't be added
+    """
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    org_controllers = vmxml.get_controllers(contr.type)
+    vmxml.add_device(contr)
+    vmxml.sync()
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    updated_controllers = vmxml.get_controllers(contr.type)
+    for one_contr in updated_controllers:
+        if one_contr not in org_controllers:
+            new_added_ctr = one_contr
+    if 'new_added_ctr' not in locals():
+        raise exceptions.TestFail("Fail to get new added controller.")
+
+
+def create_controller_xml(contr_dict):
+    """
+    Create a controller xml
 
     :param contr_dict: The dict params includs controller configurations
-    :param oper: Operation need to do, current we have 2 choice
-                1. "get_xml": controller xml according to dict
-                2. "add_controller": add controller according to dict and
-                updated in vm
-    :param vm_name: name of vm, mandatory for 'add_controller' operation
-    :return: controller xml file if oper is equal to "get_xml";
-             controller object for 'add_controller'.
+    :return: new controller created
     """
     contr_type = contr_dict.get("controller_type", 'scsi')
     contr_model = contr_dict.get("controller_model", "virtio-scsi")
     contr_index = contr_dict.get("controller_index")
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1786,15 +1787,24 @@ def create_controller_xml(contr_dict, oper="get_xml", vm_name=None):
...
     if contr_alias:
         contr.alias = dict(name=contr_alias)
+    if contr_busNr:
+        contr.target = dict(busNr=contr_busNr)
+    if contr_node:
+        contr.node = contr_node
     if contr_addr:
         contr.address = contr.new_controller_address(attrs=eval(contr_addr))
     if oper == "add_controller":
+        if vmxml_ori:
```
chunfuwen: previously this method has existed parameter vm_name.  Can vmxml_ori be provided by vm_name itself ,rather than add additional parameter?

dzhengfy: @chunfuwen , if use 'vm_name', then we have to dump vm to get guest xml, but that is not what I need. Sometimes, I need a fresh vm xml customized by myself here which I create a new controller in it.

virttest/utils_test/libvirt.py
```diff
@@ -1438,47 +1438,55 @@ def create_hostdev_xml(pci_id, boot_order=None, xmlfile=True,
...
-        for contr in updated_controllers:
-            if contr not in org_controllers:
-                new_added_ctr = contr
-        if 'new_added_ctr' not in locals():
-            raise exceptions.TestFail("Fail to get new added controller.")
-        return new_added_ctr
-    elif oper == "get_xml":
-        return contr.xml
+
+    return (contr.xml, contr)
```
chunfuwen, line 1489: I assume returning controller object only is enough since contr.xml is one attribute of controller object

virttest/utils_test/libvirt.py
```diff
@@ -1438,47 +1438,55 @@ def create_hostdev_xml(pci_id, boot_order=None, xmlfile=True,
     return hostdev_xml.xml
 
 
-def create_controller_xml(contr_dict, oper="get_xml", vm_name=None):
+def add_controller(vm_name, contr):
```
chunfuwen, line 1441: I assume we don't need return controller back  since you have newly added contr as input parameter

virttest/utils_test/libvirt.py
```diff
@@ -1438,47 +1438,55 @@ def create_hostdev_xml(pci_id, boot_order=None, xmlfile=True,
...
     """
-    Create controller xml or add controller
+    Add a specified controller to the vm xml
+    :param vm_name: The vm name to be added with a controller
+    :param contr: The controller object to be added
+    :return: The newly added controller object
+    """
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
```
chunfuwen, line 1448: If vm is running state, I assume it will not return actual vm xml at that time

virttest/utils_test/libvirt.py
```diff
@@ -1438,47 +1438,55 @@ def create_hostdev_xml(pci_id, boot_order=None, xmlfile=True,
...
+    org_controllers = vmxml.get_controllers(contr.type)
+    vmxml.add_device(contr)
+    vmxml.sync()
+    vmxml = vm_xml.VMXML.new_from_inactive_dumpxml(vm_name)
+    updated_controllers = vmxml.get_controllers(contr.type)
+    for contr in updated_controllers:
```
chunfuwen, line 1454: The block local variable has the same name with parameter contr. 
I know it prioritize over parameter variable,but it is better not override it in order to avoid confusion

virttest/utils_test/libvirt.py
dzhengfy: update

virttest/utils_test/libvirt.py
dzhengfy: This is originally copied from existing codes.
Update..

virttest/utils_test/libvirt.py
dzhengfy: Fix the original codes. Update.

smitterl:
LGTM

@dzhengfy Thank you for this :D This change is not backwards compatible but I see you are already taking care of it in https://github.com/autotest/tp-libvirt/pull/2728 and the result is a cleaner API.

PR #2495, opened 2020-03-31
Add more params for blkdeviotune

dzhengfy:
There are more options for blkdeviotune than before. So this is to
add those into this function.


dzhengfy:
This PR only has impacts on existing cases in tp-libvirt/libvirt/tests/src/migration/migrate_vm.py. I will create another tp-libvirt PR to update.

PR #2503, opened 2020-04-08
Add xml support for domain backup

lento-sun:

### DIFF
--- a/virttest/libvirt_xml/backup_xml.py
+++ b/virttest/libvirt_xml/backup_xml.py
@@ -1,0 +1,117 @@
+"""
+Module simplifying manipulation of XML described at
+http://libvirt.org/formatbackup.html
+"""
+from virttest import xml_utils
+
+from virttest.libvirt_xml import base, accessors, xcepts
+
+
+class BackupXML(base.LibvirtXMLBase):
+    """
+    Domain backup XML class
+
+    Properties:
+        mode:
+            string (push/pull), backup mode
+        incremental:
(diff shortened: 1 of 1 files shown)


lento-sun:
I'll keep this as draft for now, and when I finish the implementation of checkpoint xml and some basic test cases. I'll set this ready to review.

virttest/libvirt_xml/backup_xml.py
```diff
@@ -0,0 +1,117 @@
...
+        self.xmltreefile.write()
+
+    def del_disks(self):
+        """
+        Del all disk sub-elements form the backup xml
+        """
+        self.xmltreefile.remove_by_xpath('/disks', remove_all=True)
+        self.xmltreefile.write()
+
+    class DiskXML(base.LibvirtXMLBase):
```
chunfuwen, line 69: Can we reuse Disk under disk.py ?

lento-sun: This disk xml is different from the virtual disk xml in vm, it's mainly to indicate if backup of that disk will be executed or not, and indicate some params for backup job, will be as following:
```
<domainbackup>
  <disks>
    <disk name='vda' backup='yes'/>
    <disk name='vdb' type='file'>
      <target file='/path/to/vdb.backup'/>
      <driver type='raw'/>
    </disk>
    <disk name='vdc' backup='no'/>
  </disks>
</domainbackup>
```

PR #2508, opened 2020-04-09
Add on create_local_disk to create luks local file

jgao12345:
Add on create_local_disk to create luks local file

Signed-off-by: jgao <jgao@localhost.localdomain>

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -2961,7 +2961,7 @@ def device_exists(vm, target_dev):
 
 def create_local_disk(disk_type, path=None,
                       size="10", disk_format="raw",
-                      vgname=None, lvname=None):
+                      vgname=None, lvname=None, extra=''):
     if disk_type != "lvm" and path is None:
         raise exceptions.TestError("Path is needed for creating local disk")
     if path:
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -2961,7 +2961,7 @@ def device_exists(vm, target_dev):
 
 def create_local_disk(disk_type, path=None,
                       size="10", disk_format="raw",
-                      vgname=None, lvname=None):
+                      vgname=None, lvname=None, luks=None, password=None):
```
dzhengfy, line 2964: It is not better to hard code these parameter.

virttest/utils_test/libvirt.py
```diff
@@ -2972,7 +2972,12 @@ def create_local_disk(disk_type, path=None,
         pass
     cmd = ""
     if disk_type == "file":
-        cmd = "qemu-img create -f %s %s %s" % (disk_format, path, size)
+        if luks is None:
+            cmd = "qemu-img create -f %s %s %s" % (disk_format, path, size)
+        else:
+            cmd = "qemu-img create -f %s "\
+                  "--object secret,id=sec0,data=`printf '%s' | base64`,format=base64 "\
```
dzhengfy, line 2979: It is better not to hard code these options here.
I suggest that we add an extra param like:
"qemu-img create -f %s %s %s %s" % (disk_format, extra, path, size)
And you can pass extra with ''--object secret,id=sec..." through calling function.

PR #2511, opened 2020-04-10
Fix polkit policy

dzhengfy:
After libvirt-5.6.0, more polkit rules are needed to be setup.
Like,
if (action.id == "org.libvirt.api.network.read" &&
subject.user == "test1") {
if (action.lookup("connect_driver") == 'network') {
return polkit.Result.YES;
} else {
return polkit.Result.NO;
}
}
if (action.id == "org.libvirt.api.network-port.create" &&
subject.user == "test1") {
if (action.lookup("connect_driver") == 'network') {
return polkit.Result.YES;
} else {
return polkit.Result.NO;
}
}
...

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -35,6 +35,7 @@ from virttest import utils_net
 from virttest import utils_config
 from virttest import utils_package
 from virttest import kernel_interface
+from virttest import libvirt_version
 from virttest.staging import service
 from virttest.staging import utils_memory
 from virttest.compat_52lts import results_stderr_52lts, decode_to_text
@@ -1887,19 +1888,29 @@ class LibvirtPolkitConfig(object):
         if distro_obj.name.lower().strip() == 'ubuntu':
             self.polkit_name = "polkitd"
         self.polkitd = service.Factory.create_service(self.polkit_name)
-
-        if self.params.get("action_id"):
-            self.action_id = self.params.get("action_id").split()
-        else:
-            self.action_id = []
         self.user = self.params.get("unprivileged_user")
-        if self.params.get("action_lookup"):
-            # The action_lookup string should be separated by space and
-            # each separated string should have ':' which represent key:value
-            # for later use.
-            self.attr = self.params.get("action_lookup").split()
+        # For libvirt-5.6.0 and later, use 'action_matrix' for all rules
+        # required to setup polkit.
+        # For libvirt before 5.6.0, still use 'action_id' and 'action_lookup'
+        # for all rules required to setup polkit.
+        self.action_list = []
+        if (libvirt_version.version_compare(5, 6, 0) and
+                self.params.get("action_matrix")):
+            self.action_list = eval(self.params.get("action_matrix"))
         else:
-            self.attr = []
+            # This is for the versions before libvirt-5.6.0
+            if self.params.get("action_id"):
+                self.action_id = self.params.get("action_id").split()
+            else:
+                self.action_id = []
+
+            if self.params.get("action_lookup"):
+                # The action_lookup string should be separated by space and
+                # each separated string should have ':' which represent key:value
+                # for later use.
+                self.attr = self.params.get("action_lookup").split()
+            else:
+                self.attr = []
 
     def file_replace_append(self, fpath, pat, repl):
         """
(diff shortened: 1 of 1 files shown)


virttest/test_setup.py
```diff
@@ -1887,19 +1888,28 @@ def __init__(self, params):
...
-            # for later use.
-            self.attr = self.params.get("action_lookup").split()
+        # After libvirt-5.6.0, additional rules are needed.
+        # So case provider needs to provide action_matrix including
+        # all rules.
+        self.action_list = []
+        if (libvirt_version.version_compare(5, 6, 0) and
+                self.params.get("action_matrix")):
+            self.action_list = eval(self.params.get("action_matrix"))
         else:
```
chunfuwen, line 1899: I assume we don't if ...else based on libvirt_version.version_compare(5, 6, 0). Just move self.action_list = [] to end of this block, then update the list if libvirt_version.version_compare(5, 6, 0)

dzhengfy: @chunfuwen , I think your way is also ok. But I have one concern.
Cases will provide action_matrix (for libvirt 5.6.0 and later) and action_id/action_lookup (for libvirt 5.6.0 before). But when libvirt 5.6.0 and later is used, there is no need to handle action_id/action_lookup codes. So I prefer to my current codes. What do you think?

chunfuwen: I am misled by the comment "After libvirt-5.6.0, additional rules are needed." I assume you need add action_matrix besides previous existing action_id/action_lookup . So If the situation is matched your previous comment, I am fine with your previous solution

dzhengfy: @chunfuwen ,ok,  I should update the comments. Sorry to make you misunderstand.

virttest/test_setup.py
```diff
@@ -1887,19 +1888,29 @@ def __init__(self, params):
...
-            # The action_lookup string should be separated by space and
-            # each separated string should have ':' which represent key:value
-            # for later use.
-            self.attr = self.params.get("action_lookup").split()
+        # For libvirt-5.6.0 and later, use 'action_matrix' for all rules
+        # required to setup polkit.
+        # For libvirt before 5.6.0, still use 'action_id' and 'action_lookup'
+        # for all rules required to setup polkit.
+        self.action_list = []
+        if (libvirt_version.version_compare(5, 6, 0) and
```
kylazhang, line 1897: where should it go if version > 5.6.0 and no "action_matrix" given?

virttest/test_setup.py
```diff
@@ -1981,32 +2018,30 @@ def _set_polkit_conf(self):
...
-
-                action_tmp = ""
-                for i in range(len(action_opt)):
-                    action_tmp += action_opt[i]
-
-                # replace ACTION_LOOKUP with string from self.attr
-                handle_tmp = handle.replace('ACTION_LOOKUP', action_tmp)
-                rule_tmp = rule_tmp.replace('HANDLE', handle_tmp)
+            rules = ""
+            if libvirt_version.version_compare(5, 6, 0) and self.action_list:
```
kylazhang, line 2022: As above

virttest/test_setup.py
dzhengfy: @kylazhang , this is to be compatible to existing cases which have not provided 'action_matrix' parameter. If version > 5.6.0 and no "action_matrix" given， 'action_id' and 'action_lookup' still are used as before.  The case will pass if it does not require the new rules, or fail if it does.

kylazhang: ok, thanks

dzhengfy:
Regression test for target cases with 'action_matrix' provided:

```
# avocado run --vt-type libvirt virsh.restore.expected_option.acl_test
JOB ID     : 4bc573026758ae57f017ce0aa41a5c661467e7cb
JOB LOG    : /root/avocado/job-results/job-2020-04-10T18.06-4bc5730/job.log
 (1/4) type_specific.io-github-autotest-libvirt.virsh.restore.expected_option.acl_test.no_option: PASS (78.30 s)
 (2/4) type_specific.io-github-autotest-libvirt.virsh.restore.expected_option.acl_test.paused_option: PASS (96.95 s)
 (3/4) type_specific.io-github-autotest-libvirt.virsh.restore.expected_option.acl_test.running_option: PASS (171.74 s)
 (4/4) type_specific.io-github-autotest-libvirt.virsh.restore.expected_option.acl_test.xml_option: PASS (54.13 s)  (this case depends on tp-libvirt 2658 )
```

Regression test for other cases without 'action_matrix' provided:

```
# avocado run --vt-type libvirt 
virsh.reset.positive_test.acl_test.id_options
JOB ID     : d01e5a599fce30066bfcc815764119e4e223ddad
JOB LOG    : /root/avocado/job-results/job-2020-04-10T17.53-d01e5a5/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.reset.positive_test.acl_test.id_options: PASS (75.00 s)
JOB TIME   : 77.21 s
```

PR #2515, opened 2020-04-16
Add 'force_ipv4' to function ping

chloerh:
When 'force_ipv4' is True, ping will only use ipv4 address as source

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -880,7 +880,7 @@ def raw_ping(command, timeout, session, output_func):
 def ping(dest=None, count=None, interval=None, interface=None,
          packetsize=None, ttl=None, hint=None, adaptive=False,
          broadcast=False, flood=False, timeout=0,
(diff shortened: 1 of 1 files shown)


chloerh:
@Yingshun

virttest/utils_net.py
```diff
@@ -880,7 +880,7 @@ def raw_ping(command, timeout, session, output_func):
 def ping(dest=None, count=None, interval=None, interface=None,
          packetsize=None, ttl=None, hint=None, adaptive=False,
          broadcast=False, flood=False, timeout=0,
-         output_func=logging.debug, session=None):
+         output_func=logging.debug, session=None, force_ipv4=False):
```
Yingshun, line 883: Please update docstring for new parameter

yalzhang:
Suggest to change default value: force_ipv4=False to "True" as we use ipv4 in most cases.

PR #2518, opened 2020-04-20
Enable customized kernel hugepage file

dzhengfy:
This is to add a parameter to use customized kernel hugepage file instead of a default one.

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -375,10 +375,10 @@ class HugePageConfig(object):
         self.qemu_overhead = int(params.get("hugepages_qemu_overhead", 128))
         self.deallocate = params.get("hugepages_deallocate", "yes") == "yes"
         self.hugepage_path = '/mnt/kvm_hugepage'
-        if os.path.exists('/proc/sys/vm/nr_hugepages'):
-            self.kernel_hp_file = '/proc/sys/vm/nr_hugepages'
-        else:
-            raise exceptions.TestSkipError("System doesn't support hugepages")
+        self.kernel_hp_file = params.get("kernel_hp_file", "/proc/sys/vm/nr_hugepages")
+        if not os.path.exists(self.kernel_hp_file):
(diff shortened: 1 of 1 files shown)


virttest/test_setup.py
```diff
@@ -379,6 +379,8 @@ def __init__(self, params):
             self.kernel_hp_file = '/proc/sys/vm/nr_hugepages'
```
sathnaga, line 379: `self.kernel_hp_file = params.get("kernel_hp_file", '/proc/sys/vm/nr_hugepages') ` - better added here...

dzhengfy: @santwana Thanks. Good catch. I updated.

virttest/test_setup.py
```diff
@@ -376,7 +376,7 @@ def __init__(self, params):
         self.deallocate = params.get("hugepages_deallocate", "yes") == "yes"
         self.hugepage_path = '/mnt/kvm_hugepage'
         if os.path.exists('/proc/sys/vm/nr_hugepages'):
-            self.kernel_hp_file = '/proc/sys/vm/nr_hugepages'
+            self.kernel_hp_file = params.get("kernel_hp_file", "/proc/sys/vm/nr_hugepages")
```
sathnaga, line 379: @dzhengfy on a lighter note, you misspelled my username :-)
I guess we can move this before the check for file, so that we can achieve it for user given path aswell.

how about 
```
self.kernel_hp_file = params.get("kernel_hp_file", "/proc/sys/vm/nr_hugepages")
if not os.path.exists(self.kernel_hp_file):
    raise exceptions.TestSkipError("System doesn't support hugepages,  (user given ) hugepage config file not found: %s" % self.kernel_hp_file)
```
something like this covers all cases and makes it informational for user, feel free to add/edit logging.

dzhengfy: @sathnaga , I used a simple error message which I think it is enough. Please help review again.

PR #2524, opened 2020-04-22
guest_agent: Add function of get virtio device of vm

xiagao:
Get virtio device of windows guest.
ID: 1824430 

### DIFF
--- a/virttest/guest_agent.py
+++ b/virttest/guest_agent.py
@@ -930,3 +930,14 @@ class QemuAgent(Monitor):
         cmd = "guest-get-users"
         self.check_has_command(cmd)
         return self.cmd(cmd=cmd)
+
+    def get_virtio_device(self):
+        """
+        Get virtio device driver info of windows guest.
+
+        :return: a list of virtio device driver info, such as device-id,
+                 driver-name, driver-version, driver-date and vender-id.
+        """
+        cmd = "guest-get-devices"
+        self.check_has_command(cmd)
+        return self.cmd(cmd=cmd)


virttest/guest_agent.py
```diff
@@ -930,3 +930,14 @@ def get_users(self):
         cmd = "guest-get-users"
         self.check_has_command(cmd)
         return self.cmd(cmd=cmd)
+
+    def get_virtio_device(self):
+        """
+        Get virtio device driver info of windows guest.
+
+        :return: a list of virtio deivice driver info, such as device-id,
```
lolyu, line 938: `/deivice/device`

xiagao: updated, thanks.

xiagao:
@luckyh  Call for your review, thanks.

virttest/guest_agent.py
```diff
@@ -930,3 +930,14 @@ def get_users(self):
...
         return self.cmd(cmd=cmd)
+
+    def get_virtio_device(self):
+        """
+        Get virtio device driver info of windows guest.
+
+        :return: a list of virtio device driver info, such as device-id,
+                 driver-name, driver-version, driver-date and vender-id.
+        """
+        cmd = "guest-get-devices"
```
vivianQizhu, line 941: Since this command is named `guest-get-devices`, does it only query virtio device or all type devices, I mean, if it is proper to name this function `get_virtio_device`?

xiagao: it only query virtio devices.
From upstream patch of qga.
add command guest-get-devices for reporting VirtIO devices

PR #2530, opened 2020-04-23
Add sriov related functions

Yingshun:
It's required by sriov migration test for now.

### DIFF
--- a/virttest/utils_sriov.py
+++ b/virttest/utils_sriov.py
@@ -1,0 +1,94 @@
+"""
+SRIOV related utility functions
+"""
+
+import logging
+
+from avocado.core import exceptions
+
+from virttest import utils_misc
+from virttest import utils_net
+from virttest import utils_package
+
+
+def find_pf(driver, session=None):
+    """
+    Check if available PF exists to test
+
+    :param driver: The driver name, for example "ixgbe"
+    :param session: The session object to the host
+    :raise: exceptions.TestError when no pf found
+    :return: (interface name, pf device)
+    """
+    driver_dir = "/sys/bus/pci/drivers/%s" % driver
+    cmd = "ls -d %s/000*" % driver_dir
+    s_, output = utils_misc.cmd_status_output(cmd, shell=True,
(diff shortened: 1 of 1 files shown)


chunfuwen:
suggest change python name to sriov_utils.py to keep consistent with other function utils

virttest/sriov.py
```diff
@@ -0,0 +1,110 @@
...
+from virttest import remote
+from virttest.staging import service
+
+
+def find_pf(driver, session=None):
+    """
+    Check if available PF exists to test
+
+    :param session: The session object to the host
+    :raise: test.cancel when no pf found
```
chunfuwen, line 20: generally, we don't have test.cancel available, but just exception in utils module

virttest/sriov.py
```diff
@@ -0,0 +1,110 @@
...
+from virttest import utils_net
+from virttest import remote
+from virttest.staging import service
+
+
+def find_pf(driver, session=None):
+    """
+    Check if available PF exists to test
+
+    :param session: The session object to the host
```
chunfuwen, line 19: miss driver python doc

virttest/sriov.py
```diff
@@ -0,0 +1,110 @@
...
+    :param pci_addr: The pci address
+    :param vf_no: The value to be set
+    :param session: The session object to the host
+    :return: True if successful
+    """
+    logging.debug("pci_addr is %s", pci_addr)
+    cmd = "echo %s > %s/sriov_numvfs" % (vf_no, pci_addr)
+    s, o = utils_misc.cmd_status_output(cmd, shell=True,
+                                        verbose=True, session=session)
+    if not s:
```
chunfuwen, line 62: generally , we need keep one function or method to have one return value in whatever situation if you want them return value. 
So suggest change to :
return not s

Yingshun:
> suggest change python name to sriov_utils.py to keep consistent with other function utils

Updated the file to utils_sriov.py and fixed all your comments.

PR #2535, opened 2020-04-24
virttest.utils_test: Improvement to pip installation

anascko:
That patch provides a fix in two places.
In case to install the dependencies we get the status of the last package.
The package was installed not with pip but need to update it.

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -996,10 +996,10 @@ class AvocadoGuest(object):
             pip_pack = ['setuptools', 'netifaces', 'aexpect', 'netaddr']
             cmd = ""
             for each in pip_pack:
-                cmd += "%s install %s --upgrade;" % (self.pip_bin, each)
-            if self.session.cmd_status(cmd, timeout=self.timeout) != 0:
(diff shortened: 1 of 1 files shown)


xutian:
suggest to install required packages by command. 

pip install -r requriements.txt

xutian:
> suggest to install required packages by command.
> 
> pip install -r requriements.txt

And we can give requriements.txt for python3 and python2 envrionments.

sathnaga:
> > suggest to install required packages by command.
> > pip install -r requriements.txt
> 
> And we can give requriements.txt for python3 and python2 envrionments.

Probably we can take that up as different PR, as you'r suggesting the approach to be enhanced/changed, this PR address a valid bug fix, lets get this in, for time being.
Thanks :-)

sathnaga:
Thanks @anascko for your patch, let's merge this, feel free to improve the pip package installation further as per @xutian suggestions, 

Here is my view on that:
there is a catch to requirements.txt, we need to anyway create requirements.txt inside guest on the fly which again takes additional step to populate the file and use it.
@xutian probably you are thinking a different approach ?

PR #2540, opened 2020-04-27
fix MigrationTest.do_migration to be able to run function without params

Yingshun:
The function which will be executed during migration may not have
any parameter.

### DIFF
--- a/virttest/migration.py
+++ b/virttest/migration.py
@@ -297,7 +297,10 @@ class MigrationTest(object):
                         elif func == virsh.migrate_postcopy:
                             func(vm.name, uri=srcuri, debug=True)
                         else:
-                            func(args['func_params'])
+                            if 'func_params' in args:
+                                func(args['func_params'])
+                            else:
+                                func()
                     else:
                         logging.error("Migration failed to start for %s",
                                       vm.name)
@@ -466,6 +469,7 @@ class MigrationTest(object):
 
         pid = utils_misc.wait_for(_get_pid, 30)
         if utils_misc.safe_kill(pid, signal.SIGINT):
-            logging.info("Succeed to cancel migration: [%s].", pid)
+            logging.info("Succeed to cancel migration: [%s].", pid.strip())
         else:
-            raise exceptions.TestError("Fail to cancel migration: [%s]" % pid)
+            raise exceptions.TestError("Fail to cancel migration: [%s]"
+                                       % pid.strip())


Yingshun:
```
func_name = migrate_test.do_cancel
migrate_test.do_migration(vms, None, dest_uri, 'orderly',
                                  options, thread_timeout=900,
                                  ignore_status=True, virsh_opt=virsh_options,
                                  func=func_name, extra_opts=extra)
```


Test result:
```
2020-04-27 04:46:24,785 process          L0705 INFO | Command /bin/virsh -c 'qemu:///system' migrate --live --verbose --domain avocado-vt-vm1 --desturi qemu+ssh://10.66.85.245/system running on a thread
2020-04-27 04:46:25,025 virsh            L1471 WARNI| VM avocado-vt-vm1 does not exist
2020-04-27 04:46:26,658 migration        L0291 INFO | Migration started for avocado-vt-vm1
2020-04-27 04:46:26,658 process          L0626 INFO | Running 'ps aux |grep 'virsh .* migrate' |grep -v grep |awk '{print $2}''
2020-04-27 04:46:26,676 process          L0458 DEBUG| [stdout] 25757
2020-04-27 04:46:26,676 process          L0714 INFO | Command 'ps aux |grep 'virsh .* migrate' |grep -v grep |awk '{print $2}'' finished with 0 after 0.014486312866210938s
2020-04-27 04:46:26,696 process          L0626 INFO | Running 'kill -2 25757'
2020-04-27 04:46:26,700 process          L0714 INFO | Command 'kill -2 25757' finished with 0 after 0.0006012916564941406s
2020-04-27 04:46:26,700 migration        L0472 INFO | Succeed to cancel migration: [25757
].
2020-04-27 04:46:26,700 migration        L0308 DEBUG| start_time:1587977184, eclipse_time:2
2020-04-27 04:46:26,953 process          L0458 DEBUG| [stderr]
2020-04-27 04:46:26,953 process          L0458 DEBUG| [stderr] Migration: [ 44 %]
2020-04-27 04:46:26,953 process          L0458 DEBUG| [stderr] Migration: [ 47 %]error: operation aborted: migration out job: canceled by client
2020-04-27 04:46:26,963 process          L0458 DEBUG| [stdout]

```

dzhengfy:
@Yingshun , what is the original error without the fix?

virttest/migration.py
```diff
@@ -297,7 +297,10 @@ def do_migration(self, vms, srcuri, desturi, migration_type,
                         elif func == virsh.migrate_postcopy:
                             func(vm.name, uri=srcuri, debug=True)
                         else:
-                            func(args['func_params'])
+                            try:
```
dzhengfy, line 300: Is it better to firstly check if args['func_params'] is available instead of using try clause which is not good for performance.

Yingshun: > @Yingshun , what is the original error without the fix?
it depends on the params. 
if there is no args['func_params'], it raise KeyError, else raise TypeError.

Yingshun:
> @Yingshun , what is the original error without the fix?
it depends on params. If there is func_params, it will raise KeyError, else TypeError.

virttest/migration.py
Yingshun: But sometimes, people may set func_params incorrectly.

dzhengfy:
@Yingshun BTW, from the test result output, 
2020-04-27 04:46:26,700 migration        L0472 INFO | Succeed to cancel migration: [25757
].
You may fix it handy if you would like.

virttest/migration.py
```diff
@@ -297,7 +297,10 @@ def do_migration(self, vms, srcuri, desturi, migration_type,
...
+                                func(args['func_params'])
+                            except (TypeError, KeyError):
```
dzhengfy, line 302: This is not safe because if the func(xxx) function codes raise a TypeError or KeyError, it will also be caught here.
So I still suggest we do 
1.  If  args haskey func_params, then func(args['func_params']). else func()
2.  In tp-libvirt, current cases have two scenarios: 
     a. without func_params= provided
     b. with func_params provided 
For a, no impact because there is no 'func' is provided.
For b, no impact because there is  func_params provided.
But new cases from now on can invoke like this:
```
extra_args = {}
extra_args.update({'func_params': params})  # if one case needs it
migration_test.do_migration(vms, None, dest_uri, 'orderly',
                                            options, thread_timeout=900,
                                            ignore_status=True, virsh_opt=virsh_opt,
                                            func=func_name, extra_opts=extra,
                                           **extra_args)
```

Yingshun: updated

virttest/migration.py
```diff
@@ -297,7 +297,10 @@ def do_migration(self, vms, srcuri, desturi, migration_type,
                         elif func == virsh.migrate_postcopy:
                             func(vm.name, uri=srcuri, debug=True)
                         else:
-                            func(args['func_params'])
+                            if 'func_params' in args:
```
chunfuwen, line 300: Do you need check args['func_params'] = None situation? which may leads to unexpected result when call func(args['func_params'])

Yingshun: Some function need a argument which is None, So just check if func_params are in args here.

PR #2544, opened 2020-05-01
Support more properties for dimm devices

PaulYuuu:
In addition to some common properties, nvdimm device has some unique
properties. pseries already supports this device since qemu-5.0, but it
needs to set the "uuid" property for this dimm device on the qemu
command line.

1. qcontainer: Set nvdimm's uuid if required
2. qcontainer: Supports add other properties with "dimm_extra_params"
3. Introduce of new params in the cfg

ID: 1829139

### DIFF
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -808,3 +808,13 @@ force_reset_go_down_check = "qmp"
 
 # Add -S  to qemu by default, if you don't need it, pls set it off.
 qemu_stop = on
+
+# Set uuid property of nvdimm device (Since QEMU-5.0), it supports the string
+# representation of a UUID as a URN. // RFC 4122
+# If you use a wrong format, will generate a random UUID base on uuid5 and
+# device name.
+# nvdimm_uuid = urn:uuid:ea0ec480-6066-40bc-9709-65d0467ad024
+nvdimm_uuid = ""
+
+# Add extra params for dimm devices
+#dimm_extra_params = "label-size=128k slot=1"
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -15,6 +15,7 @@ import os
 import shutil
 import stat
 import json
+import uuid
 
 # Avocado imports
 from avocado.core import exceptions
@@ -2548,9 +2549,19 @@ class DevContainer(object):
         params = params.object_params("dimm")
         dimm_type = "nvdimm" if params.get("nv_backend") else "pc-dimm"
         attrs = qdevices.Dimm.__attributes__[:]
-        params = params.copy_from_keys(attrs)
-        dev = qdevices.Dimm(params=params, dimm_type=dimm_type)
+        dev = qdevices.Dimm(params=params.copy_from_keys(attrs),
+                            dimm_type=dimm_type)
         dev.set_param("id", "%s-%s" % ("dimm", name))
+        if dimm_type == "nvdimm" and params.get("nvdimm_uuid"):
+            try:
+                dev.set_param("uuid", uuid.UUID(params["nvdimm_uuid"]))
+            except ValueError:
+                nvdimm_uuid = params["nvdimm_uuid"]
+                if nvdimm_uuid == "<auto>":
+                    nvdimm_uuid = uuid.uuid5(uuid.NAMESPACE_OID, name)
+                dev.set_param("uuid", nvdimm_uuid)
+        for ext_k, ext_v in params.get_dict("dimm_extra_params").items():
+            dev.set_param(ext_k, ext_v)
         return dev
 
     def memory_define_by_params(self, params, name):
--- a/shared/cfg/machines.cfg
+++ b/shared/cfg/machines.cfg
@@ -38,6 +38,7 @@ variants:
         del rtc_drift
         del soundcards
         serial_type = 'spapr-vty'
+        nvdimm_uuid = <auto>
     - arm64-mmio:
         only aarch64
         auto_cpu_model = "no"


PaulYuuu:
pc-dimm:
```bash
# /usr/libexec/qemu-kvm -device pc-dimm,?
pc-dimm options:
  addr=<uint64>          -  (default: 0)
  memdev=<link<memory-backend>>
  node=<uint32>          -  (default: 0)
  size=<uint64>
  slot=<int32>           -  (default: -1)
```
nvdimm:
```bash
# /usr/libexec/qemu-kvm -device nvdimm,?
nvdimm options:
  addr=<uint64>          -  (default: 0)
  label-size=<int>
  memdev=<link<memory-backend>>
  node=<uint32>          -  (default: 0)
  size=<uint64>
  slot=<int32>           -  (default: -1)
  unarmed=<bool>         -  (default: false)
  uuid=<QemuUUID>
```

For pseries, "uuid" is required to be set:
```bash
# /usr/libexec/qemu-kvm -m 30720,slots=4,maxmem=32G -object memory-backend-file,size=1G,mem-path=/tmp/nvdimm0,share=on,id=mem-mem1 -device nvdimm,label-size=256m,id=dimm-mem1,memdev=mem-mem1
qemu-kvm: -device nvdimm,label-size=256m,id=dimm-mem1,memdev=mem-mem1: NVDIMM device requires the uuid to be set

# /usr/libexec/qemu-kvm -m 30720,slots=4,maxmem=32G -object memory-backend-file,size=1G,mem-path=/tmp/nvdimm0,share=on,id=mem-mem1 -device nvdimm,label-size=256m,id=dimm-mem1,uuid=6dae7c71-ffa2-5dec-b1a5-f2b00102766b,memdev=mem-mem1
```

PaulYuuu:
@huangyum @luckyh Could you guys help to review? Thanks a lot.

virttest/qemu_devices/qdevices.py
```diff
@@ -1298,7 +1298,7 @@ class Dimm(QDevice):
     the QOM objects 'memory-backend-ram' and 'memory-backend-file'
     """
 
-    __attributes__ = ["memdev", "slot", "addr", "node"]
+    __attributes__ = ["memdev", "slot", "addr", "node", "size"]
```
huangyum, line 1301: The 'size' option is not usable for dimm.

virttest/qemu_devices/qcontainer.py
```diff
@@ -2548,9 +2549,13 @@ def dimm_device_define_by_params(self, params, name):
         params = params.object_params("dimm")
         dimm_type = "nvdimm" if params.get("nv_backend") else "pc-dimm"
         attrs = qdevices.Dimm.__attributes__[:]
-        params = params.copy_from_keys(attrs)
-        dev = qdevices.Dimm(params=params, dimm_type=dimm_type)
+        attrs_params = params.copy_from_keys(attrs)
```
huangyum, line 2552: Could you please explain why a new variant is needed here?

shared/cfg/base.cfg
```diff
@@ -808,3 +808,9 @@ force_reset_go_down_check = "qmp"
 
 # Add -S  to qemu by default, if you don't need it, pls set it off.
 qemu_stop = on
+
+# Set uuid property of nvdimm device (Since QEMU-5.0)
+nvdimm_set_uuid = no
+
+# Add extra params for dimm devices
+#dimm_extra_params = label-size=128k
```
huangyum, line 816: Had better list two or more params as example.

virttest/qemu_devices/qcontainer.py
PaulYuuu: `params.copy_from_keys(attrs)` will redefine the `params` object, then `params.keys()` becomes `qdevices.Dimm.__attributes__[:]`.
We need to get other keys after creating the dimm device, `nvdimm_set_uuid` and `dimm_extra_params` only exist in the first `params`.
Maybe I can modify it more clearly:
```python
        attrs = qdevices.Dimm.__attributes__[:]
        dev = qdevices.Dimm(params=params.copy_from_keys(attrs),
                            dimm_type=dimm_type)
```

virttest/qemu_devices/qdevices.py
PaulYuuu: Reverted it according to your comment.

virttest/qemu_devices/qcontainer.py
```diff
@@ -2548,9 +2549,13 @@ def dimm_device_define_by_params(self, params, name):
         params = params.object_params("dimm")
         dimm_type = "nvdimm" if params.get("nv_backend") else "pc-dimm"
         attrs = qdevices.Dimm.__attributes__[:]
-        params = params.copy_from_keys(attrs)
-        dev = qdevices.Dimm(params=params, dimm_type=dimm_type)
+        dev = qdevices.Dimm(params=params.copy_from_keys(attrs),
+                            dimm_type=dimm_type)
         dev.set_param("id", "%s-%s" % ("dimm", name))
+        if dimm_type == "nvdimm" and params.get_boolean("nvdimm_set_uuid"):
+            dev.set_param("uuid", uuid.uuid5(uuid.NAMESPACE_OID, name))
```
luckyh, line 2556: Shall we let users specify uuid to nvdimm devices? Actually I suggest making the new param to be string type therefore it would support both user specified value and random ones.

PaulYuuu: Make senses, users should able to control the UUID by themself.
I refer to the documentation of [rfc4122](https://www.ietf.org/rfc/rfc4122.txt), now this param supports UUID as a URN format, if user specified a wrong format, will generate a random UUID.

virttest/qemu_devices/qcontainer.py
```diff
@@ -2548,9 +2549,18 @@ def dimm_device_define_by_params(self, params, name):
         params = params.object_params("dimm")
         dimm_type = "nvdimm" if params.get("nv_backend") else "pc-dimm"
         attrs = qdevices.Dimm.__attributes__[:]
-        params = params.copy_from_keys(attrs)
-        dev = qdevices.Dimm(params=params, dimm_type=dimm_type)
+        dev = qdevices.Dimm(params=params.copy_from_keys(attrs),
+                            dimm_type=dimm_type)
         dev.set_param("id", "%s-%s" % ("dimm", name))
+        if dimm_type == "nvdimm" and params.get("nvdimm_uuid"):
```
luckyh, line 2555: Since the option is supported by qemu-5.0 or later, shall we add a check to judge whether should express it in command line or not?

virttest/qemu_devices/qcontainer.py
```diff
@@ -2548,9 +2549,18 @@ def dimm_device_define_by_params(self, params, name):
...
+            try:
+                dev.set_param("uuid", uuid.UUID(params["nvdimm_uuid"]))
+            except ValueError as err:
+                logging.warning("%s, will generate a random UUID base on uuid5"
+                                " and '%s'", err, name)
+                dev.set_param("uuid", uuid.uuid5(uuid.NAMESPACE_OID, name))
```
luckyh, line 2561: I'd suggest calling this `<auto>` because it not actually generate "random" hexadecimal digits when using some version (uuid5), and simply setting the option with the literal string user provides (which means we allow wrong values and let qemu complain about), then the code would be like:
```python
nvdimm_uuid = params.get("nvdimm_uuid")
if nvdimm_uuid == "<auto>":
    nvdimm_uuid = uuid.uuid5(uuid.NAMESPACE_OID, name)
dev.set_param("uuid", nvdimm_uuid)
```

virttest/qemu_devices/qcontainer.py
PaulYuuu: I don't think so, although this option is supported since QEMU 5.0, it's not a required one except for pseries, so I set `nvdimm_uuid = ""` in base.cfg. If users want to set the UUID for nvdimm device, they can modify their configure file by themselves in testcase side.

virttest/qemu_devices/qcontainer.py
PaulYuuu: Okay, I got your point that if we config a wrong format of `nvdimm_uuid`, let vt forces set it to assign a negative value, and raise error when create_qemu_command.

virttest/qemu_devices/qcontainer.py
luckyh: Here is an issue, by setting this param in `machines.cfg` for pseries guests, users run tests with qemu less than 5.0 would get the failure since the option is not supported, isn't it?

PaulYuuu: This case will not happen.
pseries supports nvdimm device beginning at [qemu-5.0](https://wiki.qemu.org/ChangeLog/5.0#PowerPC), which means we need to add a filter `required_qemu = [5.0, ]` in test case cfg. If users want to launch a guest with nvdimm device less than qemu-5.0, qemu will be terminated and report that there is not this device.

**QEMU 4.2:**
```bash
$ /usr/libexec/qemu-kvm -device ? | tail -n 3
Uncategorized devices:
name "pc-dimm", desc "DIMM memory module"
name "spapr-tpm-proxy"
```

**QEMU 5.0:**
```bash
$ /usr/libexec/qemu-kvm -device ? | tail -n 4
Uncategorized devices:
name "nvdimm", desc "DIMM memory module"
name "pc-dimm", desc "DIMM memory module"
name "spapr-tpm-proxy"
```

luckyh: ok, thanks for your explanation!

luckyh:
Thanks all.

PR #2550, opened 2020-05-11
Extract guest agent state

smitterl:
In order to allow for setting guest agent state without
re-installing package, extract method safely.

Legacy behavior is left unchanged.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2938,6 +2938,15 @@ class VM(virt_vm.BaseVM):
         self.install_package('pm-utils', ignore_status=True, timeout=15)
         self.install_package('qemu-guest-agent')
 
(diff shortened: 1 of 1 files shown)


virttest/libvirt_vm.py
```diff
@@ -2938,6 +2938,15 @@ def prepare_guest_agent(self, prepare_xml=True, channel=True, start=True):
         self.install_package('pm-utils', ignore_status=True, timeout=15)
         self.install_package('qemu-guest-agent')
 
+        self.set_state_guest_agent(start)
+
+    def set_state_guest_agent(self, start):
```
chunfuwen, line 2943: this change implicitly make functions such as  _is_ga_running() to be hidden. You may need check whether any python has ever used them.

PR #2554, opened 2020-05-12
Use correct comparison between str/int and python literals

pevogam:
Reference for this change:
https://bugs.python.org/issue34850


pevogam:
@luckyh This fixes lots of Python 3.8 warnings and is correct behavior also for pre-3.8 python versions. I think nobody checks for new PRs and simply assign themselves so I hope explicit tagging is not a problem for you.

pevogam:
@sathnaga Sure, added.

luckyh:
Thanks all!

PR #2555, opened 2020-05-14
migration: prameterize signal for do_cancel

Yingshun:
Some cases require different signals.

### DIFF
--- a/virttest/migration.py
+++ b/virttest/migration.py
@@ -456,19 +456,20 @@ class MigrationTest(object):
             if int(result.exit_status) != 0:
                 raise exceptions.TestFail(result.stderr_text.strip())
 
-    def do_cancel(self):
+    def do_cancel(self, sig=signal.SIGKILL):
         """
         Kill process during migration.
 
+        :param sig: The signal to send
         :raise: test.error when kill fails
         """
         def _get_pid():
-            cmd = "ps aux |grep 'virsh .* migrate' |grep -v grep |awk '{print $2}'"
+            cmd = "ps aux |grep 'virsh .*migrate' |grep -v grep |awk '{print $2}'"
             pid = process.run(cmd, shell=True).stdout_text
             return pid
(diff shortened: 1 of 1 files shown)


Yingshun:
```
2020-05-13 22:36:52,021 process          L0714 INFO | Command 'ps aux |grep 'virsh .*migrate' |grep -v grep |awk '{print $2}'' finished with 0 after 0.039717674255371094s
2020-05-13 22:36:52,023 process          L0626 INFO | Running 'kill -9 219828'
2020-05-13 22:36:52,027 process          L0714 INFO | Command 'kill -9 219828' finished with 0 after 0.0008020401000976562s
2020-05-13 22:36:52,027 migration        L0473 INFO | Succeed to cancel migration: [219828].
```

virttest/migration.py
```diff
@@ -456,19 +456,20 @@ def check_result(self, result, params):
...
         """
         def _get_pid():
-            cmd = "ps aux |grep 'virsh .* migrate' |grep -v grep |awk '{print $2}'"
+            cmd = "ps aux |grep 'virsh .*migrate' |grep -v grep |awk '{print $2}'"
             pid = process.run(cmd, shell=True).stdout_text
             return pid
 
         pid = utils_misc.wait_for(_get_pid, 30)
-        if utils_misc.safe_kill(pid, signal.SIGINT):
+        if utils_misc.safe_kill(pid, sig):
```
dzhengfy, line 472: What is the reason to replace signal.SIGINT with signal.SIGKILL? Does it impact existing tp-libvirt cases?

Yingshun: Actually, it's a method which hasn't been called yet.  It will be used in new automation case 2663. 
The previous signal  SIGINT cannot stop migration correctly and checking on vm after cancelling will failed in this case. So update the default value to SIGKILL instead and make it configurable.

PR #2557, opened 2020-05-15
Fix the issue of duplicated adding panic device

chloerh:
The original assertion is not correct, we must compare search result
with 'None' to confirm if there is already a panic device in vm xml.


chloerh:
@Yingshun

PR #2560, opened 2020-05-18
Add access_mode setting in vmxml.set_memoryBacking_tag()

yafu-1:
Signed-off-by: Yan Fu <yafu@redhat.com>

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -1562,7 +1562,7 @@ class VMXML(VMXMLBase):
 
     @staticmethod
     def set_memoryBacking_tag(vm_name, hpgs=True, nosp=False, locked=False,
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -1562,7 +1562,7 @@ def get_blkdevio_params(vm_name, options="", virsh_instance=base.virsh):
 
     @staticmethod
     def set_memoryBacking_tag(vm_name, hpgs=True, nosp=False, locked=False,
-                              virsh_instance=base.virsh):
+                              access_mode=None, virsh_instance=base.virsh):
```
Yingshun, line 1565: It is better to always add new arguments at last for better backward compatibility or to change all the users accordingly.

yafu-1: Updated.

PR #2563, opened 2020-05-20
Move check_domjobinfo function to utils_test.libvirt

Yingshun:
It was introduced in tp-libvirt migrate_options_shared.py.
Now, a new migration script needs it, so move it
to utils_test.

### DIFF
--- a/virttest/utils_test/libvirt_domjobinfo.py
+++ b/virttest/utils_test/libvirt_domjobinfo.py
@@ -1,0 +1,65 @@
+
+"""
+Virsh domjobinfo command related utility functions
+"""
+
+import re
+import math
+import logging
+
+from avocado.core import exceptions
+
+from virttest import virsh
+
+
+# pylint: disable=E1121
+def check_domjobinfo(vm, params, option="", remote_virsh_dargs=None):
+    """
+    Check given item in domjobinfo of the guest is as expected
+
+    :param vm: the vm object
+    :param params: the parameters used
+    :param option: options for domjobinfo
+    :param remote_virsh_dargs: the parameters for remote host
+    :raise: exceptions.TestFail if the value of given item is unexpected
+    """
+    def _search_jobinfo(jobinfo, ignore_status=False):
+        """
+        Find value of given item in domjobinfo
+
+        :param jobinfo: cmdResult object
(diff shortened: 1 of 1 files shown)


Yingshun:
@dzhengfy  check_domjobinfo was implemented by you,  can you please review?

virttest/utils_test/libvirt.py
```diff
@@ -3677,3 +3678,56 @@ def check_dumpxml(vm, content, err_ignore=False):
...
+    """
+    Check given item in domjobinfo of the guest is as expected
+
+    :param vm: the vm object
+    :param params: the parameters used
+    :param option: options for domjobinfo
+    :param remote_virsh_dargs: the parameters for remote host
+    :raise: exceptions.TestFail if the value of given item is unexpected
+    """
+    def search_jobinfo(jobinfo, ignore_status=False):
```
chunfuwen, line 3694: Better inner method  begin with _ prefix

virttest/utils_test/libvirt.py
```diff
@@ -3677,3 +3678,56 @@ def check_dumpxml(vm, content, err_ignore=False):
                 raise exceptions.TestFail("Expected '%s' was not found in "
                                           "%s's xml" % (content, vm.name))
     return True
+
+
+# pylint: disable=E1121
+def check_domjobinfo(vm, params, option="", remote_virsh_dargs=None):
```
chunfuwen, line 3684: suggest create one separate utils file such as utils_domjob.py under avocado-vt/virttest to avoid putting everything into libvirt.py file

virttest/utils_test/libvirt.py
Yingshun: updated.

virttest/utils_test/libvirt.py
Yingshun: @chunfuwen If we want a separate file, I prefer libvirt_domjobinfo.py under avocado-vt/virttest/utils_test. What do you think?

chunfuwen: sound ok

dzhengfy: +1

PR #2569, opened 2020-05-25
utils_netperf: Add params to support latest netperf version

yanglei-rh:
Add params to support latest netperf version.
ID:1833873

### DIFF
--- a/virttest/utils_netperf.py
+++ b/virttest/utils_netperf.py
@@ -124,8 +124,9 @@ class NetperfPackage(remote.Remote_Package):
         build_type = {"aarch64": "aarch64-unknown-linux-gnu"}
         build_arch = self.session.cmd_output("arch", timeout=60).strip()
         np_build = build_type.get(build_arch, build_arch).strip()
-        setup_cmd = "./configure --build=%s %s > /dev/null 2>&1" % (np_build,
-                                                                    compile_option)
+        setup_cmd = "./autogen.sh > /dev/null 2>&1 &&" \
+                    " ./configure --build=%s %s > /dev/null 2>&1" % (np_build,
+                                                                     compile_option)
         setup_cmd += " && make > /dev/null 2>&1"
         self.env_cleanup(clean_all=False)
         cmd = "%s && %s " % (pre_setup_cmd, setup_cmd)


yanglei-rh:
@PaulYuuu Please help me review it,thanks.

virttest/utils_netperf.py
```diff
@@ -124,8 +124,13 @@ def pack_compile(self, compile_option=""):
         build_type = {"aarch64": "aarch64-unknown-linux-gnu"}
         build_arch = self.session.cmd_output("arch", timeout=60).strip()
         np_build = build_type.get(build_arch, build_arch).strip()
-        setup_cmd = "./configure --build=%s %s > /dev/null 2>&1" % (np_build,
-                                                                    compile_option)
+        if self.netperf_dir == "/var/tmp/netperf-2.7.1":
```
PaulYuuu, line 127: Check the absolute path is not a good idea, better to use RegExp to judge its version from `self.netperf_file` and then add `autogen.sh` before currently command if need.

PaulYuuu:
@vivianQizhu Also request for your review, thanks.

virttest/utils_netperf.py
PaulYuuu: Checking if `autogen.sh` exists is an easier method.

yanglei-rh: I have updated it,please help me review it again.Thanks.

virttest/utils_netperf.py
```diff
@@ -124,8 +124,9 @@ def pack_compile(self, compile_option=""):
         build_type = {"aarch64": "aarch64-unknown-linux-gnu"}
         build_arch = self.session.cmd_output("arch", timeout=60).strip()
         np_build = build_type.get(build_arch, build_arch).strip()
-        setup_cmd = "./configure --build=%s %s > /dev/null 2>&1" % (np_build,
-                                                                    compile_option)
+        setup_cmd = "./autogen.sh > /dev/null 2>&1 &&" \
+                    " ./configure --build=%s %s > /dev/null 2>&1" % (np_build,
+                                                                     compile_option)
```
vivianQizhu, line 129: I not sure about your change, so both versions have `autogen.sh` script? You will fail the setup if it does not exist?

yanglei-rh: Yes,  The existing netperf version will release the 'autogen.sh' script after decompression.

vivianQizhu:
@PaulYuuu Request for your ack, thanks.

PR #2572, opened 2020-05-26
Add cephfs mount and umount utils

chunfuwen:
cephfs mount and umount intends to be used for cephfs cases

### DIFF
--- a/virttest/ceph.py
+++ b/virttest/ceph.py
@@ -15,6 +15,7 @@ from avocado.utils import process
 
 from virttest import utils_numeric
 from virttest import error_context
+from virttest import utils_misc
 
 
 class CephError(Exception):
@@ -219,3 +220,61 @@ def create_config_file(ceph_monitor):
         with open(ceph_cfg, 'w+') as f:
             f.write('mon_host = %s' % ceph_monitor)
     return ceph_cfg
+
+
+@error_context.context_aware
+def cephfs_mount(ceph_uri, mount_point, options=None, verbose=False, session=None):
+    """
+    Mount ceph volume to mountpoint.
+
+    :param ceph_uri: str to express ceph uri, e.g 10.73.xx.xx:6789:/
+    :param mount_point: mount point, str
+    :param options: mount options, e.g -o name=admin,secret=AQCAgd5cFyMQLhAAHaz6w+WKy5LvmKjmxx
+    :param verbose: enable verbose or not
+    :param session: mount within the session if given
+    """
+    cephfs_umount(ceph_uri, mount_point, verbose, session)
+    mount_cmd = ['mount -t ceph']
+    if options:
+        mount_cmd.extend(['-o', options])
+    mount_cmd.extend([ceph_uri, mount_point])
+    mount_cmd = ' '.join(mount_cmd)
+    if session:
+        create_mount_point = "mkdir -p  %s" % mount_point
+        session.cmd(create_mount_point, ok_status=[0, 1], ignore_all_errors=True)
+        session.cmd(mount_cmd, ok_status=[0, 1], ignore_all_errors=True)
+    else:
+        if not os.path.exists(mount_point):
+            try:
+                utils_misc.make_dirs(mount_point)
+            except OSError as dirError:
+                logging.debug("Creation of the directory:%s failed:%s", mount_point, str(dirError))
+            else:
+                logging.debug("Successfully created the directory %s", mount_point)
+        process.system(mount_cmd, verbose=verbose)
+
+
+@error_context.context_aware
+def cephfs_umount(ceph_uri, mount_point, verbose=False, session=None):
+    """
+    Umount ceph volume from mountpoint.
+
+    :param ceph_uri: str to express ceph uri, e.g 10.73.xx.xx:6789:/
+    :param mount_point: mount point, str
+    :param verbose: enable verbose or not
+    :param session: mount within the session if given
+    """
+    umount_cmd = "umount %s" % mount_point
+    if session:
+        session.cmd(umount_cmd, ok_status=[0, 1], ignore_all_errors=True)
+        rm_mount_point = "rm -rf %s" % mount_point
+        session.cmd(rm_mount_point, ok_status=[0, 1], ignore_all_errors=True)
+    else:
+        process.system(umount_cmd, ignore_status=True, verbose=verbose)
+        if os.path.exists(mount_point):
+            try:
+                utils_misc.safe_rmdir(mount_point)
+            except OSError as dirError:
+                logging.debug("Delete of the directory:%s failed:%s", mount_point, str(dirError))
+            else:
+                logging.debug("Successfully deleted the directory %s", mount_point)


chunfuwen:
gvfsd-fuse /run/user/1000/gvfs fuse.gvfsd-fuse rw,nosuid,nodev,relatime,user_id=1000,group_id=1000 0 0
libvirtauthceph.usersys.redhat.com:6789:/ is not mounted
Running 'mount -t ceph -o name=admin,secret=AQDkY/xd2pqyLhAAPQ2Yrla/nGe1PazR4/n+IQ== libvirtauthceph.usersys.redhat.com:6789:/ /mnt/cephfs'
Command 'mount -t ceph -o name=admin,secret=AQDkY/xd2pqyLhAAPQ2Yrla/nGe1PazR4/n+IQ== libvirtauthceph.usersys.redhat.com:6789:/ /mnt/cephfs' finished with 0 after 0.1860349178314209s
End cephes mount ***********************************33
lets wait for some
Begin umount ****************************************************77
Running 'cat /proc/mounts'

libvirtauthceph.usersys.redhat.com:6789:/ /mnt/cephfs ceph rw,relatime,name=admin,secret=<hidden>,acl,wsize=16777216 0 0
libvirtauthceph.usersys.redhat.com:6789:/ is mounted
Running 'umount /mnt/cephfs'
Command 'umount /mnt/cephfs' finished with 0 after 14.391308307647705s
Successfully deleted the directory /mnt/cephfs
finished umount ****************************************************88

virttest/ceph.py
```diff
@@ -220,3 +221,65 @@ def create_config_file(ceph_monitor):
...
+    :param session: mount within the session if given
+    """
+    cephfs_umount(ceph_uri, mount_point, verbose, session)
+    if not os.path.exists(mount_point):
+        try:
+            os.mkdir(mount_point)
+        except OSError as dirError:
+            logging.debug("Creation of the directory:%s failed:%s", mount_point, str(dirError))
+        else:
+            logging.debug("Successfully created the directory %s", mount_point)
```
Yingshun, line 244: These parts only work on localhost, please update to support the creation/check of mount_point on remote host

virttest/ceph.py
```diff
@@ -220,3 +221,65 @@ def create_config_file(ceph_monitor):
...
+        if session:
+            umount_result = session.cmd_status(umount_cmd, safe=True)
+        umount_result = process.system(umount_cmd, ignore_status=False, verbose=verbose)
+    if os.path.exists(mount_point):
+        try:
+            os.rmdir(mount_point)
+        except OSError as dirError:
+            logging.debug("Delete of the directory:%s failed:%s", mount_point, str(dirError))
+        else:
+            logging.debug("Successfully deleted the directory %s", mount_point)
```
Yingshun, line 271: same as above.

virttest/ceph.py
```diff
@@ -220,3 +221,65 @@ def create_config_file(ceph_monitor):
...
+    :param mount_point: mount point, str
+    :param verbose: enable verbose or not
+    :param session: mount within the session if given
+    """
+    umount_result = 1
+    if ceph_is_mounted(ceph_uri, mount_point, verbose=verbose, session=session):
+        umount_cmd = "umount %s" % mount_point
+        if session:
+            umount_result = session.cmd_status(umount_cmd, safe=True)
+        umount_result = process.system(umount_cmd, ignore_status=False, verbose=verbose)
```
Yingshun, line 264: you'd better run 'fuser -k' before unmount, you can reuse utils_disk.umount.

virttest/ceph.py
chunfuwen: updated

virttest/ceph.py
chunfuwen: updated

virttest/ceph.py
chunfuwen: fuser -k is dangerous operations, In my local test, it crash desktop and cause all console terminated

Yingshun: what's your full command line?

chunfuwen: As I mentioned previously, the terminal was killed ,and there is no output,which shows command line that is executing. Fuser -km ,which is used in utils_disk.umount , could kill all processed ,which access certain file.

virttest/ceph.py
```diff
@@ -220,3 +221,74 @@ def create_config_file(ceph_monitor):
...
+        session.cmd(rm_mount_point, ok_status=[0, 1], ignore_all_errors=True)
+    else:
+        process.system(umount_cmd, ignore_status=True, verbose=verbose)
+        if os.path.exists(mount_point):
+            try:
+                os.rmdir(mount_point)
+            except OSError as dirError:
+                logging.debug("Delete of the directory:%s failed:%s", mount_point, str(dirError))
+            else:
+                logging.debug("Successfully deleted the directory %s", mount_point)
```
Yingshun, line 281: Can you check utils_disk.safe_rmdir and utils_misc.umount. you can reuse them I think.

virttest/ceph.py
```diff
@@ -220,3 +221,74 @@ def create_config_file(ceph_monitor):
...
+        session.cmd(mount_cmd, ok_status=[0, 1], ignore_all_errors=True)
+    else:
+        if not os.path.exists(mount_point):
+            try:
+                os.mkdir(mount_point)
+            except OSError as dirError:
+                logging.debug("Creation of the directory:%s failed:%s", mount_point, str(dirError))
+            else:
+                logging.debug("Successfully created the directory %s", mount_point)
+        process.system(mount_cmd, verbose=verbose)
```
Yingshun, line 255: I think you'd better to reuse utils_misc.make_dirs and utils_disk.mount, what do you think?

virttest/ceph.py
chunfuwen: I assume there is no safe_rmdir in utils_disk.py, but in utils_mic.py.  Actually, The method has quite limited usability, it can not handle the scenario gracefully where directory not existed.
But it can have a try.
*******************************************
>>> utils_misc.safe_rmdir('/tmp/abc')
  File "/home/chunfu/workdir/p3-v3/code/avocado-vt/virttest/utils_misc.py", line 993, in safe_rmdir
    assert check_isdir(path, session=session), "Invalid directory to remove %s" % path
AssertionError: Invalid directory to remove /tmp/ab
*******************************************************
utils_misc.umount call utils_disk.umount() underlying, which has issue as mentioned previously

virttest/ceph.py
chunfuwen: Part of utils_misc.make_dirs can be used.
For  utils_disk.mount, if you carefully check its implementation, you will find in most cases for remote session, it may throw error. The reason is that session.cmd can not handle  commands such as: ls -l /var/abc, mount -f ceph xx xx, where commands return some error message or no any output 
******************
    if session:
        return session.cmd_status(cmd, timeout=120, safe=True) == 0

chunfuwen:
Sending command: echo $?
Sending command: grep cpu /proc/cpuinfo | awk '{print $3}' | head -n 1
Sending command: echo $?
PASS 1-type_specific.io-github-autotest-libvirt.virsh.migrate_ceph.positive_testing.migrating_cephfs_backend.mount_dir_softlink.without_postcop

virttest/ceph.py
```diff
@@ -220,3 +223,74 @@ def create_config_file(ceph_monitor):
...
+        session.cmd(mount_cmd, ok_status=[0, 1], ignore_all_errors=True)
+    else:
+        if not os.path.exists(mount_point):
+            try:
+                utils_misc.make_dirs(mount_point)
+            except OSError as dirError:
+                logging.debug("Creation of the directory:%s failed:%s", mount_point, str(dirError))
+            else:
+                logging.debug("Successfully created the directory %s", mount_point)
+        process.system(mount_cmd, verbose=verbose)
```
kylazhang, line 257: I agree with yicui, they are duplicated. if it will raise error, how about use try except?

virttest/ceph.py
```diff
@@ -220,3 +223,74 @@ def create_config_file(ceph_monitor):
...
+@error_context.context_aware
+def cephfs_umount(ceph_uri, mount_point, verbose=False, session=None):
+    """
+    Umount ceph volume from mountpoint.
+
+    :param ceph_uri: str to express ceph uri, e.g 10.73.xx.xx:6789:/
+    :param mount_point: mount point, str
+    :param verbose: enable verbose or not
+    :param session: mount within the session if given
+    """
```
kylazhang, line 269: may need judge ceph_is_mounted or not.

virttest/ceph.py
```diff
@@ -220,3 +223,74 @@ def create_config_file(ceph_monitor):
...
+        rm_mount_point = "rm -rf %s" % mount_point
+        session.cmd(rm_mount_point, ok_status=[0, 1], ignore_all_errors=True)
+    else:
+        process.system(umount_cmd, ignore_status=True, verbose=verbose)
+        if os.path.exists(mount_point):
+            try:
+                utils_misc.safe_rmdir(mount_point)
+            except OSError as dirError:
+                logging.debug("Delete of the directory:%s failed:%s", mount_point, str(dirError))
+            else:
```
kylazhang, line 282: from man
```
       fuser -km /xxx
              kills all processes accessing the file system  /xxx  in  any
              way.
```
not sure why it may kill all process, or because of the mount dir is not specific?
If use utils_disk.umount, or below func will be useless? Or if we can add param to disable fuser operation? Also for the error or we can ignore with try..except

chunfuwen: Is there any reason why we need fuser -km firstly before umount ? Is there any explicit evidences that this is  best practices ,which can be promoted?

virttest/ceph.py
chunfuwen: no need since the implementation will handle them gracefully whether it is mounted or not

virttest/ceph.py
chunfuwen: As I mentioned before,due to constraints from session.cmd, below commands need to cared about in special way:ok_status=[0, 1], ignore_all_errors=True
      session.cmd(create_mount_point, ok_status=[0, 1], ignore_all_errors=True)
        session.cmd(mount_cmd, ok_status=[0, 1], ignore_all_errors=True

virttest/ceph.py
```diff
@@ -15,6 +15,9 @@
 
 from virttest import utils_numeric
 from virttest import error_context
+from virttest import utils_disk
```
kylazhang, line 18: Please also remove it

chunfuwen: done

PR #2573, opened 2020-05-27
Modify create_disk_xml function to support create image with slice

jgao12345:
Modify create_disk_xml function to support create image with slice

Signed-off-by: jgao <jgao@localhost.localdomain>


virttest/utils_test/libvirt.py
```diff
@@ -1548,7 +1548,7 @@ def alter_boot_order(vm_name, pci_id, boot_order=0):
     vmxml.sync()
 
 
-def create_disk_xml(params):
+def create_disk_xml(params, input_source_file=''):
```
dzhengfy, line 1551: Could we put this new parameter into 'params'? There is no need to add a new parameter like this in case the existing parameter can contain it.

PR #2582, opened 2020-06-02
Add TPM device support to virt-install cmdline

sathnaga:
this patchs adds support to add TPM device
to libvirt guest xml via virt-install using
user params.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -946,6 +946,31 @@ class VM(virt_vm.BaseVM):
                 item += 1
             return cputune_str.rstrip(",")
 
+        def add_tpmdevice(help_text, device_path, model=None, type=None):
+            """
+            Add TPM device to guest xml
+            :param help_text: string, virt-install help text
+            :param device_path: path to TPM device
+            :param model: tpm device model to be added
+                          tpm-tis, tpm-crb, tpm-spapr etc
+            :param type: type of device attach
+                         passthrough, emulator
+            :return: string of tpm cmdline for virt-install
+            """
+            result = ""
+            if not has_option(help_text, "tpm"):
+                logging.warning("tpm option is not supported in virt-install")
+                return result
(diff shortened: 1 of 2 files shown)


virttest/libvirt_vm.py
```diff
@@ -947,6 +947,31 @@ def add_cputune(vcpu_cputune=""):
...
+            Add TPM device to guest xml
+            :param help_text: string, virt-install help text
+            :param device_path: path to TPM device
+            :param model: tpm device model to be added
+                          tpm-tis, tpm-crb, tpm-spapr etc
+            :param type: type of device attach
+                         passthrough, emulator
+            :return: string of tpm cmdline for virt-install
+            """
+            result = None
```
chloerh, line 961: Looks like 'result' is supposed to be an empty string, right?

virttest/libvirt_vm.py
```diff
@@ -1341,7 +1366,13 @@ def add_cputune(vcpu_cputune=""):
             virt_install_cmd += " --extra-args '%s'" % kernel_params
 
         virt_install_cmd += " --noautoconsole"
-
+        # Add TPM device
+        tpm_device = params.get("tpm_device_path", None)
+        if tpm_device:
+            tpm_model = params.get("tpm_model", None)
+            tpm_type = params.get("tpm_type", None)
+            virt_install_cmd += add_tpmdevice(help_text, tpm_device, tpm_model,
```
chloerh, line 1374: And if the function does return 'None', this line would fail, since 'virt_install_cmd' is a str, and cannot be '+' by 'None'.

sathnaga: makes sense, tnx, will change it.

sathnaga:
@chloerh thanks for your review, have updated now.

sathnaga:
@chloerh tnx!,
since this does not affect any other functionality and new option included, merging.

PR #2583, opened 2020-06-04
Reuse the unified image directory upon vm image creation

pevogam:
The base image directory parameter was ignored in some cases of image
creation where it should override the default path for this.


pevogam:
Thanks a lot for you constant aid @luckyh! I hope the more active @balamuruhans and @sathnaga and willing to check it too.

pevogam:
Thanks @balamuruhans! @luckyh Since we have two reviews, can we proceed with merging?

luckyh:
@pevogam yeah, no problem.

PR #2585, opened 2020-06-05
qdevices: Add align option for memory object

huangyum:
ID: 1843844


virttest/qemu_devices/qdevices.py
```diff
@@ -1251,7 +1251,7 @@ class Memory(QObject):
 
     __attributes__ = ["size", "prealloc", "mem-path", "backend", "policy",
                       "host-nodes", "share", "merge", "dump", "pmem",
-                      "discard-data"]
+                      "discard-data", "align"]
```
yanan-fu, line 1254: Update the annotations too ?

huangyum: Yes, thanks.

PR #2588, opened 2020-06-07
qemu_monitor:  Reassign timeout default for _acquire_lock

YongxueHong:
The original timeout default maybe short in some scenarios where
send multiple commands with just one qmp monitor in parallel then
the timeout for acquiring lock need more time, so reassign timeout
default with instance attribute so that it could be modified in
the case level.

ID: 1844782
Signed-off-by: Yongxue Hong <yhong@redhat.com>

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -366,7 +366,9 @@ class Monitor(object):
             pass
         self._socket.close()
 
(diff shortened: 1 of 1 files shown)


YongxueHong:
Hi, @luckyh  please help me to review, thanks.

PR #2590, opened 2020-06-10
Use dict method instead of Params method

zhencliu:
Signed-off-by: Zhenchao Liu <zhencliu@redhat.com>

ID: 1845834

### DIFF
--- a/virttest/storage.py
+++ b/virttest/storage.py
@@ -42,7 +42,7 @@ def preprocess_images(bindir, params, env):
 
 
 def preprocess_image_backend(bindir, params, env):
-    enable_gluster = params.get_boolean("enable_gluster")
+    enable_gluster = params.get("enable_gluster") == "yes"
     gluster_image = params.get("gluster_brick")
     if enable_gluster and gluster_image:
         return gluster.create_gluster_vol(params)
@@ -70,7 +70,7 @@ def file_exists(params, filename_path):
 
     :return: True if image file exists else False
     """
-    if params.get_boolean("enable_gluster"):
+    if params.get("enable_gluster") == "yes":
         return gluster.file_exists(params, filename_path)
 
     if params.get("enable_ceph") == "yes":


YongxueHong:
@zhencliu might better to write the reason why `Use dict's method instead of Params` into the commit message to track this code change by commit log in the future.
BTW, the character `'` of commit title is redundant.
Thanks.

zhencliu:
> @zhencliu might better to write the reason why `Use dict's method instead of Params` into the commit message to track this code change by commit log in the future.
> BTW, the character `'` of commit title is redundant.
> Thanks.

@YongxueHong Agreed. Updated the commit message with more detailed information. Thanks.

XueqiangWei:
Tested related cases, all passed.
(1/2) Host_RHEL.m8.u2.product_av.spice.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.qemu_disk_img.rebase.snC.to_new: PASS (1763.40 s)
 (2/2) Host_RHEL.m8.u2.product_av.spice.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.qemu_disk_img.rebase.snC.to_new.restore_to_snB: PASS (1709.65 s)
 (1/1) Host_RHEL.m8.u2.product_av.spice.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.boot: PASS (61.94 s)

LGTM
Acked-by: xuwei@redhat.com

zhencliu:
Thank you , guys. 
@luckyh can you please help to merge code, thanks.

pevogam:
> @zhencliu might better to write the reason why `Use dict's method instead of Params` into the commit message to track this code change by commit log in the future.
> BTW, the character `'` of commit title is redundant.
> Thanks.

Yes, I was also confused about the reasons for this change at first. Even better would have been to write within the code it in a single comment line on top of the change to avoid tracking altogether. I still wonder though, couldn't the code passing the dictionary be fixed instead.

zhencliu:
> > @zhencliu might better to write the reason why `Use dict's method instead of Params` into the commit message to track this code change by commit log in the future.
> > BTW, the character `'` of commit title is redundant.
> > Thanks.
> 
> Yes, I was also confused about the reasons for this change at first. Even better would have been to write within the code it in a single comment line on top of the change to avoid tracking altogether. I still wonder though, couldn't the code passing the dictionary be fixed instead.

The original change(params.get-->params.get_boolean) was made in 7a130388122090e16c96b7a39c294689be58be49(gluster storage backend support), which led to a regression, we had a discussion then to determine how to fix it, tp-qemu or vt code? Considering there could be several test cases  passing a dict instead of a param object, we chose to fix vt code, to handle it in the same way before commit 7a130388122090e16c96b7a39c294689be58be49, please also refer to 1845834 for details.

pevogam:
> The original change(params.get-->params.get_boolean) was made in [7a13038](https://github.com/avocado-framework/avocado-vt/commit/7a130388122090e16c96b7a39c294689be58be49)(gluster storage backend support), which led to a regression, we had a discussion then to determine how to fix it, tp-qemu or vt code?

This fix is also a kind of regression as it makes a step away from a type specific parameter.

> Considering there could be several test cases passing a dict instead of a param object, we chose to fix vt code,

I guess there should be a way to do a test run that will detect all failing tests so that they could be fixed instead.

> to handle it in the same way before commit [7a13038](https://github.com/avocado-framework/avocado-vt/commit/7a130388122090e16c96b7a39c294689be58be49), please also refer to 1845834 for details.

I guess 1845834 is a commit hash? I couldn't find it in the git history though. In any case, I was just raising concerns about the test fixes, nothing too breaking though, just a small regression.

PR #2600, opened 2020-06-19
Fix VM cleanup issue when vm.start fails

xiaodwan:
1) When vm.start fails, the vm should be deleted and image should
be removed from export domain.
2) When timeout is -1, wait forever until the VM is imported to
data domain successfully.

### DIFF
--- a/virttest/ovirt.py
+++ b/virttest/ovirt.py
@@ -176,6 +176,8 @@ class VMManager(virt_vm.BaseVM):
         """
         Return MAC address of a VM.
         """
+        if not self.instance:
+            self.update_instance()
         vnet_list = self.instance.nics_service().list()
         if net_name != '*':
             vnet_list = [vnet for vnet in vnet_list if vnet.name == net_name]
@@ -377,7 +379,9 @@ class VMManager(virt_vm.BaseVM):
         :param export_name: Export domain name.
         :param storage_name: Storage domain name.
         :param cluster_name: Cluster name.
+        :param timeout: timeout value
         """
+        begin_time = time.time()
         end_time = time.time() + timeout
         vm = self.lookup_by_storagedomains(export_name)
         sds_service = self.connection.system_service().storage_domains_service()
@@ -403,7 +407,8 @@ class VMManager(virt_vm.BaseVM):
             time.sleep(1)
         if not vm_down:
             raise WaitVMStateTimeoutError("DOWN", self.state())
-        logging.info('Import %s successfully', self.name)
+        logging.info('Import %s successfully(time lapse %ds)',
+                     self.name, time.time() - begin_time)
 
     def export_from_export_domain(self, export_name, timeout=300):
         """
--- a/virttest/utils_v2v.py
+++ b/virttest/utils_v2v.py
@@ -1181,11 +1181,11 @@ def v2v_cmd(params, auto_clean=True, cmd_only=False):
         vpx_dc = None
     uri = uri_obj.get_uri(hostname, vpx_dc, esxi_host)
 
-    # Pre-process for v2v
-    _v2v_pre_cmd()
-
-    target_obj = Target(target, uri)
     try:
+        # Pre-process for v2v
+        _v2v_pre_cmd()
+
+        target_obj = Target(target, uri)
         # Return virt-v2v command line options based on 'target' and
         # 'hypervisor'
         options = target_obj.get_cmd_options(params)
@@ -1314,6 +1314,9 @@ def import_vm_to_ovirt(params, address_cache, timeout=600):
                 v2v_cmd)
     except Exception as e:
         logging.error("Start %s failed: %s", vm.name, e)
+        vm.delete()
+        if output_method != 'rhv_upload':
+            vm.delete_from_export_domain(export_name)
         return False
     return True
 


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.function_test_esx.positive_test.no-copy.esx_67.rhev.rhv_upload.it_vddk.vpx_uri: PASS (464.14 s)
(1/1) type_specific.io-github-autotest-libvirt.function_test_xen.positive_test.rhev.rhel6_2.output_mode.rhev.rhv: PASS (790.28 s)
 (1/1) type_specific.io-github-autotest-libvirt.convert_from_file.positive_test.linux.input_mode.vmx.vmx_ssh.latest8.output_mode.rhev.rhv: PASS (739.76 s)
 (1/1) type_specific.io-github-autotest-libvirt.convert_from_file.positive_test.linux.input_mode.vmx.vmx_ssh.latest8.output_mode.rhev.rhv_upload: PASS (745.05 s)
  (1/4) type_specific.io-github-autotest-libvirt.convert_vm_to_ovirt.esx.vm.6_7.linux.latest8.arch_x86_64.raw_f.NFS.rhv.it_vddk: PASS (2139.57 s)
 (2/4) type_specific.io-github-autotest-libvirt.convert_vm_to_ovirt.esx.vm.6_7.linux.latest8.arch_x86_64.raw_f.NFS.rhv_upload.rhv_direct.rhv_verifypeer.nopreallocated.it_vddk: PASS (2116.49 s)
 (3/4) type_specific.io-github-autotest-libvirt.specific_kvm.positive_test.linux.bogus_kernel.esx.esx_67.output_mode.rhev.rhv.it_default: PASS (2340.40 s)
 (4/4) type_specific.io-github-autotest-libvirt.specific_kvm.positive_test.linux.bogus_kernel.esx.esx_67.output_mode.rhev.rhv_upload.it_vddk: PASS (2055.36 s)

xiaodwan:
@chunfuwen @chloerh @yantian0 Please help to review this PR, thanks.

xiaodwan:
Thanks for your review.

Please don't be scared by the 'waiting forever' description. 

The reasons I do the change are:
1) It riches the waiting process. Users can choose waiting forever or not. 
It's just a small improvement, and the default behaviour is still timeout.
2) This makes import_vm_to_ovirt more intelligent. It just needs to wait ovirt to
break the loop. I confirmed with developer, this should be a safe way to
use it like this.
3) In the past two year, I debugged too much timeout problems and tweaked the
timeout values again and again, from 20mins, to 40min, 1h, and I'm not suspected
the value should be changed again in future. Because you never know
how much size of the disk you are testing and how bad the connection
is in the lab. They already exhausted too much energy of me. 
4) If the timeout happens, and the importing is still in process, the
VM is in locking status, The clean operations may not success. This 
will bring risks for following testing cases. Compare to this, waiting for ovirt
is a safer way.
5) Anyway, If the bad situation really happens, and the while loop really
keeps running forever, ok, we found a bug now, shouldn't we be happy?
6) IMHO, A tool should serve for users. As a tester, sometimes I don't want to
interfere any normal testing process. But now the code does do it, only because 
it's not finished in a special time. This is not a performance testing.
Shouldn't we reconsider how meaningful those *protecting* mechanism are?
Does they really protect something or just hide some potential problems? 

That's all my reasons, If you all still insist the waiting forever should not be added.
I'm going to keep the original code.

chloerh:
@xiaodwan I understand the 'blood and sweat' you have put here, I too agree that we shouldn't waste out time here. But the infinity loop is forbidden for many reasons, one of which is that it could bring unpredictable problems. 
May I suggest a compromised way here? What if we set an unreasonable long time as timeout when it's supposed to 'wait forever', like 4 hours or something? Since no matter what, we still have our own time limit in our own jenkins job, right?

xiaodwan:
> @xiaodwan I understand the 'blood and sweat' you have put here, I too agree that we shouldn't waste out time here. But the infinity loop is forbidden for many reasons, one of which is that it could bring unpredictable problems.
> May I suggest a compromised way here? What if we set an unreasonable long time as timeout when it's supposed to 'wait forever', like 4 hours or something? Since no matter what, we still have our own time limit in our own jenkins job, right?

I agree with you that wait_forever should not be used in general.
But in this case, you can't avoid it, because we are expecting another module to give us feedback. 
Based on the usage of ovirtskd4, there is no better way to get intelligent notification instead of keep
checking VM status in a loop. But we can trust the ovirt module and trust the choice of users.
IMHO, '4h' or other values which you think it should be fine has no big difference than wait_forever(
here the wait_forever means waiting the ovirt task until it finishes.). 

I guess you didn't notice my description about another problem. When the script is aborted forcely,
the VM cannot be cleanup and will be left on ovirt server. It will cause following scripts which use
same guest fail to startup the guest because of mac duplicate problem. In other words, you messed up
the testing environment by yourself. 

I really don't understand what's your concern. 
First, the change only adds another way of waiting, the old way is still exists and it's by default.
Second, this is going to be used in other testing project, users should clearly known when they choose to use it in their own script. 
Third, Even the worst situation happens, so what?  You can easily know which script has problem and then to solve it.

BTW, the ovirt.py should do more changes to break the loop when any exceptions happens.
I will fix this.

Yingshun:
@xiaodwan  I agree with @chloerh and @sathnaga test should exit on timeout. For the reason 4, how about expanding the test execution time and checking vm health status and the existence of importing process before starting vm?

xiaodwan:
Thanks all your review.  @Yingshun @chloerh @sathnaga 
I'll change back the waiting process. 
It's not a MUST HAVE change for the basic problem of this patch trying to fix.

sathnaga:
Thanks all, lets merge.

PR #2606, opened 2020-06-23
Add one operation to del_mac_address for interface xml

yalzhang:
Add one operation to del_mac_address for interface xml

Signed-off-by: yalzhang <yalzhang@redhat.com>

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -3447,6 +3447,7 @@ def modify_vm_iface(vm_name, oper, iface_dict, index=0):
     del_rom = iface_dict.get('del_rom')
     del_filter = iface_dict.get('del_filter')
     del_port = iface_dict.get('del_port')
+    del_mac = "yes" == iface_dict.get('del_mac', "no")
     if iface_type:
         iface.type_name = iface_type
     if iface_driver:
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -3436,6 +3436,7 @@ def modify_vm_iface(vm_name, oper, iface_dict, index=0):
     del_addr = iface_dict.get('del_addr')
     del_rom = iface_dict.get('del_rom')
     del_filter = iface_dict.get('del_filter')
+    del_mac = iface_dict.get('del_mac')
```
Yingshun, line 3439: del_mac in your tests is set to "yes" or "no", how about change to:
del_mac = "yes" == iface_dict.get("del_mac", "no")

yalzhang: > del_mac in your tests is set to "yes" or "no", how about change to:
> del_mac = "yes" == iface_dict.get("del_mac", "no")

no, the {"del_mac": "yes"} is an optional parameter just like the parameters above "del_rom" "del_addr", it no need to do the delete, there is no such parameter passed here.

Yingshun: Yes, I understand.  But for a case which wants to keep mac address, if it's wrongly set to "no", it will delete the mac address with your code. 
>>> a="no"
>>> if a:
...  print("ds")
... 
ds
>>>

Yingshun:
@yalzhang  Can you please rebase and send it again to pass Cirrus CI?

virttest/utils_test/libvirt.py
yalzhang: updated

Yingshun (approved):

sathnaga (approved):

PR #2610, opened 2020-06-24
Add method to construct cpu xml from domcapabilities

smitterl:
Define a static constructor method to create a
VMCPUXML intance valid for cpu-compare.

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -1899,6 +1899,24 @@ class VMCPUXML(base.LibvirtXMLBase):
             pass  # Element already doesn't exist
         self.xmltreefile.write()
 
+    @staticmethod
+    def from_domcapabilities(domcaps_xml):
+        """
+        Construct a cpu definition from domcapabilities host-model definition.
+
+        :param domcaps_xml: DomCapabilityXML with host-model definition
+        :return: None
+        """
+        cpu_xml = VMCPUXML()
+        cpu_xml['model'] = domcaps_xml.get_hostmodel_name()
+        features = domcaps_xml.get_additional_feature_list('host-model')
+        feature_names = [name for f_list in [list(d.keys()) for d in features]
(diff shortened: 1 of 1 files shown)


smitterl:
I'm not familiar with the LGTM analysis: Python check. I checked the link https://lgtm.com/projects/g/avocado-framework/avocado-vt/logs/rev/pr-d5bed950d2d20ea5b318a6b04c5cec0b6675d9b7/lang:python/stage:Preparing%20merge%20commit_6cc630b649a98a909b51f85609178cfce1f5528b but I don't see what I have to do to make this pass. Any help is highly appreciated.

virttest/libvirt_xml/vm_xml.py
```diff
@@ -1899,6 +1899,24 @@ def remove_numa_cells(self):
             pass  # Element already doesn't exist
         self.xmltreefile.write()
 
+    @staticmethod
+    def from_domcapabilities(domcaps_xml):
```
chunfuwen, line 1903: Better method name follow the convention: verb.+ noun format

smitterl: @chunfuwen I prefer this way. I believe the 'fromX' is a common pattern even in the standard library to provide class constructors. I did a [search](https://docs.python.org/3/search.html?q=from&check_keywords=yes&area=default) for confirmation
E.g.
* [datetime.fromX methods](https://docs.python.org/3/library/datetime.html?highlight=from#datetime.date.fromtimestamp)
* [fractions.from_X methods](https://docs.python.org/3/library/fractions.html?highlight=from#fractions.Fraction.from_float)

I chose underscore_style as usual in avocado. The example I found are all `@classmethod` instead of `@staticmethod`. But I prefer to keep `@staticmethod` in order to restrict its usage to non-instances, i.e.
```python
cpu = VMCPUXML.from_domcapabilities(domcapa_xml)
```
Please, let me know your thoughts.

PR #2617, opened 2020-06-30
usb_host: add usb-host device specific parameters support

ybduan:
1. hostbus and hostaddr
2. hostbus and hostport
3. venderid and productid

ID: 1847777, 1847778

Signed-off-by: ybduan <yduan@redhat.com>

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -1499,7 +1499,7 @@ class DevContainer(object):
 
     # USB Device related methods
     def usb_by_variables(self, usb_name, usb_type, controller_type, bus=None,
-                         port=None, serial=None):
+                         port=None, serial=None, dev_options=None):
         """
         Creates usb-devices by variables.
         :param usb_name: usb name
@@ -1508,6 +1508,7 @@ class DevContainer(object):
         :param bus: the bus name (my_bus.0, ...)
         :param port: port specification (4, 4.1.2, ...)
         :param serial: serial specification (1234, d1, ...)
+        :param dev_options: device options dict of usb_type
         :return: QDev device
         """
         if not self.has_device(usb_type):
@@ -1519,6 +1520,9 @@ class DevContainer(object):
             device.set_param('bus', bus)
             device.set_param('port', port)
             device.set_param('serial', serial)
+            if dev_options:
+                for opt_name, opt_value in dev_options.items():
+                    device.set_param(opt_name, opt_value)
             device.parent_bus += ({'type': controller_type},)
         else:
             if "tablet" in usb_type:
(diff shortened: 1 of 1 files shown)


ybduan:
Please help review this PR, thanks! @yanan-fu

ybduan:
Please help review this PR, thanks! @luckyh

ybduan:
Please help review this PR, thanks! @yanan-fu

yanan-fu:
It looks good for me.  
@luckyh  Could you help review ?   For the param name since it is have suffix of device name so i think it is not must to have feature keywords such as 'usb_hostbus_***'.  What's your opinion ?

CongLi:
@luckyh Could you please help review this patch ?

virttest/qemu_devices/qcontainer.py
```diff
@@ -1499,7 +1499,8 @@ def usbc_by_params(self, usb_name, params, pci_bus={"aobject": "pci.0"}):
 
     # USB Device related methods
     def usb_by_variables(self, usb_name, usb_type, controller_type, bus=None,
-                         port=None, serial=None):
+                         port=None, serial=None, hostbus=None, hostaddr=None,
+                         hostport=None, vendorid=None, productid=None):
```
luckyh, line 1503: Suggest packing the device specified into one variable, e.g. `dev_options`

virttest/qemu_devices/qcontainer.py
```diff
@@ -1519,6 +1525,32 @@ def usb_by_variables(self, usb_name, usb_type, controller_type, bus=None,
             device.set_param('bus', bus)
             device.set_param('port', port)
             device.set_param('serial', serial)
+            if hostbus and hostaddr:
```
luckyh, line 1528: And then you have to check if the given device is `usb-host` before performing the following routine, otherwise, you could simply set the device options by:
```python
for opt_name, opt_value in dev_options.items():
    device.set_param(opt_name, opt_value)
```

virttest/qemu_devices/qcontainer.py
```diff
@@ -1542,7 +1574,12 @@ def usb_by_params(self, usb_name, params):
                                      params.get("usb_controller"),
                                      params.get("bus"),
                                      params.get("port"),
-                                     params.get("usbdev_serial"))
+                                     params.get("usbdev_serial"),
+                                     params.get("hostbus"),
```
luckyh, line 1578: As @yanan-fu said, suggest adding the perfix `usbdev_option_` to these params.

virttest/qemu_devices/qcontainer.py
ybduan: Updated, thanks!

virttest/qemu_devices/qcontainer.py
ybduan: Updated, thanks!

virttest/qemu_devices/qcontainer.py
ybduan: Updated, thanks!

ybduan:
Could you please help review this patch again? @luckyh

yanan-fu:
Well, we remove the steps of checking the correctness of the user' input. I am ok with that as we discussed.
How about add one more comment to explain that user is responsible for the correctness of the dev_options ?

ybduan:
> Well, we remove the steps of checking the correctness of the user' input. I am ok with that as we discussed.
> How about add one more comment to explain that user is responsible for the correctness of the dev_options ?

Do you mean to add a comment in source code or commit message? @yanan-fu 
But I think it's reasonable for avocado-vt to accept data and do what upper level tell it to do, so I don't catch your point for this additional comment.
What's your opinion? @luckyh

yanan-fu:
> > Well, we remove the steps of checking the correctness of the user' input. I am ok with that as we discussed.
> > How about add one more comment to explain that user is responsible for the correctness of the dev_options ?
> 
> Do you mean to add a comment in source code or commit message? @yanan-fu
> But I think it's reasonable for avocado-vt to accept data and do what upper level tell it to do, so I don't catch your point for this additional comment.
> What's your opinion? @luckyh

I mean the source code, thanks!

yanan-fu:
Discussed with owner, the comment is must it, keep current status, so ACK.

@luckyh could you help review again. Thanks!

PR #2621, opened 2020-07-01
Fix the stdout error as compat_52lts removed

kylazhang:
Involved by commit d9acd3a3a19ff11e8a9987379379d3f332b3d1d0


chunfuwen:
This is caused by removal of lts, and it is urgent fix

PR #2624, opened 2020-07-03
Do not force use int type for 'fd' in qmp command

yanan-fu:
id: 1853538

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -2214,7 +2214,12 @@ class QMPMonitor(Monitor):
                     value = "=".join(opt[1:])
                     try:
                         if re.match("^[0-9]+$", value):
-                            value = int(value)
+                            # when force convert type to int,exclude 'fd' as bz
+                            # 1853538. And the int type convert is not absolute
+                            # accurate for other values, if hit problem please
+                            # check the expect type of qemu and update.
+                            if opt[0] != 'fd':
+                                value = int(value)
                         elif re.match("^[0-9]+\.[0-9]*$", value):
                             value = float(value)
                         elif re.findall("true", value, re.I):


virttest/qemu_monitor.py
```diff
@@ -2214,7 +2214,8 @@ def send_args_cmd(self, cmdlines, timeout=CMD_TIMEOUT, convert=True):
                     value = "=".join(opt[1:])
                     try:
                         if re.match("^[0-9]+$", value):
-                            value = int(value)
+                            if not (opt[0] == 'fd'):
+                                value = int(value)
```
vivianQizhu, line 2218: What's the use case for it? If it is all numbers, what's the difference of 'int' or 'str' here?

vivianQizhu: Ah, okay I see the product bz, so it can't be int in new released qemu.

yanan-fu: @vivianQizhu qemu 5.0 only accept str now, and str can work with old version too include rhel7.
You can get more details from the bz i added.

vivianQizhu:
@yanan-fu Would you explain it a bit in the commit log?

yanan-fu:
@vivianQizhu  Just updated the commit info already ~

yanan-fu:
@vivianQizhu  Sorry, fix a typo in commit info ~

vivianQizhu:
Actually I am thinking that, it does not make sense to convert the variant to string only because it's value is all numbered. Like in this case actually 'fd' can be anything besides a number, just because we use a random number so here it converts it into 'int'. But I am not sure the effort to take to correct this ill logic.

yanan-fu:
> Actually I am thinking that, it does not make sense to convert the variant to string only because its value is all numbered. Like in this case actually 'fd' can be anything besides a number, just because we use a random number so here it converts it into 'int'. But I am not sure the effort to take to correct this ill logic.

Seems you reversed that change ?  The old design is covert the variant to "int" (not strting) when the value is all numbered.
This patch is used exclude 'fd' for the old logic, to use the origin format of fd, do not force change the type.

In fact, i agree with you, i have some concern about the pre design too, but we do not know the history about the reason of using 'int' by force when the variants is all numbered, so i did not update all in case cause other side effect. So~   only exclude fd, what do you think ?

vivianQizhu:
@yanan-fu Yes, I won't request you correct the conversion logic in this PR, that might require to match all qmp variants type here if we want to make every thing accurate. So let's just fix the fd issue here since it's more urgent.

vivianQizhu:
But we might add a comment here to info further users that this type convert is not that accurate. @yanan-fu What do you think?

yanan-fu:
> But we might add a comment here to info further users that this type convert is not that accurate. @yanan-fu What do you think?

Make sense ~   Update with a comment to explain this, thanks !

yanan-fu:
@yanglei-rh  Could you please help review ? Thanks!

virttest/qemu_monitor.py
```diff
@@ -2214,7 +2214,12 @@ def send_args_cmd(self, cmdlines, timeout=CMD_TIMEOUT, convert=True):
                     value = "=".join(opt[1:])
                     try:
                         if re.match("^[0-9]+$", value):
-                            value = int(value)
+                            # when force convert type to int,exclude 'fd' as bz
+                            # 1853538. And the int type convert is not absolute
+                            # accurate for other values, if hit problem please
+                            # check the expect type of qemu and update.
+                            if not (opt[0] == 'fd'):
+                                value = int(value)
```
YongxueHong, line 2222: How about? 
```
                        if opt[0] == "cert-subject":
                            value = value.replace('/', ',')
                        if opt[0] == 'fd':
                            value = str(value)
```
since the previous codes,  ` if re.match("^[0-9]+$", value)` deal with the general situation, the specific situation is moved bellow them, such as `if opt[0] == "cert-subject":` so I think that `fd` option should be handled same as `cert-subject` to keep the previous code design.

Thanks.

yanglei-rh:
LGTM
Acked-by: Lei Yang

Test Pass:
RHEL7.9
(1/6) Host_RHEL.m7.u9.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.9.x86_64.io-github-autotest-qemu.multi_nics_hotplug.nic_8139: PASS (801.81 s)
 (2/6) Host_RHEL.m7.u9.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.9.x86_64.io-github-autotest-qemu.multi_nics_hotplug.nic_virtio: PASS (678.59 s)
 (3/6) Host_RHEL.m7.u9.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.9.x86_64.io-github-autotest-qemu.multi_nics_hotplug.nic_e1000: PASS (684.95 s)
 (4/6) Host_RHEL.m7.u9.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.9.x86_64.io-github-autotest-qemu.multi_nics_hotplug.nic_e1000e: PASS (681.13 s)
 (5/6) Host_RHEL.m7.u9.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.9.x86_64.io-github-autotest-qemu.multi_nics_hotplug.combination: PASS (937.97 s)
 (6/6) Host_RHEL.m7.u9.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.9.x86_64.io-github-autotest-qemu.multi_nics_hotplug.sameid: PASS (378.94 s)
RHEL8.3
(1/5) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.multi_nics_hotplug.nic_8139.q35:PASS (291.59 s)
 (2/5) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.multi_nics_hotplug.nic_virtio.q35: PASS (281.35 s)
 (3/5) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.multi_nics_hotplug.nic_e1000e.q35: PASS (686.05 s)
 (4/5) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.multi_nics_hotplug.combination.q35: PASS (1048.34 s)
 (5/5) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.multi_nics_hotplug.sameid.q35: PASS (172.90 s)

virttest/qemu_monitor.py
yanan-fu: @YongxueHong  Eh~  the solution above will cause `fd` go into two blocker.
First,  `value = int(value)` as it is all digital, then `str(value)` ad it is `fd`.
 
I think it is a little redundant.  I prefer to exclude `fd` when it is all digital. Thanks!

yanan-fu: @YongxueHong  Do you have other comment for this issue ?

YongxueHong: > @YongxueHong Eh~ the solution above will cause `fd` go into two blocker.
> First, `value = int(value)` as it is all digital, then `value = str(value)` as it is `fd`.
> 
> I think it is a little redundant. I prefer to exclude `fd` when it is all digital. Thanks!


If you think it is a redundant, how about` "cert-subject" `? Is the` if opt[0] == "cert-subject":` a redundant? why didn't put 
```
                        if opt[0] == "cert-subject":
                            value = value.replace('/', ',')
```
 into `if re.match("^[0-9]+$", value):` blocker ? and `"cert-subject"` also goes into two blockker.
So I think we should keep consistency.
Either keep the previous code design or keep your present code design.

yanan-fu: @YongxueHong  That is because i did not check the specific value of `cert-subject` before which is not related with this patch ~

I just checked, it is not all digital, so can not move into the `if re.match("^[0-9]+$", value):` blocker, so ~
Could you explain where you get it can be moved in to the all digital blocker ?

YongxueHong: > @YongxueHong That is because i did not check the specific value of `cert-subject` before which is not related with this patch ~
> 
> I just checked, it is not all digital, so can not move into the `if re.match("^[0-9]+$", value):` blocker, so ~
> Could you explain where you get it can be moved in to the all digital blocker ?

if `opt[0] == 'fd' `, I think the upper user must pass a string like '12334', so it will re.match("^[0-9]+$", value) when the `opt[0] == 'fd'`

yanan-fu: `cert-subject` is something include digital, letter and other character such as `=`, `/`. It does not belong to any if block before it, so only one time ~

Here is an example: `spice_x509_server_subj = /C=CZ/L=BRNO/O=SPICE/CN=my Server`
```
cert_s = clone.spice_options.get("spice_x509_server_subj",  "")                            
cert_subj = "%s" % cert_s[1:]
cert_subj += host_ip
cert_subj = "\"%s\"" % cert_subj
```

yanan-fu:
@YongxueHong  Replace `not (opt[0] == 'fd')`  to `opt[0] != 'fd'`, could you help review again, thanks ~

virttest/qemu_monitor.py
YongxueHong: > `cert-subject` is something include digital, letter and other character such as `=`, `/`. It does not belong to any if block before it, so only one time ~
> 
> Here is an example: `spice_x509_server_subj = /C=CZ/L=BRNO/O=SPICE/CN=my Server`
> 
> ```
> cert_s = clone.spice_options.get("spice_x509_server_subj",  "")                            
> cert_subj = "%s" % cert_s[1:]
> cert_subj += host_ip
> cert_subj = "\"%s\"" % cert_subj
> ```

Yes, but it belongs to else blocker:
```
                        else:
                            value = value.strip()
                        if opt[0] == "cert-subject":
                            value = value.replace('/', ',')
```
According to your design, this place should be 
```
                        else:
                            value = value.strip()
                            if opt[0] == "cert-subject":
                                value = value.replace('/', ',')
```
Although I don't have an ideat about the reason why it did not do that.

PR #2625, opened 2020-07-03
Fix error of replacing exit_status with stdout

chloerh:
The replacements of removing 52lts bring in this error. According
to context, it should be exit_status, not stdout_text.


chloerh:
PASS 1-type_specific.local.libvirt_vcpu_plug_unplug.positive_test.vcpu_set.guest.vm_operate.no_operation

PR #2631, opened 2020-07-06
Update LibvirtdDebugLog to enable debuging log for virt* daemons

Yingshun:

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -36,6 +36,7 @@ from virttest import utils_config
 from virttest import utils_package
 from virttest import kernel_interface
 from virttest import libvirt_version
+from virttest import libvirtd_decorator
 from virttest.staging import service
 from virttest.staging import utils_memory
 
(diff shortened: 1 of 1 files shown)


Yingshun:
depends on https://github.com/avocado-framework/avocado-vt/pull/2628

Yingshun:
depends on https://github.com/avocado-framework/avocado-vt/pull/2646

virttest/test_setup.py
```diff
@@ -2423,38 +2424,71 @@ def __init__(self, test, log_level="1", log_file=""):
         self.log_level = log_level
         self.log_file = log_file
         self.test = test
-        self.libvirtd = utils_libvirtd.Libvirtd()
-        self.libvirtd_conf = utils_config.LibvirtdConfig()
-        self.backupfile = "%s.backup" % self.libvirtd_conf.conf_path
+        self.daemons_dict = {}
+        self.daemons_dict["libvirtd"] = {
+            "daemon": utils_libvirtd.Libvirtd("virtqemud"),
```
kylazhang, line 2429: I think here better remove the "virtqemud" to be more clear

dzhengfy: Here the purpose to use virtqemud is to set ' virtqemud' if modular daemon is enabled, otherwise set 'libvirtd'. 
If we use utils_libvirtd.Libvirtd(), all sub daemons will be used if modular daemon is enabled. So that is not her purpose, I think.

virttest/test_setup.py
```diff
@@ -2423,38 +2424,71 @@ def __init__(self, test, log_level="1", log_file=""):
...
+            "conf": utils_config.LibvirtdConfig(),
+            "backupfile": "%s.backup" % utils_config.LibvirtdConfig().conf_path}
+        if (libvirt_version.version_compare(5, 6, 0) and
+           libvirtd_decorator.get_libvirtd_split_enable_bit()):
```
dzhengfy, line 2433: To detect modular daemon environment, I suggest we use a single function instead of checking version and enable bit everywhere. But anyway we can leave it as is for now.

virttest/test_setup.py
```diff
@@ -2423,38 +2424,71 @@ def __init__(self, test, log_level="1", log_file=""):
...
+
+        for value in self.daemons_dict.values():
+            if os.path.isfile(value.get("backupfile")):
+                os.remove(value.get("backupfile"))
+
+            # backup config files before restarting services
+            try:
+                with open(value.get("backupfile"), "w") as fd:
+                    fd.write(value.get("conf").backup_content)
+                    fd.close()
```
dzhengfy, line 2478: Remove this line because we use 'with open' clause.

virttest/test_setup.py
```diff
@@ -2423,38 +2424,71 @@ def __init__(self, test, log_level="1", log_file=""):
...
+            except IOError as info:
+                self.test.error(info)
+
+            value.get("conf")["log_level"] = self.log_level
+            value.get("conf")["log_outputs"] = '"%s:file:%s"' % (self.log_level,
+                                                                 self.log_file)
```
dzhengfy, line 2484: It is ok to just share same 'self.log_level' among all sub daemons, but for self.log_file, it seems not very reasonable.

virttest/test_setup.py
Yingshun: Yes, I agree. We still need to update get_libvirtd_split_enable_bit() later, may be at that time, we can do it.

virttest/test_setup.py
Yingshun: Oops, thanks! done.

virttest/test_setup.py
Yingshun: Actually, this is to continue the previous design which keeps libvirtd.log if test fails or errors. I've talked about feature owner, at this stage, they prefer it.

dzhengfy: Ok. Let's see if there is an issue after some tests.

PR #2635, opened 2020-07-06
VMStress: Replace session.sendline to session.cmd

PaulYuuu:
session.sendline will send the command line and ignore its return,
sometimes this will prevent the pattern check of next command, use
"session.cmd" instead of it. Also, update the parent sendline with
"process.run".

ID: 1854114

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -30,6 +30,7 @@ import time
 import subprocess
 import shutil
 import ast
+import functools
 
 import aexpect
 from avocado.core import exceptions
@@ -2153,7 +2154,8 @@ class Stress(object):
         self.dst_path = self.params.get('stress_dst_path', '/home')
         self.cmd_status_output = process.getstatusoutput
         self.cmd_output_safe = process.getoutput
-        self.cmd_status = self.sendline = process.system
+        self.cmd_status = process.system
+        self.cmd = functools.partial(process.run, ignore_status=True)
         self.cmd_launch = os.system
 
     @session_handler
@@ -2187,7 +2189,7 @@ class Stress(object):
         """
 
         def _unload_stress():
-            self.sendline(self.stop_cmd)
+            self.cmd(self.stop_cmd)
             if not self.app_running():
                 return True
             return False
@@ -2390,7 +2392,7 @@ class VMStress(Stress):
         self.session = self.get_session()
         self.cmd_status_output = self.session.cmd_status_output
         self.cmd_status = self.session.cmd_status
-        self.sendline = self.session.sendline
+        self.cmd = functools.partial(self.session.cmd, ignore_all_errors=True)
         self.cmd_output_safe = self.cmd_launch = self.session.cmd_output_safe
 
     def get_session(self):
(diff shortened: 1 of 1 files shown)


PaulYuuu:
Hello @vivianQizhu 
Can you help to review? Thanks.

vivianQizhu:
@PaulYuuu I am not sure but it might break some test case with this change. It is intended to only send command without waiting for its return, we only want to launch the stress test in this module when design it, cause it might run a long time for some stress test, and we check the status outside but do not need to use another thread. Actually the beginning of the usage of this module indeed is to avoid the multi-threading for stress test, which is hard to control the order of the command execution between different thread, there were many bugs before when we use multi-thread in such cases.
I understand the issue you are trying to resolve, but maybe we could find another way to fix it, like clear the last line or type in a `\n` before next command is executed?

PaulYuuu:
> @PaulYuuu I am not sure but it might break some test case with this change. It is intended to only send command without waiting for its return, we only want to launch the stress test in this module when design it, cause it might run a long time for some stress test, and we check the status outside but do not need to use another thread. Actually the beginning of the usage of this module indeed is to avoid the multi-threading for stress test, which is hard to control the order of the command execution between different thread, there were many bugs before when we use multi-thread in such cases.
> I understand the issue you are trying to resolve, but maybe we could find another way to fix it, like clear the last line or type in a `\n` before next command is executed?

@vivianQizhu In fact, the checkpoint is `self.app_running()`, and we still have to wait for the return of `stop_cmd`, if `stop_cmd` hangs, the next command cannot be executed, because they are executed in the same session, but we will ignore its return status. So I add `ignore_all_errors=True` and `ignore_status=True`.

This check is under `unload_stress`, when a stress test is run with high CPU/memory usage, it may take a few seconds to stop, but I don't think it will break the test case.

vivianQizhu:
@PaulYuuu You are right, I just misread it, though it's about the loading the stress. 
So it looks good to me, would you give a test result of any test case utilizing it?

PaulYuuu:
@vivianQizhu I will do a regression test later.

PaulYuuu:
(1/7) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.cpu_device_hotpluggable.hotplug.multi_vcpu.with_stress: PASS (407.38 s)
 (2/7) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.virtual_nic.with_stress: PASS (1702.12 s)
 (3/7) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.hugepage_mem_stress: PASS (237.95 s)
 (4/7) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.balloon_stress: PASS (574.62 s)
 (5/7) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.linux_stress: PASS (229.90 s)
 (6/7) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.linux_stress.stressapptest: PASS (227.82 s)
 (7/7) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.linux_stress.stress-ng: PASS (182.51 s)

vivianQizhu:
Thanks @PaulYuuu

PR #2640, opened 2020-07-07
Fix on avocado "TestSuite() class"

anascko:
These updates related to avocado implementation:
https://github.com/avocado-framework/avocado/commit/e444cc530ed06658e0820fe2dd479bba10d16524

### DIFF
--- a/avocado_vt/plugins/vt_joblock.py
+++ b/avocado_vt/plugins/vt_joblock.py
@@ -114,9 +114,9 @@ class VTJobLock(Pre, Post):
 
     def pre_tests(self, job):
         try:
-            if job.test_suite is not None:
+            if job.test_suite.tests is not None:
                 if any(test_factory[0] is VirtTest
-                       for test_factory in job.test_suite):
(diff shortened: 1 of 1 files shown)


avocado_vt/plugins/vt_joblock.py
```diff
@@ -114,9 +114,9 @@ def _lock(self, job):
 
     def pre_tests(self, job):
         try:
-            if job.test_suite is not None:
+            if job.test_suite.tests is not None:
```
beraldoleal, line 117: Hi @anascko, thanks for catching this. I'm not sure here, because looks like we have to handle here on Avocado-vt both, the current model and the previous one, right?

So maybe look for `if hasattr(test_suite, 'tests') ...` not sure.

Also, see #2639 proposed by @clebergnu.

anascko: @beraldoleal thanks)
I missed this fix #2639

anascko:
#2639

PR #2652, opened 2020-07-10
Add loadparm attribute to disk xml

smitterl:
Allow for supported attribute //disk/boot@loadparm.

Signed-off-by: Sebastian Mitterle <smitterl@redhat.com>

PR #2653, opened 2020-07-13
Add method to handle all boot attributes

smitterl:
Depends on https://github.com/avocado-framework/avocado-vt/pull/2652

Currently, only //disk/boot@order can be set.
Add a method to allow also for //disk/boot@loadparm.

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -1642,6 +1642,24 @@ class VMXML(VMXMLBase):
                     device.boot = order
         self.set_devices(devices)
 
+    def set_boot_attrs_by_target_dev(self, target, **attrs):
+        """
+        Set boot attributes by target dev
+
+        :param target: The target dev on host
+        :param attrs: Dict of boot attributes
+        """
+        devices = self.get_devices()
+        for device in devices:
+            if device.device_tag == "disk":
+                if device.target.get("dev") == target:
+                    for name, value in list(attrs.items()):
+                        if name == "order":
+                            device.boot = value
+                        elif name == "loadparm":
+                            device.loadparm = value
+        self.set_devices(devices)
+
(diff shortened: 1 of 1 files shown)


smitterl:
New method is used in system-level testing of https://github.com/autotest/tp-libvirt/pull/2919

virttest/libvirt_xml/vm_xml.py
```diff
@@ -1642,6 +1642,24 @@ def set_boot_order_by_target_dev(self, target, order):
                     device.boot = order
         self.set_devices(devices)
 
+    def set_boot_attrs_by_target_dev(self, target, **attrs):
```
chunfuwen, line 1645: I assume set_boot_attrs_by_target_dev can cover previous set_boot_order_by_target_dev, do you want to deprecate it? Actually there are only 2 files ,which use set_boot_order_by_target_dev.

[root@chunfulap2 tp-libvirt]# find ./ -name "*.py" |xargs grep -i set_boot_order_by_target_dev
./libvirt/tests/src/bios/boot_integration.py:            vmxml.set_boot_order_by_target_dev(disk_target_dev, "1")
./libvirt/tests/src/bios/virsh_boot.py:        vmxml.set_boot_order_by_target_dev(target_dev, boot_order)

smitterl: @chunfuwen In terms of "don't break working code" I hadn't thought of removing the existing method. Do you think we should?

chunfuwen: From code maintainance pointview, Better have one single method existed. But as you mentioned, it may break working code. Therefore, I am fine with any choice you had.

smitterl: @chunfuwen Can you help merge?

PR #2654, opened 2020-07-14
utils_misc: remove line-feed for function get_full_pci_id

quanwenli:
id: 1856702

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -753,7 +753,7 @@ def get_full_pci_id(pci_id):
     """
     cmd = "lspci -D | awk '/%s/ {print $1}'" % pci_id
     try:
-        return process.run(cmd, shell=True).stdout_text
+        return process.run(cmd, shell=True).stdout_text.strip()
     except process.CmdError:
         return None
 


quanwenli:
@vivianQizhu, Could you help review it, thanks. btw, the check was failed, but it seems it's not the code issue.

vivianQizhu:
Wait, I see your log the error is "TypeError: join() argument must be str or bytes, not 'NoneType'", so the value of `full_id` is `None` there, how could you resolve this by strip it? @quanwenli

quanwenli:
@vivianQizhu , you can check the job.log, it's because the command line was wrong since “carriage return character” ， so full_id is None

2020-07-14 04:52:52,855 process          L0689 INFO | Command 'lspci -D | awk '/0000:06:00.0
/ {print $1}'' finished with 1 after 0.02002429962158203s
2020-07-14 04:52:52,871 output           L0660 DEBUG| [stdout] full_id=None

vivianQizhu:
> @vivianQizhu , you can check the job.log, it's because the command line was wrong since “carriage return character” ， so full_id is None
> 
> 2020-07-14 04:52:52,855 process L0689 INFO | Command 'lspci -D | awk '/0000:06:00.0
> / {print $1}'' finished with 1 after 0.02002429962158203s

So here should be the `pci_id` who has the "return character", and `process.run(cmd, shell=True)` should still fail since you does not change this command.

> 2020-07-14 04:52:52,871 output L0660 DEBUG| [stdout] full_id=None

quanwenli:
> It should not fix the issue properly.



> > @vivianQizhu , you can check the job.log, it's because the command line was wrong since “carriage return character” ， so full_id is None
> > 2020-07-14 04:52:52,855 process L0689 INFO | Command 'lspci -D | awk '/0000:06:00.0
> > / {print $1}'' finished with 1 after 0.02002429962158203s
> 
> So here should be the `pci_id` who has the "return character", and `process.run(cmd, shell=True)` should still fail since you does not change this command.
> 
> > 2020-07-14 04:52:52,871 output L0660 DEBUG| [stdout] full_id=None

since there is "return character" behind "0000:06:00.0", the command cut into two lines, so there is no correct return.

vivianQizhu:
> > It should not fix the issue properly.
> 
> > > @vivianQizhu , you can check the job.log, it's because the command line was wrong since “carriage return character” ， so full_id is None
> > > 2020-07-14 04:52:52,855 process L0689 INFO | Command 'lspci -D | awk '/0000:06:00.0
> > > / {print $1}'' finished with 1 after 0.02002429962158203s
> > 
> > 
> > So here should be the `pci_id` who has the "return character", and `process.run(cmd, shell=True)` should still fail since you does not change this command.
> > > 2020-07-14 04:52:52,871 output L0660 DEBUG| [stdout] full_id=None
> 
> since there is "return character" behind "0000:06:00.0", the command cut into two lines, so there is no correct return.

Yes, so what you should fix is "0000:06:00.0", the `pci_id`, not the `full_id`, which is `None`, it will still be None no matter you strip it or not.

PR #2671, opened 2020-07-28
Update customize_libvirt_config to set configurations for more daemons

Yingshun:
Some of cases need special settings, so update accordingly.


Yingshun:
@chloerh @dzhengfy  would you please help review? Thanks!

PR #2674, opened 2020-07-28
Add a method to check if modular daemon is enabled

Yingshun:

### DIFF
--- a/virttest/utils_split_daemons.py
+++ b/virttest/utils_split_daemons.py
@@ -388,3 +388,21 @@ def virtproxyd_stop():
 def virtproxyd_restart():
     virtproxyd_instance = VirtProxyd()
     return virtproxyd_instance.restart()
+
+
+def is_modular_daemon(session=None):
+    """
+    Check whether modular daemon is enabled
+
+    :params session: An session to guest or remote host
+    :return: True if modular daemon is enabled
+    """
+    daemons = ["virtqemud.socket", "virtinterfaced.socket",
+               "virtnetworkd.socket", "virtnodedevd.socket",
+               "virtnwfilterd.socket", "virtsecretd.socket",
+               "virtstoraged.socket", "virtproxyd.socket"]
+    if session:
+        runner = remote.RemoteRunner(session=session).run
+    else:
+        runner = process.run
+    return any([service.Factory.create_service(d, run=runner).status() for d in daemons])


virttest/utils_split_daemons.py
```diff
@@ -388,3 +388,21 @@ def virtproxyd_stop():
...
+
+
+def is_modular_daemon(session=None):
+    """
+    Check whether modular daemon is enabled
+
+    :params session: An session to guest or remote host
+    :return: True if modular daemon is enabled
+    """
+    daemons = ["virtqemud.socket", "virtinterfaced.socket",
```
chunfuwen, line 400: As I understand, checking virtqemud.socket  status only is enough.

Yingshun: Please have a look at 'Libvirt modular deamon follow up', it needs all of them.

chunfuwen: it is used to detect which mode current libvirtd is working, modular or monothlic ?  not checking all daemons work normally under modular mode.  So only virtqemud.socket check  is enough.
@yafu-1 ,please help check this.

Yingshun: Only for modular daemons, there are several ways to check if libvirtd is working, I don't want to add a new way.
Actually, it also needs virtproxyd.socket at least for some cases, so I don't think it's enough to only check virtqemud.socket.

virttest/utils_split_daemons.py
```diff
@@ -388,3 +388,21 @@ def virtproxyd_stop():
...
+               "virtnetworkd.socket", "virtnodedevd.socket",
+               "virtnwfilterd.socket", "virtsecretd.socket",
+               "virtstoraged.socket", "virtproxyd.socket"]
+    if session:
+        runner = remote.RemoteRunner(session=session).run
+    else:
+        runner = process.run
+    return all([service.Factory.create_service(d, run=runner).status() for d in daemons])
```
dzhengfy, line 408: I am not sure the reason to use 'create_service()' here. Is it only a way to provide a service in order to check its status?

virttest/utils_split_daemons.py
yafu-1: @Yingshun @chunfuwen 
It's not robust enough if just checking virtqemud. It can not ensure other split daemons are running besides virtqemud. So it's better to check all the split daemons.

Yingshun: Thanks for your updates! @yafu-1

virttest/utils_split_daemons.py
Yingshun: Yes, it's able to create different kinds of service managers and check the status.

dzhengfy, line 408: I think we should treat it as modular daemon environment in case any of sub services is running, because we can not say that is monothetic.

Yingshun: updated.

virttest/utils_split_daemons.py
chunfuwen: @yafu-1 
@Yingshun 
Check all split daemons status doesn't allow cover future possible scenarios where split daemon is started once required.

Yingshun: @chunfuwen Hmm, I think that's why checking virtqemud only is not sufficient. Actually we only have a 'full' split daemon scenario for now and decide to restart all sub daemons if need, so I think it's ok. And the purpose of this function is to check if it's a modular daemon env, so I think any() has already fixed your problem. How do you think?

chunfuwen:
Any validated result available?

Yingshun:
**Test result:** 
_on modular daemon env:_

```
JOB ID     : de5f67d437b61e4ce13028749791251be32ac4f9
JOB LOG    : /root/avocado/job-results/job-2020-07-28T22.32-de5f67d/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_bandwidth.precopy.set_speed_before_startup: PASS (265.96 s)
JOB TIME   : 268.33 s
```

_on monolith libvirt env:_

```
JOB LOG    : /root/avocado/job-results/job-2020-07-28T22.22-f1a5eab/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_bandwidth.precopy.set_speed_before_startup: PASS (262.75 s)
JOB TIME   : 264.15 s
```

PR #2679, opened 2020-07-29
env_process: Fix wrong usage of getting cpu arch/family

chloerh:
The function get_cpu_arch() is now deprecated on master branch of
avocado, and the original usage of it is not outdated, which means
we could no longer get the info we want (e.g. power8 or power9)
by calling the function. The function that could provide the info
is 'get_family()' now, But there's no function 'get_family()'
on 69lts branch which we're currently using. This fix is a
compromised way to fit with both branches for us to get certain
cpu info.

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -935,9 +935,10 @@ def preprocess(test, params, env):
         pvr = float(a_process.system_output(pvr_cmd, shell=True).strip())
         power9_compat = "yes" == params.get("power9_compat", "no")
 
-        if "power8" in cpu_arch:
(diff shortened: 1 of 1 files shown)


chloerh:
Related with https://github.com/avocado-framework/avocado/pull/3350

chloerh:
@sathnaga Would you mind taking a look?

sathnaga:
@chloerh this change is included in overall set of change to avocado-vt, https://github.com/avocado-framework/avocado-vt/pull/2455, could you please help review/merge that PR? Thanks in advance

chloerh:
The final solution should be decided, and this change should be included in https://github.com/avocado-framework/avocado-vt/pull/2455 , so close this one.

PR #2683, opened 2020-07-31
correct the usage of parameter cache-size

XueqiangWei:
cache-size is a qcow2 option, it cannot be given to any
imgfmt but qcow2.

ID: 1862315

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -2110,6 +2110,8 @@ class DevContainer(object):
             drv_extra_params = (_.split('=', 1) for _ in
                                 drv_extra_params.split(',') if _)
             for key, value in drv_extra_params:
+                if key == 'cache-size' and imgfmt != 'qcow2':
+                    continue
                 if Flags.BLOCKDEV in self.caps:
                     if key == 'discard':
                         value = re.sub('on', 'unmap', re.sub('off', 'ignore', value))


XueqiangWei:
(1/1) Host_RHEL.m8.u3.product_av.qcow2.virtio_blk.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.cache_sizes_test.q35: PASS (942.50 s)
 (1/1) Host_RHEL.m8.u3.product_av.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.cache_sizes_test.q35: PASS (445.53 s)

XueqiangWei:
@YongxueHong Please help review it. Many thanks.

virttest/qemu_devices/qcontainer.py
```diff
@@ -2110,6 +2110,8 @@ def define_hbas(qtype, atype, bus, unit, port, qbus, pci_bus, iothread,
             drv_extra_params = (_.split('=', 1) for _ in
                                 drv_extra_params.split(',') if _)
             for key, value in drv_extra_params:
+                if key == 'cache-size' and imgfmt != 'qcow2':
+                    continue
```
YongxueHong, line 2114: As for as know, The `cache-size` just works on `qcow2` now, so the user should have such this background about this.
in other words, should correct it in the case level not in the framework.
I think it's the user's job to decide whether to add the `cache-size` parameter according to the user requirement.
Thanks.

XueqiangWei: Yes, you are right. But for windows guest, although we add \"only qcow2\" in the case level, it will add winutils.iso directly. We have to skip cdrom for windows guest.

YongxueHong: Hi @XueqiangWei 
yes , I saw the related case: `cache_sizes_test`
My suggestion that modify case line 24 to :
```
    for cache_size, tag in zip(cache_sizes, initial_tag):
        params["drv_extra_params_%s" % tag] = "cache-size=%s" % cache_size
```
to specify the device you wanna set the` cache-size`.
Thanks.

XueqiangWei: Agree with you. It's better to update it in the case level. Thank you very much.

XueqiangWei:
I will fix it in the tp-qemu

PR #2684, opened 2020-07-31
Enable disk address specified

dzhengfy:
When creating disk xml, the 'address' is allowed to be specified.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1589,8 +1589,11 @@ def create_disk_xml(params):
     # Use auth_in_source to diff whether it is placed in source or disk itself.
     auth_in_source = params.get('auth_in_source')
     input_source_file = params.get("input_source_file")
+    disk_addr = params.get("disk_address")
     if snapshot_attr:
         diskxml.snapshot = snapshot_attr
+    if disk_addr:
+        diskxml.address = diskxml.new_disk_address(attrs=eval(disk_addr))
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1589,8 +1589,11 @@ def create_disk_xml(params):
     # Use auth_in_source to diff whether it is placed in source or disk itself.
     auth_in_source = params.get('auth_in_source')
     input_source_file = params.get("input_source_file")
+    disk_addr = params.get("disk_address")
     if snapshot_attr:
         diskxml.snapshot = snapshot_attr
+    if disk_addr:
+        diskxml.address = diskxml.new_disk_address(attrs=eval(disk_addr))
```
chunfuwen, line 1596: def new_disk_address(self, type_name='drive', **dargs):
        """
        Return a new disk Address instance and set properties from dargs
        """
        new_one = self.Address(type_name=type_name, virsh_instance=self.virsh)
        for key, value in list(dargs.items()):
            setattr(new_one, key, value)
        return new_one
*******************************************************************
Do we need give type_name as one input parameter?

dzhengfy: Yes. But I found I made a mistake that autotest/tp-libvirt#2963 does not use this PR ,so the problem was not exposed. Now I prefer to close it. I will reopen it whenever there is a need and I will update it again.

PR #2686, opened 2020-07-31
Create interface xml

dzhengfy:

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1513,6 +1513,35 @@ def create_controller_xml(contr_dict):
     return contr
 
 
+def create_iface_xml(iface_dict):
+    """
+    Create interface xml file
+
+    :param iface_dict: The dict params includes interfaces' configurations
+
+    :return: the interface object
+    """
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1513,6 +1513,35 @@ def create_controller_xml(contr_dict):
     return contr
 
 
+def create_iface_xml(iface_dict):
+    """
+    Create interface xml file
+
+    :param iface_dict: The dict params includes interfaces' configurations
+
+    :return: the interface object
```
chunfuwen, line 1522: change to the interface object representing iface xml ?

chunfuwen:
Any result available?

chunfuwen:
./libvirt/tests/src/virtual_network/iface_options.py:52:    def create_iface_xml(iface_mac):

Previously there exists one method in tp-libvirt, is it possible to make them compatible ? Maybe in the future we can remove the one defined in tp-libvirt.

dzhengfy:
As there is no big requirement to add this, I decide to close it.

PR #2690, opened 2020-08-04
Add attr 'port' to Interface class

chloerh:
- Add d dict type attr of Interface
- Update function modify_vm_iface() to support newly added attr


chloerh:
@Yingshun

PR #2694, opened 2020-08-06
Change cell_numa in CPUXML to add more subelements under it

jinqi7:

Since there are some sub elements added under "cell" elements, "caches" and its
attributes and sub elements, the current dict type is need to extend.

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -820,8 +820,9 @@ class VMXML(VMXMLBase):
                 vmxml['cpu'] = vmcpu_xml
             try:
                 vmcpu_xml = vmxml['cpu']
-                if (update_numa and vmxml.cpu.numa_cell):
-                    no_numa_cell = len(vmxml.cpu.numa_cell)
+                cellxmls = vmxml.cpu.get_numa_cell(no_cell=True)
+                if (update_numa and cellxmls):
+                    no_numa_cell = len(cellxmls)
                 elif numa_number is not None:
                     numa_number = int(numa_number)
                     if 0 < int(numa_number) <= vcpus:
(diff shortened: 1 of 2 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -1749,11 +1748,7 @@ def __init__(self, virsh_instance=base.virsh):
                                  forbidden=[],
                                  parent_xpath='/',
                                  tag_name='topology')
-        accessors.XMLElementList(property_name="numa_cell",
```
chunfuwen, line 1752: if numa_cell propery is removed from vm_xml, some cases in tp-libvirt, which is dependent on it, may fail.

jinqi7: The property can be added as required.Since the xml in numa/cell is becaming more complicated for new feature added,  it's need to be defined as a class.

jinqi7: Related tp-libvirt test scripts are changed in PR ttps://github.com/autotest/tp-libvirt/pull/2976

jinqi7:
> If there is no dependence between CellCacheXML and CellXML, I suggest you separate them into two PRs :one for CellXML, the other for CellCacheXML since any change in vm_xml need intensive testing.

CellXML & CellCacheXML has denpendence, the xml format is as below.
<cell xxx>
   <cache xxx/>
</cell>

virttest/libvirt_xml/vm_xml.py
```diff
@@ -1783,29 +1777,171 @@ def __init__(self, virsh_instance=base.virsh):
...
+                                       libvirtxml=self,
+                                       forbidden=[],
+                                       parent_xpath='/',
+                                       tag_name='line',
+                                       attribute='unit')
+                super(VMCPUXML.CellXML.CellCacheXML, self).__init__(
+                    virsh_instance=virsh_instance)
+                self.xml = '<cache/>'
+
+        def get_cellcache(self):
```
chunfuwen, line 1891: maybe need add doc string to explain what the function does here?

jinqi7: Added

chunfuwen (approved):

virttest/libvirt_xml/vm_xml.py
```diff
@@ -1783,29 +1777,196 @@ def __init__(self, virsh_instance=base.virsh):
         super(VMCPUXML, self).__init__(virsh_instance=virsh_instance)
         self.xml = '<cpu/>'
 
-    @staticmethod
-    def marshal_from_cell(item, index, libvirtxml):
+    # Sub-element of cpu/numa
+    class Numa_CellXML(VMXML):
```
chloerh, line 1781: Looks like this is not the usually way we try to handle a nested xml, may I ask the reason of that? 
According to our usually way, we could change 'numa_cell' from 'XMLElementList' to 'XMLElementNest' if the 'list' type of element no longer fits our need.  I suppose that 'numa_cell' element will still be nested under 'cpu', then wouldn't it make more sense to just use the old way of setting to 'XMLElementNest'? Not mention you have already setup the xml class.

jinqi7: Since there at least two levels' list, the usual way doesn't work well as I ever tried. And because it is a list , the "XMLElementNest" does not match aslo.

chloerh (changes requested):
Please hold off merging this one, I think I have another solution for this change.

chloerh:
Please refer to https://github.com/avocado-framework/avocado-vt/pull/2749  , that way, the behavior of cpu.numa_cell is more similar with current way, maybe it would bring less modifications on tp-libvirt.

sathnaga:
@jinqi7 with merge of #2749, I hope this PR needs to be closed, pls confirm and close it, thanks!

PR #2697, opened 2020-08-09
utils_test: Add new_xml_with_object() to avoid deleting the xml file

wang-shengfang:
Add the xml object reference to the function return value
to avoid triggering the GC and deleting the xml file
after calling the create xml function.

Signed-off-by: Wang Shengfang <wangshengfang@huawei.com>
Signed-off-by: Chen Qun <kuhn.chenqun@huawei.com>

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -213,6 +213,28 @@ class LibvirtNetwork(object):
             virsh.net_undefine(self.name)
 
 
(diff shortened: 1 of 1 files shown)


dzhengfy:
We need not to introduce a new class to solve this problem. See https://github.com/avocado-framework/avocado-vt/pull/2988. 
It works if we only return an object, instead of object.xml in related function.

PR #2698, opened 2020-08-09
Add the second way to create object secret

XueqiangWei:
The following two ways should all work well:
1. --object secret,id={s.aid},data={s.data}
2. --object secret,id={s.aid},file={s.filename}

ID: 1867037

### DIFF
--- a/shared/cfg/guest-hw.cfg
+++ b/shared/cfg/guest-hw.cfg
@@ -115,6 +115,8 @@ variants:
     - luks:
         image_format = luks
         image_secret = redhat
+        # image_secret_expression = data
+        # value is data or file, the default value is data
     - qcow2v3:
         image_format = qcow2
         image_extra_params = "compat=1.1"
--- a/virttest/qemu_storage.py
+++ b/virttest/qemu_storage.py
@@ -614,7 +614,10 @@ class QemuImg(storage.QemuImg):
     def _secret_objects(self):
         """All secret objects str needed for command line."""
         secret_objects = self.encryption_config.image_key_secrets
-        secret_obj_str = "--object secret,id={s.aid},data={s.data}"
+        if self.params.get("image_secret_expression") == "file":
+            secret_obj_str = "--object secret,id={s.aid},file={s.filename}"
+        else:
+            secret_obj_str = "--object secret,id={s.aid},data={s.data}"
         return [secret_obj_str.format(s=s) for s in secret_objects]
 
     @property
@@ -781,6 +784,8 @@ class QemuImg(storage.QemuImg):
                                 "Other errors may ensue")
                 os.makedirs(image_dirname)
 
+        if self.encryption_config.key_secret:
+            self.encryption_config.key_secret.save_to_file()
         msg = "Create image by command: %s" % qemu_img_cmd
         error_context.context(msg, logging.info)
         cmd_result = process.run(
@@ -788,8 +793,6 @@ class QemuImg(storage.QemuImg):
         if cmd_result.exit_status != 0 and not ignore_errors:
             raise exceptions.TestError("Failed to create image %s\n%s" %
                                        (self.image_filename, cmd_result))
-        if self.encryption_config.key_secret:
-            self.encryption_config.key_secret.save_to_file()
         cmd_result.stdout = cmd_result.stdout_text
         cmd_result.stderr = cmd_result.stderr_text
         return self.image_filename, cmd_result


XueqiangWei:
(1/1) Host_RHEL.m8.u3.product_av.luks.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.qemu_img_negative.luks_with_non_utf8_secret.q35: PASS (3.09 s)

XueqiangWei:
@zhencliu Please help review it. Many thanks.

virttest/storage.py
```diff
@@ -336,8 +336,12 @@ def image_secret_define_by_params(cls, image, params):
     def save_to_file(self):
         """Save secret data to file."""
         _make_secret_dir()
-        with open(self.filename, "w") as fd:
-            fd.write(self.data)
+        try:
+            with open(self.filename, "w") as fd:
+                fd.write(self.data)
+        except TypeError:
+            with open(self.filename, "wb") as fd:
```
zhencliu, line 343: do we need this? this file likes a cfg file, created by VT, so we might not need to consider the binary file.

XueqiangWei: Yes, we do. For non utf8 string, we have to store it as binary file. And then create object secrect by \"--object secret,id={s.aid},file={s.filename}\". Thanks.

virttest/qemu_storage.py
```diff
@@ -561,7 +561,10 @@ def _backing_access_secrets(self):
     def _secret_objects(self):
         """All secret objects str needed for command line."""
         secret_objects = self.encryption_config.image_key_secrets
-        secret_obj_str = "--object secret,id={s.aid},data={s.data}"
+        if self.params.get("secret_obj_file") == "yes":
```
zhencliu, line 564: if we introduce a new param, just like Xu Han said, we'd better add the param's namespace, here secret_obj_file is used for image, so image_secret_obj_file might be better. Also, it's better to add it and comment it in shared/cfg/guest-hw.cfg, under image_secret.

Here is another concern from curl backend, just FYI.
You know, for http/https, we can pass cookie with a secret object, it also uses the format of '--object secret, data=xxxx' to define the secret object, but when we associate the secret object with an image in command line, we have to use 'cookie-secret=sec_id' instead of 'key-secret=id' or 'password-secret=id', what I mean is, although when defining a luks format, we now only use either 'key-secret' or 'password-secret', but we might introduce a new 'xxx-secret' in future, just like 'cookie-secret' for http, at that time, a flag param 'secret_obj_file' cannot cover all scenarios(it only covers two: yes--'password-secret', no----'key-secret'), so what about introducing a param like 'image_secret_type', now it has two values: 'key' and 'password', and it defaults to 'key' if it's not defined. we can easily enhance it a new value 'xxxx', used for 'xxxx-secret' in future.

XueqiangWei: updated, many thanks.

XueqiangWei:
@luckyh Please help review it. Many thanks.

virttest/qemu_storage.py
```diff
@@ -728,15 +731,15 @@ def create(self, params, ignore_errors=False):
                                 "Other errors may ensue")
                 os.makedirs(image_dirname)
 
+        if self.encryption_config.key_secret:
+            self.encryption_config.key_secret.save_to_file()
```
luckyh, line 788: Would you please elaborate on this change?

virttest/storage.py
luckyh, line 390: Would you please provide the error message you met? it should be either opened with or without the `b` flag, not just try one then fall back to another when the attempt failed.

shared/cfg/guest-hw.cfg
```diff
@@ -115,6 +115,10 @@ variants:
     - luks:
         image_format = luks
         image_secret = redhat
+        # image_secret_obj_file
```
luckyh, line 118: Considering that if there are other expressions of the secret data, bool values is not good to handle that, thus suggest using enum values, for example `image_secret_expression_stg1 = file; image_secret_expression_stg2 = data`

virttest/qemu_storage.py
XueqiangWei: There are two ways to create luks image.
1. \# qemu-img create --object secret,id=image1_encrypt0,data=redhat -f luks -o key-secret=image1_encrypt0 /home/test.luks 1G
2. \# qemu-img create --object secret,id=image1_encrypt0,file=/home/test.secret -f luks -o key-secret=image1_encrypt0 /home/test.luks 1G

If we use the second way to create luks image, the test.secret must be saved before image creation. Currently, saving key_secret to file after image creation.

luckyh: oh, I see, thanks for your elaboration!

virttest/storage.py
XueqiangWei: When create luks image with non_utf8_secret, hit this issue.
\# echo -n -e '\x3a\x3c\x3b\xff' > non_utf8_secret
# qemu-img create -f luks --object secret,id=sec0,file=non_utf8_secret -o key-secret=sec0 test.luks 4M

For non utf8 string, we have to store it as binary file. And then create object secrect by "--object secret,id={s.aid},file={s.filename}". So retry it with \"wb\".

shared/cfg/guest-hw.cfg
XueqiangWei: OK, I will update it. Thanks.

virttest/storage.py
luckyh: sorry, I don't quite understand your issue, what content do you want to store, bytes or texts?

XueqiangWei: The content was utf8 string, and stored as texts by default.  Just for non utf8 string testing, hit TypeError when saved it as texts. So store it as a binary file.

TypeError: write() argument must be str, not bytes

luckyh: 1. currently, the `data` attribute of the `ImageSecret` class only supports `str` values (although it is not documented), that means you have to restructure it if you want to support `bytes` values
2. `open` supports specifying encoding type on python3 ([ref](https://docs.python.org/3/library/functions.html#open)), which means you can get non-utf8 files without using binary mode, e.g.
```
$ cat test.py
with open('/tmp/secret', 'w', encoding='euc_jp') as f:
    f.write('\x3a\x3c\x3b\xff')
$ python3 test.py
$ qemu-img create -f luks --object secret,id=sec0,file=/tmp/secret -o key-secret=sec0 test.luks 4M
Formatting 'test.luks', fmt=luks size=4194304 key-secret=sec0
qemu-img: test.luks: Data from secret sec0 is not valid UTF-8
```
So, the point is, no matter what you want is 1 or 2, `ImageSecret` (or accurately `ImageSecret.data`) have to be restructured, instead of simply using the current approach.

XueqiangWei: Thanks for your detailed explanation.
Since non-utf8 file is negative test, I think set a default parameter \"encoding\" to save_to_file is OK.
```
def save_to_file(self, encoding="utf-8"):
        """Save secret data to file."""
        _make_secret_dir()
        with open(self.filename, "w", encoding=encoding) as fd:
            fd.write(self.data)
```

And add a parameter \"image_secret_encoding_type\", in order to specify encoding type in test case.  
```
if self.encryption_config.key_secret:
     encoding = params.get(\"image_secret_encoding_type\", \"utf-8\")
     self.encryption_config.key_secret.save_to_file(encoding=encoding)
```

Do you agree with the above update? thanks

XueqiangWei:
Test it with local fs, iscsi backend and do negative test, all work well.
(1/1) Host_RHEL.m8.u3.product_av.luks.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.qemu_img_negative.luks_with_non_utf8_secret.q35: PASS (3.17 s)
(1/1) Host_RHEL.m8.u3.product_av.iscsi_direct.luks.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.boot.q35: PASS (51.37 s)
(1/1) Host_RHEL.m8.u3.product_av.luks.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.boot.q35: PASS (73.88 s)

XueqiangWei:
@luckyh Please help review it again. Many thanks.

zhencliu:
hi @luckyh , in fact, @XueqiangWei and I talked about this patch on wechat, from my side, this is a negative case and there is no new case planned in future, so IMO, can we just form the qemu-img command in tp-qemu and not do VT code change? It might be too heavy for this specific negative test case.

luckyh:
> hi @luckyh , in fact, @XueqiangWei and I talked about this patch on wechat, from my side, this is a negative case and there is no new case planned in future, so IMO, can we just form the qemu-img command in tp-qemu and not do VT code change? It might be too heavy for this specific negative test case.

Agree with you, if so let's drop the VT fix, thanks!

XueqiangWei:
> > hi @luckyh , in fact, @XueqiangWei and I talked about this patch on wechat, from my side, this is a negative case and there is no new case planned in future, so IMO, can we just form the qemu-img command in tp-qemu and not do VT code change? It might be too heavy for this specific negative test case.
> 
> Agree with you, if so let's drop the VT fix, thanks!

If then, 
1. I have to write hard code to  qemu-img command in tp-qemu test case \"image_creation_luks_with_non_utf8_secret\".
2. I think just don't change save_to_file related codes. The way to create object secret by file is regular function, also need to add it.
@luckyh @zhencliu Do you agree?

zhencliu:
> > > hi @luckyh , in fact, @XueqiangWei and I talked about this patch on wechat, from my side, this is a negative case and there is no new case planned in future, so IMO, can we just form the qemu-img command in tp-qemu and not do VT code change? It might be too heavy for this specific negative test case.
> > 
> > 
> > Agree with you, if so let's drop the VT fix, thanks!
> 
> If then,
> 
> 1. I have to write hard code to  qemu-img command in tp-qemu test case "image_creation_luks_with_non_utf8_secret".
> 2. I think just don't change save_to_file related codes. The way to create object secret by file is regular function, also need to add it.
>    @luckyh @zhencliu Do you agree?

Hi @XueqiangWei , I see, a file-store secret is supported by qemu, and we can simply call QemuImg.create to create the image for your non-utf8 case, quit straightforward. Just one concern:
if we support file-store secret for qemu-img, it might be better for us to support file-store secret for qemu-kvm(keep qemu function consistent), i.e. when we create a file-store secret luks image, and use it as an image of qemu-kvm, we also should use the file-store secret instead of the data-store one, so I'm afraid we also have to make code change to qcontainer.py(images_define_by_variables). As we cannot pass 'params' to images_define_by_variables, we might have to change storage.ImageSecret to hold the 'file' or 'data'  secret, all might be a big code change. Would @luckyh please make more comments on this?

XueqiangWei:
@zhencliu thanks for your reply, I think the following codes should be OK. Let's wait for @luckyh 's comment. I will update the case under tp-qemu first. Thanks.

```
if self.params.get("image_secret_expression") == "file":
    secret_obj_str = "--object secret,id={s.aid},file={s.filename}"
else:
    secret_obj_str = "--object secret,id={s.aid},data={s.data}"
```

XueqiangWei:
Close it first, I will update it when I finish the debugging. @luckyh @zhencliu Thanks for your help.

PR #2699, opened 2020-08-10
NumaInfo: add online_nodes_cpus to collect cpus for every node

yanan-fu:
id: 1859028

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -1596,8 +1596,10 @@ class NumaInfo(object):
         self.online_nodes_withcpu = self.get_online_nodes_withcpu()
         self.online_nodes_withcpumem = list(set(self.online_nodes_withcpu) &
                                             set(self.online_nodes_withmem))
+        self.online_nodes_meminfo = self.get_all_node_meminfo()
         self.nodes = {}
         self.distances = {}
+        self.online_nodes_cpus = self.get_all_node_cpus()
 
         # ensure numactl package is available
         if not utils_package.package_install('numactl', session=self.session):
@@ -1702,6 +1704,30 @@ class NumaInfo(object):
         :rtype: string
         """
         return self.get_all_node_meminfo()[node_id][key]
+
+    def get_all_node_cpus(self):
+        """
+        Get the cpus of all online nodes.
+
+        :return: All nodes' cpus
+        :rtype: Dict
+        """
+        cpus = {}
+        cpulist_file = os.path.join(self.numa_sys_path, "node%s/cpulist")
+        for node in self.get_online_nodes():
+            numa_sys = kernel_interface.SysFS(cpulist_file % node,
+                                              session=self.session)
+            node_cpus = ""
+            convert_list = re.findall("(\d+-\d+|\d+)", str(numa_sys.sys_fs_value))
+            for cstr in convert_list:
+                start = min(int(cstr.split("-")[0]),
+                            int(cstr.split("-")[-1]))
+                end = max(int(cstr.split("-")[0]),
+                          int(cstr.split("-")[-1]))
+                for n in range(start, end + 1, 1):
+                    node_cpus += " %s" % str(n)
+            cpus[node] = node_cpus
+        return cpus
 
     def get_online_nodes_withmem(self):
         """


virttest/kernel_interface.py
```diff
@@ -82,7 +82,7 @@ def proc_fs_value(self):
         try:
             return int(self.fs_value)
         except ValueError:
-            return self.fs_value
+            return self.fs_val
```
huangyum, line 85: Do you need add self.fs_val in init function?

yanan-fu: > Do you need add self.fs_val in init function?

I think no need,  the function `fs_value` do it.

huangyum: I see. Had better update other places as well, like L98 and L154.

yanan-fu: > I see. Had better update other places as well, like L98 and L154.

We can't do that,  L98 and L154 is to check if the new value be setted successfully ~

huangyum: Yeah, you are right.

yanan-fu:
@vivianQizhu  Could you help review ? thanks ~

virttest/kernel_interface.py
vivianQizhu, line 85: Good catch.

virttest/utils_misc.py
```diff
@@ -1697,7 +1700,7 @@ def read_from_node_meminfo(self, node_id, key):
         :return: The value in KB
         :rtype: string
         """
-        return self.get_all_node_meminfo()[node_id][key]
+        return self.online_nodes_meminfo[node_id][key]
```
vivianQizhu, line 1703: If it possible that these data are dynamically changed something, during some test, like hot plug, unplug something, thus maybe some users need these functions provide them the real time data?

yanan-fu: > If it possible that these data are dynamically changed something, during some test, like hot plug, unplug something, thus maybe some users need these functions provide them the real time data?

A good point ~ I think it is possible no matter such a case exist or not.  @huangyum  What's your opinion ?

huangyum: Yes, that's possible as well as online_nodes_withmem and online_nodes_withcpu. In that case, we have to call `self.online_nodes_meminfo = self.get_all_node_meminfo()` first to get latest status.

yanan-fu: With this, this origin purpose of this patch has gone ~
Let's fix the kernel_interface issue with this patch, thanks!

huangyum: I don't understand. Why do you revert read_from_node_meminfo?

yanan-fu: > I don't understand. Why do you revert read_from_node_meminfo?

The all_node_meminfo may change after hotplug/unplug,  to get the latest info, we have to call `get_all_node_meminfo` to get the latest online node. Isn't it ?

huangyum: Yes, but it only happens in hotplug/unplug senario. In most cases, it doesn't change.

yanan-fu: > Yes, but it only happens in hotplug/unplug senario. In most cases, it doesn't change.

The new change is to cover hotplug/unplug, so i revert it.

huangyum: That's not very convincible to me.  The scenario can still be covered without revert.

yanan-fu: > That's not very convincible to me. The scenario can still be covered without revert.

How can it be covered with hotplug ?  After hotplug/unplug the online nodes may changed, means the `online_nodes_meminfo` may be changed too.

If you are talking about use `get_all_node_meminfo()` in test case, and get special node info from it in test case, that is another usage i think.

huangyum: No, I meat to refresh online_nodes_meminfo by `self.online_nodes_meminfo = self.get_all_node_meminfo()` after hotplug/unplug.
The process is like this:

numaInfo = class NumaInfo()
hotplug/unplug..
numaInfo.online_nodes_meminfo = numaInfo.get_all_node_meminfo()
mem_info_node = numaInfo.read_from_node_meminfo(node, xx)

huangyum: Or we could create a new instance for class NumaInfo after hotplug/unplug.

yanan-fu: > No, I meat to refresh online_nodes_meminfo by `self.online_nodes_meminfo = self.get_all_node_meminfo()` after hotplug/unplug.
> The process is like this:
> 
> numaInfo = class NumaInfo()
> hotplug/unplug..
> numaInfo.online_nodes_meminfo = numaInfo.get_all_node_meminfo()
> mem_info_node = numaInfo.read_from_node_meminfo(node, xx)

Why refresh `online_nodes_mem_info`, without the refresh we still can get the latest info with `read_from_node_meminfo` ?
For this, my expect usage after hotplug/unplug is:
```
numaInfo = class NumaInfo()
hotplug/unplug
mem_info_node = numaInfo.read_from_node_meminfo(node, xx)
```

huangyum: Then it comes to the original question. You will call get_all_node_meminfo() many times as node number, which is unnecessary.

yanan-fu: > Then it comes to the original question. You will call get_all_node_meminfo() many times as node number, which is unnecessary.

Eh~  after considering the hotplug/unplug scenario, IMO, this usage already be dropped, that is why i revert the change. and do not plan to use `read_from_node_meminfo`:

Not only one way to avoid the repeated call, and here, i think we need keep consistent, the functions should all use the latest.
In test case, without use `read_from_node_meminfo` is not a great loss:
```
numa_info = NumaInfo(***)
numa_info.online_nodes_meminfo[node_id][key]
```

huangyum: That could work too. I'm not sure which is better. Up to you.

yanan-fu: > That could work too. I'm not sure which is better. Up to you.

Thanks, let's keep consistent in this class, all use latest, thanks~
Do you have other comments ?

virttest/utils_misc.py
```diff
@@ -1699,6 +1701,19 @@ def read_from_node_meminfo(self, node_id, key):
...
+        """
+        Get the cpus of all online nodes.
+
+        :return: All nodes' cpus
+        :rtype: Dict
+        """
+        cpus = {}
+        node = NumaNode(session=self.session)
+        for node_id in self.get_online_nodes():
+            cpus[node_id] = node.get_node_cpus(node_id)
```
huangyum, line 1714: This function will run 'numactl -H'. To get all node cpus info, we only need run it one time. I suggest either not to call get_nodes_cpus() or update it.

virttest/utils_misc.py
huangyum: Pls see L1714.

virttest/utils_misc.py
yanan-fu: Then it go back to the initial version of the tp-qemu one, but just move part of them from tp-qemu to vt...

I am sure i will not update the class NumaNode which effect many other test while has little to do with this test.
Update this function to get the all the numa cpus without call `get_nodes_cpus()` is another kind of duplication.
Base on the current design we can not avoid all repeated call, which one is acceptable for you ?

yanan-fu: I choose rewrite function `get_all_node_cpus`, do not call `get_nodes_cpus()`. 
Both `numactl -H` and `cpulist` can achieve the purpose and i choose the second one, please help review again, thanks!

virttest/utils_misc.py
```diff
@@ -1703,6 +1705,30 @@ def read_from_node_meminfo(self, node_id, key):
         """
         return self.get_all_node_meminfo()[node_id][key]
 
+    def get_all_node_cpus(self):
```
xutian, line 1708: rename to get_online_cpus ?

yanan-fu: > rename to get_online_cpus ?

Yes, you are right, it is about `online` cpus,  and it is a node and cpu mapping, i want to "node". 
And keep consistent with `get_all_node_meminfo`, i prefer the current name more ~   how about keep it ?

sathnaga (approved):
LGTM

vivianQizhu:
@huangyum Request for your re-review, thanks.

huangyum (approved):
ACK

vivianQizhu:
Thanks all.

PR #2702, opened 2020-08-12
Add run_tmux_cmd() to execute some shell-command in tmux

Yingshun:
Some cases need to run command line in tmux, so add a method.

### DIFF
--- a/virttest/utils_tmux.py
+++ b/virttest/utils_tmux.py
@@ -1,0 +1,26 @@
+"""
+The tmux command related utility functions
+"""
+
+import logging
+
+from virttest import utils_misc
+from virttest import utils_package
+
+
+def run_tmux_cmd(cmd, session=None, ignore_status=False):
+    """
+    Execute shell-commands in tmux on localhost or remote
+
+    :param cmd: The command line to run
+    :param session: The session object to the host
+    :param ignore_status: Whether to raise an exception when command fails
+    :return: command status and output
+    """
+    if not utils_package.package_install("tmux", session):
+        logging.error("Failed to install required package - tmux!")
+    tmux_cmd = 'tmux -c "{}"'.format(cmd)
+
+    return utils_misc.cmd_status_output(tmux_cmd, shell=True, verbose=True,
+                                        ignore_status=ignore_status,
+                                        session=session)


Yingshun:
**Test result:**
` (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_network.ovs_br.without_postcopy: PASS (316.58 s)`

virttest/utils_tmux.py
```diff
@@ -0,0 +1,23 @@
...
+
+def run_tmux_cmd(cmd, session=None):
+    """
+    Execute shell-commands in tmux on localhost or remote
+
+    :param cmd: The command line to run
+    :param session: The session object to the host
+    """
+    if not utils_package.package_install("tmux", session):
+        logging.error("Failed to install required package - tmux!")
```
chunfuwen, line 19: If tmux is necessary package , better raise exception directly here.

Yingshun: Actually, I didn't not raise an exception here on purpose, the output of tmux command line will show the details.

chunfuwen: a little confused here, if tmux is not installed successfully and program doesn't terminate here, that means utils_misc.cmd_status_output will be executed without tmux command existence

Yingshun: Yes, I'd like to check the result of utils_misc.cmd_status_output's return.

virttest/utils_tmux.py
```diff
@@ -0,0 +1,26 @@
...
+    Execute shell-commands in tmux on localhost or remote
+
+    :param cmd: The command line to run
+    :param session: The session object to the host
+    :param ignore_status: Whether to raise an exception when command fails
+    :return: command status and output
+    """
+    if not utils_package.package_install("tmux", session):
+        logging.error("Failed to install required package - tmux!")
+    tmux_cmd = 'tmux -c "{}"'.format(cmd)
```
dzhengfy, line 22: The target is to add 'tmux -c' before a command. So I do not think we need to add a new function. What we need to do is only to combine 'tmux -c' with your target command together.

dzhengfy: Generally, I indent to close this PR.

Yingshun: Ok, since it's a simple function, I will do it in my tp cases.

PR #2704, opened 2020-08-12
Add one method to clean up existed nwfilter binding list

chunfuwen:

### DIFF
--- a/virttest/utils_libvirt/libvirt_nwfilter.py
+++ b/virttest/utils_libvirt/libvirt_nwfilter.py
@@ -1,0 +1,30 @@
+"""
+Accommodate libvirt nwfilter utility functions.
+
+:copyright: 2020 Red Hat Inc.
+"""
+
+import logging
+
+from virttest import virsh
+
+
+def clean_up_nwfilter_binding(ignore_status=False):
+    """
+    Clean up existed nwfilter binding.
+
+    :param ignore_status: default value True to allow silent failure.
+    """
+    cmd_result = virsh.nwfilter_binding_list(debug=True)
+    binding_list = cmd_result.stdout_text.strip().splitlines()
+    binding_list = binding_list[2:]
+    result = []
+    # If binding list is not empty.
+    if binding_list:
+        for line in binding_list:
+            # Split on whitespace, assume 1 column
+            linesplit = line.split(None, 1)
+            result.append(linesplit[0])
+    logging.info("existed nwfilter binding list is: %s", result)
+    for binding_uuid in result:
+        virsh.nwfilter_binding_delete(binding_uuid, ignore_status=ignore_status)


chunfuwen:
Running virsh command: nwfilter-binding-list 
Running '/bin/virsh nwfilter-binding-list '
[stdout]  Port Dev              Filter               
[stdout] ------------------------------------------------------------------
Command '/bin/virsh nwfilter-binding-list ' finished with 0 after 0.01506805419921875s
[stdout]  vnet0                 clean-traffic       
[stdout] 
status: 0
stdout: Port Dev              Filter               
------------------------------------------------------------------
 vnet0                 clean-traffic
stderr: 
existed nwfilter binding list is: ['vnet0']
new interface xml is: <?xml version='1.0' encoding='UTF-8'?>
<interface type="bridge">
      <mac address="52:54:00:f3:74:25" />
      <source bridge="virbr0" />
      <model type="virtio" />
      <address bus="0x00" domain="0x0000" function="0x0" slot="0x03" type="pci" />
    <filterref filter="allow-arp" /></interface>
Undefine VM avocado-vt-vm1
Define VM from /var/tmp/xml_utils_temp_16q314gd.xml



[stdout] 
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.nwfilter_update_vm_running.update_arp_rule

virttest/utils_test/libvirt.py
```diff
@@ -3742,3 +3742,27 @@ def check_dumpxml(vm, content, err_ignore=False):
...
+    # If binding list is not empty.
+    if binding_list:
+        for line in binding_list:
+            # Split on whitespace, assume 1 column
+            linesplit = line.split(None, 1)
+            result.append(linesplit[0])
+    logging.info("existed nwfilter binding list is: %s", result)
+    for binding_uuid in result:
+        try:
+            virsh.nwfilter_binding_delete(binding_uuid)
```
kylazhang, line 3765: I think it should add ignore_status=False?

virttest/utils_test/libvirt.py
```diff
@@ -3742,3 +3742,27 @@ def check_dumpxml(vm, content, err_ignore=False):
...
+        except Exception as e:
+            logging.error("Exception thrown while undefining nwfilter-binding: %s", str(e))
+            raise
```
kylazhang, line 3768: Could just use
```
for binding_uuid in result:
    virsh.nwfilter_binding_delete(binding_uuid, ignore_status=False)
```
meet your need? As it will call process.run and give clear cmd and result error log

virttest/utils_test/libvirt.py
chunfuwen: updated

virttest/utils_test/libvirt.py
chunfuwen: remove raise statement in order to allow it silently failed once happened.
The reason behind it is that it is just clean up functionality, we can not stop or fail case execution due to clean up failures.

chunfuwen:
The code is now moved to new file :virttest/utils_libvirt/utils_libvirt_nwfilter.py .

virttest/utils_libvirt/utils_libvirt_nwfilter.py
```diff
@@ -0,0 +1,32 @@
...
+
+:copyright: 2020 Red Hat Inc.
+"""
+
+import logging
+
+from virttest import virsh
+
+
+def clean_up_nwfilter_binding():
```
sathnaga, line 12: consider providing the option for ignore_status to user via param here and default it it False and return the processed result.

sathnaga (changes requested):

dzhengfy:
@chunfuwen , please help add some PR description and commit message, thanks.

virttest/utils_libvirt/utils_libvirt_nwfilter.py
```diff
@@ -0,0 +1,32 @@
...
+        for line in binding_list:
+            # Split on whitespace, assume 1 column
+            linesplit = line.split(None, 1)
+            result.append(linesplit[0])
+    logging.info("existed nwfilter binding list is: %s", result)
+    for binding_uuid in result:
+        try:
+            virsh.nwfilter_binding_delete(binding_uuid, ignore_status=False)
+        except Exception as e:
+            logging.error("Exception thrown while undefining nwfilter-binding: %s", str(e))
```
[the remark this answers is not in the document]
chunfuwen: no, see my explanation previously

virttest/utils_libvirt/utils_libvirt_nwfilter.py
```diff
@@ -0,0 +1,32 @@
+"""
```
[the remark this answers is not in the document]
chunfuwen: updated

virttest/utils_libvirt/utils_libvirt_nwfilter.py
chunfuwen: add one parameter ignore_status,but default value is True to allow silent failure. 
The reason behind the scene is that it is just clean up functionality, better not stop or fail case  execution at the beginning due to clean up failures

virttest/utils_libvirt/utils_libvirt_nwfilter.py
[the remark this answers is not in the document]
chunfuwen: see previous sathnaga's  comment

chunfuwen:
@dzhengfy , commit message updated accordingly.

virttest/utils_libvirt/utils_libvirt_nwfilter.py
sathnaga: it is better to notify the user and fail this testcase, rather leaving the unstable environment and subsequent test may result in failure or unknown behaviour.

If we expect the test case to clean the env it created it should do so, otherwise we should call it as error/failure based on the test steps.

chunfuwen: I understand there are two scenarios it may match:
1)Clean up dirty environment left by last case as pre-caution step, and I assume silent failure is reasonable
2)Clean up  environment created by this case itself, it should fail /error this 
So as you suggest, it provide ignore_status parameter to allow user to have flexisbility

kylazhang (approved):

dzhengfy (approved):
LGTM

dzhengfy:
@sathnaga Could you help review again?

virttest/utils_libvirt/libvirt_nwfilter.py
```diff
@@ -0,0 +1,30 @@
...
+
+:copyright: 2020 Red Hat Inc.
+"""
+
+import logging
+
+from virttest import virsh
+
+
+def clean_up_nwfilter_binding(ignore_status=True):
```
sathnaga, line 12: I believe it is better to have default option as `ignore_status=False` and can be overridden by user with known precautions.

sathnaga (changes requested):

virttest/utils_libvirt/libvirt_nwfilter.py
sathnaga: API by default should not silently ignore the failure..

chunfuwen: updated accordingly

sathnaga: Tnx!

PR #2705, opened 2020-08-12
Add port to network xml

chloerh:
- Add dict-attr 'port' to network xml
- Update function create_net_xml() to support 'port'
- Fix error that net_name being overwritten by params['net_name'].
Before fix, if we provide a name by setting var 'net_name' but not
set on params['net_name'], the var will be overwritten to 'default'


chloerh:
@kylazhang Would you mind taking a look?

yalzhang (approved):

chunfuwen:
Any test result available ?

chloerh:
> Any test result available ?

Yes, share test result with  https://github.com/autotest/tp-libvirt/pull/2993

PR #2709, opened 2020-08-17
Add migration_template.py, also update base.cfg

fangge1212:
This module is used for cross feature testing with migration.
With this module, testing migration with specified feature can
be implemented easily. Before, when a specified feature is
automated, migration part is always left behind because of its
complexity. Now with this module, when we automate a specified
feature, we can implement the migration part by inheriting the
MigrationTemplate and copy&&paste the code related to specified
feature here.

Signed-off-by: Fangge Jin <fjin@redhat.com>

PR #2713, opened 2020-08-18
migration: Re-init the instance variables before do_migration

fangge1212:
In case do_migration is called twice, if the first migration
succeeded but second migration failed, "ret" will not be written
during second migration, then check_result will get wrong result.
So re-init the instance variables before do_migration.

### DIFF
--- a/virttest/migration.py
+++ b/virttest/migration.py
@@ -36,6 +36,12 @@ class MigrationTest(object):
         # The CmdResult returned from virsh migrate command
         self.ret = None
 
+    def _reinit(self):
+        """
+        Re-init the instance variables
+        """
+        self.__init__()
+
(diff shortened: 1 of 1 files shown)


fangge1212:
https://github.com/avocado-framework/avocado-vt/pull/2709 depends on this pr

virttest/migration.py
```diff
@@ -36,6 +36,12 @@ def __init__(self):
         # The CmdResult returned from virsh migrate command
         self.ret = None
 
+    def _reinit(self):
```
dzhengfy, line 39: I could not see the requirement for us to create this new function. Could you give a detailed explanation or example?

harvey0100:

PR #2717, opened 2020-08-20
qemu_vm.py: Enable multi-threads support for cpu model EPYC

vivianQizhu:
id: 1867453

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1509,6 +1509,25 @@ class VM(virt_vm.BaseVM):
         add_memorys(devices, params)
         mem = int(params.get("mem", 0))
 
+        # Get cpu model, before add smp, to determine cpu topology
+        cpu_model = params.get("cpu_model", "")
+        use_default_cpu_model = True
+        if cpu_model:
+            use_default_cpu_model = False
+            for model in re.split(",", cpu_model):
+                model = model.strip()
+                if model not in support_cpu_model:
+                    continue
+                cpu_model = model
+                break
+            else:
+                cpu_model = model
+                logging.error("Non existing CPU model %s will be passed "
+                              "to qemu (wrong config or negative test)", model)
+
+        if use_default_cpu_model:
+            cpu_model = params.get("default_cpu_model", "")
+
         # Add smp
         smp = params.get_numeric('smp')
         vcpu_maxcpus = params.get_numeric("vcpu_maxcpus")
@@ -1533,10 +1552,11 @@ class VM(virt_vm.BaseVM):
         if not amd_vendor_string:
             amd_vendor_string = "AuthenticAMD"
         if amd_vendor_string == cpu.get_cpu_vendor():
-            # AMD cpu do not support multi threads.
-            if params.get("test_negative_thread", "no") != "yes":
+            # AMD cpu do not support multi threads besides EPYC
+            if (params.get("test_negative_thread", "no") != "yes" and
+                    not cpu_model.startswith('EPYC')):
                 vcpu_threads = 1
-                txt = "Set vcpu_threads to 1 for AMD cpu."
+                txt = "Set vcpu_threads to 1 for AMD non-EPYC cpu."
                 logging.warn(txt)
 
         smp_err = ""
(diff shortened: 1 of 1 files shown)


vivianQizhu:
@nanliu-r @PaulYuuu Would you help review it? Thanks.

virttest/qemu_vm.py
```diff
@@ -1509,6 +1509,25 @@ def sort_key(dev):
...
+        if cpu_model:
+            use_default_cpu_model = False
+            for model in re.split(",", cpu_model):
+                model = model.strip()
+                if model not in support_cpu_model:
+                    continue
+                cpu_model = model
+                break
+            else:
+                cpu_model = model
```
xutian, line 1524: suggest to set  cpu_mode  to default cpu module here.

vivianQizhu: @xutian Thanks for your review. It's moved from Line 1649 and I don't like to modify it in this patch. And Line 1529 has set to default cpu model too.

virttest/qemu_vm.py
```diff
@@ -1533,10 +1552,11 @@ def sort_key(dev):
         if not amd_vendor_string:
             amd_vendor_string = "AuthenticAMD"
         if amd_vendor_string == cpu.get_cpu_vendor():
-            # AMD cpu do not support multi threads.
-            if params.get("test_negative_thread", "no") != "yes":
+            # AMD cpu do not support multi threads besides EPYC
+            if (params.get("test_negative_thread", "no") != "yes" and
+                    not cpu_model.startswith('EPYC')):
```
PaulYuuu, line 1557: `cpu_model` maybe `None` on other architectures, so you would better to check at first or use `(cpu_model or '').startswith` to handle this problem.

vivianQizhu: @PaulYuuu Thanks, I have noticed that so updated the definition of `cpu_model` to `cpu_model = params.get("cpu_model", "")` at Line 1513, that should okay.

PaulYuuu: Okay, I see. Then this line also needs to be modified: `cpu_model = params.get("default_cpu_model")`

vivianQizhu: @PaulYuuu Yes you are right, modifying. Thanks.

PaulYuuu (approved):

vivianQizhu:
@nanliu-r Would you help review? Thanks.

nanliu-r:
> @nanliu-r Would you help review? Thanks.

Hi, do need to guarantee that the `cpu_model` is not "" before we configure the parameters in Line 1658?

vivianQizhu:
> > @nanliu-r Would you help review? Thanks.
> 
> Hi, do need to guarantee that the `cpu_model` is not "" before we configure the parameters in Line 1658?

Sorry, I don't get your point, Line 1658 is `if cpu_model:` so it already assures that `cpu_model` is not `""`. What do you expect more?

vivianQizhu:
@nanliu-r If you mean that `cpu_model` should not be "" for ever, I think here it allows it to be None and thus the qemu command line it generates will not has this line, which is also allowed by qemu-kvm. That's all depends on the users. Thanks.

nanliu-r:
> @nanliu-r If you mean that `cpu_model` should not be "" for ever, I think here it allows it to be None and thus the qemu command line it generates will not has this line, which is also allowed by qemu-kvm. That's all depends on the users. Thanks.

ok，I have request one epyc machine now and will update test results once I get it. Thanks.

nanliu-r:
I need to check more since guest will fail to boot on Milan EPYC machine for : 
```
 /usr/libexec/qemu-kvm -cpu EPYC-Rome -M q35 -smp 162,maxcpus=180,cores=10,threads=2,dies=1,sockets=9
qemu-kvm: current -smp configuration requires Extended Interrupt Mode enabled. You can add an IOMMU using: -device intel-iommu,intremap=on,eim=on
```

```
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              256
On-line CPU(s) list: 0-255
```

vivianQizhu:
@nanliu-r You might add `intel_iommu=yes` in your configuration in such case.

nanliu-r:
ack.

vivianQizhu:
Thanks @PaulYuuu @nanliu-r

PR #2721, opened 2020-08-22
Support suppress exceptions when create monitor

yanan-fu:
id: 1871156

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -132,13 +132,14 @@ def get_monitor_filenames(vm):
     return [get_monitor_filename(vm, m) for m in vm.params.objects("monitors")]
 
 
-def create_monitor(vm, monitor_name, monitor_params):
+def create_monitor(vm, monitor_name, monitor_params, suppress_exceptions):
     """
     Create monitor object and connect to the monitor socket.
 
     :param vm: The VM object which has the monitor.
     :param monitor_name: The name of this monitor object.
     :param monitor_params: The dict for creating this monitor object.
+    :param suppress_exceptions: Whether to suppress exceptions when create monitor
     """
     MonitorClass = HumanMonitor
     if monitor_params.get("monitor_type") == "qmp":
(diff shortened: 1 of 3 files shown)


yanan-fu:
@vivianQizhu @luckyh  could you help review ? Thanks a lot!

virttest/qemu_monitor.py
```diff
@@ -803,10 +805,10 @@ def __init__(self, vm, name, monitor_params, suppress_exceptions=False):
                 self.on_str, self.off_str = "on", "off"
 
         except MonitorError as e:
-            self._close_sock()
             if suppress_exceptions:
                 logging.warn(e)
             else:
+                self._close_sock()
```
sathnaga, line 811: no reference for this change in commit msg, could you pls add the reason for avoid closing the socket.

sathnaga (changes requested):

virttest/qemu_monitor.py
```diff
@@ -150,13 +150,14 @@ def create_monitor(vm, monitor_name, monitor_params):
...
-    monitor = MonitorClass(vm, monitor_name, monitor_params)
+    monitor = MonitorClass(vm, monitor_name, monitor_params, suppress_exceptions)
     monitor.verify_responsive()
 
     return monitor
 
 
-def wait_for_create_monitor(vm, monitor_name, monitor_params, timeout):
+def wait_for_create_monitor(vm, monitor_name, monitor_params,
+                            suppress_exceptions, timeout):
```
vivianQizhu, line 160: Please append the new parameter to the end.

vivianQizhu:
Please update the doc string when introduce new parameter, thanks.

virttest/qemu_monitor.py
yanan-fu: @vivianQizhu  You are right, i understand your point, to avoid this change effect other calling.
I checked tp-libvirt, tp-spice, and tp-qemu, no call for this function. The only one call is in `qemu_vm.py` which i have update it.  Prefer to put timeout at the end of the parameter a bit more, Is it acceptable for you ?

virttest/qemu_monitor.py
yanan-fu: > no reference for this change in commit msg, could you pls add the reason for avoid closing the socket.

I prefer that this is an existing bug, when `suppress_exceptions = True`, close the socket will cause monitor connection failed too, this is not the expect result.   Let me explain this in the commit message.

virttest/qemu_monitor.py
vivianQizhu: @yanan-fu Yes it's also okay for me.

virttest/qemu_monitor.py
yanan-fu: @sathnaga  updated, could you help review again, thanks!

virttest/qemu_monitor.py
yanan-fu: Thanks~

yanan-fu:
> Please update the doc string when introduce new parameter, thanks.

Updated, thanks!

vivianQizhu (approved):

luckyh (changes requested):
Would you please explain more on this proposition? It looks several side-effects would be introduced by this patch, which should be avoided.

yanan-fu:
Thanks all!

This solution is a little rough, my original purpose was to ignore some nonfatal exceptions, suppress all exceptions have risk, let me close it.
Thanks @vivianQizhu @sathnaga @luckyh

PR #2724, opened 2020-08-24
kernel_interface: check fs existing or not with correct target

yanan-fu:
1. kernel_interface: Avoid repeated calls to 'fs_value'
2. kernel_interface: check fs existing or not with correct target

id: 1871870

### DIFF
--- a/virttest/kernel_interface.py
+++ b/virttest/kernel_interface.py
@@ -82,7 +82,7 @@ class ProcFS(FS):
         try:
             return int(self.fs_value)
         except ValueError:
-            return self.fs_value
+            return self.fs_val
 
     @proc_fs_value.setter
     def proc_fs_value(self, value):
@@ -133,9 +133,9 @@ class SysFS(FS):
         if output:
             return str(output.group()).strip()
         try:
-            return int(self.fs_value)
+            return int(self.fs_val)
         except ValueError:
-            return self.fs_value
+            return self.fs_val
 
     @sys_fs_value.setter
     def sys_fs_value(self, value):
--- a/virttest/kernel_interface.py
+++ b/virttest/kernel_interface.py
@@ -17,12 +17,21 @@ class FS(object):
         :param session: ShellSession object of remote or VM
         """
         self.fs = fs
-        if not os.path.isfile(self.fs):
-            raise AttributeError("%s is not available" % self.fs)
         self.func = process.getstatusoutput
         self.session = session
+        if not self._check_isfile():
+            raise AttributeError("%s is not available" % self.fs)
         if self.session:
             self.func = self.session.cmd_status_output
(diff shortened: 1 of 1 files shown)


yanan-fu:
@vivianQizhu  Could you help review ?
I move the fix of repeated call from https://github.com/avocado-framework/avocado-vt/pull/2699 to this one, since i want to fix the second issue "check fs existing or not with correct target", they need modify the same file.

Leave https://github.com/avocado-framework/avocado-vt/pull/2699 to fix the NumaInfo issue only. Thanks a lot!

yanan-fu:
The CI failing is not related with the fix:
```
avocado list: error: unrecognized arguments: --vt-type=libvirt
Perhaps a plugin is missing; run 'avocado plugins' to list the installed ones
```

Will wait and force update again to see if it will gone.

sathnaga (approved):
LGTM

sathnaga:
```
>>> from virttest import kernel_interface
>>> testobj = kernel_interface.ProcFS("/home/sath/tests/avocado-vt/test")
>>> testobj.proc_fs_value
'testing'
>>> testobj.proc_fs_value=1
>>> testobj.proc_fs_value
1
>>>
```

yanan-fu:
@sathnaga Thanks, i update the patch to fix the cyclic import issue since:
https://github.com/avocado-framework/avocado-vt/blob/2cc74315c7eb4f05d052428e79a4f4739754a601/virttest/utils_misc.py#L85

 Could you help review again ? Thanks!

vivianQizhu:
@yanan-fu Why not make it a class member so that you don't need to pass the session object?

yanan-fu:
> @yanan-fu Why not make it a class member so that you don't need to pass the session object?

Same for filename(self.fs) ?  Then no need pass any param, do you mean this ?

vivianQizhu:
@yanan-fu Yes.

yanan-fu:
> @yanan-fu Yes.

Updated, could you help review again, thanks~

virttest/kernel_interface.py
```diff
@@ -17,13 +17,23 @@ def __init__(self, fs, session=None):
...
         self.session = session
+        if not self._check_isfile():
+            raise AttributeError("%s is not available" % self.fs)
         if self.session:
             self.func = self.session.cmd_status_output
 
+    def _check_isfile(self):
+        """
+        check whether the fs exist in local/remote host/VM
+
```
vivianQizhu, line 30: This blank line might be removed.

yanan-fu: > This blank line might be removed.

Ah, yes, delete the blank line now, thanks!

vivianQizhu (approved):

PR #2726, opened 2020-08-25
qemu_vm: Add device intel-iommu before any virtio devices

vivianQizhu:
id: 1866233

### DIFF
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -88,6 +88,20 @@ netdst = virbr0
 # present in VM, default is 'no'
 flexible_nic_index = no
 
+# Add device intel-iommu
+# intel_iommu = yes
+# Set intel-iommu options
+# iommu_intremap = on/off/auto  (on by default)
+# iommu_device_iotlb = on/off  (on by default)
+# iommu_caching_mode = on/off
+# iommu_eim = on/off/auto
+# iommu_x_buggy_eim = on/off
+# iommu_version = <uint32>
+# iommu_x_scalable_mode = on/off
+# iommu_dma_drain = on/off
+# iommu_pt = on/off
+# iommu_aw_bits = <uint8>
+
 # Enable guest iommu (intel_iommu=on) in guest kernel cmd line
 # This parameter only support intel platform so the default
 # value is set to None.
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1484,6 +1484,22 @@ class VM(virt_vm.BaseVM):
         if params.get("defconfig", "yes") == "no":
             devices.insert(StrDev('nodefconfig', cmdline=" -nodefconfig"))
 
+        # Add device intel-iommu, it must be added before any virtio devices
+        if (params.get('intel_iommu') == 'yes' and
+                devices.has_device('intel-iommu')):
+            iommu_params = {
+                'intremap': params.get('iommu_intremap', 'on'),
+                'device-iotlb': params.get('iommu_device_iotlb', 'on'),
+                'caching-mode': params.get('iommu_caching_mode'),
+                'eim': params.get('iommu_eim'),
+                'x-buggy-eim': params.get('iommu_x_buggy_eim'),
+                'version': params.get('iommu_version'),
+                'x-scalable-mode': params.get('iommu_x_scalable_mode'),
+                'dma-drain': params.get('iommu_dma_drain'),
+                'pt': params.get('iommu_pt'),
+                'aw-bits': params.get('iommu_aw_bits')}
+            devices.insert(QDevice('intel-iommu', iommu_params))
+
         vga = params.get("vga")
         if vga:
             devices.insert(add_vga(devices, vga))


vivianQizhu:
@yanan-fu Please help review, thanks.

vivianQizhu:
@quanwenli Please help review, thanks.

virttest/qemu_vm.py
```diff
@@ -1687,6 +1687,25 @@ def sort_key(dev):
                 vcpu_dev = devices.vcpu_device_define_by_params(params, vcpu)
                 devices.insert(vcpu_dev)
 
+        # Add device intel-iommu, it must be added before any virtio devices
+        if params.get('virtio_dev_iommu_platform') == 'on':
```
PaulYuuu, line 1691: The namespace of each virtio device should have its own `virtio_dev_iommu_platform` and it's only used to add `iommu_platform` option on virtio devices, so I suggest to introduce a new parameter to define `intel-iommu`.
Another issue is `intel-iommu` only available on x86, this change will break the initialization of other architectures.

PaulYuuu: BTW, VGA device can be virtio-gpu-pci (aarch64 is using it), so `intel-iommu` must be added earlier than VGA.
https://github.com/avocado-framework/avocado-vt/blob/f425ead735daa1ccca408d8f62bdeb65c4b1f688/virttest/qemu_vm.py#L1487-L1489
https://github.com/avocado-framework/avocado-vt/blob/f425ead735daa1ccca408d8f62bdeb65c4b1f688/virttest/qemu_vm.py#L1109-L1120

yanan-fu: > The namespace of each virtio device should have its own `virtio_dev_iommu_platform` and it's only used to add `iommu_platform` option on virtio devices, so I suggest to introduce a new parameter to define `intel-iommu`.

@vivianQizhu I have same idea with @PaulYuuu , 'iommu_platform' is a virtio device option, use prefix `virtio_dev`.
Add "intel-iommu" device but without set 'iommu_platform' for virtio device is possible. It is better to introduce a new parameter to define `intel-iommu` device.

yanan-fu:
@vivianQizhu `intel-iommu` device work with q35 and need work with machine type extra parameter `kernel-irqchip=split`, It is for x86, not confident about others, may be we could consider this together ?

virttest/qemu_vm.py
vivianQizhu: > The namespace of each virtio device should have its own `virtio_dev_iommu_platform` and it's only used to add `iommu_platform` option on virtio devices, so I suggest to introduce a new parameter to define `intel-iommu`.
> Another issue is `intel-iommu` only available on x86, this change will break the initialization of other architectures.

@PaulYuuu Thanks for your review. Actually all test cases that specify this `virtio_dev_iommu_platform` variable they specify `only q35` together, so it wont affect other architectures, but yes, it should be restrict here. However, even I name a new parameter, it have same issue that it should be restrict to intel, so I would like to add a condition check here, instead of introduce a new parameter, what do you think?

vivianQizhu: > BTW, VGA device can be virtio-gpu-pci (aarch64 is using it), so `intel-iommu` must be added earlier than VGA.

Yes, good catch.

vivianQizhu: > > The namespace of each virtio device should have its own `virtio_dev_iommu_platform` and it's only used to add `iommu_platform` option on virtio devices, so I suggest to introduce a new parameter to define `intel-iommu`.
> 
> @vivianQizhu I have same idea with @PaulYuuu , 'iommu_platform' is a virtio device option, use prefix `virtio_dev`.
> Add "intel-iommu" device but without set 'iommu_platform' for virtio device is possible. It is better to introduce a new parameter to define `intel-iommu` device.

Well, okay, let me introduce a new parameter then. Thanks. cc @PaulYuuu

vivianQizhu:
> @vivianQizhu `intel-iommu` device work with q35 and need work with machine type extra parameter `kernel-irqchip=split`, It is for x86, not confident about others, may be we could consider this together ?

Not sure if it exists some scenario like negative case to test intel-iommu without `machine_type_extra_params = "kernel-irqchip=split"`, if we specify another parameter besides `virtio_dev_iommu_platform`, let's just keep everything independent(actually there is `enable_guest_iommu` too),  what do you think?

virttest/qemu_vm.py
PaulYuuu: > > The namespace of each virtio device should have its own `virtio_dev_iommu_platform` and it's only used to add `iommu_platform` option on virtio devices, so I suggest to introduce a new parameter to define `intel-iommu`.
> > Another issue is `intel-iommu` only available on x86, this change will break the initialization of other architectures.
> 
> @PaulYuuu Thanks for your review. Actually all test cases that specify this `virtio_dev_iommu_platform` variable they specify `only q35` together, so it wont affect other architectures, but yes, it should be restrict here. However, even I name a new parameter, it have same issue that it should be restrict to intel, so I would like to add a condition check here, instead of introduce a new parameter, what do you think?

@vivianQizhu Maybe we can check if the device in the qemu device category has `intel-iommu`?
ppc:
```
Uncategorized devices:
name "nvdimm", desc "DIMM memory module"
name "pc-dimm", desc "DIMM memory module"
name "spapr-tpm-proxy"
```

x86:
```
Uncategorized devices:
name "edu", bus PCI
name "intel-iommu", bus System
name "nvdimm", desc "DIMM memory module"
name "pc-dimm", desc "DIMM memory module"
```

And `virtio_dev_iommu_platform` should work on other archs.
```
# /usr/libexec/qemu-kvm -M pseries -device virtio-net-pci,iommu_platform=on,disable-legacy=on
```

vivianQizhu: @PaulYuuu yes, good idea. Will update later, thanks.

yanan-fu:
> > @vivianQizhu `intel-iommu` device work with q35 and need work with machine type extra parameter `kernel-irqchip=split`, It is for x86, not confident about others, may be we could consider this together ?
> 
> Not sure if it exists some scenario like negative case to test intel-iommu without `machine_type_extra_params = "kernel-irqchip=split"`, if we specify another parameter besides `virtio_dev_iommu_platform`, let's just keep everything independent(actually there is `enable_guest_iommu` too), what do you think?

@vivianQizhu Yeah, i though about this issue again, separate the configuration of different devices (machine, intel-iommu, virtio device,  guest kernel and so on...) is more reasonable, more easier to cover different test combinations, agree with you ~

vivianQizhu:
@yanan-fu @PaulYuuu Updated per your comments, please help re-review, thanks.

virttest/qemu_vm.py
```diff
@@ -1484,6 +1484,26 @@ def sort_key(dev):
...
 
+        # Add device intel-iommu, it must be added before any virtio devices
+        if (params.get('intel_iommu') == 'yes' and
+                devices.has_device('intel-iommu')):
+            iommu_params = {
+                'intremap': params.get('iommu_intremap', 'on'),
+                'device-iotlb': params.get('iommu_device-iotlb', 'on')}
+            iommu_dev = QDevice('intel-iommu', iommu_params)
+            iommu_dev.set_param(
+                'caching-mode', params.get('iommu_caching_mode'))
```
yanan-fu, line 1495: @vivianQizhu  Both can work, but could you explain why don't set all the params first then create QDevice ?

vivianQizhu: @yanan-fu Because `set_param` can handle it when certain parameter is `None`, to define them all before create QDevice might hit error if the user does not specify all of them.

yanan-fu: Qdevice use `set_param` too ?
https://github.com/avocado-framework/avocado-vt/blob/2cc74315c7eb4f05d052428e79a4f4739754a601/virttest/qemu_devices/qdevices.py#L89

I tested with set all params before create QDevice, it can pass too.

yanan-fu: If it is not must, i prefer use same method for all the parameter a bit more ~

vivianQizhu:
@yanan-fu Updated, thanks.

shared/cfg/base.cfg
```diff
@@ -88,6 +88,20 @@ netdst = virbr0
 # present in VM, default is 'no'
 flexible_nic_index = no
 
+# Add device intel-iommu
+# intel_iommu = yes
+# Set intel-iommu options
+# iommu_intremap = on/off/auto  (on by default)
+# iommu_device-iotlb = on/off  (on by default)
```
yanan-fu, line 95: Keep consistent, no matter what it is in the option, all param use `_` ?   `iommu_device_iotlb` ?
Or, use the exact option name, both are ok but keep consistent ~

virttest/qemu_vm.py
```diff
@@ -1484,6 +1484,22 @@ def sort_key(dev):
         if params.get("defconfig", "yes") == "no":
             devices.insert(StrDev('nodefconfig', cmdline=" -nodefconfig"))
 
+        # Add device intel-iommu, it must be added before any virtio devices
+        if (params.get('intel_iommu') == 'yes' and
+                devices.has_device('intel-iommu')):
+            iommu_params = {
+                'intremap': params.get('iommu_intremap', 'on'),
+                'device-iotlb': params.get('iommu_device-iotlb', 'on'),
```
PaulYuuu, line 1492: No other comments, only change `iommu_device-iotlb` to `iommu_device_iotlb` to align other parameters.

vivianQizhu: @PaulYuuu Ah yes, updated. Thanks.

yanan-fu (approved):

virttest/qemu_vm.py
```diff
@@ -1484,6 +1484,22 @@ def sort_key(dev):
...
+                'intremap': params.get('iommu_intremap', 'on'),
+                'device-iotlb': params.get('iommu_device_iotlb', 'on'),
+                'caching-mode': params.get('iommu_caching_mode'),
+                'eim': params.get('iommu_eim'),
+                'x-buggy-eim': params.get('iommu_x_buggy_eim'),
+                'version': params.get('iommu_version'),
+                'x-scalable-mode': params.get('iommu_x_scalable_mode'),
+                'dma-drain': params.get('iommu_dma_drain'),
+                'pt': params.get('iommu_pt'),
+                'aw-bits': params.get('iommu_aw_bits')}
```
PaulYuuu, line 1500: In older qemu versions, some arguments are different. Please note to handle it if users want to test them in future.

$ /usr/libexec/qemu-kvm -version
QEMU emulator version 2.12.0 (qemu-kvm-rhev-2.12.0-33.el7_7.5)
Copyright (c) 2003-2017 Fabrice Bellard and the QEMU Project developers

$ /usr/libexec/qemu-kvm -device intel-iommu,?
intel-iommu.eim=OnOffAuto (on/off/auto)
**intel-iommu.x-aw-bits=uint8**
intel-iommu.x-buggy-eim=bool
intel-iommu.intremap=bool
intel-iommu.version=uint32
intel-iommu.pt=bool
intel-iommu.device-iotlb=bool
intel-iommu.caching-mode=bool

PaulYuuu (approved):

virttest/qemu_vm.py
vivianQizhu: @PaulYuuu Thanks, I guess it is less possible for us to test it.

vivianQizhu:
@quanwenli Waiting for your test, thanks.

vivianQizhu:
Thanks all.

PR #2741, opened 2020-09-03
Add a new function for multiple pkg version checking

xiaodwan:
Sometime a case may requires multiple pkgs are at special versions.
This new function can help you specify those pkgs and their versions
in cases cfg files.

### DIFF
--- a/virttest/utils_v2v.py
+++ b/virttest/utils_v2v.py
@@ -1836,8 +1836,37 @@ def compare_version(interval, version=None, cmd=None):
     """
     if not version:
         if not cmd:
-            cmd = 'rpm -q libguestfs'
+            cmd = 'rpm -q virt-v2v'
         # exit status checking will be done in process.run
         version = process.run(cmd, shell=True).stdout_text.strip()
 
     return check_version(version, interval)
+
+
+def multiple_versions_compare(interval):
+    """
+    Mutiple pkgs can be specified by ';', e.g.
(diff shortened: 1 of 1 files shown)


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.function_test_esx.positive_test.log_size_decrease.esx_67.rhev.rhv_upload.it_vddk.vpx_uri: PASS (1743.84 s)

xiaodwan:
@chunfuwen @chloerh @yantian0 Please help to  review this PR, thanks.

virttest/utils_v2v.py
```diff
@@ -1836,8 +1836,37 @@ def compare_version(interval, version=None, cmd=None):
     """
     if not version:
         if not cmd:
-            cmd = 'rpm -q libguestfs'
+            cmd = 'rpm -q virt-v2v'
```
yantian0, line 1839: Is it required to distinguish the rpm based on the release version?

xiaodwan: No, This is a default behaviour change. 
Users should always tell the api what cmd they are using like the new function does.

chunfuwen (approved):

chloerh (approved):

virttest/utils_v2v.py
yantian0: Ok, got it. Thank you.

yantian0 (approved):

PR #2749, opened 2020-09-08
XMLElementList:support list of xml elements with child elements

chloerh:
- XMLElementList doesn't use to support a list of xml objects which
have child elements, this commit will fix it.
- numa_cell under 'cpu' tag of libvirt xml is now updated according
to libvirt.org. It used to be one-line xml element without child-elements,
it now could have child elements, therefore an update of marshal function
is necessary.

Signed-off-by: Haijiao Zhao <haizhao@redhat.com>

PR #2756, opened 2020-09-11
Revert "XMLElementList:support list of xml elements with child elements"

chunfuwen:
Reverts avocado-framework/avocado-vt#2749
This need tp-libvirt cases changes at the same time,otherwise leads to many cases failures
revert it firstly to avoid cases failures


chloerh:
There's no need to revert this pr, I'll update to fix those errors

sathnaga:
@chloerh @chunfuwen should we close this as @chloerh agreed to send a test case fix soon.

dzhengfy:
Close it as we will fix the test cases.

PR #2765, opened 2020-09-17
Change default value of 'mem' from empty string to 'None'

chloerh:
When running with Avocado 82.0, is_registering_settings_required()
returns True, settings will be registered, default value of mem is ''.
Setting of mem in 'base.cfg' or other config files from test provider
will not take effect if 'mem' is registered as empty string, changing
to 'None' could fix it.

### DIFF
--- a/avocado_vt/plugins/vt_init.py
+++ b/avocado_vt/plugins/vt_init.py
@@ -65,7 +65,7 @@ if hasattr(plugin_interfaces, 'Init'):
                         'and test provider specific files')
             settings.register_option(section, 'mem',
                                      help_msg=help_msg,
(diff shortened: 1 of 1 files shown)


chloerh:
@clebergnu Could you help review this pr? Thanks.

clebergnu (approved):
I realize now that this was *always* setting `mem` to the empty string because it would satisfy the `if mem is not None` on `options.py:VirtTestOptionsProcess._process_mem()`.  And because it's done very late, it'd override the values on cartesian configuration files.

So, yes, this LGTM.  Thanks a lot!

kylazhang (approved):

PR #2768, opened 2020-09-19
shared: add openEuler-20.03 guest

wsf0824:
Signed-off-by: Wang Shengfang <wangshengfang@huawei.com>


luckyh:
@wsf0824 thank you for your contribution, would you please help provide the result of the installation test with using the latest code? thanks!

PR #2771, opened 2020-09-22
Handle the failures when getting host cpu family

luckyh:
Some routines would request the information of host cpu family for
branch selection, so we need to suppress all the exceptions to ensure
that the whole procedure won't be interrupted.

Reported-by: Qinghua Cheng <qcheng@redhat.com>

ID: 1878675


yanan-fu:
@luckyh  let's add id `1878675`

luckyh:
> @luckyh let's add id `1878675`

Sure, thanks.

sathnaga (approved):

yanan-fu (approved):
Ack

PR #2772, opened 2020-09-23
Add protected key support

smitterl:
1. Add helper class to handle commands for working with protected keys.
2. Add keywrap element to XML to control keywrap configuration.

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -162,13 +162,14 @@ class VMXMLBase(base.LibvirtXMLBase):
     """
 
     # Additional names of attributes and dictionary-keys instances may contain
-    __slots__ = ('hypervisor_type', 'vm_name', 'uuid', 'title', 'vcpu', 'max_mem',
-                 'current_mem', 'dumpcore', 'numa_memory', 'numa_memnode',
-                 'devices', 'seclabel', 'cputune', 'placement', 'cpuset',
-                 'current_vcpu', 'vcpus', 'os', 'cpu', 'pm', 'on_poweroff',
-                 'on_reboot', 'on_crash', 'features', 'mb', 'max_mem_unit',
-                 'current_mem_unit', 'memtune', 'max_mem_rt', 'max_mem_rt_unit',
-                 'max_mem_rt_slots', 'iothreads', 'iothreadids', 'memory', 'perf')
+    __slots__ = ('hypervisor_type', 'vm_name', 'uuid', 'title', 'vcpu',
+                 'max_mem', 'current_mem', 'dumpcore', 'numa_memory',
+                 'numa_memnode', 'devices', 'seclabel', 'cputune', 'placement',
+                 'cpuset', 'current_vcpu', 'vcpus', 'os', 'cpu', 'pm',
+                 'on_poweroff', 'on_reboot', 'on_crash', 'features', 'mb',
+                 'max_mem_unit', 'current_mem_unit', 'memtune', 'max_mem_rt',
+                 'max_mem_rt_unit', 'max_mem_rt_slots', 'iothreads',
+                 'iothreadids', 'memory', 'perf', 'kewyrap')
 
     __uncompareable__ = base.LibvirtXMLBase.__uncompareable__
 
@@ -347,6 +348,13 @@ class VMXMLBase(base.LibvirtXMLBase):
                                  parent_xpath='/',
                                  tag_name='features',
                                  subclass=VMFeaturesXML,
+                                 subclass_dargs={
+                                     'virsh_instance': virsh_instance})
+        accessors.XMLElementNest(property_name='keywrap',
+                                 libvirtxml=self,
+                                 parent_xpath='/',
+                                 tag_name='keywrap',
+                                 subclass=VMKeywrapXML,
                                  subclass_dargs={
                                      'virsh_instance': virsh_instance})
         accessors.XMLElementNest(property_name='mb',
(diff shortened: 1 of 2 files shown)


smitterl:
Manually tested by running tp-libvirt test cases: https://github.com/autotest/tp-libvirt/pull/3085#issuecomment-697473409

virttest/libvirt_xml/vm_xml.py
```diff
@@ -3325,3 +3333,39 @@ def __init__(self, virsh_instance=base.virsh):
                                attribute='unit')
         super(VMFeaturesHptXML, self).__init__(virsh_instance=virsh_instance)
         self.xml = '<hpt/>'
+
+
+class VMKeywrapXML(base.LibvirtXMLBase):
+    """
+    Keywrap class for s390x ciphers on QEMU
+    xpath: /domain/keywrap
```
chunfuwen, line 3341: It is better add one simple example show how VMKeywrapXML looks like

virttest/utils_libvirt/libvirt_keywrap.py
```diff
@@ -0,0 +1,45 @@
+"""
+Module to contain logic testing protected key module 'pkey'
+"""
+import logging
+import os
+
+from virttest.utils_misc import cmd_status_output
```
chunfuwen, line 7: I am thinking how can we prevent cirrus-loop import by avoiding one utils refer to method in other utils. 
Shall we just use process.run in avocado project?

virttest/libvirt_xml/vm_xml.py
smitterl: @chunfuwen Do you mean doc string should contain 
```xml
<keywrap>
  <cipher name='aes' state='on'/>
</keywrap>
```
or do you mean an actual usage in code like e.g. https://github.com/autotest/tp-libvirt/blob/e0d140e83b90505c5de25e73545317d7126f6f3a/libvirt/tests/src/libvirt_keywrap.py#L29 ?

virttest/utils_libvirt/libvirt_keywrap.py
smitterl: @chunfuwen I'm not familiar with the cirrus-loop import issue, I'm sorry. Could you show me what you mean?

If this has to be avoided, I really would want to maintain usage of `cmd_status_output` to handle running a command in guest or on host in the same way without duplicating code. From what you mention `cmd_status_output` should then maybe be moved to `avocado.utils`?

Another way I can think of is to have the test code (e.g.  in tp-libvirt) pass the method to use, s. https://github.com/smitterl/avocado-vt/blob/f3f55f362e3991b3963b30df7b3ac32116de6bce/virttest/utils_zcrypt.py#L27

What do you think?

virttest/libvirt_xml/vm_xml.py
chunfuwen: I   mean doc string better contain sample VMKeywrapXML

virttest/utils_libvirt/libvirt_keywrap.py
chunfuwen: sorry, typo issue, should be circular-imports. 
see https://stackabuse.com/python-circular-imports/
Normally, if avocado-vt refer to some modules in avocado, there is no problem here because avocado never refer to module in avocado-vt. 
However, if avocado-vt refer to avocado-vt modules, there are potential risk here because we don't have good Hierarchical 
design at the beginning. And We met a few cases before.
I find there is one method:getstatusoutput in process.py, can you try whether it can satisfy your requirement?

smitterl: > sorry, typo issue, should be circular-imports.
> see https://stackabuse.com/python-circular-imports/
> Normally, if avocado-vt refer to some modules in avocado, there is no problem here because avocado never refer to module in avocado-vt.
> However, if avocado-vt refer to avocado-vt modules, there are potential risk here because we don't have good Hierarchical
> design at the beginning. And We met a few cases before.
> I find there is one method:getstatusoutput in process.py, can you try whether it can satisfy your requirement?

Unfortunately https://github.com/avocado-framework/avocado/blob/master/avocado/utils/process.py#L1233 is not an option because this doesn't let me execute the command in the guest.

However, I believe the import chain is not circular:
virttest.utils_libvirt.libvirt_keywrap
  -> virttest.utils_misc
    -> avocado.utils.process
    -> aexpect.client.ShellSession
and
utils_misc does not import utils_libvirt.
 
I'm not sure if I'm missing something?

virttest/libvirt_xml/vm_xml.py
smitterl: > I mean doc string better contain sample VMKeywrapXML

Got it. So for example the following?
```python
class VMKeywrapXML(base.LibvirtXMLBase):
    """
    Keywrap class for s390x ciphers on QEMU
    xpath: /domain/keywrap

    Example:
    
        vmxml = VMXML.new_from_dumpxml(vm_name)
        kw = VMKeywrapXML()
        kw.set_cipher("aes", "off")
        vmxml.set_keywrap(kw)

    :param ....
```

chunfuwen (approved):
LGTM

virttest/libvirt_xml/vm_xml.py
```diff
@@ -3325,3 +3333,46 @@ def __init__(self, virsh_instance=base.virsh):
...
+        kw.set_cipher("aes", "off")
+        vmxml.set_keywrap(kw)
+    """
+    def __init__(self, virsh_instance=base.virsh):
+        super(VMKeywrapXML, self).__init__(virsh_instance=virsh_instance)
+        self.xml = '<keywrap/>'
+
+    def get_cipher(self, name):
+        """
+        Gets cipher 'name' if it exists, else None
```
dzhengfy, line 3356: Could you add a new blank line under this line?

virttest/libvirt_xml/vm_xml.py
```diff
@@ -3325,3 +3333,46 @@ def __init__(self, virsh_instance=base.virsh):
...
+        """
+        root = self.__dict_get__('xml').getroot()
+        for cipher in root.findall('cipher'):
+            if cipher.name == name:
+                return cipher
+        return None
+
+    def set_cipher(self, name, state):
+        """
+        Sets cipher state for name; adds a new cipher if it doesn't exist yet.
```
dzhengfy, line 3368: Same as above.

virttest/utils_libvirt/libvirt_keywrap.py
```diff
@@ -0,0 +1,41 @@
...
+    Helper class to test for protected key support on s390x
+    """
+    def __init__(self, session=None):
+        self.session = session
+        self.sysfs = "/sys/devices/virtual/misc/pkey/protkey"
+        self.module_name = "pkey"
+
+    def load_module(self):
+        """
+        Loads pkey module
```
dzhengfy, line 21: Could you add a new blank line under this line? And same to the following functions.

dzhengfy (commented):
Others look fine to me.

smitterl:
@dzhengfy Thank you very much for your review. I added the missing blank between descriptions and parameter list. Please, could you review again?

dzhengfy (approved):

PR #2779, opened 2020-09-29
qemu_vm: Refactor code to add virtio options

PaulYuuu:
1. Delete 'add_virtio_option' since it will execute qemu each time it is
called.
2. Get supported properties to decide whether to add options.

ID: 1879359

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1356,14 +1356,6 @@ class VM(virt_vm.BaseVM):
                 dev.set_param(key, value)
             devices.insert(dev)
 
-        def add_virtio_option(option, value, devices, dev, dev_type):
-            """
-            This function is used to add given option for virtio pci device
-            """
-            options = devices.execute_qemu("-device %s,?" % dev_type)
-            if option in options:
-                dev.set_param(option, value)
-
         def add_pci_controllers(devices, params):
             """
             Insert pci controllers into qcontainer by order.
@@ -2461,25 +2453,25 @@ class VM(virt_vm.BaseVM):
             attr_info = [None, params["keyboard_layout"], None]
             add_qemu_option(devices, "k", [attr_info])
 
-        # Add disable_legacy and disable_modern options
-        virtio_devices = filter(lambda x: x.get_param('driver', '').startswith('virtio-'), devices)
+        # Add options for all virtio devices
+        virtio_devices = filter(lambda x: re.search(r"(?:^virtio-)|(?:^vhost-)",
+                                                    x.get_param('driver', '')),
+                                devices)
         for device in virtio_devices:
             dev_type = device.get_param("driver")
             # Currently virtio1.0 behaviour on latest RHEL.7.2/RHEL.7.3
             # qemu versions is default, we don't need to specify the
             # disable-legacy and disable-modern options explicitly.
-            disable_legacy = params.get("virtio_dev_disable_legacy", None)
-            disable_modern = params.get("virtio_dev_disable_modern", None)
-            iommu_platform = params.get("virtio_dev_iommu_platform", None)
-            ats = params.get("virtio_dev_ats", None)
-            if disable_legacy:
-                add_virtio_option("disable-legacy", disable_legacy, devices, device, dev_type)
-            if disable_modern:
-                add_virtio_option("disable-modern", disable_modern, devices, device, dev_type)
-            if iommu_platform:
-                add_virtio_option("iommu_platform", iommu_platform, devices, device, dev_type)
-            if ats:
-                add_virtio_option("ats", ats, devices, device, dev_type)
+            dev_info = devices.execute_qemu("-device %s,help" % dev_type)
+            dev_properties = re.findall(r"([a-zA-Z0-9_-]+)=\S+", dev_info)
+            properties_to_be_set = {
+                "disable-legacy": params.get("virtio_dev_disable_legacy"),
+                "disable-modern": params.get("virtio_dev_disable_modern"),
+                "iommu_platform": params.get("virtio_dev_iommu_platform"),
+                "ats": params.get("virtio_dev_ats")}
+            for key, value in properties_to_be_set.items():
+                if key in dev_properties:
+                    device.set_param(key, value)
 
         # Add extra root_port at the end of the command line only if there is
         # free slot on pci.0, discarding them otherwise


PaulYuuu:
Hello @yanan-fu @vivianQizhu, please help to review, thanks.
One discussion point: Would it be more stronger to remove the device filter below? This means if other devices support those options or virtio devices have a new naming, the code can handle them directly.
```python
        virtio_devices = filter(lambda x: re.search(r"(?:^virtio-)|(?:^vhost-)",
                                                    x.get_param('driver', '')),
                                devices)
```

vivianQizhu:
> Hello @yanan-fu @vivianQizhu, please help to review, thanks.
> One discussion point: Would it be more stronger to remove the device filter below? This means if other devices support those options or virtio devices have a new naming, the code can handle them directly.
> 
> ```python
>         virtio_devices = filter(lambda x: re.search(r"(?:^virtio-)|(?:^vhost-)",
>                                                     x.get_param('driver', '')),
>                                 devices)
> ```

I agree this if the above judgement totally equals to what you added for device property check, I mean, if there is no such condition like other device has the property but we don't want it be turned on. I am not totally confident about this, will check it later.

vivianQizhu:
Checked by:
```
devices = process.getoutput("/usr/libexec/qemu-kvm -device ?|cut -d ' ' -f 2", shell=True).split()
for dev in devices:
    property = process.getoutput("/usr/libexec/qemu-kvm -device %s,help" % dev.strip('",'))
    if "iommu_platform" in property:
        print(dev)
```
"vhost-user-fs-device",
"vhost-user-fs-pci",
"virtio-blk-device",
"virtio-blk-pci",
"virtio-blk-pci-non-transitional",
"virtio-blk-pci-transitional",
"virtio-scsi-device",
"virtio-scsi-pci",
"virtio-scsi-pci-non-transitional",
"virtio-scsi-pci-transitional",
"virtio-net-device",
"virtio-net-pci",
"virtio-net-pci-non-transitional",
"virtio-net-pci-transitional",
"virtio-input-host-device",
"virtio-input-host-pci",
"virtio-keyboard-device",
"virtio-keyboard-pci",
"virtio-mouse-device",
"virtio-mouse-pci",
"virtio-serial-device",
"virtio-serial-pci",
"virtio-serial-pci-non-transitional",
"virtio-serial-pci-transitional",
"virtio-tablet-device",
"virtio-tablet-pci",
"virtio-gpu-device",
"virtio-gpu-pci",
"virtio-vga",
"vhost-vsock-device",
"vhost-vsock-pci",
"vhost-vsock-pci-non-transitional",
"vhost-vsock-pci-transitional",
"virtio-balloon-device",
"virtio-balloon-pci",
"virtio-balloon-pci-non-transitional",
"virtio-balloon-pci-transitional",
"virtio-rng-device",
"virtio-rng-pci",
"virtio-rng-pci-non-transitional",
"virtio-rng-pci-transitional",

So yes, I think the filter could be removed, @yanan-fu what do you think?

yanan-fu:
@PaulYuuu  @vivianQizhu  Sorry for miss this patch before ...
It is ok to remove the filter with the current qemu, but i just have a bit concern: @vivianQizhu had commented above,  will other device have the same property but was not expected to be set in the future ? Not confident about this...    the param name with prefix `virtio_dev`  was designed for virtio device only.

I prefer to keep the filter a bit more:
1. Without the filter,  traverse all of devices looks a little redundant ?
2. It is easier to extend the filter if other devices with new name format come out than exclude devices ?

PaulYuuu:
> @PaulYuuu @vivianQizhu Sorry for miss this patch before ...
> It is ok to remove the filter with the current qemu, but i just have a bit concern: @vivianQizhu had commented above, will other device have the same property but was not expected to be set in the future ? Not confident about this... the param name with prefix `virtio_dev` was designed for virtio device only.
> 
> I prefer to keep the filter a bit more:
> 
> 1. Without the filter,  traverse all of devices looks a little redundant ?
> 2. It is easier to extend the filter if other devices with new name format come out than exclude devices ?

OK, let's keep it current. BTW, some virtio devices do not yet support iommu, we can consider modifying it in the future so that virtio properties can be added individually on each device.

virttest/qemu_vm.py
```diff
@@ -2460,25 +2452,29 @@ def sort_key(dev):
...
+            dev_info = devices.execute_qemu("-device %s,help" % dev_type)
+            dev_properties = re.findall(r"([a-zA-Z0-9_-]+)=\S+", dev_info)
+            disable_legacy = params.get("virtio_dev_disable_legacy")
+            disable_modern = params.get("virtio_dev_disable_modern")
+            iommu_platform = params.get("virtio_dev_iommu_platform")
+            ats = params.get("virtio_dev_ats")
+            if "disable-legacy" in dev_properties:
+                device.set_param("disable-legacy", disable_legacy)
+            if "disable-modern" in dev_properties:
+                device.set_param("disable-legacy", disable_modern)
```
vivianQizhu, line 2473: Here should be `"disable-modern"`, right?

virttest/qemu_vm.py
```diff
@@ -2460,25 +2452,29 @@ def sort_key(dev):
...
+            if "iommu_platform" in dev_properties:
+                device.set_param("disable-legacy", iommu_platform)
```
vivianQizhu, line 2475: And `"iommu_platform"`? Seems you wrongly copy `"disable-legacy"` for all of them?

virttest/qemu_vm.py
PaulYuuu: Yes, it's a mistake. Already updated.

virttest/qemu_vm.py
PaulYuuu: +1

virttest/qemu_vm.py
```diff
@@ -2461,25 +2453,29 @@ def sort_key(dev):
...
+            iommu_platform = params.get("virtio_dev_iommu_platform")
+            ats = params.get("virtio_dev_ats")
+            if "disable-legacy" in dev_properties:
+                device.set_param("disable-legacy", disable_legacy)
+            if "disable-modern" in dev_properties:
+                device.set_param("disable-modern", disable_modern)
+            if "iommu_platform" in dev_properties:
+                device.set_param("iommu_platform", iommu_platform)
+            if "ats" in dev_properties:
+                device.set_param("ats", ats)
```
vivianQizhu, line 2478: How about:
```
properties_to_be_set = {"disable-legacy": params.get("virtio_dev_disable_legacy"),
                        "disable-modern": params.get("virtio_dev_disable_modern"),
                        "iommu_platform": params.get("virtio_dev_iommu_platform"),
                        "ats": params.get("virtio_dev_ats")}
for key, value in properties_to_be_set.items():
    if key in dev_properties:
        device.set_param(key, value)
```

PaulYuuu: OK for me.

vivianQizhu (approved):

yanan-fu (approved):
Ack

vivianQizhu:
@yanan-fu Request for your ACK. Thanks.

yanan-fu:
> @yanan-fu Request for your ACK. Thanks.

Ah~  almost done at the same time, already acked, let's  merge it ~

PR #2780, opened 2020-09-30
virttest: add s390x crypto device support

smitterl:
1. utils_zcrypt: add s390x crypto device helpers

CryptoDeviceInfoXXX:
Add classes to store crypto device info on s390x systems
obtained from 'lszcrypt' for later passthrough setup.

load_vfio_ap:
Add method to assure passthrough driver is available.

APMaskHelper/MatrixDevice:
Add classes to control device availability and the matrix device
to be attached to the guest (details s. kernel doc).

2. libvirt_xml/devices/hostdev: add support for uuid address
Add support for `<address uuid="..."/>`.
Add support for `<hostdev model="..."/>`.

### DIFF
--- a/virttest/utils_zcrypt.py
+++ b/virttest/utils_zcrypt.py
@@ -1,0 +1,380 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Red Hat Inc. 2020
+# Author: Sebastian Mitterle <smitterl@redhat.com>
+
+"""
+Module for crypto device management on s390x aka IBM Z
+
+Example:
+
+    # get available domains and set up host
+    lszcrypt_info = CryptoDeviceInfoBuilder.get()
+    mask_helper = APMaskHelper.from_infos(lszcrypt_info.domains)
+    matrix_device = MatrixDevice.from_infos(lszcrypt_info.domains)
+    ...
+    # Use matrix_device.uuid for guest definition/device attachment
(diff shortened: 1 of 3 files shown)


smitterl:
Unit tests added were found reasonable (de-/serialization) and used in tp-libvirt system level test:
https://github.com/autotest/tp-libvirt/pull/3104#issuecomment-701265142

dzhengfy (approved):
LGTM

virttest/utils_zcrypt.py
```diff
@@ -0,0 +1,378 @@
...
+        self.session = session
+
+    def get_info(self):
+        """
+        Gets crypto device info
+
+        :return: CryptoDeviceInfo instance
+        """
+        info = CryptoDeviceInfo()
+        e, o = cmd_status_output("lszcrypt -V", shell=True,
```
chunfuwen, line 67: e,o can be more meaningful variables rather than single letter

virttest/utils_zcrypt.py
```diff
@@ -0,0 +1,378 @@
...
+    def domain(self):
+        """
+        The domain property
+
+        :return: domain value
+        """
+        return self._domain
+
+    @staticmethod
+    def from_string(line):
```
chunfuwen, line 187: May I ask whether those member variables can be wrapped into __init__(), rather than introduce additional static method from_string ?

virttest/utils_zcrypt.py
smitterl: @chunfuwen Yep, I changed to "err, out".

virttest/utils_zcrypt.py
smitterl: @chunfuwen I'm sorry, I don't understand. Which member variable do you mean could be wrapped into **init()**?

Having `from_string` is a common constructor pattern in python, see for example comment https://github.com/avocado-framework/avocado-vt/pull/2610#discussion_r445400984

If you mean that the `__init__` should have signature `__init__(self, line)` then I prefer not to do that. By having the initialization (with None) and the setting of fields from a string separate, I can still decide to manually define an instance `CryptoDeviceInfoEntry` without the need to first codify the desired entry into a string.

chunfuwen (approved):
LGTM

PR #2781, opened 2020-10-10
env_process:check value of cpu_family in case it's None

chloerh:
cpu_family could be 'None' according to function get_family(), and
'NoneType' is not iterable. Therefore check whether it's None or not
before string comparing.


pevogam (approved):
Awesome fix, thanks!

clebergnu (approved):
LGTM, thanks!

xutian:
LGTM

PR #2789, opened 2020-10-19
utils_net: Update to start openvswitch if needed

Yingshun:
If the openvswitch service is not running on the host, we should
start it before testing.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -35,7 +35,9 @@ from virttest import arch
 from virttest import utils_selinux
 from virttest import utils_package
 
+from virttest.remote import RemoteRunner
 from virttest.staging import utils_memory
+from virttest.staging import service
 from virttest.versionable_class import factory
 from virttest.utils_windows import virtio_win
 from virttest.utils_windows import system
@@ -4259,6 +4261,12 @@ def create_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
     :param ignore_status: Whether to raise an exception when command fails
     :return: The command status and output
     """
+    runner = process.run
+    if session:
+        runner = RemoteRunner(session=session).run
+    openvswitch_serv = service.Factory.create_service("openvswitch", runner)
+    if not openvswitch_serv.status():
+        openvswitch_serv.start()
     runner = local_runner
     if session:
         runner = session.cmd
@@ -4289,6 +4297,12 @@ def delete_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
     :param ignore_status: Whether to raise an exception when command fails
     :return: The command status and output
     """
+    runner = process.run
+    if session:
+        runner = RemoteRunner(session=session).run
+    openvswitch_serv = service.Factory.create_service("openvswitch", runner)
+    if not openvswitch_serv.status():
+        openvswitch_serv.start()
     runner = local_runner
     if session:
         runner = session.cmd


Yingshun:
**Test result:**
```

2020-10-19 05:25:02,398 process          L0626 INFO | Running 'systemctl status openvswitch'
2020-10-19 05:25:02,425 process          L0458 DEBUG| [stdout] * openvswitch.service - Open vSwitch
2020-10-19 05:25:02,426 process          L0714 INFO | Command 'systemctl status openvswitch' finished with 3 after 0.020351171493530273s
2020-10-19 05:25:02,427 process          L0458 DEBUG| [stdout]    Loaded: loaded (/usr/lib/systemd/system/openvswitch.service; disabled; vendor preset: disabled)
2020-10-19 05:25:02,427 process          L0458 DEBUG| [stdout]    Active: inactive (dead) (thawing) since Mon 2020-10-19 05:24:46 EDT; 15s ago
2020-10-19 05:25:02,428 process          L0458 DEBUG| [stdout]   Process: 91294 ExecStop=/bin/true (code=exited, status=0/SUCCESS)
2020-10-19 05:25:02,428 process          L0458 DEBUG| [stdout]  Main PID: 78494 (code=exited, status=0/SUCCESS)
2020-10-19 05:25:02,428 process          L0458 DEBUG| [stdout]
... ... 
systemd[1]: Stopped Open vSwitch.
2020-10-19 05:25:02,455 service          L0472 DEBUG| Setting ignore_status to True.
2020-10-19 05:25:02,456 process          L0626 INFO | Running 'systemctl start openvswitch'
2020-10-19 05:25:02,964 process          L0714 INFO | Command 'systemctl start openvswitch' finished with 0 after 0.50032639503479s

```

Yingshun:
It's introduced by 2e361394efde1468f.

kylazhang (approved):

virttest/utils_net.py
```diff
@@ -4259,6 +4261,12 @@ def create_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
     :param ignore_status: Whether to raise an exception when command fails
     :return: The command status and output
     """
+    runner = process.run
+    if session:
+        runner = RemoteRunner(session=session).run
```
pevogam, line 4266: Just wondering, is there a way to achieve the same without using the a remote runner? This class was planned to be deprecated and I am not sure what is it that it provides here that a normal remote session won't provide.

Yingshun: No, I think RemoteRunner.run() is the only way to manage services with session via create_service() for now. It requires a process.CmdResult instance, so if you want to remove virttest.remote.RemoteRunner, maybe need to update service.FactoryHelper() or add some new one.

BTW, may I ask why you want to deprecate RemoteRunner? We have 11 files in tp-libvirt which call that, some cases will be impacted if you do that. 
So I wonder if we can get more information about the virttest.remote changes. Is there an issue or something to track the changes?

Yingshun: @pevogam  Would you please review it again? Thanks a lot!

pevogam: Hi, sorry I somehow missed your last comment.

> No, I think RemoteRunner.run() is the only way to manage services with session via create_service() for now.

If you mean system services than we are just adding a [utility](https://github.com/avocado-framework/aexpect/pull/54/files#diff-a5831b347e80eb0db7ab3201c357ec5244c2d053d3641457aefe6db58c219147R515) doing this as well (I guess for an unrelated reason).

> BTW, may I ask why you want to deprecate RemoteRunner? We have 11 files in tp-libvirt which call that, some cases will be impacted if you do that.

If we go anywhere near the plan we will make sure to make everyone know and deprecate it slowly so don't expect any breakage any time soon. That doesn't mean however that there aren't better options to move to even right now.

> So I wonder if we can get more information about the virttest.remote changes. Is there an issue or something to track the changes?

The basic PR where the `remote` module was moved to aexpect (where it belongs together with all session handling) was the one where we first wondered how many people are using the RemoteRunner and decided to keep it as a leftover utility in VT that we might one day deprecate.

In your particular case I still don't see anything special about using the RemoteRunner other than the additional `CmdResult` which could either be added as a shared wrapper of a session result or the API of a few places could simply be adapted to use a command status and output as two separate variables.

Yingshun: @pevogam Thanks for you reply! I understand your concerns, but I'm wondering if we can merge this pr if the schedule of deprecating RemoteRunner is not fixed.  It has been created for one month and a high priority case's failed by it.

@dzhengfy I know you have a lot of experience in remote session operations, could you please share your thought?

pevogam (approved):
Hi @Yingshun, this looks good. As long as you are acknowledging that support for the remote runner might not be forever and are clear on the alternatives we should look at in the future, I am not going to block this change.

Yingshun:
> Hi @Yingshun, this looks good. As long as you are acknowledging that support for the remote runner might not be forever and are clear on the alternatives we should look at in the future, I am not going to block this change.

Yes, got it. Thank you @pevogam !

PR #2791, opened 2020-10-21
utils_cgroup: update 'get_load_per_cpu' to get all types of cpu times

yanan-fu:
Gather all types of cpu times instead of only one type, so that
user can decide which one to use.

id: 1875683

### DIFF
--- a/virttest/staging/utils_cgroup.py
+++ b/virttest/staging/utils_cgroup.py
@@ -614,21 +614,23 @@ def get_load_per_cpu(_stats=None):
     """
     Gather load per cpu from /proc/stat
     :param _stats: previous values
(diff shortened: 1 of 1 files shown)


yanan-fu:
@luckyh Could you help review ? Thanks !

yanan-fu:
@luckyh  As we discussed, do not do data conversion, return the origin data format. Please help review again, thanks!

luckyh (approved):
LGTM

yanan-fu:
@xiangchunfu  Could you please help review ? Thanks!

xiangchunfu:
@yanan-fu  reviewed and tested this patch. It works.
Acked-by:Xiangchun Fu xfu@redhat.com

PR #2805, opened 2020-10-28
Update virsh secret set value with file option

chunfuwen:
As per https://bugzilla.redhat.com/show_bug.cgi?id=1826636,
virsh secret-set-value will throw error if pass secret value by --base64 option
Read the secret from a file is right choice.

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -23,6 +23,7 @@ for non-existant keys.
 
 import signal
 import logging
+import os
 import re
 import weakref
 import time
@@ -40,6 +41,7 @@ from six.moves import urllib
 
 from virttest import propcan
 from virttest import utils_misc
+from virttest import data_dir
 
 
 # list of symbol names NOT to wrap as Virsh class methods
@@ -3794,7 +3796,7 @@ def secret_get_value(uuid, options=None, **dargs):
     return command(cmd, **dargs)
 
 
-def secret_set_value(uuid, password, options=None, encode=False, **dargs):
+def secret_set_value(uuid, password, options=None, encode=False, use_file=False, **dargs):
     """
     Set a secret value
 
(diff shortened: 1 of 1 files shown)


chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.nbd.coldplug.enable_tls.private_key_encryption.disable_export.lifecycle_operate

dzhengfy (approved):
LGTM

virttest/virsh.py
```diff
@@ -3808,10 +3810,14 @@ def secret_set_value(uuid, password, options=None, encode=False, **dargs):
     if password:
         if encode:
             encoding = locale.getpreferredencoding()
-            cmd += (" --base64 %s"
```
lento-sun, line 3811: '--base64' is still an available choice.  When use it, libvirt will return 'error: Passing secret value as command-line argument is insecure!'. But actually it's just a warning instead of a 'error', and value could be set correctly. BZ1826636 was to ask devel to change the error message to a warning message (but it's not worth fixing due to devel's comment, so it's closed). 
So this PR will remove ability of using '--base64' and introduce '--file'. My suggestion is to keep them both, and use a new param such as 'secret_set_value(..., use_file=False, ...)'. Pls have a check, thx.

chunfuwen: I add one parameter use_base64=False since --file is recommend way.

lento-sun: @chunfuwen The '--file' is a recent change in rhel-av build. See the following comit is IN origin/next-av-8.3.0 but NOT IN origin/rhel-8.4.0 branch
```
commit dbbc74e4ce2b76598b3c0fa6e41ab0da143621f0
Author: Peter Krempa <pkrempa@redhat.com>
Date:   Fri Jan 10 15:54:05 2020 +0100

    virsh: secret: Add --file 'filename' support for secret-set-value
    
    The necessity to specify the secret value as command argument is
    insecure. Allow reading the secret from a file.
    
    Signed-off-by: Peter Krempa <pkrempa@redhat.com>
    Reviewed-by: Daniel P. Berrangé <berrange@redhat.com>
```
So I guess if we use '--file' as default setting, some cases will be failed in following situtaion:
1. Non AV builds
2. AV builds < rhelav8.3

chunfuwen: thanks for reminding, updated according

lento-sun (approved):
LGTM

PR #2813, opened 2020-11-04
adapt inittab pattern for more OSes

zhijianli88:
set_console_getty() will update /etc/inittab to allow user login from a specific
device(ttyS0 for example) if /etc/inittab takes effect.

We check the specific pattern in /etc/inittab to decide if /etc/inittab is still
used. However for modern OSes, the patterns are a bit difference, so we should
need to update the pattern to adapt more OSes.

RHEL7: inittab is no longer used when using systemd
RHEL8/Fedora27: inittab is no longer used

Signed-off-by: Li Zhijian <lizhijian@cn.fujitsu.com>

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -1826,8 +1826,8 @@ class VM(virt_vm.BaseVM):
             try:
                 # Only configurate RHEL5 and below
                 regex = "gettys are handled by"
-                # As of RHEL7 systemd message is displayed
-                regex += "|inittab is no longer used when using systemd"
+                # As of RHEL7/8 systemd message is displayed
+                regex += "|inittab is no longer used"
                 output = session.cmd_output("cat /etc/inittab")
                 if re.search(regex, output):
                     logging.debug("Skip setting inittab for %s", device)


chloerh (commented):
Could you provide some test result?

zhijianli88:
> Could you provide some test result?

this patch will no impact the result in my environment, it doesn't matter  though something are written to /etc/initttab
```
$ cat /etc/inittab 
# inittab is no longer used.
#
# ADDING CONFIGURATION HERE WILL HAVE NO EFFECT ON YOUR SYSTEM.
#

```

dzhengfy:
@zhijianli88 Please rebase your branch and push it again forcely. The centos 7 failures has been fixed already.

zhijianli88:
> @zhijianli88 Please rebase your branch and push it again forcely. The centos 7 failures has been fixed already.

Done

chloerh (approved):

chloerh:
Just one small thing, how about we rewrite the commit message to make it more readable? I'm embarrassed that I didn't understand the message the first time :sweat_smile:

zhijianli88:
> Just one small thing, how about we rewrite the commit message to make it more readable? I'm embarrassed that I didn't understand the message the first time

Sure!! fell free to update it.

chloerh:
> > Just one small thing, how about we rewrite the commit message to make it more readable? I'm embarrassed that I didn't understand the message the first time
> 
> Sure!! fell free to update it.

:joy:  I mean could you help rewrite it? I don't think I can edit the commit message.

chloerh:
```
As content of file /etc/inittab on RHEL8/Fedora27 differs from RHEL7, update regex to match with it.
RHEL7: inittab is no longer used when using systemd
RHEL8/Fedora27: inittab is no longer used
```
Something like this

PR #2815, opened 2020-11-05
libvirt_vm: Add method set_state to change VM state

qiankehan:
For virsh, all the VM states in defined at virshDomainState[1]. Except
abnormal states "no state", "idle", "crashed" and transient state "in
shutdown", the state "running", "paused", "pmsuspended" could be
switched to by virsh commands. This method also has a state "os-booted". That
state will be useful to some cases requiring a booted OS, like live
device attach/detach.

[1]: https://gitlab.com/libvirt/libvirt/-/blob/master/tools/virsh-domain-monitor.c#L134

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -281,6 +281,123 @@ class VM(virt_vm.BaseVM):
         except process.CmdError:
             return False
 
+    def set_state(self, state, shutoff_method="destroy", pm_target="mem"):
+        """
+        Set VM state.
+        The supported target states are listed in the method local variable "target_states"
+        Return True if the VM state is set to target state. Otherwise return False.
+
+        param state, the target state of VM
+        param shutoff_method: The method to convert VM state to shutoff. Supported methods
+            are listed in the local variable "shutoff_methods"
+        param pm_target: The targets for pmsuspended. Supported targets are listed in the local
+            variable "pm_targets"
+        """
+        target_states = ["running", "os-booted", "paused", "pmsuspended", "shutoff"]
+        abnormal_states = ["no state", "idle", "crashed"]
+        shutoff_methods = ["shutdown", "destroy"]
+        pm_targets = ["mem", "disk", "hybrid"]
+        cur_state = self.state()
+        if cur_state == "running" and self.wait_for_login():
+            cur_state = "os-booted"
+
+        if cur_state == "in shutdown":
+            if utils_misc.wait_for(lambda: self.state() == "shutoff", timeout=360, first=10.0,
+                                   step=5.0, text="Waiting for VM shutdown",
+                                   ignore_errors=False) is None:
+                raise AssertionError("VM %s keeps in '%s' state for more than 6 minutes"
+                                     % (self.name, cur_state))
+
+        if cur_state in abnormal_states:
+            raise AssertionError("The VM %s's state %s is abnormal" % (self.name, cur_state))
+
+        if state not in target_states:
+            raise AssertionError("Unable to set VM %s's state from %s to %s: the target state is not"
+                                 "one of %s" % (self.name, cur_state, state, target_states))
+
+        if shutoff_method not in shutoff_methods:
+            raise AssertionError("Invalid shutoff method %s: it should be one of %s"
+                                 % (shutoff_method, shutoff_methods))
+
+        if pm_target not in pm_targets:
+            raise AssertionError("Invalid pmsuspend target %s: it should be one of %s"
+                                 % (pm_target, pm_targets))
+
+        if cur_state == state or (cur_state == "os-booted" and state == "running"):
+            logging.info("Current VM %s's state %s is the same with the target state",
+                         self.name, cur_state)
+            return True
+
+        if (cur_state == "running" and state == "os-booted"):
+            logging.error("Unable to set the state of VM %s from %s to %s",
+                          self.name, cur_state, state)
+            return False
+
+        if cur_state in ["os-booted", "running"]:
+            if state == "paused":
+                self.pause()
+            elif state == "shutoff":
+                if shutoff_method == "shutdown":
+                    self.shutdown()
+                else:
+                    self.destroy()
+            elif state == "pmsuspended":
+                self.pmsuspend(target=pm_target)
+        elif cur_state == "paused":
+            if state == "running":
+                self.resume()
+            elif state == "os-booted":
+                self.resume()
+                if self.wait_for_login():
+                    return True
+                else:
+                    logging.error("Fail to login %s when set VM state from %s to %s",
+                                  self.name, cur_state, state)
+                    return False
+            elif state == "shutoff":
+                self.destroy()
+            elif state == "pmsuspended":
+                self.resume()
+                self.pmsuspend(target=pm_target)
+        elif cur_state == "shutoff":
+            if state == "running":
+                self.start()
+            elif state == "os-booted":
+                self.start()
+                if self.wait_for_login():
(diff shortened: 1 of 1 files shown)


virttest/libvirt_vm.py
```diff
@@ -281,6 +281,97 @@ def is_autostart(self):
...
+            if self.state() == "in shutdown":
+                logging.error("VM %s still keep in 'in shutdown' state after 30s waiting",
+                              self.name)
+
+        if cur_state in abnormal_states:
+            logging.error("The VM %s's state %s is abnormal", self.name, cur_state)
+            return
+
+        if states not in target_states:
+            logging.error("Unable to set VM %s's state to %s", self.name, state)
```
chloerh, line 310: How about we add more info here, to let users know which states are supported?
Something like `'state xxx is not one of %s' % target_states`

virttest/libvirt_vm.py
```diff
@@ -281,6 +281,97 @@ def is_autostart(self):
...
+                virsh.dompmwakeup(self.name, uri=self.connect_uri, debug=debug)
+                if not self.wait_for_login():
+                    logging.error("Fail to loggin %s when set VM state from %s to %s",
+                                  self.name, cur_state, state)
+            elif state == "paused":
+                virsh.dompmwakeup(self.name, uri=self.connect_uri, debug=debug)
+                virsh.suspend(self.name, uri=self.connect_uri, debug=debug)
+            elif state == "shutoff":
+                virsh.destroy(self.name, uri=self.connect_uri, debug=debug)
+
```
chloerh, line 374: How to make sure that the virsh commands don't fail when executing them? How about we set `ignore_status=False`?
Or we could add state checking after setting part is done, maybe a `wait_for(vm state is xxx)` ?

virttest/libvirt_vm.py
```diff
@@ -281,6 +281,97 @@ def is_autostart(self):
...
+            time.sleep(30)
+            if self.state() == "in shutdown":
+                logging.error("VM %s still keep in 'in shutdown' state after 30s waiting",
+                              self.name)
+
+        if cur_state in abnormal_states:
+            logging.error("The VM %s's state %s is abnormal", self.name, cur_state)
+            return
+
+        if states not in target_states:
```
chloerh, line 309: Should be 'state'

chloerh (commented):
This function could be very useful, nice work! And could you rebase and push again? I think there's something wrong with code check.

virttest/libvirt_vm.py
smitterl: Isn't it rather `"Setting VM %s to state %s is not supported."`?

virttest/libvirt_vm.py
```diff
@@ -281,6 +281,97 @@ def is_autostart(self):
...
+            return
+
+        if (cur_state == "running" and state == "os-booted"):
+            logging.error("Unable to set the state of VM %s from %s to %s",
+                          self.name, cur_state, state)
+            return
+
+        if cur_state in ["os-booted", "running"]:
+            if state == "paused":
+                virsh.suspend(self.name, uri=self.connect_uri, debug=debug)
```
smitterl, line 325: Why don't you use the existing `virsh.pause()`, `virsh.resume()` etc.? Wouldn't this remove code duplication `(self.name, uri=self.connect_uri, debug=debug)`?

virttest/libvirt_vm.py
smitterl: If we set `ignore_status=False`, the test code would have to test if the target state would have to test if the expected target state was actually established, shouldn't that be the responsibility of this very method `set_state`?

IMO, it makes sense to use `wait_for`, e.g. the pattern
```
wait_for(vm state is xxx, ignore_error=True, ...)
if vm.state() != expected_state:
 raise AssertionError("Couldn't establish state %s....)
```

smitterl (changes requested):
Thank you @qiankehan, IMO this is a useful addition. I added some comments. What I would want to insist on changing:

a) use `wait_for` instead of `time.sleep`
b) the `set_state` should raise an exception if the target state cannot be established so that the test code doesn't have to implement that check for each test.

virttest/libvirt_vm.py
qiankehan: Thanks for your suggestions. I will combine these two together, then the error message will be like:
"Unable to set state %s from %s: the target state is not the one of %s" %(cur_state, state, target_states)

virttest/libvirt_vm.py
qiankehan: > Why don't you use the existing `virsh.pause()`, `virsh.resume()` etc.? Wouldn't this remove code duplication 
`(self.name, uri=self.connect_uri, debug=debug)`?

Not such method `virsh.pause()` but `vm.pause()`. `vm.pause()` seems better because of the error handling. I will use it.

qiankehan:
> Thank you @qiankehan, IMO this is a useful addition. I added some comments. What I would want to insist on changing:
> 
> a) use `wait_for` instead of `time.sleep`
> b) the `set_state` should raise an exception if the target state cannot be established so that the test code doesn't have to implement that check for each test.

For a), I will use `wait_for` to replace `time.sleep()`
For b), for the abnormal vm states and the long time *in shutdown*, it will raise exceptions. Otherwise if the state is not set to target state, the `set_state()` will return false.

qiankehan:
V2 updates:
1. Replace `time.sleep()` with `wait_for()`
2. Update return values, `True` for VM state changed to target state, `False` for not changed to target status. And add some assertions
3. Some trivial updates from comments above

smitterl:
> > Thank you @qiankehan, IMO this is a useful addition. I added some comments. What I would want to insist on changing:
> > a) use `wait_for` instead of `time.sleep`
> > b) the `set_state` should raise an exception if the target state cannot be established so that the test code doesn't have to implement that check for each test.
> 
> For a), I will use `wait_for` to replace `time.sleep()`
> For b), for the abnormal vm states and the long time _in shutdown_, it will raise exceptions. Otherwise if the state is not set to target state, the `set_state()` will return false.

IMO, this is more complex than it needs too be. It seems that you have two different behaviors now for the function `set_state`. If you use both control patterns - exceptions and return value - to both mean "we could not establish the desired state" the user code would always have to handle both cases:
```python
try:
  state_could_be_set = vm.set_state(some_state)
  if not state_could_be_set:
     # HANDLE e.g.
     test.error("could not set state")
except:
  # HANDLE e.g.
  test.error("could not set state")
```
Therefore, I would prefer you decided for one of both ways, either exceptions or a boolean return value. If you  choose a boolean return value, I believe that introduces more client code OR clients missing to handle the situation:
A)
```python
  # handle
  state_could_be_set = vm.set_state(some_state)
  if not state_could_be_set:
     # HANDLE e.g.
     test.error("could not set state")
```
will lead to test case result `ERROR: could not set state ...`

B)
```python
 # don't handle
 vm.set_state(some_state)
 # some instructions down the line fail because the machine is not in the right state
 # e.g. update some configuration on disk
 update_disk_config(vm)
```
could lead to test case result `FAIL: could not update disk configuration`, and the test log might give the additional info "Cannot update disk on avocado-vt-vm1. Domain is not running.", that is, the root cause and the actual failure are not close and the result requires more human analysis.

If, OTOH, you always used a `AssertionError` or maybe even a `exceptions.TestError`, I expect the test result should be `ERROR: AssertionError: couldn't set vm state`, resp. `ERROR couldn't set vm state`.

Therefore, I personally would prefer to use always an `Error` instead of return value.

PR #2823, opened 2020-11-11
unattended.py: Add new parameters for kickstart script

6-dehan:
Add parameters for kickstart script

ID: 1890839

### DIFF
--- a/virttest/tests/unattended_install.py
+++ b/virttest/tests/unattended_install.py
@@ -399,6 +399,11 @@ class UnattendedInstallConfig(object):
             raise ValueError("Unexpected installation medium %s" % self.url)
         contents = re.sub(dummy_medium_re, content, contents)
 
+        dummy_rh_virt_qga_re = r'\bRH_VIRT_QGA\b'
+        if re.search(dummy_rh_virt_qga_re, contents):
+            rh_virt_qga = self.params.get("rh_virt_qga", "")
+            contents = re.sub(dummy_rh_virt_qga_re, rh_virt_qga, contents)
+
         dummy_rh_system_stream_id_re = r'\bRH_SYSTEM_STREAM_ID\b'
         if re.search(dummy_rh_system_stream_id_re, contents):
             rh_system_stream_id = self.params.get("rh_system_stream_id", "")


vivianQizhu:
@yanan-fu Would you please help review this one? Thanks.

yanan-fu:
@6-dehan  I think, it is not must to introduce two parameters `RH_GUEST_NAME` and `RH_VIRT_MODULE_STREAM`.   Use one to specify the required information for your test is ok.

virttest/tests/unattended_install.py
```diff
@@ -399,6 +399,16 @@ def answer_kickstart(self, answer_path):
             raise ValueError("Unexpected installation medium %s" % self.url)
         contents = re.sub(dummy_medium_re, content, contents)
 
+        dummy_guest_name_re = r'\bRH_GUEST_NAME\b'
```
pevogam, line 402: Is there are reason for the `RH_` prefix? The guest could be any distro AFAIK.

virttest/tests/unattended_install.py
```diff
@@ -399,6 +399,16 @@ def answer_kickstart(self, answer_path):
...
+        if re.search(dummy_guest_name_re, contents):
+            guest_name = self.params.get("guest_name")
+            contents = re.sub(dummy_guest_name_re, guest_name, contents)
```
pevogam, line 405: Won't this error out if `guest_name` is not defined?

virttest/tests/unattended_install.py
```diff
@@ -399,6 +399,16 @@ def answer_kickstart(self, answer_path):
...
+
+        dummy_virt_module_stream_re = r'\bRH_VIRT_MODULE_STREAM\b'
+        if re.search(dummy_virt_module_stream_re, contents):
+            virt_module_stream = self.params.get("virt_module_stream", "")
+            contents = re.sub(dummy_virt_module_stream_re, virt_module_stream, contents)
```
pevogam, line 410: Same here.

virttest/tests/unattended_install.py
vivianQizhu: @pevogam Thanks for your review. These piece of code is for internal RHEL build(not released version), other distributions should not need this parameter, it will skip this step by default since Line 403 would be False. Since there is no similar requirement from other distributions so far, I suppose we could adjust it if there is in the future.

virttest/tests/unattended_install.py
vivianQizhu: @pevogam It has to be defined together with `dummy_guest_name_re`, so it is intended to raise error if it is not, but should not be here, the above line should be `guest_name = self.params["guest_name"]`, @6-dehan

yanan-fu: @pevogam @vivianQizhu IMO, it is not must to define `guest_name` together with `dummy_guest_name_re`. Miss `guest_name` should not effect the test.   But you are right, the code need be updated, i suggest `params.get("guest_name", '')`.  And i suggest to use one param to cover this as the pre comment, that is another story ~

1. No `guest_name`, but `dummy_guest_name_re` in the `content`, it will be replaced with '',  and some test was expected to be skipped in ks.
2. No `dummy_guest_name_re` in the `content`, but `guest_name` was specified, it will not take effect in the ks, nothing ~

virttest/tests/unattended_install.py
yanan-fu: > Same here.

same with above,  this line is correct i think, since it use default `""`

virttest/tests/unattended_install.py
vivianQizhu: @yanan-fu Yes you are right.

vivianQizhu:
> @6-dehan I think, it is not must to introduce two parameters `RH_GUEST_NAME` and `RH_VIRT_MODULE_STREAM`. Use one to specify the required information for your test is ok.

@6-dehan What do you think about this? Prefer to keep it?

6-dehan:
> > @6-dehan I think, it is not must to introduce two parameters `RH_GUEST_NAME` and `RH_VIRT_MODULE_STREAM`. Use one to specify the required information for your test is ok.
> 
> @6-dehan What do you think about this? Prefer to keep it?

I prefer to keep guest_name. you know the guest_name will get string sort of 'rhel830' and virt_module_stream will get '8.2' or '8.3',  but for rhel840, we change the 'virt_module_stream' as 'av', so it's not related 8.3 or 8.4 even 8.5 in the future any more, maybe not, but for now,we need a stable string to help us query target variable. guest_name won't change as my opinion. at lease it's more stable than virt_module_stream.

6-dehan:
> > > @6-dehan I think, it is not must to introduce two parameters `RH_GUEST_NAME` and `RH_VIRT_MODULE_STREAM`. Use one to specify the required information for your test is ok.
> > 
> > 
> > @6-dehan What do you think about this? Prefer to keep it?
> 
> I prefer to keep guest_name. you know the guest_name will get string sort of 'rhel830' and virt_module_stream will get '8.2' or '8.3', but for rhel840, we change the 'virt_module_stream' as 'av', so it's not related 8.3 or 8.4 even 8.5 in the future any more, maybe not, but for now,we need a stable string to help us query target variable. guest_name won't change as my opinion. at lease it's more stable than virt_module_stream.

sorry not changed 'virt_module_stream', I should explain that change  "virt_module_stream = 8.3"  as "virt_module_stream = av". more clear and understandable.

yanan-fu:
@6-dehan  Both guest name and virt module stream was used to generate the tag. That is the current implement that need use guest name, but it is not the only solution to get the expected tag.

ex:
For slow: `***= rhel:840`, 
For fast: `*** = av:840`, or `***= 8.3:831`

6-dehan:
> @6-dehan Both guest name and virt module stream was used to generate the tag. That is the current implement that need use guest name, but it is not the only solution to get the expected tag.
> 
> ex:
> For slow: `***= rhel:840`,
> For fast: `*** = av:840`, or `***= 8.3:831`
if I didn't misunderstand you

virttest/tests/unattended_install.py
pevogam: Using `params.get("guest_name", '')` sounds much better indeed. And using more specific parameter name than `guest_name` is definitely welcome too.

6-dehan:
@vivianQizhu @yanan-fu . the parameter has changed as "module_stream_n_release_ver", which binds original two paramters we talked about, and its usage changed as following command. if you guys don't think this new name or usage is suitable for our test name rules, please let me know, I'll appreciate it. 
python3 ConfigTest.py xxxx  --customsparams="module_stream_n_release_ver = 8.3:831"

yanan-fu:
@6-dehan About the parameter name, how do you think about `virt_guest_agent` or `virt_qga` ?
Some of my thoughts: it is used to get the expected qemu-guest-agent, the method to get the info may be different for different product, and request for different input `virt_qga`.

6-dehan:
> @6-dehan About the parameter name, how do you think about `virt_guest_agent` or `virt_qga` ?
> Some of my thoughts: it is used to get the expected qemu-guest-agent, the method to get the info may be different for different product, and request for different input `virt_qga`.

yes, you're right, It's a better way to set this parameter reply on our requirement , updated it. thanks for your suggestions.

6-dehan:
> > @6-dehan About the parameter name, how do you think about `virt_guest_agent` or `virt_qga` ?
> > Some of my thoughts: it is used to get the expected qemu-guest-agent, the method to get the info may be different for different product, and request for different input `virt_qga`.
> 
> yes, you're right, It's a better way to set this parameter reply on our requirement , updated it. thanks for your suggestions.

and according this suggestion, test result is still passed, so you can have a look with this now.

virttest/tests/unattended_install.py
```diff
@@ -399,6 +399,11 @@ def answer_kickstart(self, answer_path):
             raise ValueError("Unexpected installation medium %s" % self.url)
         contents = re.sub(dummy_medium_re, content, contents)
 
+        dummy_virt_module_stream_re = r'\bRH_VIRT_QGA\b'
```
yanan-fu, line 402: Could you update the `dummy_virt_module_stream_re` together.

6-dehan: oh, sorry, my bad

virttest/tests/unattended_install.py
```diff
@@ -399,6 +399,11 @@ def answer_kickstart(self, answer_path):
...
+        if re.search(dummy_virt_module_stream_re, contents):
+            virt_module_stream = self.params.get("virt_qga", "")
```
yanan-fu, line 404: It is better to keep consistent `rh_virt_qga`

6-dehan: @yanan-fu please have a review.

yanan-fu:
Ack

yanan-fu (approved):
@vivianQizhu  Could you help review too? Thanks!

vivianQizhu (approved):

vivianQizhu:
Thanks all.

PR #2825, opened 2020-11-16
Update customize_libvirt_config for modular daemons

Yingshun:
There are 8 sub daemons instead of a libvirtd when modular daemons
are enabled. In previous code, it changed 'libvirtd' to 'virtqemud'
for modular daemons env which fixed some problems. But for now, we
still need to support other daemons. So it is extended to all
modular daemons.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -3587,7 +3587,7 @@ def check_machine_type_arch(machine_type):
 
 
 def customize_libvirt_config(params,
-                             config_type="libvirtd",
+                             config_type="virtqemud",
                              remote_host=False,
                              extra_params=None,
                              is_recover=False,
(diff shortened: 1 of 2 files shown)


Yingshun:
@fangge1212 @chloerh  Could you please help review it? Thank You!

virttest/utils_test/libvirt.py
```diff
@@ -3611,8 +3611,14 @@ def customize_libvirt_config(params,
     :return: utils_config.LibvirtConfigCommon object
     """
     # Hardcode config_type to virtqemud under modularity daemon mode when config_type="libvirtd"
+    # On the contrary, hardcode it to "libvirtd" when config_type="virt*d".
     # Otherwise accept config_type as it is.
-    if utils_split_daemons.is_modular_daemon():
+    if not utils_split_daemons.is_modular_daemon():
+        if config_type in ["virtqemud", "virtproxyd", "virtnetworkd",
+                           "virtstoraged", "virtinterfaced", "virtnodedevd",
+                           "virtnwfilterd"]:
```
fangge1212, line 3619: Missed virtsecretd here?

virttest/utils_test/libvirt.py
```diff
@@ -3611,8 +3611,14 @@ def customize_libvirt_config(params,
     :return: utils_config.LibvirtConfigCommon object
     """
     # Hardcode config_type to virtqemud under modularity daemon mode when config_type="libvirtd"
+    # On the contrary, hardcode it to "libvirtd" when config_type="virt*d".
```
fangge1212, line 3614: Maybe it would be more clear if move the above two comment lines into corresponding if/else body

virttest/utils_test/libvirt.py
Yingshun: Yes, you're right. Updated.

virttest/utils_test/libvirt.py
Yingshun: I've updated the code's order as the comment. I think it should be better now.

fangge1212 (approved):

chloerh (approved):

PR #2836, opened 2020-11-25
Fix ProtectedKeyHelper

smitterl:
1. All methods were run on host. Pass guest session to assure
commands are run in guest if session is present.

2. The utils_misc.cmd_status_output function tries to decode
protected key token as utf-8 which can cause an error. Define a local
version just passing the stdout as is.

### DIFF
--- a/virttest/utils_libvirt/libvirt_keywrap.py
+++ b/virttest/utils_libvirt/libvirt_keywrap.py
@@ -4,7 +4,7 @@ Module to contain logic testing protected key module 'pkey'
 import logging
 import os
 
-from virttest.utils_misc import cmd_status_output
+from avocado.utils import process
 
 
 class ProtectedKeyHelper(object):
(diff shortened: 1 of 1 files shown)


virttest/utils_libvirt/libvirt_keywrap.py
```diff
@@ -36,8 +37,29 @@ def get_some_aes_key_token(self):
...
+    :param cmd: Command to issue.
+    :param session: Guest session. If empty, command is exeucted on host.
+    """
+
+    status = None
+    stdout = None
+    if session:
+        status, stdout = session.cmd_status_output(cmd, timeout=timeout)
+    else:
+        result = process.run(cmd, shell=False, ignore_status=True,
```
chunfuwen, line 61: suggest ignore_status may be pulled up to cmd_status_output

smitterl: @chunfuwen I would like to keep it this way the function `cmd_status_output` is not supposed to be consumed outside of the module. Actually, it could better be prefixed by "_" to make this clearer.
Let me know, thanks.

chunfuwen (approved):
LGTM

chunfuwen:
@smitterl , By the way , could you upload any available test result?

smitterl:
> @smitterl , By the way , could you upload any available test result?

@chunfuwen Sure, the test runs are on the tp-libvirt PR comment: https://github.com/autotest/tp-libvirt/pull/3174#issuecomment-733562821

chloerh (approved):

PR #2847, opened 2020-12-03
utils_time: Add a timeout parameter

YanhuiMa:
Adding a timeout parameter for logging in with host load

ID: 1903945
Signed-off-by: yama <yama@redhat.com>

### DIFF
--- a/virttest/utils_time.py
+++ b/virttest/utils_time.py
@@ -49,13 +49,14 @@ def verify_timezone_linux(session):
 
 
 @error_context.context_aware
-def sync_timezone_linux(vm):
+def sync_timezone_linux(vm, login_timeout=360):
     """
     Sync linux guest's timezone
 
     :param vm: Virtual machine object
+    :param login_timeout: Time (seconds) to keep trying to log in.
     """
-    session = vm.wait_for_login(serial=True)
+    session = vm.wait_for_login(timeout=login_timeout, serial=True)
     error_context.context("Sync guest's timezone", logging.info)
     set_timezone_cmd = "timedatectl set-timezone %s"
     if not verify_timezone_linux(session):
@@ -117,13 +118,14 @@ def verify_timezone_win(session):
 
 
 @error_context.context_aware
-def sync_timezone_win(vm):
+def sync_timezone_win(vm, login_timeout=360):
     """
     Verify and sync windows guest's timezone
 
     :param vm: Virtual machine object
+    :param login_timeout: Time (seconds) to keep trying to log in
     """
-    session = vm.wait_for_login(serial=True)
+    session = vm.wait_for_login(timeout=login_timeout, serial=True)
     set_timezone_cmd = 'tzutil /s "%s"'
     (ver_result, output) = verify_timezone_win(session)
 


YanhuiMa:
(1/1) type_specific.io-github-autotest-qemu.timerdevice.boot_test.with_boot.with_host_load.clksource_kvm-clock.base_utc.clock_host: PASS

virttest/utils_time.py
```diff
@@ -49,13 +49,14 @@ def verify_timezone_linux(session):
 
 
 @error_context.context_aware
-def sync_timezone_linux(vm):
+def sync_timezone_linux(vm, timeout=600):
```
yanan-fu, line 52: The meaning is not clear, it is login timeout for `vm.wait_for_login`, or it is the timeout for `verify_timezone_linux` ?
Also,  `sync_timezone_win`  have the same issue i think.

YanhuiMa: No problem with sync_timezone_win, because the host load are different from linux.

YanhuiMa: changing timeout to login_timeout.

yanan-fu: > No problem with sync_timezone_win, because the host load are different from linux.

Eh~  It is not for the specific test scenario that you hit problem at the current stage.

YanhuiMa: session = vm.wait_for_login(serial=True)  I mean, if no special requirement, the default timeout for logging in is enough.

yanan-fu: > session = vm.wait_for_login(serial=True) I mean, if no special requirement, the default timeout for logging in is enough.

We can not assume there "no special requirement" in the future, so i suggest apply the same rule for both linux and windows.

YanhuiMa: If you think it is necessary, I can add it for windows. But should we submit a new patch for it? Because it doesn't have relationship with the bug, for windows, it belongs to enhance the code, not a bug.

yanan-fu: I think no need,  it is just add a param for both windows and linux together.

YanhuiMa: @yanan-fu Updated
win2012r2:
PASS 2-type_specific.io-github-autotest-qemu.timerdevice.boot_test.with_boot.with_host_load.clksource_unchanged.base_localtime.clock_host

yanan-fu:
@YanhuiMa  Isn't `600s` too long as default ?

@vivianQizhu could you help review this too, thanks~

vivianQizhu:
> @YanhuiMa Isn't `600s` too long as default ?
> 
> @vivianQizhu could you help review this too, thanks~

@yanan-fu If the failure is only because it takes a while to boot up, I think it's okay to extend the timeout to wait for it, in case it is not a performance test. Just to make sure it will not be a product bug from feature owner's perspective @YanhuiMa.

YanhuiMa:
> > @YanhuiMa Isn't `600s` too long as default ?
> > @vivianQizhu could you help review this too, thanks~
> 
> @yanan-fu If the failure is only because it takes a while to boot up, I think it's okay to extend the timeout to wait for it, in case it is not a performance test. Just to make sure it will not be a product bug from feature owner's perspective @YanhuiMa.

Thanks for confirmation. No product bug, serial login just needs to take a longer time with host 50% or 100% load.

yanan-fu:
@vivianQizhu  It is not product issue, not performance issue as @YanhuiMa mentioned.

Many other cases use this function without specifiy the `login_timeout`, and `600` is too long i think.
For the case that use stress and need extend the login_timeout value which @YanhuiMa is working on now,  600 is ok. But it should not be the default value.

vivianQizhu:
> @vivianQizhu It is not product issue, not performance issue as @YanhuiMa mentioned.
> 
> Many other cases use this function without specifiy the `login_timeout`, and `600` is too long i think.
> For the case that use stress and need extend the login_timeout value which @YanhuiMa is working on now, 600 is ok. But it should not be the default value.

@yanan-fu Sounds make sense. And I will prefer `vm.wait_for_login(params['login_timeout'])` instead of specifically defining it for the function, but not a strong opinion, anyway they all work.

YanhuiMa:
@yanan-fu I will remove the default value, just like this sync_timezone_linux(vm, login_timeout), is it ok?

yanan-fu:
> @yanan-fu I will remove the default value, just like this sync_timezone_linux(vm, login_timeout), is it ok?

I am afraid not,  many other call for this function does not pass the second argument, no default value means you need update all the call. Do you plan to do that ? 
If you plan to update all the call, use `params['login_timeout']` is more better as @vivianQizhu comment above , but if do not want to effect other call, i suggest `sync_timezone_linux(vm, login_timeout=360)`.

YanhuiMa:
@yanan-fu Updated

yanan-fu:

@vivianQizhu  Could you help re-review ? Thanks!

vivianQizhu (approved):

PR #2852, opened 2020-12-08
Add encryption support for domain backup xml

lento-sun:
With the implementation of libvirt RFE bz1811906, luks encryption
was supported during domain backup process. Now, the SCRATCH
of the pull-mode backup and the TARGET of the push-mode can be
encrypted with LUKS. And the usage is to add a 'encrytion' subelement
under 'scratch' or 'target' elements. This commit is to support it.

### DIFF
--- a/virttest/libvirt_xml/backup_xml.py
+++ b/virttest/libvirt_xml/backup_xml.py
@@ -9,7 +9,39 @@ from virttest.libvirt_xml import base, accessors, xcepts
 
 class BackupXML(base.LibvirtXMLBase):
     """
-    Domain backup XML class
+    Domain backup XML class. Following are 2 samples for push mode backup xml
+    and pull mode backup xml.
+    Push mode:
+    <domainbackup mode="push">
+      <incremental>checkpoint_0</incremental>
+        <disks>
+            <disk name='vda' backup='no'/>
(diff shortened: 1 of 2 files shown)


lento-sun:
This will effect exisitng tp-libvirt cases, will set 'ready for review' after update these cases

virttest/libvirt_xml/backup_xml.py
```diff
@@ -116,11 +115,96 @@ def __init__(self, virsh_instance=base.virsh):
...
+                                     subclass=self.DiskScratch,
+                                     subclass_dargs={
+                                         'virsh_instance': virsh_instance})
+
             super(self.__class__, self).__init__(virsh_instance=virsh_instance)
             self.xml = "<disk/>"
+
+        class DiskTarget(base.LibvirtXMLBase):
+            """
+            Backup disk target XML class
```
chunfuwen, line 136: suggest add example what DiskTarget looks like

lento-sun: thx for the suggestion, just added two xml samples in class BackupXML's docstring, this may be more clear to instruct the usage. pls have a check.

chunfuwen (approved):
LGTM

chloerh (approved):

PR #2859, opened 2020-12-10
`--vt-type` default

richtja:
In the commit f54b602 was introduced a configuration of command line options
compatible with avocado Job API. Unfortunately there missing default value for
`--vt-type` option on avocado versions 80 and lower. This commit fixing this
problem.


rdkdd (approved):
Tested - lgtm

PR #2862, opened 2020-12-14
Enable to config other elements in prepare_guest_agent

iccaszhulili:
Enable config source name and target path when preparing
guest agent

Signed-off-by: Lili Zhu <lizhu@redhat.com>

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2954,7 +2954,8 @@ class VM(virt_vm.BaseVM):
             logging.error("Removal of package %s failed", name)
         session.close()
 
-    def prepare_guest_agent(self, prepare_xml=True, channel=True, start=True):
+    def prepare_guest_agent(self, prepare_xml=True, channel=True, start=True,
+                            source_path=None, target_name='org.qemu.guest_agent.0'):
         """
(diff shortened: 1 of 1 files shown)


virttest/libvirt_vm.py
```diff
@@ -2954,23 +2954,40 @@ def remove_package(self, name, ignore_status=False):
...
+            # whether the agent is existing.
+            is_existing = False
+            ga_channels = vmxml.get_agent_channels()
+            for chnl in ga_channels:
+                name = chnl.find('./target').get('name')
+                try:
+                    path = chnl.find('./source').get('path')
+                except AttributeError:
+                    path = None
+                if (name == target_name and path == source_path):
```
chunfuwen, line 2983: Do we need check whether source_path is not None? otherwise if source_path is None, and path is None, it may falsely set is_existing = True.

iccaszhulili: @chunfuwen if source_path==path==None, and path==source_path, then it is the agent xml
I needed. is_existing is True indeed. May I did not get exactly what you mean, please help to check.

chunfuwen (approved):
LGTM

dzhengfy (approved):
LGTM

PR #2866, opened 2020-12-18
[libvirt.attach_disks] set correct disk_type for lvm

smitterl:
disk_type is used both for disk definition xml and for create_local_disk function
call. Since commit avocado-vt 18fd732 create_local_disk won't
default to lvm for value `block` anymore. Therefore, any call to attach_disks with disk_type
block is currently failing. Those calls should be fixed to pass `lvm` now to be more specific
as later changes to create_local_disk might add other block disk types besides lvm.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -2927,7 +2927,7 @@ def attach_disks(vm, path, vgname, params):
             set_controller_multifunction(vm.name, disk_target)
 
         disk_params = {}
-        disk_params['type_name'] = disk_type
+        disk_params['type_name'] = disk_type if not disk_type == 'lvm' else 'block'
         disk_params['target_dev'] = target_dev
         disk_params['target_bus'] = disk_target
         disk_params['device_type'] = params.get("device_type", "disk")


smitterl:
Code and test runs using this fix: https://github.com/autotest/tp-libvirt/pull/3216

virttest/utils_test/libvirt.py
```diff
@@ -2927,7 +2927,7 @@ def generate_disks_index(count, target="virtio"):
             set_controller_multifunction(vm.name, disk_target)
 
         disk_params = {}
-        disk_params['type_name'] = disk_type
+        disk_params['type_name'] = disk_type if not disk_type == 'lvm' else 'block'
```
chunfuwen, line 2930: I assume  we can explicitly set  disk_type == 'lvm'  or  'block', rather than do judgement.

smitterl: @chunfuwen I don't understand, sorry. Do you mean to avoid the `not`?
```python
disk_params['type_name'] = 'block' if disk_type == 'lvm' else disk_type
```

smitterl: @chunfuwen IIUC, unfortunately we can't do that because currently
a.) when passing disk_type = lvm to attach_disks, then this very line is set to 'lvm' which is an invalid value on the domain xml
b.) when passign disk_type = block to attach_disks then the logic for create_local_disk - called by attach_disks - will fail
This change maintains api for create_local_disks but adapts to create_local_disks api (needing lvm explicitly since at least https://github.com/avocado-framework/avocado-vt/commit/18fd732a30147cc60f4abc0b9d87e23ee37b901c )
Although this change changes the attach_disks api the risk is controlled, see details on comment https://github.com/avocado-framework/avocado-vt/pull/2866#issuecomment-748496864

chunfuwen: thanks for updates!

chunfuwen (approved):
LGTM

PR #2875, opened 2021-01-04
virt_vm: avoid of vm's unstable status

dzhengfy:
This function usually is invoked after vm starts. But sometimes on specific host, like
in openstack, the vm starts a little slow, so serial console login may get unexpected
result when recogizing the console output message. This is to make sure the vm is in
proper state before console message is prompted and recoginized correctly.

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -1380,6 +1380,7 @@ class BaseVM(object):
             eg. during reboot or pause)
         :return: ConsoleSession instance.
         """
+        time.sleep(60) # Guest needs more time to start sometimes.
         logging.debug("Attempting to log into '%s' via serial console "
                       "(timeout %ds)", self.name, timeout)
         end_time = time.time() + timeout


virttest/virt_vm.py
```diff
@@ -1380,6 +1380,7 @@ def wait_for_serial_login(self, timeout=LOGIN_WAIT_TIMEOUT,
             eg. during reboot or pause)
         :return: ConsoleSession instance.
         """
+        time.sleep(60) # Guest needs more time to start sometimes.
```
sathnaga, line 1383: this is going to affect and add every test a blind minute wait which would be costly when running tests in batches, I would say make it a controlled via user param and control it in your local test config and by default it will behaves as before without wait.

sathnaga (changes requested):

virttest/virt_vm.py
dzhengfy: OK, thanks

PR #2876, opened 2021-01-04
guest_agent: Add new command 'guest-get-disks' for checking disks info

6-dehan:
Add command 'guest-get-disk' for checking disks info.

ID: 1906654


vivianQizhu:
ACK.

PR #2886, opened 2021-01-15
XMLElementList: create parent xpath if not exists

chloerh:
It fails if trying to assign an XMLElementList if its parent xpath
not exists. This commit will try to create the missing parent xpath.

### DIFF
--- a/virttest/libvirt_xml/accessors.py
+++ b/virttest/libvirt_xml/accessors.py
@@ -786,7 +786,12 @@ class XMLElementList(AccessorGeneratorBase):
             # Allow other classes to generate parent structure
             parent = self.xmltreefile().find(self.parent_xpath)
             if parent is None:
-                raise xcepts.LibvirtXMLNotFoundError
+                # Create parent xpath if not exists
+                self.xmltreefile().create_by_xpath(self.parent_xpath)
+                parent = self.xmltreefile().find(self.parent_xpath)
+                if parent is None:
+                    raise xcepts.LibvirtXMLNotFoundError(
+                        'Parent xpath %s not found.' % self.parent_xpath)
             # Remove existing by calling accessor method, allowing
             # any "untouchable" or "filtered" elements (by marshal)
             # to be ignored and left as-is.


chloerh:
Before
---
```
from virttest.libvirt_xml.vm_xml import VMCPUXML
... cpuxml = VMCPUXML()
... cells = cpuxml.dicts_to_cells([{'id':'0','cpus':'0-1','memory':'1024000','unit':'KiB'}])
... cpuxml.numa_cell = cells
... print(cpuxml)
  File "<input>", line 4, in <module>
  File "/root/avocado-vt/virttest/propcan.py", line 235, in __setattr__
    return self.__setitem__(key, value)
  File "/root/avocado-vt/virttest/propcan.py", line 196, in __setitem__
    return accessor(value)
  File "/root/avocado-vt/virttest/libvirt_xml/accessors.py", line 789, in __call__
    raise xcepts.LibvirtXMLNotFoundError
virttest.libvirt_xml.xcepts.LibvirtXMLNotFoundError
```

After
---
```
from virttest.libvirt_xml.vm_xml import VMCPUXML
... cpuxml = VMCPUXML()
... cells = cpuxml.dicts_to_cells([{'id':'0','cpus':'0-1','memory':'1024000','unit':'KiB'}])
... cpuxml.numa_cell = cells
... print(cpuxml)
<cpu><numa><cell cpus="0-1" id="0" memory="1024000" unit="KiB" /></numa></cpu>
```

Yingshun (approved):
LGTM

virttest/libvirt_xml/accessors.py
```diff
@@ -786,7 +786,12 @@ def __call__(self, value):
             # Allow other classes to generate parent structure
             parent = self.xmltreefile().find(self.parent_xpath)
             if parent is None:
-                raise xcepts.LibvirtXMLNotFoundError
+                # Create parent xpath if not exists
+                self.xmltreefile().create_by_xpath(self.parent_xpath)
```
chunfuwen, line 790: This is super class for all libvirtXML, can we test more types of xml other than VMCPUXML ?

chloerh: @chunfuwen  That's a good idea, but unfortunately, we do not have that kind of test cases for now. Would you like to give some other suggestions about the tests? Thanks.

chunfuwen: something similar  like above VMCPUXML, for example:create xml such as disk, network

chloerh: @chunfuwen  Again, I don't think there is that kind of tests now. Could you help provide the test cases you mentioned? I have tried, but cannot find them by myself. Thank you very much.

chloerh:
@dzhengfy Would you mind taking a look?

virttest/libvirt_xml/accessors.py
dzhengfy, line 790: It is a little hard to find a suitable test case to verify this change. 
Considering this is a common enhancement, and similar workaround has already been tested in vm_xml.py like below, I evaluate this change has low risk. So I approve it and would like to observe if any other cases will fail against this.
                     ```
   if numa_number is not None and numa_number > 0:
                            vmcpu_xml.xmltreefile.create_by_xpath('/numa')
                            vmcpu_xml.numa_cell = vmcpu_xml.dicts_to_cells(nodexml_list)
```

dzhengfy (approved):

PR #2897, opened 2021-01-20
ip_sniffing: Support tshark (wireshark) 3.x

luckyh:
From tshark 3.0.0, the field name - bootp [1] is no longer supported
and users should use dhcp [2] instead.

    tshark: Some fields aren't valid:
        bootp.option.dhcp
        bootp.ip.your
        ...
    (Process terminated with status 1)

Updating the module accordingly.

References:
[1] https://www.wireshark.org/docs/dfref/b/bootp.html
[2] https://www.wireshark.org/docs/dfref/d/dhcp.html

ID: 1917682

### DIFF
--- a/virttest/ip_sniffing.py
+++ b/virttest/ip_sniffing.py
@@ -10,10 +10,12 @@ from collections import Iterable
 import aexpect
 from aexpect.remote import handle_prompts
 from avocado.utils import path as utils_path
+from avocado.utils import process
 
 import six
 
 from virttest.utils_misc import log_line
+from virttest.utils_version import VersionInterval
 
 
 class AddrCache(object):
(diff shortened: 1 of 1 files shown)


luckyh:
@yanan-fu @PaulYuuu request for your review, thanks!

luckyh:
@kylazhang @chunfuwen @chloerh would you please help review the patch as you may be interested as well? thanks!

luckyh:
Test result:

tshark 3.4.0
```
2021-01-20 03:10:31: Running as user "root" and group "root". This could be dangerous.
2021-01-20 03:10:31: ip.src ip.dst dhcp.type dhcp.id dhcp.hw.mac_addr dhcp.ip.your dhcp.option.dhcp ipv6.src ipv6.dst
2021-01-20 03:10:31: Capturing on 'any'
2021-01-20 03:10:34: 0.0.0.0 255.255.255.255 1 0xcd508cbd 9a:55:1b:xx:xx:xx 0.0.0.0 3
2021-01-20 03:10:34: 0.0.0.0 255.255.255.255 1 0xcd508cbd 9a:55:1b:xx:xx:xx 0.0.0.0 3
...
```

tshark 2.6.2
```
2021-01-20 03:27:17: Running as user "root" and group "root". This could be dangerous.
2021-01-20 03:27:17: ip.src ip.dst bootp.type bootp.id bootp.hw.mac_addr bootp.ip.your bootp.option.dhcp ipv6.src ipv6.dst
2021-01-20 03:27:17: Capturing on 'any'
2021-01-20 03:27:21:        fe80::9eb6:54xx:xxxx:xxxx ff02::1:2
2021-01-20 03:27:22:        fe80::fe15:b4xx:xxxx:xxxx ff02::1:2
...
```

yanan-fu:
Test with both tshark 3.4.0 and tshark 2.6.2 env,  can get ip address successfully, and ip-sniffer.log is normally now with this change.  And the change LGTM

yanan-fu (approved):

chloerh:
Running an acceptance test job with this pr on both x86 and ppc. 
Will get x86 result within 1 hour. PPC job is queuing for machine.

chloerh (approved):
All 27 cases on x86 passed.

virttest/ip_sniffing.py
```diff
@@ -323,5 +346,35 @@ def _output_handler(self, line):
             return True
 
 
+class TShark1To2(TSharkSniffer):
+
+    """
+    TShark sniffer class for version 1.0.0 - 2.x.x.
+    """
+
+    options = ("-npi any -T fields -E separator=/s -E occurrence=f "
```
PaulYuuu, line 355: I see that except for `bootp` and `dhcp`, other options are the same. Is it better to extract the options and pass `bootp`/`dhcp`?

luckyh: I'm not sure if it is possible, would you like to give an example? thanks.

PaulYuuu: Em, it's possible but need to put it into the `__init__` magic method, which is not worth it. Let's keep it as it is.

PaulYuuu (approved):

virttest/ip_sniffing.py
luckyh: Yep, thanks.

yanan-fu:
@luckyh  @PaulYuuu Do you have any more comment ?   If not, could we merge it since there are two acks already ? Thanks!

luckyh:
> @luckyh @PaulYuuu Do you have any more comment ? If not, could we merge it since there are two acks already ? Thanks!

Sure, thanks all!

PR #2903, opened 2021-01-22
Add ceph setup utility module

chunfuwen:
Enable add ceph backend disk into VM

### DIFF
--- a/virttest/utils_libvirt/libvirt_ceph_utils.py
+++ b/virttest/utils_libvirt/libvirt_ceph_utils.py
@@ -1,0 +1,174 @@
+"""
+High-level libvirt ceph utility functions.
+
+This module is meant to reduce code size by performing common ceph setup procedures.
+:author: Chunfu Wen <chwen@redhat.com>
+:copyright: 2021 Red Hat Inc.
+"""
+
+import logging
+import os
+
+from avocado.core import exceptions
+from avocado.utils import process
+
+from virttest import ceph
+from virttest import data_dir
+from virttest import utils_package
+from virttest import virsh
+
+from virttest.utils_test import libvirt
+from virttest.utils_libvirt import libvirt_disk
+
+from virttest.libvirt_xml import vm_xml
+
+
+def _create_secret(auth_sec_usage_type, ceph_auth_key):
+    """
+    Setup secret with auth
+
+    :param auth_sec_usage_type: auth secret usage
+    :param ceph_auth_key: ceph auth key
+    :return: auth secret uuid
+    """
+    auth_sec_dict = {"sec_usage": auth_sec_usage_type,
+                     "sec_name": "ceph_auth_secret"}
+    auth_sec_uuid = libvirt.create_secret(auth_sec_dict)
+    virsh.secret_set_value(auth_sec_uuid, ceph_auth_key,
+                           debug=True)
+    return auth_sec_uuid
+
+
+def _create_image(device_format, img_file, vm_name, storage_size,
+                  ceph_disk_name, ceph_mon_ip, key_opt,
+                  ceph_auth_user, ceph_auth_key):
+    """
+    Creates image on ceph
+    :param device_format: device format
+    :param img_file: image file
+    :param vm_name: vm name
+    :param storage_size: storage size
+    :param ceph_disk_name: ceph disk name
+    :param ceph_mon_ip: ceph host ip
+    :param key_opt: key option
+    :param ceph_auth_user: ceph auth user
+    :param ceph_auth_key: ceph auth key
+    """
+    #Create necessary image file if not exists
+    if img_file is None:
+        img_file = os.path.join(data_dir.get_data_dir(),
+                                "%s_test.img" % vm_name)
+        # Create an local image and make FS on it.
+        disk_cmd = ("qemu-img create -f %s %s %s" %
+                    (device_format, img_file, storage_size))
+        process.run(disk_cmd, ignore_status=False, shell=True)
+    # Convert the image to remote ceph storage
+    disk_path = ("rbd:%s:mon_host=%s" %
+                 (ceph_disk_name, ceph_mon_ip))
+    if ceph_auth_user and ceph_auth_key:
+        disk_path += (":id=%s:key=%s" %
+                      (ceph_auth_user, ceph_auth_key))
+    rbd_cmd = ("rbd -m %s %s info %s 2> /dev/null|| qemu-img convert -O"
+               " %s %s %s" % (ceph_mon_ip, key_opt, ceph_disk_name,
+                              device_format, img_file, disk_path))
+    process.run(rbd_cmd, ignore_status=False, shell=True, verbose=True)
+
+
+def create_or_cleanup_ceph_backend_vm_disk(vm, params, is_setup=True):
+    """
+    Setup vm ceph disk with given parameters
+
+    :param vm: the vm object
+    :param params: dict, dict include setup vm disk xml configurations
+    :param is_setup: one parameter indicate whether setup or clean up
+    """
+    vmxml = vm_xml.VMXML.new_from_dumpxml(vm.name)
+    logging.debug("original xml is: %s", vmxml)
+
+    # Device related configurations
+    device_format = params.get("virt_disk_device_format", "raw")
+    device_bus = params.get("virt_disk_device_bus", "virtio")
+    device = params.get("virt_disk_device", "disk")
+    device_target = params.get("virt_disk_device_target", "vdb")
+    hotplug = "yes" == params.get("virt_disk_device_hotplug", "no")
+    keep_raw_image_as = "yes" == params.get("keep_raw_image_as", "no")
+
+    # Ceph related configurations
+    ceph_mon_ip = params.get("ceph_mon_ip", "EXAMPLE_MON_HOST")
+    ceph_host_port = params.get("ceph_host_port", "EXAMPLE_PORTS")
+    ceph_disk_name = params.get("ceph_disk_name", "EXAMPLE_SOURCE_NAME")
+    ceph_client_name = params.get("ceph_client_name")
+    ceph_client_key = params.get("ceph_client_key")
+    ceph_auth_user = params.get("ceph_auth_user")
+    ceph_auth_key = params.get("ceph_auth_key")
+    auth_sec_usage_type = params.get("ceph_auth_sec_usage_type", "ceph")
+    storage_size = params.get("storage_size", "1G")
+    img_file = params.get("ceph_image_file")
+    attach_option = params.get("virt_device_attach_option", "--live")
+    key_file = os.path.join(data_dir.get_tmp_dir(), "ceph.key")
+    key_opt = ""
+    is_local_img_file = True if img_file is None else False
+    rbd_key_file = None
+
+    # Prepare a blank params to confirm if delete the configure at the end of the test
+    ceph_cfg = ""
+    disk_auth_dict = None
+    auth_sec_uuid = None
+    names = ceph_disk_name.split('/')
+    pool_name = names[0]
+    image_name = names[1]
+    if not utils_package.package_install(["ceph-common"]):
+        raise exceptions.TestError("Failed to install ceph-common")
+
+    # Create config file if it doesn't exist
+    ceph_cfg = ceph.create_config_file(ceph_mon_ip)
+    # If enable auth, prepare a local file to save key
+    if ceph_client_name and ceph_client_key:
+        with open(key_file, 'w') as f:
+            f.write("[%s]\n\tkey = %s\n" %
+                    (ceph_client_name, ceph_client_key))
+        key_opt = "--keyring %s" % key_file
+        rbd_key_file = key_file
+    if is_setup:
+        # If enable auth, prepare disk auth
+        if ceph_client_name and ceph_client_key:
+            auth_sec_uuid = _create_secret(auth_sec_usage_type, ceph_auth_key)
+            disk_auth_dict = {"auth_user": ceph_auth_user,
+                              "secret_type": auth_sec_usage_type,
+                              "secret_uuid": auth_sec_uuid}
+        # clean up image file if exists
+        ceph.rbd_image_rm(ceph_mon_ip, pool_name,
+                          image_name, keyfile=rbd_key_file)
+
+        #Create necessary image file if not exists
+        _create_image(device_format, img_file, vm.name, storage_size,
+                      ceph_disk_name, ceph_mon_ip, key_opt,
+                      ceph_auth_user, ceph_auth_key)
+
+        # Disk related config
+        disk_src_dict = {"attrs": {"protocol": "rbd",
+                                   "name": ceph_disk_name},
+                         "hosts":  [{"name": ceph_mon_ip,
+                                     "port": ceph_host_port}]}
+        # Create network disk
+        disk_xml = libvirt_disk.create_primitive_disk_xml("network", device, device_target, device_bus,
+                                                          device_format, disk_src_dict, disk_auth_dict)
+        if not keep_raw_image_as:
+            if hotplug:
+                virsh.attach_device(vm.name, disk_xml.xml,
+                                    flagstr=attach_option, ignore_status=False, debug=True)
+            else:
+                vmxml = vm_xml.VMXML.new_from_dumpxml(vm.name)
+                vmxml.add_device(disk_xml)
+                vmxml.sync()
+    else:
+        ceph.rbd_image_rm(ceph_mon_ip, pool_name,
+                          image_name, keyfile=rbd_key_file)
+        # Remove ceph config and key file if created.
+        for file_path in [ceph_cfg, key_file]:
+            if os.path.exists(file_path):
+                os.remove(file_path)
+        if is_local_img_file and img_file and os.path.exists(img_file):
+            libvirt.delete_local_disk("file", img_file)
+        if auth_sec_uuid:
+            virsh.secret_undefine(auth_sec_uuid, ignore_status=True)


chunfuwen:
Test result  uploading is underway

virttest/utils_test/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,152 @@
...
+    img_file = params.get("ceph_image_file")
+    key_file = os.path.join(data_dir.get_tmp_dir(), "ceph.key")
+    key_opt = ""
+    is_local_img_file = False
+
+    # Prepare a blank params to confirm if delete the configure at the end of the test
+    ceph_cfg = ""
+    disk_auth_dict = None
+    if not utils_package.package_install(["ceph-common"]):
+        test.error("Failed to install ceph-common")
```
Yingshun, line 62: It should raise exceptions.TestError.

virttest/utils_test/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,152 @@
...
+from virttest import utils_package
+from virttest import ceph
+
+from virttest.utils_test import libvirt
+
+from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml.devices.disk import Disk
+
+
+def create_ceph_backend_vm_disk(vm, params, is_setup=True):
```
Yingshun, line 24: We'd better build a new function for disk removal. How do you think?

virttest/utils_test/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,152 @@
...
+        cmd = ("rbd -m {0} {1} info {2} && rbd -m {0} {1} rm "
+               "{2}".format(ceph_mon_ip, key_opt, ceph_disk_name))
+        cmd_result = process.run(cmd, ignore_status=True, shell=True)
+        logging.debug("result of rbd removal: %s", cmd_result)
+        # Remove ceph configure file if created.
+        if ceph_cfg:
+            os.remove(ceph_cfg)
+        # Remove ceph key file if has.
+        if os.path.exists(key_file):
+            os.remove(key_file)o
```
Yingshun, line 150: What does 'o' stand for?

Yingshun (commented):
BTW, it's better put this file under utils_libvirt.

virttest/utils_test/libvirt_ceph_utils.py
chunfuwen: updated

virttest/utils_test/libvirt_ceph_utils.py
chunfuwen: The logic and pattern  are very similar to existed utils such as :libvirt.setup_or_cleanup_iscsi,setup_or_cleanup_gluster.
change name to create_or_cleanup_ceph_backend_vm_disk

virttest/utils_test/libvirt_ceph_utils.py
chunfuwen: removed

chunfuwen:
The test result is under going

chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.blockcommit.normal_test.network_disk.ceph.relative_path.pause_domain.shallow.top_active

Yingshun:
We'd better to move  to utils_libvirt. how do you think?

virttest/utils_test/libvirt_ceph_utils.py
Yingshun: Yes, that's right, but that shouldn't be best practice. We'd better to ask a function to do one thing to make the code clean.

chunfuwen: @kylazhang 
@dzhengfy 

Do we have guidelines that this is not best practices?
Since if we go through utils module, most of functions or methods achieve  more than one single function, otherwise we may have much more numbers of functions or methods maintained? I assume we need keep balance among them

chunfuwen:
> We'd better to move to utils_libvirt. how do you think?

updated

chunfuwen:
@kylazhang 
@dzhengfy 

please help take a look.

chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.luks.positive_test.coldplug.device_disk.ceph_test.enable_auth.auth_in_source.encryption_in_source

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,154 @@
...
+import logging
+
+from avocado.utils import process
+from avocado.core import exceptions
+
+from virttest import data_dir
+from virttest import virsh
+from virttest import utils_package
+from virttest import ceph
+from virttest import utils_secret
```
chloerh, line 18: module names better be imported alphabetically.

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,154 @@
...
+def create_or_cleanup_ceph_backend_vm_disk(vm, params, is_setup=True):
+    """
+    Setup vm ceph disk with given parameters
+
+    :param vm: the vm object
+    :param params: dict, dict include setup vm disk xml configurations
+    :param is_setup: one parameter indicate whether setup or clean up
+    """
+    vmxml = vm_xml.VMXML.new_from_dumpxml(vm.name)
+    logging.debug("original xml is: %s", vmxml.xmltreefile)
```
chloerh, line 35: Why we're logging vmxml.xmltreefile instead of vmxml ?

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,154 @@
...
+            device_source = "rbd:%s:mon_host=%s:keyring=%s" % (ceph_disk_name,
+                                                               ceph_mon_ip,
+                                                               key_file)
+        else:
+            device_source = "rbd:%s:mon_host=%s" % (ceph_disk_name, ceph_mon_ip)
+        logging.debug("device source is: %s", device_source)
+        # clean up image file if exists
+        cmd = ("rbd -m {0} {1} info {2} && rbd -m {0} {1} rm "
+               "{2}".format(ceph_mon_ip, key_opt, ceph_disk_name))
+        process.run(cmd, ignore_status=True, shell=True)
```
chloerh, line 94: Better add verbose=True to print more info

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,154 @@
...
+
+        #Create necessary image file if not exists
+        if img_file is None:
+            img_file = os.path.join(data_dir.get_tmp_dir(),
```
chloerh, line 98: Considering the current tmpfs situation, I strongly suggest we run this test on out latest product, to verify.

chloerh: Whether the 'tmp' dir is suitable or not in  the future

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        # Convert the image to remote ceph storage
+        disk_path = ("rbd:%s:mon_host=%s" %
+                     (ceph_disk_name, ceph_mon_ip))
+        if ceph_client_name and ceph_client_key:
+            disk_path += (":id=%s:key=%s" %
+                          (ceph_auth_user, ceph_auth_key))
+        rbd_cmd = ("rbd -m %s %s info %s 2> /dev/null|| qemu-img convert -O"
+                   " %s %s %s" % (ceph_mon_ip, key_opt, ceph_disk_name,
+                                  device_format, img_file, disk_path))
+        process.run(rbd_cmd, ignore_status=False, shell=True)
```
chloerh, line 113: verbose=True

virttest/utils_test/libvirt_ceph_utils.py
chloerh: I agree with @Yingshun , the old way might not be the best. Splitting the function to 2 would make the calling of functions much more clear.

chunfuwen: Regarding API exposed by module, we need look at them from different levels(low level and high level).
We position utilities as high level. which provides quite complex or complicated functionality.
If someone can look at tensorflow or multiprocessing in python, both of them provides very simple API with very high level abstraction to users, but actually implement quite complicated logiic  in underlying logic. 
Without high level abstraction, we hardly imagin that we can achieve machine leaning in just a few lines. Or implement parellel handliing by calling  a few lines code of multiple processing

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: > module names better be imported alphabetically.

I observed there is guideline mentioning that import need to be categorized as system, thirty party ,and local .
I assume The import is more naturally arranged as where it is needed
Could you help share the guide or any reference that module names better be imported alphabetically ?

chloerh: Of course, I just learnt that knowledge a few weeks ago in 'Clean Code'. I'm sure you could find the info there. I won't paste the link here.

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: I see both are used somewhere ,and both can achieve full xml ouput

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: updated

virttest/utils_libvirt/libvirt_ceph_utils.py
chloerh: I know that, but we tend to keep our code short and clear. So I'd suggest we use vmxml instead of vmxml.xmltreefile.

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: That is good question. 
In any cases, we need one common utility data_dir.get_tmp_dir(), which quite large numbers of   case python file depends on heavily.
If data_dir.get_tmp_dir() has issue in new environment, the answer is clear and simple that we need fix in that, rather than painfully fix in each case python file

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: updated

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        if not keep_raw_image_as:
+            if hotplug:
+                attach_option = params.get("virt_device_attach_option", "--config")
+                virsh.attach_device(vm.name, disk_xml.xml,
+                                    flagstr=attach_option, ignore_status=False, debug=True)
+            else:
+                vmxml.add_device(disk_xml)
+                vmxml.sync()
+    else:
+        cmd = ("rbd -m {0} {1} info {2} && rbd -m {0} {1} rm "
```
Yingshun, line 142: we already have some ceph related utils in virttest.ceph, can we replace such kind of shell command lines with common utilities to improve readability?

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,154 @@
...
+            device_source = "rbd:%s:mon_host=%s:keyring=%s" % (ceph_disk_name,
+                                                               ceph_mon_ip,
+                                                               key_file)
+        else:
+            device_source = "rbd:%s:mon_host=%s" % (ceph_disk_name, ceph_mon_ip)
+        logging.debug("device source is: %s", device_source)
+        # clean up image file if exists
+        cmd = ("rbd -m {0} {1} info {2} && rbd -m {0} {1} rm "
+               "{2}".format(ceph_mon_ip, key_opt, ceph_disk_name))
+        process.run(cmd, ignore_status=True, shell=True, verbose=True)
```
Yingshun, line 94: Same as above

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        process.run(rbd_cmd, ignore_status=False, shell=True, verbose=True)
+
+        # Disk related config
+        disk_src_dict = {"attrs": {"protocol": "rbd",
+                                   "name": ceph_disk_name},
+                         "hosts":  [{"name": ceph_mon_ip,
+                                     "port": ceph_host_port}]}
+        # Create disk xml with given config
+        vmxml = vm_xml.VMXML.new_from_dumpxml(vm.name)
+        disk_xml = Disk(type_name='network')
```
Yingshun, line 122: Can you reuse libvirt.create_disk_xml() or libvirt.set_vm_disk() for this part?

chunfuwen: I don't think libvirt.create_disk_xml() or libvirt.set_vm_disk()  suit here since both of them involve in too many parameters which are not related, Here we only need one simple diskxml. For example create_disk_xml

            source_protocol = params.get("source_protocol")
            source_name = params.get("source_name")
            source_host_name = params.get("source_host_name").split()
            source_host_port = params.get("source_host_port").split()
            transport = params.get("transport")
            source_attrs = {'protocol': source_protocol, 'name': source_name}

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: > I know that, but we tend to keep our code short and clear. So I'd suggest we use vmxml instead of vmxml.xmltreefile.

isn't  vmxml.xmltreefile more meaningful than vmxml when talking about dumpxing files.

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: need add verbose=True, do you mean this ?

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: I can not understand what do you mean 'readiablity' here.
If we look at real execution logic in test cases, that is actual case that we use rbd command to execute that.
Doesn't it native and original command make readability friendly and easy to debug?

virttest/utils_libvirt/libvirt_ceph_utils.py
Yingshun: No, I'd like to propose to call ceph.rbd_image_rm.

chunfuwen:
@lento-sun 
please help take a look

virttest/utils_libvirt/libvirt_ceph_utils.py
chloerh: I don't think so. But keep it the way it is if you'd like. 
I just prefer shorter code which is more elegant.

virttest/utils_libvirt/libvirt_ceph_utils.py
chloerh: Please refer to https://github.com/avocado-framework/avocado-vt/pull/2990
Or if you'd like to fix it another way.

virttest/utils_libvirt/libvirt_ceph_utils.py
Yingshun: You might be familiar with rbd command lines, but it might be not easy for others to understand these shell command lines(eg, L92~94 & L142~144). Since the function already exists and can be used, I think the proper way is to try to use the existing way instead of repeating the operations by executing the shell command lines.

BTW, can you please reply the comment instead of edit mine? So I'll get a notification...

chunfuwen: 1)That is quote reply, not edit your comment.
2)process.run(verbose=True) implemented above will output both command line and result.  Don't forget process.run itself is one utility.
3)Better not assume that others should  use existed utility if there is one existed. It depends on some factors, such as robustness or flexability, complexisity the utility have.  in many cases, one utility can be called 'nice utility' if 80% scenarios can be covered.
4)Specifically to this utility, it separate rbd_pool_name, rbd_image_name as two parameters .Practically   in our cfg ,we only provides "pool_name/img_name" as one whole parameter. Do we need painfully split the parameters and rewrap them? Moreover , namespace parameter which rbd_image has , I don't think in rbd command area ,we have an mapping concept for this.

Yingshun: @dzhengfy @kylazhang @chloerh @smitterl @kamilvarga Would you mind sharing your ideas?

kylazhang: I agree with using the existing function instead of running cmd. If we do not use it, then why we make it as common function for ceph?

virttest/utils_libvirt/libvirt_ceph_utils.py
kylazhang: Agree to use get_data_dir especially on new written case

virttest/utils_libvirt/libvirt_ceph_utils.py
kylazhang: I think it support for any xml configuration, if you only need to give some of the parameters, it will generate xml with only what you are given. Please begin to use it as it is really useful and can make the code clean. It is our rule to write new cases that using the common function as much as possible. Thanks

virttest/utils_libvirt/libvirt_ceph_utils.py
kylazhang: I agree with yingshun as the common function will show more clear what you are doing and can make any update more easily, and maintain easily. Please use the common function as much as possible, thanks

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: > I think it support for any xml configuration, if you only need to give some of the parameters, it will generate xml with only what you are given. Please begin to use it as it is really useful and can make the code clean. It is our rule to write new cases that using the common function as much as possible. Thanks

@kylazhang 
In principal, I totally agree we need call current utility as we can as possible.
Let's elaborate more additional burdens If I use create_disk_xml:
Firstly I need rewrappped below parameters since parameter name is not consistent:
```
    params.update({"type_name" : 'network'})
    params.update({"target_dev" : device_target})
    params.update({"target_bus", device_bus})
    params.update({"device_type", device})
     params.update({"ceph_user", xx})
  .   ...
     ceph-related
secret-related
```
Secondly, auth_attrs is used to diff auth in source or out source in create_disk_xml, so I need prepare for another paramter

Thirdly, since I use netwrok, I need prepare more parmeter
'''
            source_protocol = params.get("source_protocol")
            source_name = params.get("source_name")
            source_host_name = params.get("source_host_name").split()
            source_host_port = params.get("source_host_port").split()
            transport = params.get("transport")
            source_attrs = {'protocol': source_protocol, 'name': source_name}
'''

So my question, so I need pull so many ruduants works ,but actually only 6 lines  can work out in this PR.
Hopefully you can understand more now.

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: > I agree with yingshun as the common function will show more clear what you are doing and can make any update more easily, and maintain easily. Please use the common function as much as possible, thanks

@kylazhang 
let us find seconds offline to discuss why one rule can not apply all over world

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: updated

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: > I agree with using the existing function instead of running cmd. If we do not use it, then why we make it as common function for ceph?

@kylazhang 
see previous explanation

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: updated

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,153 @@
...
+        disk_xml.driver = driver_dict
+        disk_source = disk_xml.new_disk_source(**disk_src_dict)
+        if disk_auth_dict:
+            logging.debug("disk auth dict is: %s" % disk_auth_dict)
+            disk_xml.auth = disk_xml.new_auth(**disk_auth_dict)
+        disk_xml.source = disk_source
+        logging.debug("new disk xml is: %s", disk_xml)
+        if not keep_raw_image_as:
+            if hotplug:
+                attach_option = params.get("virt_device_attach_option", "--config")
```
lento-sun, line 136: why we use '--config' here as the default "virt_device_attach_option"? we often call this kind of attach-device as 'cold plug' if '--config' added. Such as:
```
Bug 1216046 - guest which cold-plug too much memory device will disappear after restart libvirt
Bug 1559867 - When COLD PLUG multiple scsi hostdevs to an ACTIVE vm, their addresses will be duplicated
```

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,153 @@
...
+                          ceph_disk_name.split('/')[1], keyfile=rbd_key_file)
+        # Remove ceph configure file if created.
+        if os.path.exists(ceph_cfg):
+            os.remove(ceph_cfg)
+        # Remove ceph key file if has.
+        if os.path.exists(key_file):
+            os.remove(key_file)
+        if is_local_img_file and img_file and os.path.exists(img_file):
+            libvirt.delete_local_disk("file", img_file)
+        utils_secret.clean_up_secrets()
```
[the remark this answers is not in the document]
chunfuwen: it makes sense, updated

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: virt_device_attach_option is need to be specified in cfg dependent on hot or cold plug.  If not "--config" will be set as default value.
But anyway, I can change to "--live"

virttest/utils_libvirt/libvirt_ceph_utils.py
kylazhang: > @kylazhang
> In principal, I totally agree we need call current utility as we can as possible.
> Let's elaborate more additional burdens If I use create_disk_xml:
> Firstly I need rewrappped below parameters since parameter name is not consistent:
> 
> ```
>     params.update({"type_name" : 'network'})
>     params.update({"target_dev" : device_target})
>     params.update({"target_bus", device_bus})
>     params.update({"device_type", device})
>      params.update({"ceph_user", xx})
>   .   ...
>      ceph-related
> secret-related
> ```

I think you can update them all with 1 line, or just use the same name as in create_disk_xml, then this line is also not needed and may reduce some params.get ...

> 
> Secondly, auth_attrs is used to diff auth in source or out source in create_disk_xml, so I need prepare for another paramter
> 
I saw there are auth_attrs setting in different place in this func.
```
            if auth_in_source:
                disk_source.auth = diskxml.new_auth(**auth_attrs)
            else:
                diskxml.auth = diskxml.new_auth(**auth_attrs)
```
And just 1 param in cfg with the same name "auth_in_source" and should be no update needed in py file, I think it should be acceptable

> Thirdly, since I use netwrok, I need prepare more parmeter
> '''
> source_protocol = params.get("source_protocol")
> source_name = params.get("source_name")
> source_host_name = params.get("source_host_name").split()
> source_host_port = params.get("source_host_port").split()
> transport = params.get("transport")
> source_attrs = {'protocol': source_protocol, 'name': source_name}
> '''
No, seems it just meet your need
```
source_attrs = {'protocol': source_protocol, 'name': source_name}
for host_name, host_port in list(
            zip(source_host_name, source_host_port)):
     source_host.append({'name': host_name,
                                        'port': host_port})

```
just give 'source_protocol', 'source_name','source_host_name','source_host_type' in params(update params or just setting in cfg file), and may no needed for your setting in code
```
        disk_src_dict = {"attrs": {"protocol": "rbd",
                                   "name": ceph_disk_name},
                         "hosts":  [{"name": ceph_mon_ip,
                                     "port": ceph_host_port}]}
```
> 
> So my question, so I need pull so many ruduants works ,but actually only 6 lines can work out in this PR.
> Hopefully you can understand more now.

Please correct me if I have some misunderstanding as I do not update disk too much.

And more, the lines count in code may not prove everything, and the most we want to implement is to make the code more clear, tidy, easy to update and maintain, and for common functions are more widely used. Thanks

kylazhang (commented):
Please check the comment inline, thanks

kylazhang (commented):
Please check the comments inline, thanks

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: @kylazhang 
Quote "And more, the lines count in code may not prove everything, and the most we want to implement is to make the code more clear, tidy, easy to update and maintain, and for common functions are more widely used. Thanks"

I will invite you to join virtual disk cases automation. Look forward to your practice "And more, the lines count in code may not prove everything, and the most we want to implement is to make the code more clear, tidy, easy to update and maintain, and for common functions are more widely used. Thanks""

chunfuwen: In order to move this PR forward, I will attempt to create separate method to create disk.

virttest/utils_libvirt/libvirt_ceph_utils.py
dzhengfy: I also agree with using existing function instead of hardcoding the commands in every ceph related tp-libvirt cases. I remember I did same suggestion before in some ceph related cases before.

chunfuwen: > I also agree with using existing function instead of hardcoding the commands in every ceph related tp-libvirt cases. I remember I did same suggestion before in some ceph related cases before.

@dzhengfy 
it is ceph utiliy to be exposed ,not specific tp-libvirt cases
rbd command is changed to ceph.rbd_image_rm ,except that convert command,which is not supported in current ceph.py

virttest/utils_libvirt/libvirt_ceph_utils.py
kylazhang: Then that's OK, if the purpose is to avoid the circle import issues by not using create_disk_xml, I can accept the current implement method here

kylazhang (approved):

chunfuwen:
@chloerh , do you have more comments on this PR?

virttest/utils_test/libvirt_ceph_utils.py
Yingshun: environment setup and clean up are different operations and will be executed in different stages. We usually call a function to setup it and call a function to clean up it, no matter how big the project is.
I still think we should split them, but it's up to you.

chloerh:
> @chloerh , do you have more comments on this PR?

@chunfuwen I see that most of my comments have been resolved/replied, except for one about splitting the funtion into 2. But I was more about supporting @Yingshun  's suggestion, so it depends on her. I would be happy to approve when you guys resolve that one.

smitterl (commented):
@chunfuwen Hi! Thank you for this. This is useful.
TBH, I had a hard time reviewing this. The run method is very long and difficult to read.
I refactored it a bit and I believe this way it's easier to read and review. Furthermore, I could reduce some duplication in the code (`_remove_image` and `_Config.is_auth_case`).

The main function now simply reads:

```python
def create_or_cleanup_ceph_backend_vm_disk(vm, params, is_setup=True):
    """
    Setup vm ceph disk with given parameters
    :param vm: the vm object
    :param params: dict, dict include setup vm disk xml configurations
    :param is_setup: one parameter indicate whether setup or clean up
    """

    vmxml = vm_xml.VMXML.new_from_dumpxml(vm.name)
    logging.debug("original xml is: %s", vmxml)

    if not utils_package.package_install(["ceph-common"]):
        raise exceptions.TestError("Failed to install ceph-common")

    cfg = _Config(params)

    if is_setup:
        cfg.define_auth_config_and_log()
        _remove_image(cfg)
        _create_image(cfg, vm.name)
        _update_vm(cfg, vm.name)
    else:
        _remove_image(cfg)
        _cleanup_files(cfg)
        _cleanup_secret(cfg)
```

Although I have to admit that obviously after the refactor the variable `cfg` is used quite often but I think this disadvantage is outweighed by the advantage of better readability. Also, IDEs usually offer auto-completion, so by typing "cfg." it should actually list all available parameters which could then speed up typing. Actually, doing this, I identified a variable definition further below in the code (`attach_option`) that according to the tp-libvirt code guidelines [1] should be at the very top always. This also helped avoid the use of 'locals' in the clean up code (`auth_sec_uuid`).

I don't have ceph storage available to test if I didn't break anything. I opened a PR to your change in case you want to try it out https://github.com/chunfuwen/avocado-vt/pull/1 - The structure changed quite a lot so maybe you would want to read the new version without diff - https://github.com/chunfuwen/avocado-vt/blob/d7207bd07cb92e9cd691a7dae96e98c8790b16ad/virttest/utils_libvirt/libvirt_ceph_utils.py

I see that you have copy-pasted some test results. I don't know those tests. I had expected this code would be tested by using it in two tp-libvirt tests: one with auth, the other without (because of the `is_auth_case` switch that's being used several times in the code). Do you have a link to some tp-libvirt PR that covers this?

Finally, I would like to suggest you add "Author: Chunfu Wen <chwun@redhat.com>" after the Copyright line.

[1] https://github.com/autotest/tp-libvirt/blob/master/tp-libvirt_review_comment_summary.rst

dzhengfy:
@chunfuwen Is there any tp-libvirt PR depending on this one? If yes, could you help add it for reference. If no, could you help provide a Polarion case ID which uses this PR's function? This information will be helpful for reviewers to get more test coverage requirement. Thanks.

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: > @chloerh @chunfuwen I agree it should be non-blocking but I would prefer not using xmltreefile. I believe the intention of VMXML is to use vmxml directly not the xmltreefile instance, I deduce this because the VMXML implements the `__str__` method which is used for string representation of an object:
> https://github.com/avocado-framework/avocado-vt/blob/master/virttest/libvirt_xml/base.py#L54
> https://docs.python.org/3/library/stdtypes.html#textseq
> And, as pointed out by Haijiao, it's also shorter.

@smitterl , although I has different opinion with you, I will update this.

chunfuwen:
> @chunfuwen Is there any tp-libvirt PR depending on this one? If yes, could you help add it for reference. If no, could you help provide a Polarion case ID which uses this PR's function? This information will be helpful for reviewers to get more test coverage requirement. This is also a concern from @smitterl Thanks.

@smitterl 
@dzhengfy 
#https://github.com/autotest/tp-libvirt/pull/3384
This PR depends on it

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,153 @@
...
+                                   debug=True)
+            disk_auth_dict = {"auth_user": ceph_auth_user,
+                              "secret_type": auth_sec_usage_type,
+                              "secret_uuid": auth_sec_uuid}
+            device_source = "rbd:%s:mon_host=%s:keyring=%s" % (ceph_disk_name,
+                                                               ceph_mon_ip,
+                                                               key_file)
+        else:
+            device_source = "rbd:%s:mon_host=%s" % (ceph_disk_name, ceph_mon_ip)
+        logging.debug("device source is: %s", device_source)
```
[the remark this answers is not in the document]
chunfuwen: > @chunfuwen `device_source` is only used for debug message but nowhere else in the code. Is that intentional?

@smitterl 
Yes, it is just for debugging

chunfuwen:
@smitterl 

Thank you for your reviewing.
I can figure out general trend in your comments that we wrap piece of code into small method as much as we can, and avoid big code in one method.
Generally, I agree, But when we separate method, we may need consider more factors, not just readability , reuseability , how much efforts we need maintain that, etc.
I can do some refactor as per your suggestions.

virttest/utils_libvirt/libvirt_ceph_utils.py
dzhengfy: After discussion, please help create new function in virttest/utils_libvirt/libvirt_disk.py to create the function for creating a disk.

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: updated

virttest/utils_libvirt/libvirt_ceph_utils.py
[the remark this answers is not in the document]
chunfuwen: now, I remove them

chunfuwen:
new test result:
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.luks.positive_test.coldplug.device_disk.ceph_test.enable_auth.auth_in_source.encryption_in_source

cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.luks.positive_test.coldplug.device_disk.ceph_test.disable_auth.encryption_in_source


cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.blockcommit.normal_test.network_disk.ceph.relative_path.pause_domain.shallow.top_active

chunfuwen:
#https://github.com/avocado-framework/avocado-vt/pull/3007
create_disk part is moved to libvirt_disk.py as one API

chunfuwen:
@smitterl 

I refactor code as per your comments, please help take a look.

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,173 @@
...
+        disk_src_dict = {"attrs": {"protocol": "rbd",
+                                   "name": ceph_disk_name},
+                         "hosts":  [{"name": ceph_mon_ip,
+                                     "port": ceph_host_port}]}
+        # Create network disk
+        disk_xml = libvirt_disk.create_primitive_disk("network", device, device_target, device_bus,
+                                                      device_format, disk_src_dict, disk_auth_dict)
+        if not keep_raw_image_as:
+            if hotplug:
+                attach_option = params.get("virt_device_attach_option", "--live")
```
smitterl, line 154: Please, move all variable definition to the top of `create_or_cleanup_ceph_backend_vm_disk`.

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,173 @@
...
+                vmxml.sync()
+    else:
+        ceph.rbd_image_rm(ceph_mon_ip, ceph_disk_name.split('/')[0],
+                          ceph_disk_name.split('/')[1], keyfile=rbd_key_file)
+        # Remove ceph config file if created.
+        if os.path.exists(ceph_cfg):
+            os.remove(ceph_cfg)
+        # Remove ceph key file if has.
+        if os.path.exists(key_file):
+            os.remove(key_file)
```
smitterl, line 169: I'd like to suggest optionally you avoid code duplication, you're running `if ..exists: ..remove` twice here.

virttest/utils_libvirt/libvirt_ceph_utils.py
```diff
@@ -0,0 +1,173 @@
...
+                attach_option = params.get("virt_device_attach_option", "--live")
+                virsh.attach_device(vm.name, disk_xml.xml,
+                                    flagstr=attach_option, ignore_status=False, debug=True)
+            else:
+                vmxml = vm_xml.VMXML.new_from_dumpxml(vm.name)
+                vmxml.add_device(disk_xml)
+                vmxml.sync()
+    else:
+        ceph.rbd_image_rm(ceph_mon_ip, ceph_disk_name.split('/')[0],
+                          ceph_disk_name.split('/')[1], keyfile=rbd_key_file)
```
smitterl, line 163: I'd like to suggest optionally you split `ceph_disk_name` once, this way it's easier to understand the code and spot issues.
```python
    names = ceph_disk_name.split('/')
    pool_name = names[0]
    image_name = names[1]
    ceph.rbd_image_rm(ceph_mon_ip, pool_name, image_name, rbd_key_file)
```

smitterl (changes requested):
Hi @chunfuwen Thanks for this. Please, can you move the definition of `attach_option` to the top of `create_or_cleanup_ceph_backend_vm_disk`. IIUC, this is a requirement on tp-libvirt side.
Rest LGTM. Thank you.

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: updated

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: updated

virttest/utils_libvirt/libvirt_ceph_utils.py
chunfuwen: Do you means combine two files deletion together? I have updated code as below
```
        # Remove ceph config and key file if created.
        for file_path in [ceph_cfg, key_file]:
            if os.path.exists(file_path):
                os.remove(file_path)
```

smitterl (approved):

Yingshun (approved):

dzhengfy (approved):

chloerh (approved):

lento-sun (approved):
LGTM

dzhengfy:
All comments from @lento-sun  have been updated and solved. So I think it is ok to merge it now although we have not got @lento-sun 's approval.

PR #2909, opened 2021-01-30
Make the python3 RPM the default subpackage and python2 optional

pevogam:
Python 2 is no longer included in the CI and slowly falling out of favor,
let's make it optional and make Python 3 the RPM package of choice.

### DIFF
--- a/avocado-plugins-vt.spec
+++ b/avocado-plugins-vt.spec
@@ -21,9 +21,9 @@
 %endif
 
 %if 0%{?rhel}
-    %global with_python3 0
+    %global with_python2 1
 %else
-    %global with_python3 1
+    %global with_python2 0
 %endif
 
 # The Python dependencies are already tracked by the python2
@@ -38,7 +38,7 @@
 Summary: Avocado Virt Test Plugin
 Name: avocado-plugins-vt
 Version: 85.0
-Release: 1%{?gitrel}%{?dist}
+Release: 2%{?gitrel}%{?dist}
 License: GPLv2
 Group: Development/Tools
 URL: http://avocado-framework.readthedocs.org/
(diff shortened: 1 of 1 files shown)


pevogam:
@luckyh, @sathnaga: What is your opinion about this? Is it natural or you think it will cause someone some breakage?

luckyh:
> @luckyh, @sathnaga: What is your opinion about this? Is it natural or you think it will cause someone some breakage?

That makes sense to me, since I suppose all upstream users have already move to Python 3 now.

pevogam:
@PaulYuuu, @YongxueHong, @chloerh, @chunfuwen, what do you think about this change?

clebergnu (changes requested):
Hi @pevogam , I like the idea, and these changes LGTM.

Could you, please, only "rebase" this patch so that the SPEC file changelog is kept in chronological and version order?  We're now at 85.0, so this would bring the version down.

Thanks!

pevogam:
> Hi @pevogam , I like the idea, and these changes LGTM.
> 
> Could you, please, only "rebase" this patch so that the SPEC file changelog is kept in chronological and version order? We're now at 85.0, so this would bring the version down.
> 
> Thanks!

Hi @clebergnu pushed with updated changelog.

The PR was originally pushed from the previous Avocado VT version (83.0) which is why it had conflicts now.

sathnaga (approved):
Makes perfect sense to me, even if any issues we should look to fix it as we encounter.

YongxueHong (approved):
It makes sense to me, LGTM.

pevogam:
ping @clebergnu

clebergnu (approved):
LGTM, thanks!

PR #2910, opened 2021-02-01
[virttest/migration] Set default value to avoid TypeError

smitterl:
`migrate_options` might not have been set. Return empty string to avoid
tests failing with
"TypeError: argument of type 'NoneType' is not iterable"

### DIFF
--- a/virttest/migration.py
+++ b/virttest/migration.py
@@ -56,7 +56,7 @@ class MigrationTest(object):
                                           "state at destination" % vm_state)
             logging.info("Guest state is '%s' at destination is as expected",
                          vm_state)
-            if "offline" not in params.get("migrate_options"):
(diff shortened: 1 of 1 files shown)


smitterl:
After checking tp-qemu/qemu and tp-libvirt/libvirt only found usage of these options as string:
https://github.com/autotest/tp-libvirt/blob/b7c5530acb845d87fd77f5a62844764af1aef8bc/libvirt/tests/cfg/virsh_cmd/domain/virsh_migration.cfg
https://github.com/autotest/tp-libvirt/blob/b7c5530acb845d87fd77f5a62844764af1aef8bc/libvirt/tests/cfg/virsh_cmd/domain/virsh_migrate_stress.cfg

Yingshun (approved):

dzhengfy (approved):
LGTM

PR #2914, opened 2021-02-03
Tmp fix of snapshot job failure

lento-sun:


lento-sun:
This is only a tmp fix and will be added to ci within pr-extend part for snapshot task. no need to review.

PR #2922, opened 2021-02-23
[RFC] Support the use of avocado tags

luckyh:
Avocado allows tests to be given tags, which can be used to create
test categories. As VT users, we would like to benefit from this
feature as well.

Instead of using docstring directives, expressing tags by a param
is a more natural way for VT. By introducing the `tags` param,
test writers can use it to apply test tags to VT tests,
for example:

```
- boot:
    virt_test_type = qemu libvirt
    type = boot
    tags = vm_lifecycle:start fast
```

then users can list or run only the tagged tests via the
`--filter-by-tags` option:

```
$ avocado list --filter-by-tags=fast
```

### DIFF
--- a/avocado_vt/loader.py
+++ b/avocado_vt/loader.py
@@ -79,6 +79,34 @@ def arch_listing(config, guest_name_parser=None):
     LOG.debug("")
 
 
+def get_vt_test_tags(vt_params):
+    """
+    Get the test categories based on a `tags` VT parameter.
+
+    :param vt_params: currently parsed VT parameters
+    :type vt_params: dict
+    :returns: test categories based on a `tags` vt param
+    :type: dict
+    """
+    result = {}
+    # XXX: Using whitespace as the delimiter is not a standard
+    #      implementation of Avocado's tags facility, this can
+    #      function properly until whitespace being considered
+    #      as a valid character for the tag identifier.
+    for tag in vt_params.get('tags', '').split():
+        if not tag:
+            continue
+        if ':' in tag:
+            key, val = tag.split(':', 1)
+            if key in result:
+                result[key].add(val)
+            else:
+                result[key] = set([val])
+        else:
+            result[tag] = set()
+    return result
+
+
 class NotAvocadoVTTest(object):
 
     """
@@ -217,6 +245,9 @@ class VirtTestLoader(loader.TestLoader):
             params['id'] = test_name
             test_parameters = {'name': test_name,
                                'vt_params': params}
+            tags = get_vt_test_tags(params)
+            if tags:
+                test_parameters['tags'] = tags
             test_suite.append((VirtTest, test_parameters))
         if which_tests is loader.DiscoverMode.ALL and not test_suite:
             return self._report_bad_discovery(url, "No matching tests",


luckyh:
FYI:
https://avocado-framework.readthedocs.io/en/85.0/guides/user/chapters/tags.html
https://avocado-framework.readthedocs.io/en/85.0/guides/writer/chapters/writing.html#categorizing-tests

pevogam:
Hi @luckyh, what is the difference with `avocado list only=fast` for an example `fast` variant which is our main way of restricting and filtering tests? Clearly the `tags` parameter offers some extra freedom but I am afraid it will make the configuration messier and introduce at least some redundancies.

luckyh:
> Hi @luckyh, what is the difference with `avocado list only=fast` for an example `fast` variant which is our main way of restricting and filtering tests? Clearly the `tags` parameter offers some extra freedom but I am afraid it will make the configuration messier and introduce at least some redundancies.

@pevogam ok, let me explain further. From my perspective 'tags' is metadata of tests, while the variants of cartesian configuration is not, it's a part of test name. Although you can put metadata in variants, it would look very odd: 
```
- boot:
    variants:
        - fast:
```
it's hard to know that if the 'fast' represents one of the sub-tests of 'boot' or it is just a label without test instance attached to, is that right?

Even if we can let the variant name to be `[fast] boot` (assume that we've modified the cartesian module, and now it is a valid name), we still need additional features (i.e. filtering with wildcards or a new option to specify tags) to support this thing, and it would just make the syntax more complicated that I'm not intent to do so.

So, introducing a new interface would be a good choice then.  In order to create the 1:1 mapping between tags and VT tests, 'tags' is implemented as a param, which is different with what Avocado defined for its instrumented tests. And also, there would be some advantages, for instance: users can set private tags to tests for their own use cases.
```
# subtests.cfg
- boot:
    tags = fast

# subtests-private.cfg
boot:
    tags += ,<private_tag>

$ avocado list --filter-by-tags=<private_tag>
```

luckyh:
Overall, the proposal is aimed at giving users the flexibility of test case management.

pevogam:
@luckyh The experimental idea is great and it doesn't seem to harm the currently existing interface. My only concern is about redundancies and people avoiding the core ideas of the Cartesian configuration for lazier tag management that includes what would otherwise be variants as well but I guess it is worth giving it a try. and see where it gets us.

pevogam:
> is not, it's a part of test name. Although you can put metadata in variants, it would look very odd:
> 
> ```
> - boot:
>     variants:
>         - fast:
> ```
> 
> it's hard to know that if the 'fast' represents one of the sub-tests of 'boot' or it is just a label without test instance attached to, is that right?

Not sure what you mean by "test instance attached to" - neither a tag nor a variant is associated in any 1:1 way to any test instances.  A variant could result in one or a hundred tests depending only on the final products.

> Even if we can let the variant name to be `[fast] boot` (assume that we've modified the cartesian module, and now it is a valid name), we still need additional features (i.e. filtering with wildcards or a new option to specify tags) to support this thing, and it would just make the syntax more complicated that I'm not intent to do so.

Not sure why we make the assumption of having to drop the tag from the variant name at all. My original point was that lots of features from the tags are made redundant by the variant definitions so here it is better to point out what benefits could tags bring to the table and variants not. I guess what you mean here is that tags are easily dropped from the test name (something we could do with @ operator on the variants) but I am not sure what is the advantage from dropping them on the first place.

> So, introducing a new interface would be a good choice then. In order to create the 1:1 mapping between tags and VT tests, 'tags' is implemented as a param, which is different with what Avocado defined for its instrumented tests.

Is such 1:1 mapping even possible? I mean if I define
```
- boot:
    tags=a,b,c
    variants:
        - grub:
            version = 1
        - grub2:
            version = 2
```
then the tags will in fact be mapped to at least two tests. The test script in Avocado VT is not associated with a single test and I don't see why we should try to achieve form of 1:1 mapping even if we could.

luckyh:
@pevogam 
> Not sure what you mean by "test instance attached to" - neither a tag nor a variant is associated in any 1:1 way to any test instances. A variant could result in one or a hundred tests depending only on the final products.

Sorry I just had no idea how to say that in english, and yes there is no 1:1 mapping between a tag/variant and a test which I hold the same opinion. Anyway, what I really want to point out is, implementing 'tags' by using variants is not a good idea, because 'tags' should be an attribute of a test as metadata rather than a part of test name.

> Not sure why we make the assumption of having to drop the tag from the variant name at all. My original point was that lots of features from the tags are made redundant by the variant definitions so here it is better to point out what benefits could tags bring to the table and variants not. I guess what you mean here is that tags are easily dropped from the test name (something we could do with @ operator on the variants) but I am not sure what is the advantage from dropping them on the first place.

The critical is how to understand the so called "redundant". If you mean by the use of filtering tests, either 'tags' or 'variants' can be used for that usage since both are available for filtering, however it is different (at least from my perspective). Supposing that there is an empty test database, it will be filled with tests by giving a cartesian configuration, for example:
```
variants:
    - a1:
    - a2:
        variants:
            - b1:
            - b2:
```
the test db would look like:
```
+ test db -+
| a1       |
| a2.b1    |
| a2.b2    |
+----------+
```
Then if you "filter" the tests with `no b2`, it would look like:
```
+ test db -+
| a1       |
| a2.b1    |
+----------+
a2.b2
```
and this means you intend to exclude the certain test(s) from the db. So why? because **'variants' is used for constructing tests**, and all the generate tests would be filled into the test db expect those who are discarded.

While, as I said before, 'tags' is an attribute, to help describe a test and allows it to be found again by browsing or searching. By modified the above configuration as:
```
variants:
    - a1:
        tags = foo
    - a2:
        variants:
            - b1:
                tags = foo
            - b2:
```
now the db would look like:
```
+ test db --------+
| a1       (#foo) |
| a2.b1    (#foo) |
| a2.b2           |
+-----------------+
```
And then if you "filter" the tests with `tags=foo`, it would look like:
```
+ test db --------+
|+ test set -----+|
||a1       (#foo)||
||a2.b1    (#foo)||
|+---------------+|
| a2.b2           |
+-----------------+
```
this means you intend to create a test set (we have to explicitly point out this concept to distinguish with test db) by selecting tests with the given tags.

The above two examples showed the different ways for filtering tests. Although sometimes they could give the same outcome, I couldn't say they are "redundant", because they just provide two different ways of test case management: one for the constructing phase, the other for the use phase, and users can realize which one meets their demands better.
(Btw, the relationship between these two is kind of like the margin and padding's in CSS.)

> Is such 1:1 mapping even possible? I mean if I define
> 
> ```
> - boot:
>     tags=a,b,c
>     variants:
>         - grub:
>             version = 1
>         - grub2:
>             version = 2
> ```
> 
> then the tags will in fact be mapped to at least two tests. The test script in Avocado VT is not associated with a single test and I don't see why we should try to achieve form of 1:1 mapping even if we could.

The 1:1 mapping can be achieved by like this:
```
boot.grub:
   tags=foo
boot.grub2:
   tags=bar
```
Anyway, the idea is to provide fine grained control to users. 

And also, there is an apparent defeat by implementing tags in VT test script: mutually exclusive tags would associate with one test script, which is impossible. And that's why I didn't choose this approach.
```
# wrong:
# reboot.py
def run(...):
    """
    :tags: fast,slow
    """
    for i in range(repeats):
         vm.reboot()

# correct:
# reboot.cfg
- reboot:
    type = reboot
    variants:
        - 1_time:
            tags = fast
            repeats = 1
        - 1000_times:
            tags = slow
            repeats = 1000
```

pevogam:
> @pevogam
> 
> > Not sure what you mean by "test instance attached to" - neither a tag nor a variant is associated in any 1:1 way to any test instances. A variant could result in one or a hundred tests depending only on the final products.
> 
> Sorry I just had no idea how to say that in english, and yes there is no 1:1 mapping between a tag/variant and a test which I hold the same opinion. Anyway, what I really want to point out is, implementing 'tags' by using variants is not a good idea, because 'tags' should be an attribute of a test as metadata rather than a part of test name.

Alright, this makes sense. However, the test name is also an attribute of a test as metadata and instead of `tags=tag1,tag2` one could simply say that tags could be easily extracted as `name=tag1.tag2` with variants that are not tags marked as `@not_a_tag_variant`.

> The critical is how to understand the so called "redundant". If you mean by the use of filtering tests, either 'tags' or 'variants' can be used for that usage since both are available for filtering, however it is different (at least from my perspective). Supposing that there is an empty test database, it will be filled with tests by giving a cartesian configuration, for example:
> 
> ```
> variants:
>     - a1:
>     - a2:
>         variants:
>             - b1:
>             - b2:
> ```
> 
> the test db would look like:
> 
> ```
> + test db -+
> | a1       |
> | a2.b1    |
> | a2.b2    |
> +----------+
> ```
> 
> Then if you "filter" the tests with `no b2`, it would look like:
> 
> ```
> + test db -+
> | a1       |
> | a2.b1    |
> +----------+
> a2.b2
> ```
> 
> and this means you intend to exclude the certain test(s) from the db. So why? because **'variants' is used for constructing tests**, and all the generated tests would be filled into the test db expect those who are discarded.
> 
> While, as I said before, 'tags' is an attribute, to help describe a test and allows it to be found again by browsing or searching. By modified the above configuration as:
> 
> ```
> variants:
>     - a1:
>         tags = foo
>     - a2:
>         variants:
>             - b1:
>                 tags = foo
>             - b2:
> ```
> 
> now the db would look like:
> 
> ```
> + test db --------+
> | a1       (#foo) |
> | a2.b1    (#foo) |
> | a2.b2           |
> +-----------------+
> ```
> 
> And then if you "filter" the tests with `tags=foo`, it would look like:
> 
> ```
> + test db --------+
> |+ test set -----+|
> ||a1       (#foo)||
> ||a2.b1    (#foo)||
> |+---------------+|
> | a2.b2           |
> +-----------------+
> ```
> 
> this means you intend to create a test set (we have to explicitly point out this concept to distinguish with test db) by selecting tests with the given tags.

The term "test db" is more complex and new to me and so far many of us have used "test set" to mean exactly the restriction one obtains when constructing tests to be run. These examples are good descriptions of the potential redundancy of how we could obtain identical test sets (to run) from both approaches, having difference only in stages where this is done, i.e. construction vs later stage. Do you perhaps have an example of a case where we could create a certain behavior with tags but not with variants?

> The above two examples showed the different ways for filtering tests. Although sometimes they could give the same outcome, I couldn't say they are "redundant", because they just provide two different ways of test case management: one for the constructing phase, the other for the use phase, and users can realize which one meets their demands better.

I am not sure different phases (i.e. different implementation and something most uses won't differentiate) is significant difference here, especially since the goal of filtering tests is always the same - running a smaller set of tests. So letting the users realize which method meets their demands better is rather saying "let's implement something to deprecate the variants definitions that will be easier for users to use and give them the choice to prove us right". There is nothing wrong with the latter of course but then we could simply be explicit about it as part of the motivation here. An example of a filtering functionality that is not given from the test variants would be more helpful motivation to show why test tags are not necessarily a replacement but a complementary functionality to test variants.

One more thing regarding the difference in stages: in fact variants could be more memory efficient way to deal with test restrictions since the Cartesian parser will only construct the filtered tests while using test tags will rely on a potential explosion of Cartesian products and filter on top of it.

> (Btw, the relationship between these two is kind of like the margin and padding's in CSS.)

Hehe, not quite sure how to compare it with thinking inside the box and outside of the box ^^

> > Is such 1:1 mapping even possible? I mean if I define
> > ```
> > - boot:
> >     tags=a,b,c
> >     variants:
> >         - grub:
> >             version = 1
> >         - grub2:
> >             version = 2
> > ```
> > 
> > 
> > then the tags will in fact be mapped to at least two tests. The test script in Avocado VT is not associated with a single test and I don't see why we should try to achieve form of 1:1 mapping even if we could.
> 
> The 1:1 mapping can be achieved by like this:
> 
> ```
> boot.grub:
>    tags=foo
> boot.grub2:
>    tags=bar
> ```
> 
> Anyway, the idea is to provide fine grained control to users.
> 
> And also, there is an apparent defeat by implementing tags in VT test script: mutually exclusive tags would associate with one test script, which is impossible. And that's why I didn't choose this approach.
> 
> ```
> # wrong:
> # reboot.py
> def run(...):
>     """
>     :tags: fast,slow
>     """
>     for i in range(repeats):
>          vm.reboot()
> 
> # correct:
> # reboot.cfg
> - reboot:
>     type = reboot
>     variants:
>         - 1_time:
>             tags = fast
>             repeats = 1
>         - 1000_times:
>             tags = slow
>             repeats = 1000
> ```

Yes, so to clarify my terminology here - test script really means the test python script here and a mapping of test scripts and variants is never 1:1 considering we could use `type = reboot` rather arbitrarily there. So the tests in VT are always related to the final produced variants and we do not worry about "filtering test scripts" or anything like this. So your choice not to use tags in test scripts makes complete sense.

luckyh:
> > Sorry I just had no idea how to say that in english, and yes there is no 1:1 mapping between a tag/variant and a test which I hold the same opinion. Anyway, what I really want to point out is, implementing 'tags' by using variants is not a good idea, because 'tags' should be an attribute of a test as metadata rather than a part of test name.
> 
> Alright, this makes sense. However, the test name is also an attribute of a test as metadata and instead of `tags=tag1,tag2` one could simply say that tags could be easily extracted as `name=tag1.tag2` with variants that are not tags marked as `@not_a_tag_variant`.

I think that 'name' and 'tags' are mutually exclusive in the logic model of test, that means they should not contain each other. But one can just combine the two things as a new concept, saying 'title' or something else. Then the key point is what the right definition of the final produced variants is - 'name' or 'title', I'd vote for the former.

> > this means you intend to create a test set (we have to explicitly point out this concept to distinguish with test db) by selecting tests with the given tags.
> 
> The term "test db" is more complex and new to me and so far many of us have used "test set" to mean exactly the restriction one obtains when constructing tests to be run. These examples are good descriptions of the potential redundancy of how we could obtain identical test sets (to run) from both approaches, having difference only in stages where this is done, i.e. construction vs later stage. Do you perhaps have an example of a case where we could create a certain behavior with tags but not with variants?

For example if I want to build a smoke test, with the tags interface now I can achieve it by assigning the `smoke` tag to the target tests and executing the job with the cmd `avocado run --filter-by-tags=smoke`, or else I have to use a list file to record all the name of the target tests and pass them to the test runner later (the current solution).

Honestly, this can also be done by using variants, but just not that well because it looks pretty odd.

> One more thing regarding the difference in stages: in fact variants could be more memory efficient way to deal with test restrictions since the Cartesian parser will only construct the filtered tests while using test tags will rely on a potential explosion of Cartesian products and filter on top of it.

Yes, exactly, that's why we should prevent the misuse of variants, as it is so powerful.

avocado_vt/loader.py
```diff
@@ -79,6 +79,28 @@ def arch_listing(config, guest_name_parser=None):
     LOG.debug("")
 
 
+def get_vt_test_tags(vt_params):
+    """
+    Returns the test categories based on a `tags` vt param
+
+    :rtype: dict
```
pevogam, line 86: This could benefit from a complete docstring like:
```
Get the test categories based on a `tags` VT parameter.

:param vt_params: currently parsed VT parameters
:type vt_params: dict
:returns: test categories based on a `tags` vt param
:type: dict
```

avocado_vt/loader.py
```diff
@@ -79,6 +79,28 @@ def arch_listing(config, guest_name_parser=None):
...
+    """
+    tags = vt_params.get('tags', '').strip()
+    result = {}
+    for tag in tags.split(','):
```
pevogam, line 90: How about defining the tags in a more Cartesian config friendly manner through intervals? E.g.
```
tags = A B C
```
with `vt_params.objects('tags')` instead?

avocado_vt/loader.py
```diff
@@ -79,6 +79,28 @@ def arch_listing(config, guest_name_parser=None):
...
+        if not tag:
+            continue
+        if ':' in tag:
```
pevogam, line 93: Will we have the syntax and meaning of colon here documented somewhere in the project docs?

avocado_vt/loader.py
```diff
@@ -79,6 +79,28 @@ def arch_listing(config, guest_name_parser=None):
...
+            key, val = tag.split(':', 1)
+            if key in result:
+                result[key].add(val)
+            else:
+                result[key] = set([val])
+        else:
+            result[tag] = None
```
pevogam, line 100: How about using an empty set here? It will make the types more uniform.

pevogam (changes requested):
Alright, since it is useful to add this feature at least for an experimental phase, I am leaving some comments to the current PR diff I found.

avocado_vt/loader.py
luckyh: That's the first idea I got, but then I changed to the current one in order to strictly respect the tags syntax defined by Avocado, because VT is only responsible for providing the data of `tags` and fully rely on the tags implementation given by Avocado [1] [2] [3]. 

Btw, by the current tags implementation of Avocado, IMO, it should provide a function to plugins for parsing tags strings:
```python
from avocado.core.tags import parse_tags

def get_vt_test_tags(vt_params):
    tags = vt_params.get('tags', '').strip()
    return parse_tags(tags)
```
otherwise each plugin have to implement the same thing again and again. Or else, Avocado should allow plugins to redefine the tags syntax and give the implementation of the tags interpreter (include both parser and resolver).

Now back to the original topic, using `' '` as the delimiter is something not logic proper but harmless in the current stage, since `' '` is not a valid character for the tag identifier [4], so it's fine to adopt your suggestion. What do you think?

References:
[1] https://github.com/avocado-framework/avocado/blob/master/avocado/core/tags.py
[2] https://github.com/avocado-framework/avocado/blob/4a6641ab0b7c3f0bb1c568d2929d3955ceda3aba/avocado/core/suite.py#L93-L122
[3] https://github.com/avocado-framework/avocado/blob/4a6641ab0b7c3f0bb1c568d2929d3955ceda3aba/avocado/core/safeloader.py#L268-L290
[4] https://github.com/avocado-framework/avocado/blob/4a6641ab0b7c3f0bb1c568d2929d3955ceda3aba/avocado/core/safeloader.py#L232

avocado_vt/loader.py
luckyh: It's given by [this doc](https://avocado-framework.readthedocs.io/en/85.0/guides/user/chapters/tags.html#using-further-categorization-with-keys-and-values), do you think we also need to highlight it in VT?

avocado_vt/loader.py
luckyh: Yes, makes sense, will update in the next version, thanks.

avocado_vt/loader.py
luckyh: Ok, thanks for your suggestion, will update in the next version.

avocado_vt/loader.py
pevogam: I think that it is best if we use VT specific syntax (`tags = A B C`) which is also more compatible and readable together with our Cartesian configuration, then satisfy the Avocado requirements right before passing the tags further to it. I think in this way we get the best from both worlds.

avocado_vt/loader.py
pevogam: Oh, then it is fine. I forgot this is well documented on the Avocado side, VT should be fine since it follows the same rules :+1:

avocado_vt/loader.py
luckyh: ok, I'll update it per your suggestion and leave a comment to explain the issue I mentioned. Btw, we can't use `objects()`, it complaints as:
```
$ avocado list --filter-by-tags=fast
Test discovery plugin <avocado_vt.loader.VirtTestLoader object at 0x7f8a4676f490> failed: 'dict' object has no attribute 'objects'
```
will replace with `get().split()`, thanks.

luckyh:
@pevogam updated this patch per your suggestions, thanks.

avocado_vt/loader.py
pevogam: Yes good point, I guess it doesn't receive our widely used `Params` object here. How is `vt_params` parsed? Aren't the tags obtained from the Cartesian config?

avocado_vt/loader.py
```diff
@@ -79,6 +79,34 @@ def arch_listing(config, guest_name_parser=None):
...
+    :param vt_params: currently parsed VT parameters
+    :type vt_params: dict
+    :returns: test categories based on a `tags` vt param
+    :type: dict
+    """
+    result = {}
+    # XXX: Using whitespace as the delimiter is not a standard
+    #      implementation of Avocado's tags facility, this can
+    #      function properly until whitespace being considered
+    #      as a valid character for the tag identifier.
```
pevogam, line 95: Trying to understand your added comment here: what do you mean by "this can function properly until whitespace being considered as a valid character for the tag identifier"? This sounds a bit like a TODO and I don't think the plan is for Avocado to add support for them. Unless you are sure it is their limitation in which case opening a trackable issue might be better. There are some inherent differences between VT and pure Avocado so the standard usage of spaces for lists in VT shouldn't necessarily bother us too much and be changed to avocado-style or vice versa.

luckyh: In fact, the tags resolver [1] does support whitespace characters. For example, just by changing the delimiter to comma and assigning the `super fast` tag to the boot test, users can filter the test correctly.

```
$ avocado list --filter-by-tags='super fast'
VT io-github-autotest-qemu.boot
```

That's because the limitation is only applied to the parser in Avocado [2]. I cannot ensure Avocado would not revise the tags syntax forever, that's why I have the concern. And I said, Avocado should provide something like `tags.parse_tags` by the current implementation, if that's true then I have no chance to pick the delimiter.

Also, the truth is there is no built-in list type support in Cartesian config, "values are strings" [3]. One only can do serialization/deserialization to make the conversion between strings and lists, and the implementation of `str.split()` is too simple to use for deserialization.

[1] https://github.com/avocado-framework/avocado/blob/4a6641ab0b7c3f0bb1c568d2929d3955ceda3aba/avocado/core/tags.py#L19-L43
[2] https://github.com/avocado-framework/avocado/blob/4a6641ab0b7c3f0bb1c568d2929d3955ceda3aba/avocado/core/safeloader.py#L232
[3] https://avocado-vt.readthedocs.io/en/latest/CartesianConfig.html#keys-and-values

avocado_vt/loader.py
luckyh: The `Params` wrapper is provided by `virttest.utils_params`, `loader.py` doesn't import that therefore it can't take the benefit.

pevogam: Yes, I was surprised to see that Cartesian parser's `get_dicts()` doesn't in fact return a `Params` object, something I would find much more reasonable considering we use the finally parsed test parameters as its ubiquitous instance. :eyes:

avocado_vt/loader.py
pevogam: > In fact, the tags resolver [1] does support whitespace characters. For example, just by changing the delimiter to comma and assigning the `super fast` tag to the boot test, users can filter the test correctly.
> 
> ```
> $ avocado list --filter-by-tags='super fast'
> VT io-github-autotest-qemu.boot
> ```
> 
> That's because the limitation is only applied to the parser in Avocado [2].

So perhaps it is better in the above comment to mention specifically "the limitation is only applied to the parser in Avocado" instead of talking generally about the "Avocado's tags facility". 

> And I said, Avocado should provide something like `tags.parse_tags` by the current implementation, if that's true then I have no chance to pick the delimiter.

It would still be great to track somewhere where you could express your comments for the long term, e.g. an Avocado issue. At least regarding this part.

> Also, the truth is there is no built-in list type support in Cartesian config, "values are strings" [3]. One only can do serialization/deserialization to make the conversion between strings and lists, and the implementation of str.split() is too simple to be used for deserialization.

I guess this depends on the interpretation. I consider the test parameters obtained via `Params` and in particular the `get_list` [function](https://github.com/avocado-framework/avocado-vt/blob/6142776eac1ea69f0b2461759fecf4899f91fb1f/virttest/utils_params.py#L139) to already achieve this. I have even called it with actual remote serialization using Pyro4, no manual parsing/deserialization of such a potential list was needed along the way.

luckyh: > So perhaps it is better in the above comment to mention specifically "the limitation is only applied to the parser in Avocado" instead of talking generally about the "Avocado's tags facility".

That is something unclarified by Avocado, I just wrote it based on my understanding, anyway I will update the comment per your suggestion if you think it is necessary.

> It would still be great to track somewhere where you could express your comments for the long term, e.g. an Avocado issue. At least regarding this part.

Yes, makes sense, let's track it by https://github.com/avocado-framework/avocado/issues/4452, thanks.

pevogam (approved):
Hi @luckyh, LGTM.

PR #2923, opened 2021-02-23
get_additional_feature_list: Update to get complete features

Yingshun:
The method get_additional_feature_list() got the features except
invtsc. But some cases need it for now, so add a parameter to
set whether to get it or not.

### DIFF
--- a/virttest/libvirt_xml/domcapability_xml.py
+++ b/virttest/libvirt_xml/domcapability_xml.py
@@ -35,7 +35,8 @@ class DomCapabilityXML(base.LibvirtXMLBase):
         result = self.__dict_get__('virsh').domcapabilities()
         self['xml'] = result.stdout_text.strip()
 
-    def get_additional_feature_list(self, cpu_mode_name):
+    def get_additional_feature_list(self, cpu_mode_name,
+                                    ignore_features=("invtsc",)):
         """
         Get additional CPU features which explicitly specified by <feature>
         tag in cpu/mode[@name='host-model'] part of virsh domcapabilities.
@@ -65,6 +66,7 @@ class DomCapabilityXML(base.LibvirtXMLBase):
         </mode>
 
         :param cpu_mode_name: cpu mode name, must be 'host-model' since libvirt3.9
+        :param ignore_features: features that need to be ignored
         :return: list of features, feature is dict-like, feature name is set to dict key,
                  feature policy is set to dict value.
                  returen is like [{'ss': 'require'}, {'pdpe1gb', 'require'}]
@@ -78,9 +80,9 @@ class DomCapabilityXML(base.LibvirtXMLBase):
                     for feature in mode_node.findall('feature'):
                         item = {}
                         item[feature.get('name')] = feature.get('policy')
-                        # Feature invtsc doesn't support migration, so not add it to feature list
-                        if 'invtsc' not in item:
-                            feature_list.append(item)
+                        if ignore_features and item in ignore_features:
+                            continue
+                        feature_list.append(item)
         except AttributeError as elem_attr:
             logging.warn("Failed to find attribute %s" % elem_attr)
             feature_list = []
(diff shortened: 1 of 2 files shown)


Yingshun:
**Before fix:**
` (1/1) type_specific.io-github-autotest-libvirt.virsh.hypervisor_cpu_compare.cpu_xml.f_domcapa_xml: FAIL: Not find expect key word 'identical' in command output 'The CPU provided by hypervisor on the host is a superset of CPU described in /tmp/avocado_g_etoq_j/cpu.xml' (9.57 s)
`

**After fix:**
```
JOB LOG    : /root/avocado/job-results/job-2021-02-23T03.40-b341618/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.hypervisor_cpu_compare.cpu_xml.f_domcapa_xml: PASS (9.13 s)
JOB TIME   : 9.94 s

```

Yingshun:
@smitterl  Would you please help review this PR? Thanks!

Yingshun:
@dzhengfy @chloerh could you please help review it?

virttest/libvirt_xml/domcapability_xml.py
```diff
@@ -78,9 +79,11 @@ def get_additional_feature_list(self, cpu_mode_name):
                     for feature in mode_node.findall('feature'):
                         item = {}
                         item[feature.get('name')] = feature.get('policy')
-                        # Feature invtsc doesn't support migration, so not add it to feature list
-                        if 'invtsc' not in item:
-                            feature_list.append(item)
+                        if ignore_invtsc and 'invtsc' in item:
```
chloerh, line 82: Is there any chance we ignore other features too? If there is, are we considering adding an unsupported list instead of ignoring feature one by one ?

Yingshun: Thank you @chloerh for reviewing it. I just found the behavior of this case has been changed, not sure what the problem was.
So I will convert it soon for further investigation.

Yingshun: Weird, the problem is gone on the latest version. Anyway, updated.

chloerh (approved):

virttest/libvirt_xml/domcapability_xml.py
```diff
@@ -35,7 +35,8 @@ def __init__(self, virsh_instance=base.virsh):
         result = self.__dict_get__('virsh').domcapabilities()
         self['xml'] = result.stdout_text.strip()
 
-    def get_additional_feature_list(self, cpu_mode_name):
+    def get_additional_feature_list(self, cpu_mode_name,
+                                    ignore_features=("invtsc",)):
```
smitterl, line 39: @Yingshun As tuples are immutable, I'd rather have a list here because I expect it to be easier to construct a list of `ignore_features` than a tuple. But I might be wrong.

smitterl (approved):
Thanks @Yingshun I confirmed the `from_domcapabilities` was only used in the virsh.hypervisor_cpu_compare..f_domcapa_xml. I re-ran with your patch applied just in case
```bash
# avocado run --vt-type libvirt --vt-machine-type s390-virtio --vt-connect-uri qemu:///system virsh.hypervisor_cpu_compare..f_domcapa_xml
JOB ID     : 33d225b1f45d2fef1699e696bcdf69235754a78e
JOB LOG    : /root/avocado/job-results/job-2021-03-23T12.53-33d225b/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.hypervisor_cpu_compare.cpu_xml.f_domcapa_xml: PASS (5.18 s)
JOB TIME   : 5.71 s
```

virttest/libvirt_xml/domcapability_xml.py
Yingshun: @smitterl  Hmm, that's the reason why I use a tuple type. I need to set an argument that includes 'invtsc' as a default value for consistency, in other words, I can't set it to None as a default value. In this case, using a list is not recommended, the value of default value would be changed if someone wrote 'append' function.

Example:
```
>>> def mytest(mylist=["invtsc"]):
...  mylist.append("ds")
...  print(mylist)
... 
>>> mytest()
['invtsc', 'ds']
>>> mytest()
['invtsc', 'ds', 'ds']
>>> mytest()
['invtsc', 'ds', 'ds', 'ds']
>>> 

```

PR #2927, opened 2021-02-24
libvirt_pcicontr: Get free pci slot

dzhengfy:
This is to provide an utility function to get a free slot for a given bus.
This function is useful to select an usable slot when attaching a pci
device to avoid of slot conflict.

### DIFF
--- a/virttest/utils_libvirt/libvirt_pcicontr.py
+++ b/virttest/utils_libvirt/libvirt_pcicontr.py
@@ -28,3 +28,26 @@ def get_max_contr_indexes(vm_xml, cntlr_type, cntlr_model, cntl_num=1):
                                                          cntlr_model,
                                                          usable_indexes[:cntl_num]))
     return usable_indexes[:cntl_num]
+
+
+def get_free_pci_slot(vm_xml, max_slot=31):
+    """
+    Get a free slot on pcie-root controller
+
+    :param vm_xml The guest xml
+    :param max_slot: the maximum of slot to be selected
+
+    :return: str,the first free slot or None
+    """
+    used_slot = []
+    pci_devices = vmxml.xmltreefile.find('devices').getchildren()
+    for dev in pci_devices:
+        address = dev.find('address')
+        if (address is not None and address.get('bus') == '0x00'):
+            used_slot.append(address.get('slot'))
+    logging.debug("Collect used slot:%s", used_slot)
+    for slot_index in range(1, max_slot + 1):
+        slot = "%0#4x" % slot_index
+        if slot not in used_slot:
+            return slot
+    return None


virttest/utils_libvirt/libvirt_pcicontr.py
```diff
@@ -28,3 +28,27 @@ def get_max_contr_indexes(vm_xml, cntlr_type, cntlr_model, cntl_num=1):
...
+    :return: str,the first free slot or None
+    """
+    used_slot = []
+    pci_devices = vmxml.xmltreefile.find('devices').getchildren()
+    for dev in pci_devices:
+        address = dev.find('address')
+        if (address is not None and address.get('bus') == target_bus):
+            used_slot.append(address.get('slot'))
+    logging.debug("Collect used slot:%s", used_slot)
+    for slot_index in range(1, max_slot):
```
chloerh, line 50: Just one small thing: the last item of range(1, 30) is 29, is that what we want?

virttest/utils_libvirt/libvirt_pcicontr.py
```diff
@@ -28,3 +28,27 @@ def get_max_contr_indexes(vm_xml, cntlr_type, cntlr_model, cntl_num=1):
                                                          cntlr_model,
                                                          usable_indexes[:cntl_num]))
     return usable_indexes[:cntl_num]
+
+
+def get_free_pci_slot(vm_xml, target_bus, max_slot=30):
```
chunfuwen, line 33: Do you need controller  type and model as filter conditions?

virttest/utils_libvirt/libvirt_pcicontr.py
dzhengfy: @chloerh , 30 is only a default value. The user needs to change it according to his needs.

virttest/utils_libvirt/libvirt_pcicontr.py
dzhengfy: @chunfuwen currently, it seems this function only applies to the controller 'pcie-root' because pcie-root-port which is common only can accept 1 pci device. Only pcie-root can has 31 slots available.  I will update the docstring to make it clear.

virttest/utils_libvirt/libvirt_pcicontr.py
chloerh: I understand, it's just..I'm worrying the number could cause confusions, like when someone tries to get before 10, but only get the 9.
Anyway, just my opinion.

dzhengfy: Updated the max_slot for pcie-root controller.

chloerh (approved):

chunfuwen (approved):
LGTM

PR #2933, opened 2021-02-25
v2v: protect original guest name to be miss-changed when it has mutip…

xiaodwan:
…le blanks

The guest name may has mutiple blanks. During tidy up the finnal v2v command
line, those blanks in original guests were removed by mistake.
This patch fixes the problem.

### DIFF
--- a/virttest/utils_v2v.py
+++ b/virttest/utils_v2v.py
@@ -1304,8 +1304,17 @@ def v2v_cmd(params, auto_clean=True, cmd_only=False, interaction=False):
             for i in range(int(params.get('_disk_count', 0))):
                 options += ' -oo rhv-disk-uuid=%s' % str(uuid.uuid4())
 
+        # Protect the blanks in original guest name
+        safe_vm_name = ''
+        BLANK_REPLACEMENT = '%20'
+        if ' ' in vm_name and vm_name in options:
+            safe_vm_name = vm_name.replace(' ', BLANK_REPLACEMENT)
+            options = options.replace(vm_name, safe_vm_name)
         # Construct a final virt-v2v command and remove redundant blanks
         cmd = ' '.join(('%s %s' % (V2V_EXEC, options)).split())
+        # Restore the real original guest name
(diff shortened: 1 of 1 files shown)


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.function_test_esx.positive_test.local_storage.esx_70.libvirt.it_vddk.vpx_uri: PASS (689.03 s)

xiaodwan:
@chloerh @yantian0 Please help to review this PR, thanks.

chunfuwen (approved):
LGTM

yantian0 (approved):

virttest/utils_v2v.py
```diff
@@ -1304,8 +1304,16 @@ def _v2v_post_cmd():
...
+        # Protect the blanks in original guest name
+        safe_vm_name = ''
+        if ' ' in vm_name and vm_name in options:
+            safe_vm_name = vm_name.replace(' ', '%20')
+            options = options.replace(vm_name, safe_vm_name)
         # Construct a final virt-v2v command and remove redundant blanks
         cmd = ' '.join(('%s %s' % (V2V_EXEC, options)).split())
+        # Restore the real original guest name
+        if safe_vm_name:
+            cmd = cmd.replace(safe_vm_name, vm_name)
```
chloerh, line 1316: Oh, I see this replacing is a temporary move. Then I would recommend an simpler string which could be understood better than '%20' which seems unnecessary since we're not trying to use it in a url. Just my opinion

xiaodwan: I don't think it will bring readability problem here, as you said it just a temp replacement, users could understand it by the context and code comments, and also '%20' is a famous replacement in URL, we could borrow the idea from it.

xiaodwan: Anyway, I did a bit change about this.

chloerh: It's quite understandable.

chloerh (approved):

PR #2939, opened 2021-03-02
utils_conn: Extend timeout for qemu-kvm.pp installation

Yingshun:
It takes more time to install and uninstall qemu-kvm.pp on some
machines.

### DIFF
--- a/virttest/utils_conn.py
+++ b/virttest/utils_conn.py
@@ -2117,19 +2117,20 @@ class UNIXSocketConnection(ConnectionBase):
         for port_to_del in [self.desturi_port, self.migrateuri_port, self.disks_uri_port]:
             firewall_cmd.remove_port(port_to_del, 'tcp')
 
-    def install_qemu_kvm_pp(self, session, qemu_kvm_pp_path):
+    def install_qemu_kvm_pp(self, session, qemu_kvm_pp_path, timeout=240):
         """
         Install qemu-kvm module by semodule
 
         :param session: Session object
         :param qemu_kvm_pp_path: Path of qemu-kvm.pp file
+        :param timeout: Timeout to execute command lines
         """
         cmd = "semodule -l|grep qemu-kvm"
         status, output = session.cmd_status_output(cmd)
         if status:
             logging.debug("Active qemu-kvm policy.")
             cmd = "semodule -i %s" % qemu_kvm_pp_path
-            status, output = session.cmd_status_output(cmd)
+            status, output = session.cmd_status_output(cmd, timeout=timeout)
             if status:
                 logging.error("Unable to active SELinux policy module - "
                               "qemu-kvm! cmd: {} output: {}"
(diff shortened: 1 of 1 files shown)


Yingshun:
**Before fix:**
```
JOB ID     : 35eec8ac790032ee2e2b4c94a8f37d985511cb87
JOB LOG    : /root/avocado/job-results/job-2021-03-01T19.56-35eec8a/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_over_unix.positive_testing.without_copy_storage.migrate_uri.p2p_live_migration.with_postcopy: ERROR: Timeout expired while waiting for shell command to complete: 'semodule -i /var/tmp/libvirt-ci/runtest/avocado-vt/avocado-vt/virttest/shared/scripts/pmsocat/qemu-kvm.pp'    (output: '') (198.94 s)
```


**After fix:**
```
JOB ID     : 1e3b87e714ab8e36d636a702be1a507ea93886af
JOB LOG    : /root/avocado/job-results/job-2021-03-01T21.07-1e3b87e/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virsh.migrate_over_unix.positive_testing.without_copy_storage.migrate_uri.p2p_live_migration.with_postcopy: PASS (566.90 s)
JOB TIME   : 569.10 s

```

virttest/utils_conn.py
```diff
@@ -2117,37 +2117,39 @@ def del_firewall_ports(self, session):
...
         :param qemu_kvm_pp_path: Path of qemu-kvm.pp file
+        :param timeout: Timeout to execute command lines
         """
         cmd = "semodule -l|grep qemu-kvm"
         status, output = session.cmd_status_output(cmd)
         if status:
             logging.debug("Active qemu-kvm policy.")
             cmd = "semodule -i %s" % qemu_kvm_pp_path
-            status, output = session.cmd_status_output(cmd)
+            status, output = session.cmd_status_output(cmd, timeout=timeout)
```
chunfuwen, line 2133: a little wonder two items:
1)How can ensure for any session, its session.cmd_status_output() support timeout parameter considering session (as one input parameter)may vary
2)generally, if without specific timeout ,it should be never timeout. So by specifying 240 timeout , it sound like it may not fix the issue.

Yingshun: > a little wonder two items:
> 1)How can ensure for any session, its session.cmd_status_output() support timeout parameter considering session (as one input parameter)may vary
In this case, session is aexpect.client.RemoteSession, so cmd_status_output supports timeout. 
In other codes, sorry, I have no idea, I think you have to check the code...
> 2)generally, if without specific timeout ,it should be never timeout. So by specifying 240 timeout , it sound like it may not fix the issue.
No, the default value of timeout here is 60s, and shell command line needs more than 60s on some machines. And after extending it to 240s, the problem is fixed, please check the test result above.

chloerh (approved):

Yingshun:
@chunfuwen @dzhengfy could you please help review this PR?

virttest/utils_conn.py
chunfuwen: For 1) Do we need considering session may not be aexpect.client.RemoteSession?
try:
  `status, output = session.cmd_status_output(cmd, timeout=timeout)
except:
   status, output = session.cmd_status_output(cmd,)

`

Yingshun: No, I don't think so. 
The class, UNIXSocketConnection, is to prepare unix socket environment on a local and remote host. So the session here always support time parameter. Hence we shouldn't use try-except.

dzhengfy: @chunfuwen , In my POV, this is only a simple fix to enable specific timeout value for installing/uninstalling. And this is normal usage for an aexpect session. So it is ok to take it easy, I think.

dzhengfy (approved):

chunfuwen (approved):
LGTM

PR #2940, opened 2021-03-02
utils_net: Install dhcp-client before creating/deleting the ovs bridge

Yingshun:
The package "dhcp-client" was not installed on some hosts, so the
network connection would be interrupted.


dzhengfy (approved):
LGTM

dzhengfy:
This is only to add a new package to install. So I merged it.

PR #2943, opened 2021-03-03
Future Test class independence on Job class in avocado

richtja:
In the avocado 86.0 will be removed the test -> job dependency. This
commit will ensure the compatibility of avocado-vt with avocado>=86

Reference:https://github.com/avocado-framework/avocado/pull/4429

### DIFF
--- a/avocado_vt/test.py
+++ b/avocado_vt/test.py
@@ -84,17 +84,20 @@ class VirtTest(test.Test):
 
     env_version = utils_env.get_env_version()
 
-    def __init__(self, methodName='runTest', name=None, params=None,
-                 base_logdir=None, job=None, runner_queue=None,
-                 vt_params=None):
+    def __init__(self, **kwargs):
         """
-        :note: methodName, name, base_logdir, job and runner_queue params
-               are inherited from test.Test
+        :note: methodName, name, base_logdir, job/config and runner_queue
+               params are inherited from test.Test
+               From the avocado 86 the test.Test uses config instead of job
+               instance. Because of the compatibility with avocado 82.0 LTS we
+               can't remove the job instance. For avocado < 86 job instance is
+               used and for avocado=>86 config is used.
         :param params: avocado/multiplexer params stored as
                        `self.avocado_params`.
         :param vt_params: avocado-vt/cartesian_config params stored as
                           `self.params`.
         """
+        vt_params = kwargs.pop("vt_params", None)
         self.__params_vt = None
         self.__avocado_params = None
         self.bindir = data_dir.get_root_dir()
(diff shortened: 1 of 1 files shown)


clebergnu (approved):
LGTM, extended testing with the proposed change in the Avocado side can be seen here:

https://cirrus-ci.com/build/6148588276285440

More specifically in jobs such as:

https://cirrus-ci.com/task/5809965270040576
https://cirrus-ci.com/task/5247015316619264

richtja:
Hi @pevogam, this is the part of work on avcado-vt compatibility with nrunner. It's connected with  [avocado #4429](https://github.com/avocado-framework/avocado/pull/4429). I am letting you know as you wanted to be part of this effort.

beraldoleal (approved):
LGTM

avocado_vt/test.py
```diff
@@ -84,17 +84,16 @@ class VirtTest(test.Test):
...
         """
-        :note: methodName, name, base_logdir, job and runner_queue params
+        :note: methodName, name, base_logdir, job/config and runner_queue params
                are inherited from test.Test
         :param params: avocado/multiplexer params stored as
                        `self.avocado_params`.
         :param vt_params: avocado-vt/cartesian_config params stored as
                           `self.params`.
         """
+        vt_params = kwargs.pop("vt_params", None)
```
pevogam, line 96: What is the planned difference between `vt_params` and `params_vt`? I think the presence of both could be confusing for people and perhaps it is best to provide a single instance of the `Params` class for all usage within VT.

pevogam (commented):
Hi @richtja, thanks for including me here. As this PR is touching code job/test API it actually had breaking changes for our plugin and environment so it was good to take a look at it while simultaneously doing a review and fixing our own calls.

I left a few comments for you here, tests seem to be passing now so unless I write with more comments our tests can be considered green.

willianrampazzo (approved):

avocado_vt/test.py
```diff
@@ -84,17 +84,16 @@ class VirtTest(test.Test):
 
     env_version = utils_env.get_env_version()
 
-    def __init__(self, methodName='runTest', name=None, params=None,
-                 base_logdir=None, job=None, runner_queue=None,
-                 vt_params=None):
+    def __init__(self, **kwargs):
         """
-        :note: methodName, name, base_logdir, job and runner_queue params
+        :note: methodName, name, base_logdir, job/config and runner_queue params
```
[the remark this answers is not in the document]
richtja: This PR is about to make avocado-vt compatible with new version of avocado where Test class uses config, but avocado-vt still have to be compatible with  avocado LTS version where Test class uses job object. I understand that this might be confusing, so I will write better docstring with explanation.

avocado_vt/test.py
richtja: I just follow the naming from the master branch. I agree with you that this can be change, but I think it is not related to this PR. And because I am planing to use this as an example, for other extensions to keep compatibility with Avocado 86.0 and later, to the community, so I would like to make it as simple as possible.

avocado_vt/test.py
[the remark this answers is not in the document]
pevogam: Oh I see. Yes, this is a good idea.

avocado_vt/test.py
pevogam: Yes no worries, I was suspecting that and I would do the same. I was rather asking if you know more about these and if my comment makes sense to you for a future PR. In any case it is something to consider.

pevogam (approved):

Issue #2912, opened 2021-02-01
Avocado VT settings broken again


pevogam:
It seems that some of the settings of Avocado are once again not applicable:
```
# avocado config | grep tmp
    plugins.vtjoblock.dir                    /tmp
    run.keep_tmp                             False
    vt.common.tmp_dir                        /tmp
```
However, code like this in `utils_net`:
```python
ADDRESS_POOL_FILENAME = os.path.join(data_dir.get_tmp_dir(), "address_pool")
ADDRESS_POOL_LOCK_FILENAME = ADDRESS_POOL_FILENAME + ".lock"
```
accesses `data_dir.get_tmp_dir()`  which performs
```python
def get_tmp_dir(public=True):
    """
    Get the most appropriate tmp dir location.

    :param public: If public for all users' access
    """
    persistent_dir = get_settings_value('vt.common', 'tmp_dir',
                                        default="")
    if persistent_dir != "":
        return persistent_dir
```
where `persistent_dir` ends up empty. If I try to print the settings in this location:
```
[...]

PR #2944, opened 2021-03-04
Compatibility with config files in early stages

richtja:
In early stage of avocado run, there is a time when the vt options are
created, but they are not synchronized with values from config files.
This can affect some of the global variables which are set during this
time. This commit ensures the synchronization right after the options
are registered.

Reference: #2912

### DIFF
--- a/avocado_vt/plugins/vt_init.py
+++ b/avocado_vt/plugins/vt_init.py
@@ -270,6 +270,8 @@ if hasattr(plugin_interfaces, 'Init'):
                                      help_msg=help_msg,
                                      default='/tmp')
 
+            settings.merge_with_configs()
+
             virt_loader = getattr(importlib.import_module('avocado_vt.loader'),
                                   'VirtTestLoader')
             loader.register_plugin(virt_loader)
--- a/selftests/unit/test_vt_init.py
+++ b/selftests/unit/test_vt_init.py
@@ -1,0 +1,43 @@
+import importlib
+import os
+import tempfile
+import unittest
+
+from selftests import BASEDIR
+
+
+class VtInitTest(unittest.TestCase):
+
+    @staticmethod
+    def _swap():
+        data_dir = os.path.join(BASEDIR, "selftests", ".data")
+        config_file = os.path.join(BASEDIR, "avocado_vt", "conf.d", "vt.conf")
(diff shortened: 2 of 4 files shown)


richtja:
@pevogam please let me know if it helps.

pevogam:
Sure, will run a full test now.

pevogam:
Yes, the tests pass now :green_apple:

avocado_vt/plugins/vt_init.py
```diff
@@ -270,6 +270,8 @@ def initialize(self):
                                      help_msg=help_msg,
                                      default='/tmp')
 
+            settings.merge_with_configs()
```
pevogam, line 273: Do you think we can add some sort of unit tests to catch this issue in the future? So far this behavior was broken and fixed twice, once in September 2020 and once here. Perhaps it is a rarely tested path and might benefit from a simple unit test to alarm other breaking future merges.

richtja: Sure, I will figure it out and force-pushed here.

selftests/unit/test_vt_init.py
```diff
@@ -0,0 +1,38 @@
...
+        config_test_file = os.path.join(data_dir, "vt_test.conf")
+        tmp, tmp_name = tempfile.mkstemp(suffix=".conf", dir=data_dir)
+        os.rename(config_test_file, tmp_name)
+        os.rename(config_file, config_test_file)
+        os.rename(tmp_name, config_file)
+
+    def setUp(self):
+        self._swap()
+
+    def test_something(self):
```
pevogam, line 24: Perhaps a better name would be `test_early_settings` or something of the like.

pevogam: A minimal docstring might also be useful.

pevogam (commented):
Thanks for adding the unit test @richtja

selftests/unit/test_vt_init.py
richtja: Oh, of course, sorry I totally forget to rename that :D

pevogam (approved):
Alright, I won't terrorize you any further ^^ LGTM

willianrampazzo (approved):

beraldoleal (approved):
Thanks for this @richtja. Much appreciated. LGTM.

PR #2952, opened 2021-03-09
Fix driverMetadata is not defined error

chunfuwen:
The methold __eq__ override super class with DriverMetadata,which cause
vm_xml.del_device() throw not define error

### DIFF
--- a/virttest/libvirt_xml/devices/disk.py
+++ b/virttest/libvirt_xml/devices/disk.py
@@ -635,9 +635,6 @@ class Disk(base.TypedDeviceBase):
                 setattr(new_one, key, value)
             return new_one
 
-        def __eq__(self, other):
-            return super(DriverMetadata, obj).__eq__(other)
-
         class MetadataCache(base.base.LibvirtXMLBase):
             """
             Source of MetadataCache xml class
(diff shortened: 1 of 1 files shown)


chunfuwen:
>>> from virttest.libvirt_xml import vm_xml
>>> vmxml = vm_xml.VMXML.new_from_inactive_dumpxml('avocado-vt-vm1')
>>> disks = vmxml.get_devices(device_type="disk")
>>> print(disks[0])
<disk device="disk" type="file">
      <driver name="qemu" type="qcow2" />
      <source file="/home/storage/RHEL-8.4-x86_64-latest.qcow2" />
      <target bus="virtio" dev="vda" />
      <address bus="0x04" domain="0x0000" function="0x0" slot="0x00" type="pci" />
    </disk>
>>> vmxml.del_device(disks[0])

Yingshun:
It works well, some migration cases are passed with it.
But I'm not sure if we can ignore 2 new alerts, @dzhengfy @kylazhang would you please help check it?

chunfuwen:
> It works well, some migration cases are passed with it.
> But I'm not sure if we can ignore 2 new alerts, @dzhengfy @kylazhang would you please help check it?

new alerts are from  LGTM.com checking. Why I add override __eq__ last time is to fix alerts from LGTM.com, so without __eq__ or with __eq__  will both have alerts.  This may involves in complex situation ,where current-existed classes are not well defined as LGTM.com rule.

chloerh (approved):
Let's just ignore this kind of alerts from LGTM, it happens to every new xml class we try to add

Yingshun:
> Let's just ignore this kind of alerts from LGTM, it happens to every new xml class we try to add

OK, I see. LGTM.

Yingshun (approved):

dzhengfy (approved):
LGTM

PR #2953, opened 2021-03-09
Add global param for schedinfo

dzhengfy:
This is only to add the support for global_period and global_quota for
schedinfo command.

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -2735,12 +2735,15 @@ class VMCPUTuneXML(base.LibvirtXMLBase):
         emulator_quota:       int
         iothread_period:      int
         iothread_quota:       int
+        global_period:        int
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/vm_xml.py
```diff
@@ -2733,14 +2733,17 @@ class VMCPUTuneXML(base.LibvirtXMLBase):
         quota:                int
         emulator_period:      int
         emulator_quota:       int
+        global_period:        int
```
chunfuwen, line 2736: the parameter in docstring doesn't match with ones in __slots__

chloerh (approved):

Yingshun (approved):
LGTM

virttest/libvirt_xml/vm_xml.py
dzhengfy: @chunfuwen updated.

PR #2955, opened 2021-03-11
Fix 'memory_unit' missing from slots of VMXMLBase

chloerh:
'memory_unit' accessor was defined, but not added to slots,
this commit will fix it.


Yingshun (approved):

dzhengfy (approved):
It seems this is missed unexpectedly. LGTM.

dzhengfy:
@chloerh, we'd better add a comment saying there is no logic code change, so it is no need to provide test result. Thanks.

PR #2956, opened 2021-03-12
Unification of vt parameters naming

richtja:
Specific parameters for avcoad-vt are named as `vt_params` but inside
`VirtTest` those parameters are named as `params_vt`. This might be
confusing because it's still the same data. This PR unifies the
naming.


clebergnu (approved):
LGTM, thanks!

willianrampazzo (approved):

PR #2957, opened 2021-03-17
Add a function to check if the feature is supported

Yingshun:
Automation of a new feature always needs comparing libvirt version
to check if it's supported in the installed libvirt version, so
implement a common function in libvirt_misc.

### DIFF
--- a/virttest/libvirt_version.py
+++ b/virttest/libvirt_version.py
@@ -5,6 +5,7 @@ Shared code for tests that need to get the libvirt version
 import re
 import logging
 
+from avocado.core import exceptions
 from avocado.utils import path
 from avocado.utils import process
 from avocado.utils.astring import to_text
@@ -64,3 +65,38 @@ def version_compare(major, minor, update, session=None):
     elif LIBVIRT_LIB_VERSION >= compare_version:
         return True
     return False
+
+
+def is_libvirt_feature_supported(params, ignore_error=False):
+    """
+    Check whether the function is supported in this libvirt version by comparing
+    the installed libvirt version
+
+    :param params: Dictionary with the test parameters
+    :param ignore_error: Whether to raise an exception
+    :raise: When ignore_error is set False, raise TestCancel if the feature is
+        not supported
+    :return: True if the feature is supported;
+        False if the feature is not supported when ignore_error is set to True
+
+    Example:
+    ::
+    params={'func_supported_since_libvirt_ver':'(6,8,0)'}
+
+    NOTE: The value of 'func_supported_since_libvirt_ver' is a string of
+        libvirt's (major, minor, update) version.
+    """
+    func_supported_since_libvirt_ver = eval(
+        params.get("func_supported_since_libvirt_ver", '()'))
+    unspported_err_msg = params.get("unspported_err_msg",
+                                    "This libvirt version doesn't support "
+                                    "this function.")
+
+    if func_supported_since_libvirt_ver:
+        if not version_compare(*func_supported_since_libvirt_ver):
+            if ignore_error:
+                logging.error(unspported_err_msg)
+                return False
+            else:
+                raise exceptions.TestCancel(unspported_err_msg)
+    return True


Yingshun:
**Test results:**
```
>>> params={'func_supported_since_libvirt_ver':'(6,8,0)'}
>>> libvirt_misc.is_libvirt_feature_supported(params)
True
>>> libvirt_misc.is_libvirt_feature_supported(params, True)
True
>>> params={'func_supported_since_libvirt_ver':'(8,8,0)'}
>>> libvirt_misc.is_libvirt_feature_supported(params)
  File "/var/tmp/libvirt-ci/runtest/avocado-vt/avocado-vt/virttest/utils_libvirt/libvirt_misc.py", line 33, in is_libvirt_feature_supported
    raise exceptions.TestCancel(unspported_err_msg)
avocado.core.exceptions.TestCancel: This libvirt version doesn't support this function.
>>> libvirt_misc.is_libvirt_feature_supported(params, True)
ERROR:root:This libvirt version doesn't support this function.
False

```

Yingshun:
@dzhengfy @chloerh  Could you please help review this PR? Thanks!

virttest/utils_libvirt/libvirt_misc.py
```diff
@@ -0,0 +1,36 @@
...
+
+    :param params: Dictionary with the test parameters
+    :param ignore_error: Whether to raise an exception
+    :raise: When ignore_error is set False, raise TestCancel if the feature is
+        not supported
+    :return: True if the feature is supported;
+        False if the feature is not supported when ignore_error is set to True
+    """
+    func_supported_since_libvirt_ver = eval(
+        params.get("func_supported_since_libvirt_ver", '()'))
```
dzhengfy, line 24: Could you provide an example value for 'func_supported_since_libvirt_ver' in docstring for reference?

Yingshun: Updated.

virttest/utils_libvirt/libvirt_misc.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def is_libvirt_feature_supported(params, ignore_error=False):
+    """
+    Check whether the function is supported in this libvirt version by comparing
+    the installed libvirt version
+
+    :param params: Dictionary with the test parameters
+    :param ignore_error: Whether to raise an exception
+    :raise: When ignore_error is set False, raise TestCancel if the feature is
```
chunfuwen, line 18: I assume 'TestCancel" means that execption since TestCancel apply tp-libvirt project scope only

virttest/utils_libvirt/libvirt_misc.py
```diff
@@ -0,0 +1,43 @@
...
+libvirt related misc functions
+"""
+import logging
+
+from avocado.core import exceptions
+
+from virttest import libvirt_version
+
+
+def is_libvirt_feature_supported(params, ignore_error=False):
```
chunfuwen, line 11: it only use one parameter :func_supported_since_libvirt_ver, do we really need pass in whole dict?

virttest/utils_libvirt/libvirt_misc.py
```diff
@@ -0,0 +1,43 @@
...
+    Example:
+    ::
+    params={'func_supported_since_libvirt_ver':'(6,8,0)'}
+
+    NOTE: The value of 'func_supported_since_libvirt_ver' is a string of
+        libvirt's (major, minor, update) version.
+    """
+    func_supported_since_libvirt_ver = eval(
+        params.get("func_supported_since_libvirt_ver", '()'))
+    unspported_err_msg = params.get("unspported_err_msg",
```
chunfuwen, line 32: The error msg is too generalized, better output specific feature

chunfuwen:
if one python name with  "misc" word, that means that people will put any categorized or not categorized contents into this file.
Better have specific keyword in python file.

virttest/utils_libvirt/libvirt_misc.py
Yingshun: You can configure it as you want by passing unspported_err_msg.

Yingshun:
> if one python name with "misc" word, that means that people will put any categorized or not categorized contents into this file.
> Better have specific keyword in python file.

Do you have any suggestions?

virttest/utils_libvirt/libvirt_misc.py
Yingshun: I'm using a dict type argument because:
1. the values can be set in cfg
2. easy to extend the arguments

chunfuwen:
> > if one python name with "misc" word, that means that people will put any categorized or not categorized contents into this file.
> > Better have specific keyword in python file.
> 
> Do you have any suggestions?

How about libvirt_sys_utils.py?

Moreover, I am a little wonder whether we need this kind of utils, since in most cases, we only need two lines code to handle same logic. This looks like more straightforward. On contrast , if we use this ,firstly we need import this module, then call this function, there are at least two lines.
*********************************************
if not libvirt_version.version_compare(5, 3, 0):
    test.cancel(xx)

Yingshun:
> > > if one python name with "misc" word, that means that people will put any categorized or not categorized contents into this file.
> > > Better have specific keyword in python file.
> > 
> > 
> > Do you have any suggestions?
> 
> How about libvirt_sys_utils.py?
is it good to use 'sys' and 'utils' under utils_libvirt? @dzhengfy would you please share your option? 
 
> 
> Moreover, I am a little wonder whether we need this kind of utils, since in most cases, we only need two lines code to handle same logic. This looks like more straightforward. On contrast , if we use this ,firstly we need import this module, then call this function, there are at least two lines.
> 
> if not libvirt_version.version_compare(5, 3, 0):
> test.cancel(xx)

I think we'd better to config the value of the supported libvirt version in cfg file instead of hard-coding it in py script. Furthermore, in the case of having test cases that supported from different version, we don't need to add lines in the py file one by one.

dzhengfy (commented):
Move this function to virttest/libvirt_version.py because they are closely related.

Yingshun:
> Move this function to virttest/libvirt_version.py because they are closely related.

Done.

virttest/libvirt_version.py
```diff
@@ -64,3 +65,38 @@ def version_compare(major, minor, update, session=None):
...
+                                    "this function.")
+
+    if func_supported_since_libvirt_ver:
+        if not version_compare(*func_supported_since_libvirt_ver):
+            if ignore_error:
+                logging.error(unspported_err_msg)
+                return False
+            else:
+                raise exceptions.TestCancel(unspported_err_msg)
+        return True
```
dzhengfy, line 102: When func_supported_since_libvirt_ver is (), should we always return True ?

dzhengfy (commented):
Others look good to me

virttest/libvirt_version.py
Yingshun: Updated.

dzhengfy (approved):
LGTM now.

chunfuwen (approved):
lgtm

chloerh (approved):

PR #2960, opened 2021-03-17
Add function to get iface info inside vm

chloerh:
Get inside-vm info of certain interface with given mac address
Add a new module libvirt_interface

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -16,6 +16,7 @@ import platform
 import uuid
 import hashlib
 import shutil
+import json
 
 import aexpect
 from aexpect import remote
@@ -3703,6 +3704,34 @@ def get_linux_ifname(session, mac_address=""):
     # If we came empty handed, let's raise an error
     raise exceptions.TestError("Failed to determine interface name with "
                                "mac %s" % mac_address)
+
+
+def get_linux_iface_info(mac, session=None):
+    """
+    Get host/guest info of certain interface with given mac address.
+
+    :param mac: string type, mac address of given interface
+    :param session: session of given vm, default to None
+    :return: dict-type info of interface, None if not get any
+    """
+    ip_cmd = 'ip -json a'
+
+    try:
+        if session:
+            ip_output_str = session.cmd_output(ip_cmd).strip()
+        else:
+            ip_output_str = process.run(ip_cmd).stdout_text.strip()
+        ip_info = json.loads(ip_output_str)
+        logging.debug('interfaces inside vm:\n %s', ip_info)
+    except Exception as why:
+        logging.error('Failed to get interfaces inside vm. Reason: %s',
+                      str(why))
+        return None
+
+    for iface in ip_info:
+        if iface.get('address') == mac:
+            return iface
+    return None
 
 
 def update_mac_ip_address(vm, timeout=240):


chloerh:
Test result 
https://github.com/autotest/tp-libvirt/pull/3379#issuecomment-803703517

pevogam:
Hi @chloerh, is the plan to add utility for any possible operation through a vm session? There is a myriad of possible commands, many of which distribution and utility specific, so I wonder how meaningful this is.

chloerh:
> Hi @chloerh, is the plan to add utility for any possible operation through a vm session? There is a myriad of possible commands, many of which distribution and utility specific, so I wonder how meaningful this is.

We're not adding utility for any possible operation though a vm session. About this pr, I was adding a new test case which requires checking iface info inside vm: https://github.com/autotest/tp-libvirt/pull/3379#issuecomment-803703517  , and I though this kind of info could be used by other future test case, therefore I chose to add the function to `utils_libvirt`

Yingshun (commented):
It supports session, so I wonder if we need to consider getting iface on local/remote host.
And we have a similar function virttest.utils_net.get_linux_ifname, does it meet your requirement?
Or how about adding it to utils_net? How do you think?

chloerh:
> It supports session, so I wonder if we need to consider getting iface on local/remote host.
> And we have a similar function virttest.utils_net.get_linux_ifname, does it meet your requirement?
> Or how about adding it to utils_net? How do you think?

get_linux_ifname seems not fitting my request, but moving it to utils_net seems a nice idea, I'll try and update code.

kamilvarga (approved):
LGTM.

virttest/utils_net.py
```diff
@@ -4322,3 +4323,30 @@ def delete_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
     return utils_misc.cmd_status_output(tmux_cmd, shell=True, verbose=True,
                                         ignore_status=ignore_status,
                                         session=session)
+
+
+def get_iface_info(mac, session=None):
```
pevogam, line 4328: Considering that this is still similar but a more detailed version of `get_linux_ifname`, could we place it right after `get_linux_ifname` instead of at the end of the module? As this module contains windows methods too we should also give it more comparable naming like `get_linux_iface_info`.

virttest/utils_net.py
```diff
@@ -4322,3 +4323,30 @@ def delete_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
...
+    """
+    Get host/inside-vm info of certain interface with given mac address
```
pevogam, line 4330: This is better written as "Get host/guest info of ..." ending with a full stop.

virttest/utils_net.py
```diff
@@ -4322,3 +4323,30 @@ def delete_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
...
+
+    :param mac: mac address of given interface
+    :param session: session of given vm, default to None
+    :return: dict-type info of interface, None if not get any
+    """
+
```
pevogam, line 4336: Docstring linters detect this and (rightfully) complain that there shouldn't be empty lines after function docstrings.

virttest/utils_net.py
```diff
@@ -4322,3 +4323,30 @@ def delete_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
     return utils_misc.cmd_status_output(tmux_cmd, shell=True, verbose=True,
                                         ignore_status=ignore_status,
                                         session=session)
+
+
+def get_iface_info(mac, session=None):
+    """
+    Get host/inside-vm info of certain interface with given mac address
+
+    :param mac: mac address of given interface
```
pevogam, line 4332: Can you also include `:type mac` definitions here? IS the mac address supposed to be string?

pevogam (changes requested):
There is still some better integration into the existing code to be done.

virttest/utils_net.py
chloerh: @pevogam Thanks for all the good suggestions, all updated.

Yingshun (approved):
LGTM

virttest/utils_net.py
```diff
@@ -3705,6 +3706,32 @@ def _process_output(cmd, reg_pattern):
                                "mac %s" % mac_address)
 
 
+def get_linux_iface_info(mac, session=None):
+    """
+    Get host/guest info of certain interface with given mac address
```
pevogam, line 3711: I final nitpick and I am happy: can you add a full stop here?
```diff
-Get host/guest info of certain interface with given mac address
+Get host/guest info of certain interface with given mac address.
```

pevogam (commented):
Almost done, just one of my previous comments was missed.

virttest/utils_net.py
chloerh: Of course :grin:

pevogam (approved):
Alrighty, all good to go.

virttest/utils_net.py
```diff
@@ -3705,6 +3706,32 @@ def _process_output(cmd, reg_pattern):
...
+
+    try:
+        if session:
+            ip_output_str = session.cmd_output(ip_cmd).strip()
+        else:
+            ip_output_str = process.run(ip_cmd).stdout_text.strip()
+        ip_info = json.loads(ip_output_str)
+        logging.debug('interfaces inside vm:\n %s', ip_info)
+    except Exception:
+        logging.error('Failed to get interfaces inside vm.')
```
dzhengfy, line 3727: Is it better to log the detailed exception message, too?

virttest/utils_net.py
```diff
@@ -3705,6 +3706,32 @@ def _process_output(cmd, reg_pattern):
...
+        return
+
+    for iface in ip_info:
+        if iface.get('address') == mac:
+            return iface
+
```
dzhengfy, line 3733: Should we return None if no mac is matched at the end of this function?

virttest/utils_net.py
```diff
@@ -3705,6 +3706,32 @@ def _process_output(cmd, reg_pattern):
...
+    try:
+        if session:
+            ip_output_str = session.cmd_output(ip_cmd).strip()
+        else:
+            ip_output_str = process.run(ip_cmd).stdout_text.strip()
+        ip_info = json.loads(ip_output_str)
+        logging.debug('interfaces inside vm:\n %s', ip_info)
+    except Exception:
+        logging.error('Failed to get interfaces inside vm.')
+        return
```
dzhengfy, line 3728: None can be omitted?

virttest/utils_net.py
chloerh: Of course, nice catch!

virttest/utils_net.py
chloerh: I think you are right. Though it's the same, but return None is actually more clear.

virttest/utils_net.py
chloerh: Similar reason with last one, I'll update code.

dzhengfy (approved):
LGTM

PR #2967, opened 2021-03-23
Fix: pylint check not working

chloerh:
Version of pylint is too low and causing errors like:
`module 'isort' has no attribute 'SortImports'` which blocks pylint
check for all modules.
This commit should fix it.


dzhengfy (approved):
LGTM

kylazhang (approved):

PR #2973, opened 2021-03-24
Make wait for writing file finish as common func

kylazhang:

```
# avocado run --vt-type libvirt virtual_disks.ceph.hot_plug.without_auth.disk_attach.attach_device
JOB ID     : dd52e9e386c235827b0ff429eadcd546ca4051c0
JOB LOG    : /root/avocado/job-results/job-2021-03-24T03.19-dd52e9e/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virtual_disks.ceph.hot_plug.without_auth.disk_attach.attach_device: PASS (66.51 s)
JOB TIME   : 67.71 s

# avocado run --vt-type libvirt virtual_network.iface_bridge.default.shared_physical_network
JOB ID     : 23af5080a2cdbcf7c849843a4fef0ef6716db79e
JOB LOG    : /root/avocado/job-results/job-2021-03-24T07.50-23af508/job.log
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.iface_bridge.default.shared_physical_network: PASS (133.00 s)
JOB TIME   : 134.59 s

```

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1592,6 +1592,22 @@ def alter_boot_order(vm_name, pci_id, boot_order=0):
     vmxml.sync()
 
 
+def wait_for_file_over(patt, xml_file, timeout=10):
+    """
+    Wait for file write over with pattern
+
+    :param patt: pattern for grep in file to wait, eg '</disk>'
+    :param xml_file: file for writing xml
+    :param timeout: wait_for timeout
+    """
+
+    def file_write_over():
+        if not process.run("grep %s %s" % (patt, xml_file),
+                           ignore_status=True).exit_status:
+            return True
+    utils_misc.wait_for(file_write_over, timeout)
+
+
 def create_disk_xml(params):
     """
     Create a disk configuration file.
@@ -1757,11 +1773,7 @@ def create_disk_xml(params):
     utils_misc.wait_for(file_exists, 5)
 
     # Wait for file write over with '</disk>' keyword at the file end
-    def file_write_over():
-        if not process.run("grep '</disk>' %s" % diskxml.xml,
-                           ignore_status=True).exit_status:
-            return True
-    utils_misc.wait_for(file_write_over, 10)
+    wait_for_file_over('</disk>', diskxml.xml)
     return diskxml.xml
 
 
@@ -2074,6 +2086,7 @@ def create_nwfilter_xml(params):
 
         filterxml.xmltreefile.write()
         logging.info("The network filter xml is:\n%s" % filterxml)
+        wait_for_file_over('</filter>', filterxml.xml)
         return filterxml
 
     except Exception as detail:
@@ -3571,6 +3584,7 @@ def modify_vm_iface(vm_name, oper, iface_dict, index=0, virsh_instance=virsh):
         vmxml.sync()
     elif oper == "get_xml":
         logging.info("iface xml is %s", iface)
+        wait_for_file_over('</interface>', iface.xml)
         return iface.xml
 
 


virttest/utils_test/libvirt.py
```diff
@@ -1592,6 +1592,22 @@ def alter_boot_order(vm_name, pci_id, boot_order=0):
     vmxml.sync()
 
 
+def wait_for_file_over(patt, xml_file, timeout=10):
```
chunfuwen, line 1595: maybe put xml_file as first parameter,then patt as second

kylazhang: I don't see any difference as both are needed without default value...

Yingshun, line 1595: I have no problem with the code changes, I just wonder if we need to move it to utils_misc...I think this function can be used widely if do so. How do you think?

virttest/utils_test/libvirt.py
```diff
@@ -1592,6 +1592,22 @@ def alter_boot_order(vm_name, pci_id, boot_order=0):
...
+    :param patt: pattern for grep in file to wait, eg '</disk>'
+    :param xml_file: file for writing xml
+    :param timeout: wait_for timeout
+    """
+
+    def file_write_over():
+        if not process.run("grep %s %s" % (patt, xml_file),
+                           ignore_status=True).exit_status:
+            return True
+    utils_misc.wait_for(file_write_over, timeout)
```
chloerh, line 1608: Are we considering returning what `utils_misc.wait()` returns?

kylazhang: I think as it just wait for the file write finish, there should be no need to do judgement when calling it.

virttest/utils_test/libvirt.py
kylazhang: As most of the usage is for xml writing finish judgement and most of the usage should be in tp-libvirt and in the same file libvirt.py, then I think put it here should be OK.

virttest/utils_test/libvirt.py
Yingshun, line 1608: We'd better to log an error or raise exception here if there is no pattern in the file. 
for example, it could be :
```
if not utils_misc.wait_for(file_write_over, timeout):
    xxx
```

virttest/utils_test/libvirt.py
Yingshun: Ok.

chunfuwen: functionally it is no difference , but from the  human natural  language understanding point,  first we have whole thing (xml_file),then we apply some rule (patt) on that.
But it just comments, doesn't block PR merge or not merge

kylazhang: I think it should be personality issue. From my point of view, as "re" operation also set the pattern as first parameter so it is comfortable for me. Anyway, thanks for your comments

virttest/utils_test/libvirt.py
kylazhang: If it can not find the pattern, just wait timeout, and maybe raise the same error as before like xxx file can not be find. Then what we should do is to increase the timeout. I don't want to deal with the return error here as it is meaningless and the purpose is just to wait. what do you think?

chloerh: OK, I understand. Thanks for explaining.

chloerh (approved):

virttest/utils_test/libvirt.py
Yingshun: Okay, got it.

Yingshun (approved):

chunfuwen (approved):
lgtm

PR #2975, opened 2021-03-24
 Preserve monitor connections for any preserved vms across test runs

pevogam:
This is an addition to the original implementation of vm persistence
that preserves the vm objects, i.e. PR: Optional vm persistence #2105.

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -489,6 +489,9 @@ class Monitor(object):
         """
         Return True if the monitor is responsive.
         """
+        if self._socket.fileno() < 0:
+            logging.warning("Monitor socket is already closed")
+            return False
         try:
             self.verify_responsive()
             return True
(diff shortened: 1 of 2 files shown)


pevogam:
@luckyh Would you mind sharing your opinion here?

virttest/qemu_monitor.py
```diff
@@ -492,7 +492,7 @@ def is_responsive(self):
         try:
             self.verify_responsive()
             return True
-        except MonitorError:
+        except (MonitorError, ValueError):
```
luckyh, line 495: Can we simply make it by checking the `fileno`, in this case?
```python
if self._socket.fileno() < 0:
    return False
```

pevogam: Good idea, will test and come back to you here.

pevogam: This suggestion works and is more refined, thanks! I pushed the changes and have tested them locally already.

luckyh (approved):
LGTM

pevogam:
@luckyh I have rebased the branch on top of the most recent tag, all CI passes now and we have two acks.

luckyh:
> @luckyh I have rebased the branch on top of the most recent tag, all CI passes now and we have two acks.

Thanks!

PR #2985, opened 2021-03-31
Add EventTracker to enable virsh commands to wait for events

chloerh:
By defining function with this decorator, it allows virsh commands
to wait for certain events to show up before returning. Exception
will be raised if timeout and event is not found.

----
Test log:
```
Sending "event avocado-vt-vm1 --all --loop" to virsh shell
Running virsh command: change-media avocado-vt-vm1 hdc  --update --live /tmp/avocado_rzvye8ng/change_media_new.iso 
Running '/bin/virsh change-media avocado-vt-vm1 hdc  --update --live /tmp/avocado_rzvye8ng/change_media_new.iso '
[stdout] Successfully updated media.
Command '/bin/virsh change-media avocado-vt-vm1 hdc  --update --live /tmp/avocado_rzvye8ng/change_media_new.iso ' finished with 0 after 0.03598618507385254s
status: 0
stdout: Successfully updated media.
stderr: 
Welcome to virsh, the virtualization interactive terminal.

Type:  'help' for help with commands
       'quit' to quit

virsh # event 'tray-change' for domain avocado-vt-vm1 disk scsi0-0-0-2: closed

Command output:
Successfully updated media.
```
Test result:
```
PASS 1-type_specific.local.virsh.change_media_matrix.pre_vm_state_paused.cdrom_test.current.action_twice.live.insert_update
```

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -30,6 +30,9 @@ import time
 import select
 import locale
 import base64
+import inspect
+
+from functools import wraps
 
 import aexpect
 from aexpect import remote
@@ -619,6 +622,110 @@ class VirshConnectBack(VirshPersistent):
             bool(uri.count("127."))
         ]
         return True not in all_false
+
+
+class EventNotFoundError(Exception):
+    """
+    Error when certain event cannot be found.
+    """
+
+    def __init__(self, details=''):
+        self.details = details
+
+    def __str__(self):
+        return str(self.details)
+
+
+class EventTracker(object):
+
+    @staticmethod
+    def start_get_event(vm_name):
+        """
+        Use a virsh session with subcommand 'event' to catch events
+
+        :param vm_name: name of the vm to be catched
+        :return: the virsh session with 'event'
+        """
+        virsh_session = aexpect.ShellSession(VIRSH_EXEC)
+        event_cmd = 'event %s --all --loop' % vm_name
+        logging.info('Sending "%s" to virsh shell', event_cmd)
+        virsh_session.sendline(event_cmd)
+
+        return virsh_session
+
+    @staticmethod
+    def finish_get_event(virsh_session):
+        """
+        Stop virsh session and return the event output
+
+        Usage:
+        virsh_session = start_get_event(vm_name)
+        ####
+        virsh commands or other operations
+        ####
+        event_output = finish_get_event(virsh_session)
+
+        :param virsh_session: virsh session to catch events
+        :return: actual event output catched
+        """
+        virsh_session.send_ctrl("^C")
+        time.sleep(5)
+        event_output = virsh_session.get_stripped_output()
+        virsh_session.close()
+        logging.debug('Event output is %s:', event_output)
+
+        return event_output
+
+    @staticmethod
+    def wait_event(func):
+        """
+        Decorator aiming to return until certain type of virsh event defined
+        by func was found
+
+        Usage:
+        @EventTracker.wait_event
+        def func(xx, xx, ... wait_for_event=False,
+                 event_type='tray-change', event_timeout=7)
+
+        """
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+
+            def _get_arg_value(arg):
+                return kwargs.get(arg) if arg in kwargs else \
+                    inspect.signature(func).parameters[arg].default
+
+            def _get_event_output(session):
+                output = session.get_stripped_output()
+                logging.debug(output)
+                return output
+
+            wait_for_event = _get_arg_value('wait_for_event')
+            event_type = _get_arg_value('event_type')
+            event_timeout = _get_arg_value('event_timeout')
+
+            if wait_for_event is True and event_type is not None:
+                virsh_session = EventTracker.start_get_event(str(args[0]))
+                ret = func(*args, **kwargs)
+
+                if ret.exit_status:
+                    logging.error('Command execution failed. Skip waiting for event')
+                    virsh_session.close()
+                    return ret
+
+                if not utils_misc.wait_for(
+                        lambda: re.search(
+                            event_type,
+                            _get_event_output(virsh_session)
+                        ), event_timeout):
+                    raise EventNotFoundError('Not found event %s after %s seconds' %
+                                             (event_type, event_timeout))
+                virsh_session.close()
+                return ret
+            else:
+                return func(*args, **kwargs)
+
+        return wrapper
 
 
 # virsh module functions follow (See module docstring for API) #####
@@ -3189,14 +3296,17 @@ def cpu_stats(name, options, **dargs):
     return command(cmd, **dargs)
 
 
-def change_media(name, device, options, wait_change_event=False, event_timeout=7, **dargs):
+@EventTracker.wait_event
+def change_media(name, device, options, wait_for_event=False,
+                 event_type='tray-change', event_timeout=7, **dargs):
     """
     Change media of CD or floppy drive.
 
     :param name: VM's name.
     :param path: Fully-qualified path or target of disk device
     :param options: command change_media options.
-    :param wait_change_event: wait until device_change event comes
+    :param wait_for_event: wait until device_change event comes
+    :param event_type: type of the event
     :param event_timeout: timeout for virsh event command
     :param dargs: standardized virsh function API keywords
     :return: CmdResult instance
@@ -3205,8 +3315,6 @@ def change_media(name, device, options, wait_change_event=False, event_timeout=7
     if options:
         cmd += " %s " % options
     change_media_result = command(cmd, **dargs)
-    if wait_change_event:
-        event(domain=name, event='tray-change', event_timeout=event_timeout, **dargs)
 
     return change_media_result
 


Yingshun (commented):
LGTM. It can also fix detach_device problem, awesome!

virttest/virsh.py
```diff
@@ -30,13 +30,16 @@
...
 import base64
+import inspect
 
 import aexpect
 from aexpect import remote
 
 from avocado.utils import path
 from avocado.utils import process
 
+from functools import wraps
```
chunfuwen, line 41: this is system packages, please arrange imports by orders: system package,thirty party,then local .

chloerh: Nice catch! updated

virttest/virsh.py
```diff
@@ -621,6 +624,93 @@ def kosher_args(remote_ip, uri):
         return True not in all_false
 
 
+class EventNotFoundError(Exception):
+    """
+    Error when certain event cannot be found.
+    """
+
+    def __init__(self, details=''):
```
chunfuwen, line 632: details= '' will hide silent failure where none of parameters is given __init__

virttest/virsh.py
```diff
@@ -621,6 +624,93 @@ def kosher_args(remote_ip, uri):
...
+        self.details = details
+        Exception.__init__(self)
```
chunfuwen, line 634: I assume super is more straightforward when call parent __init__

virttest/virsh.py
```diff
@@ -621,6 +624,93 @@ def kosher_args(remote_ip, uri):
...
+    @staticmethod
+    def start_get_event(vm_name):
+        """
+        Use a virsh session with subcommand 'event' to catch events
+
+        :param vm_name: name of the vm to be catched
+        :return: the virsh session with 'event'
+        """
+        virsh_session = aexpect.ShellSession(VIRSH_EXEC)
+        event_cmd = 'event %s --all --loop' % vm_name
```
chunfuwen, line 651: In many cases, we need specify event_type ,and timeout as constraints, can this implement handle them gracefully?

virttest/virsh.py
chloerh: Ok, then what do you suggest we should put here?

virttest/virsh.py
```diff
@@ -3189,14 +3279,17 @@ def cpu_stats(name, options, **dargs):
     return command(cmd, **dargs)
 
 
-def change_media(name, device, options, wait_change_event=False, event_timeout=7, **dargs):
+@EventTracker.wait_event
+def change_media(name, device, options, wait_for_event=False,
```
chunfuwen, line 3283: change_media is primitive function ,which is the same with others in this file, so we may keep it as it is , and create another such as change_media_xx() as alternative one.

virttest/virsh.py
chloerh: Let me check, actually I got this definition from LibvirtXMLError, but it seems not necessary now.

virttest/virsh.py
chloerh: Could you provide examples? How gracefully did you mean?

virttest/virsh.py
chloerh: I don't think so. The whole point is to make sure the command will return only if the target event shows. If we create another function, there would be tons of new functions to be created, we don't want that, right?

virttest/virsh.py
chloerh: It is unnecessary. I'll remove it since it's useless.

virttest/virsh.py
chunfuwen: def __init__(self, details):
      self.details = details
In this way, if error will throw if no parameter is given when object  is constructed.
As zen of python says, Errors should never pass silently.

virttest/virsh.py
chunfuwen: > Could you provide examples? How gracefully did you mean?

At least, it need allow user to pass in event_type and timeout parameters  as command line, rather than hardcode --loop when start  virsh event

virttest/virsh.py
chunfuwen: @kylazhang 
@dzhengfy 
please provide your inputs

virttest/virsh.py
```diff
@@ -621,6 +624,93 @@ def kosher_args(remote_ip, uri):
...
+
+        :param vm_name: name of the vm to be catched
+        :return: the virsh session with 'event'
+        """
+        virsh_session = aexpect.ShellSession(VIRSH_EXEC)
+        event_cmd = 'event %s --all --loop' % vm_name
+        logging.info('Sending "%s" to virsh shell', event_cmd)
+        virsh_session.sendline(event_cmd)
+
+        return virsh_session
```
chunfuwen, line 655: I am a little why you return virsh_session when start_get_event, versus virsh_session as pass in parameter in finish_get_event.

virttest/virsh.py
```diff
@@ -621,6 +624,93 @@ def kosher_args(remote_ip, uri):
...
+        Exception.__init__(self)
+
+    def __str__(self):
+        return str(self.details)
+
+
+class EventTracker(object):
+
+    @staticmethod
+    def start_get_event(vm_name):
```
chunfuwen, line 643: Is this method dedicated to  wait_event for use or exposed as public API?
something like we can use :
EventTracker.start_get_event(vm)
....
EventTracker.start_finish_event(vm)

virttest/virsh.py
chloerh: @chunfuwen Of course, what do you suggest we put here? Seems you still didn't provide any suggestion?

virttest/virsh.py
chloerh: But why, why is --loop so bad? I think it's pretty good to logging out other events as output when there might be, it's easier to debug, don't you think?

virttest/virsh.py
chloerh: BTW, I didn't want to say this, but after second thought, maybe it could be helpful to you, so I'll just say it:
Seems you misunderstood the meaning of **Errors should never pass silently** which means we should always deal with errors, so....any more comments?

> def **init**(self, details):
> self.details = details
> In this way, error will throw if no parameter is given when object is constructed.
> As zen of python says, Errors should never pass silently.

virttest/virsh.py
chloerh: @chunfuwen that's because we need virsh session to get event

virttest/virsh.py
dzhengfy: @chunfuwen, I suggest you think more. From my POV, aexpect.ShellSession is a thirdparty package, we'd better not add test specific functions into it.

virttest/virsh.py
chunfuwen: There is one assumption here that  you agree that obj = EventNotFoundError() is bad practice since there is no any meaning  input ,but object can created without any error. Otherwise there is no need to process this conversation. 
Below comments is based on above assumption.
```
if we define init as below, when you attempt to obj = EventNotFoundError() ,it will throw error immediately. That is part of zen of python says, Errors should never pass silently.
      def  __init__(self, details):
             self.details = details
```

virttest/virsh.py
chunfuwen: I think we need support both --loop --event-type and  --event-type  --timeout ,this is test case requirement.

virttest/virsh.py
chloerh: > There is one assumption here that you agree that obj = EventNotFoundError() is bad practice since there is no any meaning input ,but object can created without any error. Otherwise there is no need to process this conversation.

No I don't, so.....

> Below comments is based on above assumption.
> if we define init as below, when you attempt to obj = EventNotFoundError() ,it will throw error immediately. That is part of zen of python says, Errors should never pass silently.
> 
> ```
>       def  __init__(self, details):
>              self.details = details
> ```

So, any more comment?

virttest/virsh.py
chloerh: @chunfuwen though I don't agree, but how do you propose we to do that to support?

virttest/virsh.py
chunfuwen: Generally, when people see start_get_event() , the first impression occurred to people is that it need return True or False indicating whether event is started or not , rather than return one session. 
Moreover , start_get_event and finish_get_event are similar functions, but it make more consistently if both of them accept the same parameter ,e.g vm_name.

virttest/virsh.py
chloerh: I just want to say again, since clearly you missed it, **Errors should never pass silently**  means we should always deal with errors.... not what you stated....

virttest/virsh.py
chloerh: Then how do you propose we should edit this function?

virttest/virsh.py
chunfuwen: > @chunfuwen, From my POV, aexpect.ShellSession is a thirdparty package, we'd better not add test specific functions into it.
aexpect is maintained by lukas and clebergnu from Red Hat, part of avocado projects
https://github.com/avocado-framework/aexpect

virttest/virsh.py
chloerh: The current functions are working well, I have tested. Feel free to suggest better implementations.

chloerh: Also, how necessary do we need to modify the function?

virttest/virsh.py
chloerh: @chunfuwen I agree with @dzhengfy .
But, @chunfuwen feel free to add these 2 functions to aexpect if you'd like. But that should not be reason to block this pr.
So, I'd be glad to modify these functions after your commit to aexpect got merged. Thanks.

virttest/virsh.py
chunfuwen: > silently

I do believe "Errors should never pass silently" cover more than what you stated

chloerh: @chunfuwen of course, I can get that from your comments. We have different understanding about this sentence. But having thought about your suggestion, I'd keep current implementation.

chunfuwen:
@pevogam 
If you can participate in our review, we are very grateful

pevogam:
> @pevogam
> If you can participate in our review, we are very grateful

Hi @chunfuwen, unfortunately we are only using the qemu side of the Avocado VT and not the libvirt side so I am afraid I can't test your changes here and am not much of an expert in these alternative implementation to aid with review. But I hope the others will review, there are probably more users of the libvirt functionality then there are of the qemu functionality (I think).

chunfuwen:
> > @pevogam
> > If you can participate in our review, we are very grateful
> 
> Hi @chunfuwen, unfortunately we are only using the qemu side of the Avocado VT and not the libvirt side so I am afraid I can't test your changes here and am not much of an expert in these alternative implementation to aid with review. But I hope the others will review, there are probably more users of the libvirt functionality then there are of the qemu functionality (I think).

@pevogam , sorry to misunderstand you since I observed that you reviewed some PRs from chloerh previously. But anyway, thanks for your feedbacks.

virttest/virsh.py
chunfuwen: how about adding three additional parameters: one is event_type ,second is option(--loop or --timeout), and third one is **kargs(e.g provide timeout value) ?  I think there are lots of examples in virsh.py how to compose them together.

virttest/virsh.py
chunfuwen: Can we pass  virsh_session  as one parameter to those methods?

virttest/virsh.py
```diff
@@ -621,6 +624,93 @@ def kosher_args(remote_ip, uri):
...
+        """
+        virsh_session = aexpect.ShellSession(VIRSH_EXEC)
+        event_cmd = 'event %s --all --loop' % vm_name
+        logging.info('Sending "%s" to virsh shell', event_cmd)
+        virsh_session.sendline(event_cmd)
+
+        return virsh_session
+
+    @staticmethod
+    def finish_get_event(virsh_session):
```
chunfuwen, line 658: In my understanding, what you want to expose is wait_event decorator , but wait_event doesn't call this method at all. 
Could you elaborate more what  use case finish_get_event want to achieve, how it interact with other methods?

virttest/virsh.py
chloerh: Sorry that I just don't get it. How would you like to provide an example of what you want this function to be? I think it might be easier for me to understand your suggestion.

virttest/virsh.py
chloerh: I don't think it's quite necessary right now. But feel free to perfect the implementation in the future. Thanks.

virttest/virsh.py
chloerh: @chunfuwen yes, finish_get_event() has nothing to do with wait_for_event(). Please refer to https://github.com/autotest/tp-libvirt/pull/3371

virttest/virsh.py
chunfuwen: I would like this utility can implement --event-type --timeout rather than --loop if it need make choice . 
If we look at detach_device, attach_device, or  even change-media  currently n virsh.py, what they need is just to wait for certain type event with specific timeout value.

virttest/virsh.py
chunfuwen: Rough workflow follows as:

1)virsh_session = aexpect.ShellSession(VIRSH_EXEC)
2)start_get_event(virsh_session,xx)
3)finish_get_event(virsh_session,xx)

virttest/virsh.py
chunfuwen: Better has usage guideline in your top part of your EventTracker class

chloerh: @chunfuwen oh? What kind of usage guideline?

virttest/virsh.py
chloerh: @chunfuwen sure, feel free to do the modification after it got merged. Thanks.

virttest/virsh.py
chloerh: After considering your proposal, I'd like to keep the function the way it is now, if it's not unacceptable to you.

virttest/virsh.py
chunfuwen: https://github.com/autotest/tp-libvirt/pull/3363/files
I think part of  pr is related to this 3363.
Kyla suggested wait_for_event under change-media is limited to specific sets of cases, which clues can be gotten from 3363,
We don't want to apply wait_for_event on cases with using change-media command

virttest/virsh.py
chunfuwen: If I understand your API correctly, somehow it looks like below:

```
Usage: 

options:
  @wait_for_events decorator : what kind of this can achieve
  @start_wait_event, finish_wait_event:
      start_wait_event()
           .....
           func1()
      finish_wait_event()
```

chunfuwen:
@smitterl , I would like invite you to join this review if you don't mind

virttest/virsh.py
chloerh: Oh? But I didn't see usage guideline in https://github.com/avocado-framework/avocado-vt/pull/2442 
So, seems function description is quite enough, agreed?
And I thought this thread of comments is about function finish_get_event() , right? Oddly we're discussing usage... really odd.

virttest/virsh.py
chloerh: @chunfuwen After talk to kyla, I think I understand what did you mean, and I'm very glad to explain the question you had:
See this arg 'wait_for_event=False' ? It's a sign of whether we need to wait for event or not, and it's False by default. 
I'm pretty sure you will understand the logic after my explanation.  You're welcome~

virttest/virsh.py
chunfuwen: > Oh? But I didn't see usage guideline in #2442
> So, seems function description is quite enough, agreed?
> And I thought this thread of comments is about function finish_get_event() , right? Oddly we're discussing usage... really odd.

@chloerh 
this is just enhancement to your exposed API.
You export two APIs and one decorator.
Until now I still need guest how we can use those API to achieve something except decorator on change_media

chloerh: @chunfuwen I'd be happy to explain to you. Would you like to tell me which part/line of code confuses you? And would you make your question more specific? It will help me understand which part is confusing you. Thanks.

chloerh:
@dzhengfy @Yingshun Hi, I have updated test log and test result of an example case. Would you like to review again? Thank you very much.

virttest/virsh.py
chunfuwen: commands  implementation in virsh.py is primitive. If apply one decorator on change_media, it will go decorator inner when virsh.event is called each time.  It sounds it works from functionality point,but  It is against the style of virsh.py(primitive). 
Considering only few cases  in change media need wait for events, Special cases aren't special enough to break the rules(zen of python)

kylazhang: I think there is no need to write new cmd for change_media as in the cases there will be only 1 cmd for change_media if only need once operation, we don't want to add if else in the code and just an parameter with different value should be ok in cfg file. Also checking the wrapper if wait_for_event=False it will return the function itself directly. We can do some test to make sure it can work. And the advantage of decorator is that when other cmd also need this function, it will be easy to update.

dzhengfy (approved):
LGTM

chunfuwen:
please pending for merge before some agreements are reached.

chloerh:
@dzhengfy Thank u for the review.

chloerh:
> please pending for merge before some agreements are reached.

@chunfuwen Please specify on which part/line of code there's disagreement, or there's no reason to block this PR.

virttest/virsh.py
```diff
@@ -621,6 +624,93 @@ def kosher_args(remote_ip, uri):
...
+                logging.debug(output)
+                return output
+
+            wait_for_event = _get_arg_value('wait_for_event')
+            event_type = _get_arg_value('event_type')
+            event_timeout = _get_arg_value('event_timeout')
+
+            if wait_for_event is True and event_type is not None:
+                virsh_session = EventTracker.start_get_event(str(args[0]))
+                ret = func(*args, **kwargs)
```
chunfuwen, line 697: ret=func(*args, **kwargs) need always executed ,so pull level up before if wait_for_event.
In this way, you can have consistent return :ret

chloerh: @chunfuwen way ahead of you....
Please see line 709. Maybe you need to wait a little bit longer for the page to fully loaded. Because github is a little bit slow recently.

virttest/virsh.py
```diff
@@ -621,6 +624,93 @@ def kosher_args(remote_ip, uri):
...
+
+                if not utils_misc.wait_for(
+                        lambda: re.search(
+                            event_type,
+                            _get_event_output(virsh_session)
+                        ), event_timeout):
+                    raise EventNotFoundError('Not found event %s after %s seconds' %
```
chunfuwen, line 704: I assume we can not raise exception here to mark case failure if no events are received.
Regarding change-media, whether it received event or not is to determin how fast it will return virsh.change_media result, but not final 'pass' or 'fail' ,which should be judged on virsh.change_media result

Yingshun: Sorry, I don't quite understand you. so now, you are using wait_for_event to wait for the event but you don't want to get it,right? If so, I think what you need is time.sleep.

Yingshun (approved):
LGTM

virttest/virsh.py
chunfuwen: > Sorry, I don't quite understand you. so now, you are using wait_for_event to wait for the event but you don't want to get it,right? If so, I think what you need is time.sleep.

@Yingshun 
Let's just look at primitive implementation change_media in virsh.py
What wait_change_event() bring is to just avoid dead sleep, but not break return of change_media_result regardless whether expected events is received or not.

```
def change_media(name, device, options, wait_change_event=False, event_timeout=7, **dargs):
    """
    Change media of CD or floppy drive.

    :param name: VM's name.
    :param path: Fully-qualified path or target of disk device
    :param options: command change_media options.
    :param wait_change_event: wait until device_change event comes
    :param event_timeout: timeout for virsh event command
    :param dargs: standardized virsh function API keywords
    :return: CmdResult instance
    """
    cmd = "change-media %s %s " % (name, device)
    if options:
        cmd += " %s " % options
    change_media_result = command(cmd, **dargs)
    if wait_change_event:
        event(domain=name, event='tray-change', event_timeout=event_timeout, **dargs)

    return change_media_result

```

chunfuwen: @kylazhang, I assume somehow you know the  event () background

chloerh: @chunfuwen There is a very simple solution here: If you don't want to wait for an event, just set wait_for_event=False. This just perfectly meets your need.

chunfuwen: @kylazhang 
Maybe you can help elaborate a little bit about what original PR:https://github.com/autotest/tp-libvirt/pull/3363 want to achieve after discussion.
@hahakiki2010 
Maybe, you can help describe more about original purposes why add wait_for_event into detach_device,detach_disk,detach_device_alias,detach_interface those methods

chunfuwen: > @chunfuwen There is a very simple solution here: If you don't want to wait for an event, just set wait_for_event=False. This just perfectly meets your need.

@chloerh 
NO, that's not what current change_media,detach_device,detach_disk,detach_device_alias,detach_interface achieved.

chunfuwen: > @chunfuwen There is a very simple solution here: If you don't want to wait for an event, just set wait_for_event=False. This just perfectly meets your need.

@chloerh 
NO, that's not what current change_media,detach_device,detach_disk,detach_device_alias,detach_interface achieved.

chloerh: @chunfuwen It is. You would find that I'm right about it after reading the code. Thanks.

chloerh:
New test result:

```
JOB LOG    : /root/avocado/job-results/job-2021-04-13T16.01-c5096a0/job.log
 (1/3) type_specific.local.virsh.change_media_matrix.pre_vm_state_paused.cdrom_test.current.action_twice.live.insert_update: PASS (43.01 s)
 (2/3) type_specific.local.virsh.change_media_matrix.pre_vm_state_paused.cdrom_test.force.action_twice.live.insert_update: PASS (48.29 s)
 (3/3) type_specific.local.virsh.change_media_matrix.pre_vm_state_paused.cdrom_test.config.action_twice.default.eject_update: PASS (45.68 s)
RESULTS    : PASS 3 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 137.38 s
```

virttest/virsh.py
```diff
@@ -621,6 +624,110 @@ def kosher_args(remote_ip, uri):
...
+        ####
+        virsh commands or other operations
+        ####
+        event_output = finish_get_event(virsh_session)
+
+        :param virsh_session: virsh session to catch events
+        :return: actual event output catched
+        """
+        virsh_session.send_ctrl("^C")
+        time.sleep(5)
```
kylazhang, line 672: After talk with haizhao with the sleep time here, it is acceptable as it will not blocking the session get the output as it is just to wait the event coming out totally. And it is also used in other place for waiting the event.

virttest/virsh.py
```diff
@@ -621,6 +624,110 @@ def kosher_args(remote_ip, uri):
...
+        """
+        Decorator aiming to return until certain type of virsh event defined
+        by func was found
+
+        Usage:
+        @EventTracker.wait_event
+        def func(xx, xx, ... wait_for_event=False,
+                 event_type='tray-change', event_timeout=7)
+
+        """
```
kylazhang, line 690: Thanks for adding the usage here to make others usage more clear.

virttest/virsh.py
kylazhang: Chloerh has already updated the usage in the function so I think is comments should be resolved. Thanks

virttest/virsh.py
```diff
@@ -621,6 +624,110 @@ def kosher_args(remote_ip, uri):
...
+            event_timeout = _get_arg_value('event_timeout')
+
+            if wait_for_event is True and event_type is not None:
+                virsh_session = EventTracker.start_get_event(str(args[0]))
+                ret = func(*args, **kwargs)
+
+                if ret.exit_status:
+                    logging.error('Command execution failed. Skip waiting for event')
+                    virsh_session.close()
+                    return ret
```
kylazhang, line 714: @chunfuwen here adding the return when virsh command failed so it will not wait for the event, which logic is reasonable for all existing cases.

virttest/virsh.py
kylazhang: I think this comment should be resolved by my last comments...

kylazhang:
@chunfuwen @chloerh I am on running all change_media_matrix cases in jenkins job to see if it may match all cases.

kylazhang:
> @chunfuwen @chloerh I am on running all change_media_matrix cases in jenkins job to see if it may match all cases.

I have test on 506 cases for all change_media_matrix and change_media cases, and all passed

kylazhang (approved):

virttest/virsh.py
chunfuwen: ok

chunfuwen (approved):
approve first, let's keep eye on what happen next.

virttest/virsh.py
chunfuwen: still there is gap with what we talk about last meeting

chloerh: @chunfuwen would you say that this thread is resolved?

chunfuwen: @chloerh 

primitive change_media follow below logic:
1)primitive function never judge whether case pass or fail based on  event execution result
2)whether call event(xx) depends on wait_change_event value

```
def change_media(name, device, options, wait_change_event=False, event_timeout=7, **dargs):
    """
    Change media of CD or floppy drive.

    :param name: VM's name.
    :param path: Fully-qualified path or target of disk device
    :param options: command change_media options.
    :param wait_change_event: wait until device_change event comes
    :param event_timeout: timeout for virsh event command
    :param dargs: standardized virsh function API keywords
    :return: CmdResult instance
    """
    cmd = "change-media %s %s " % (name, device)
    if options:
        cmd += " %s " % options
    change_media_result = command(cmd, **dargs)
    if wait_change_event:
        event(domain=name, event='tray-change', event_timeout=event_timeout, **dargs)

    return change_media_result
```

PR #2993, opened 2021-04-08
Remove any sub elem in <cpu>

dzhengfy:
This function is to provide a common support to remove sub element
in <cpu> configuration.

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -2091,6 +2091,20 @@ class VMCPUXML(base.LibvirtXMLBase):
             pass  # Element already doesn't exist
         self.xmltreefile.write()
 
+    def remove_elem_by_xpath(self, xpath_to_remove, remove_all=True):
+        """
+        Remove a specified sub element from cpu configuration
+
+        :param xpath_to_remove:  str, like '/model' '/numa'
+        :param remove_all: bool, True to remove all findings,
+                                 False to remove only one finding
+        """
+        try:
+            self.xmltreefile.remove_by_xpath(xpath_to_remove, remove_all)
+        except (AttributeError, TypeError):
+            logging.info("Element '%s' already doesn't exist", xpath_to_remove)
+        self.xmltreefile.write()
+
     @staticmethod
     def from_domcapabilities(domcaps_xml):
         """


virttest/libvirt_xml/vm_xml.py
```diff
@@ -2091,6 +2091,18 @@ def remove_numa_cells(self):
...
+    def remove_elem_by_xpath(self, xpath_to_remove):
+        """
+        Remove a specified sub element from cpu configuration
+
+        :param xpath_to_remove:  str, like '/model' '/numa'
+        """
+        try:
+            self.xmltreefile.remove_by_xpath(xpath_to_remove, remove_all=True)
+        except (AttributeError, TypeError):
+            pass  # Element already doesn't exist
```
kamilvarga, line 2103: What if wrong xpath was given to method? There is no indication to higher level method which will call this method. It should be logged at least, I believe.

virttest/libvirt_xml/vm_xml.py
```diff
@@ -2091,6 +2091,18 @@ def remove_numa_cells(self):
...
         self.xmltreefile.write()
 
+    def remove_elem_by_xpath(self, xpath_to_remove):
+        """
+        Remove a specified sub element from cpu configuration
+
+        :param xpath_to_remove:  str, like '/model' '/numa'
+        """
+        try:
+            self.xmltreefile.remove_by_xpath(xpath_to_remove, remove_all=True)
```
kamilvarga, line 2101: I think it would be good to have remove_all parameter configurable for this method.

kamilvarga (changes requested):

virttest/libvirt_xml/vm_xml.py
dzhengfy: Simply copied from existing codes. But you are right, I updated.

virttest/libvirt_xml/vm_xml.py
dzhengfy: Good catch. I enhanced it. Thanks.

virttest/libvirt_xml/vm_xml.py
kamilvarga: The pass statement could be deleted, but it is not harmful, so it can stay there as well.

kamilvarga (approved):
LGTM

virttest/libvirt_xml/vm_xml.py
dzhengfy: Yes, you are right, I forgot to remove it. Done.

dzhengfy:
Test result, see https://github.com/autotest/tp-libvirt/pull/3423#issuecomment-815501101

kylazhang (approved):

PR #2999, opened 2021-04-12
[get_tmp_dir] Document intended usage.

smitterl:
tmp_dir per default points to /tmp which can be tmpfs
leading to issues when use e.g. for images. Instead, the data_dir
should be used.

### DIFF
--- a/virttest/data_dir.py
+++ b/virttest/data_dir.py
@@ -144,6 +144,26 @@ def get_backend_cfg_path(backend_type, cfg_basename):
     return os.path.join(get_backend_dir(backend_type), 'cfg', cfg_basename)
 
 
+def get_fs_type(abs_path):
+    """
+    Determines the file system type for an existing
+    absolute path inspecting /proc/mounts
+    """
+    res = None
+    dev = os.lstat(abs_path).st_dev
+    with open("/proc/mounts", "r") as f:
+        for line in f:
+            # lines are device, mountpoint, filesystem, <rest>
+            line = [s for s in line.split()[:3]]
+            try:
+                if dev == os.lstat(line[1]).st_dev:
+                    res = line[2]
+                    break
+            except OSError as err:
+                continue
+    return res
+
+
 def get_deps_dir(target=None):
     """
     For a given test provider, report the appropriate deps dir.
@@ -194,6 +214,8 @@ def get_deps_dir(target=None):
 def get_tmp_dir(public=True):
     """
     Get the most appropriate tmp dir location.
+    For heavy usages like test images or requiring direct access
+    please consider using some path below data_dir instead.
 
     :param public: If public for all users' access
     """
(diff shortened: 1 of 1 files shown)


smitterl:
This PR is intended to help cover discussion on https://github.com/avocado-framework/avocado-vt/pull/2972

virttest/data_dir.py
```diff
@@ -144,6 +144,26 @@ def get_backend_cfg_path(backend_type, cfg_basename):
     return os.path.join(get_backend_dir(backend_type), 'cfg', cfg_basename)
 
 
+def get_fs_type(abs_path):
```
luckyh, line 147: Suggest it being a private function because ideally this should be provided by a utility module (e.g. `filesystem.py`).

virttest/data_dir.py
```diff
@@ -215,6 +237,11 @@ def get_tmp_dir(public=True):
         tmp_dir_st = os.stat(tmp_dir)
         os.chmod(tmp_dir, tmp_dir_st.st_mode | stat.S_IXUSR |
                  stat.S_IXGRP | stat.S_IXOTH | stat.S_IRGRP | stat.S_IROTH)
+    if "tmpfs" == get_fs_type(tmp_dir):
+        logging.warning("tmp_dir is type tmpfs. This might lead to issues"
+                        " e.g. when requiring direct access. Please, consider"
+                        " using data_dir or configuring another path in vt.common")
```
pevogam, line 243: What if we simply use the tmp_dir for storing a few files with faster IO? Should this warning always be shown? Or is it planned to be temporary?

virttest/data_dir.py
```diff
@@ -194,6 +214,8 @@ def get_deps_dir(target=None):
 def get_tmp_dir(public=True):
     """
     Get the most appropriate tmp dir location.
+    For heavy usages like test images or requiring direct access
```
pevogam, line 217: It is better if the main docstring statement is separated from this explanation by an empty line (i.e. empty line after "Get the most appropriate tmp dir location..").

virttest/data_dir.py
pevogam: Is there a way that this is taken a step further and placed in a separate module focused on filesystems (perhaps even an existing one considering the vastness of utilities)?

virttest/data_dir.py
```diff
@@ -215,6 +237,11 @@ def get_tmp_dir(public=True):
         tmp_dir_st = os.stat(tmp_dir)
         os.chmod(tmp_dir, tmp_dir_st.st_mode | stat.S_IXUSR |
                  stat.S_IXGRP | stat.S_IXOTH | stat.S_IRGRP | stat.S_IROTH)
+    if "tmpfs" == get_fs_type(tmp_dir):
+        logging.warning("tmp_dir is type tmpfs. This might lead to issues"
```
chunfuwen, line 241: May I ask add tmp_dir="/var/tmp" here?

smitterl:
Thanks for your comments @luckyh and @pevogam 
I don't see a need for this PR anymore for the reasons listed below, so closing this

1. Setting the configuration value to non-tmpfs path works
2. https://github.com/avocado-framework/avocado-vt/pull/3011 fixed known issues in the code
3. Any path even data_dir could be tmpfs - theoretically so we would have to add a debug logging message to all of them and checking this in every test introduces overhead


CC: @chunfuwen

PR #2988, opened 2021-04-02
Remove retuning xml tmp file

dzhengfy:
Sometimes it triggers GC when returning xml tmp file path which causes
the following referrence fails. So it should be removed. This PR can
fix the problem by directly returning whole object.

The error is like:
`CmdError: Command '/bin/virsh attach-device avocado-vt-vm1 /tmp/xml_utils_temp_9exbbws7.xml --current' failed.stdout: b'\n'stderr: b"error: Failed to open file '/tmp/xml_utils_temp_9exbbws7.xml': No such file or directory\n"additional_info: None`

Signed-off-by: Dan Zheng <dzheng@redhat.com>

PR #3000, opened 2021-04-13
Add extra file to avoid "Failed to open file" error

kylazhang:

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -28,6 +28,7 @@ import aexpect
 import platform
 import random
 import string
+import shutil
 
 from aexpect import remote
 
@@ -2710,7 +2711,9 @@ def attach_additional_device(vm_name, targetdev, disk_path, params, config=True)
     params['target_dev'] = targetdev
 
     # Create a file of device
-    xmlfile = create_disk_xml(params)
+    xmlfile = os.path.join(data_dir.get_tmp_dir(), "xml_file")
+    tmpfile = create_disk_xml(params)
+    shutil.copyfile(tmpfile, xmlfile)
 
     # To confirm attached device do not exist.
     if config:


virttest/utils_test/libvirt.py
```diff
@@ -2710,7 +2711,9 @@ def attach_additional_device(vm_name, targetdev, disk_path, params, config=True)
     params['target_dev'] = targetdev
 
     # Create a file of device
-    xmlfile = create_disk_xml(params)
+    xmlfile = os.path.join(data_dir.get_tmp_dir(), "xml_file")
+    tmpfile = create_disk_xml(params)
+    shutil.copyfile(tmpfile, xmlfile)
```
dzhengfy, line 2716: This problem might be caused by create_disk_xml(). Could you test the similar fix in https://github.com/avocado-framework/avocado-vt/pull/2988 to see if it solves your problem?

kylazhang: @dzhengfy Just checking the patch for #2988 but I don't think change the return value is a good way as it is widely used.

chunfuwen: From running test result on latest release, it sound likes there are a few places to throw this error.
We may need investigate root clause and  check whether there is uniform place to fix that.

chunfuwen (approved):
let's approve and have a try.

virttest/utils_test/libvirt.py
dzhengfy: > @dzhengfy Just checking the patch for #2988 but I don't think change the return value is a good way as it is widely used.

Yes, we have to change all the tp-libvirt cases which are using it. Anyway if they are working, then it is worth of doing it.

dzhengfy (approved):
This workaround fix LGTM

PR #3001, opened 2021-04-14
Add custom metadata disk into libvirt disk

chunfuwen:


chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.metadatacache.cold_plug.cdrom_qcow2.attach-device

chunfuwen:
This PR is separated from https://github.com/autotest/tp-libvirt/pull/3350 to move piece of code into libvirt_disk.py

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -40,3 +44,70 @@ def create_disk(disk_type, path=None, size="500M", disk_format="raw", extra='',
...
+
+    :param disk_path: the path of disk
+    :param disk_format: the format to disk image
+    :param disk_device: the disk device type
+    :param device_target: the target of disk
+    :param device_bus: device bus
+    :param max_size: metadata_cache max size
+    :return: disk object if created successfully
+    """
+    custom_disk = Disk(type_name='file')
```
dzhengfy, line 62: Here I suggest we use a parameter to pass the Disk(or its subclass) object instead of hard coding Disk(). If that, the codes will have stronger adaptability.  For example, in https://github.com/autotest/tp-libvirt/pull/3350 around line 88, you can reuse this function.

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -40,3 +44,70 @@ def create_disk(disk_type, path=None, size="500M", disk_format="raw", extra='',
...
+    return dirty_images
+
+
+def clean_images_with_xattr(dirty_images):
+    """
+    Clean the images' xattr
+
+    :param dirty_images: The image list to be cleaned
+    """
+    clean_cmd = ("for i in $(getfattr -m trusted.libvirt.security -d %s |"
```
dzhengfy, line 108: It is not good to use bash script here. I suggest we translate it to python codes.

virttest/utils_libvirt/libvirt_disk.py
chunfuwen: updated

virttest/utils_libvirt/libvirt_disk.py
chunfuwen: remove it from this PR

chunfuwen:
cleaning libvirtd logs...
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.metadatacache.cold_plug.disk_qcow2.batch_block_copy_snapshot_commit

chunfuwen:
@dzhengfy 

The patch was updated to pass in 'disk xml' as one parameter if needed

chunfuwen:
@kylazhang 
Please help take a look at this PR.

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -80,3 +80,43 @@ def create_primitive_disk_xml(type_name, disk_device, device_target, device_bus,
...
+    else:
+        custom_disk = Disk(type_name='file')
+    if disk_device:
+        custom_disk.device = disk_device
+    source_dict = {}
+    if disk_path:
+        source_dict.update({'file': disk_path})
+    custom_disk.source = custom_disk.new_disk_source(
+        **{"attrs": source_dict})
+    if device_target:
```
kylazhang, line 110: For the above 2 params, disk_path & device_target, are they must be given? If not, I think the default value in function params should be given with like disk_path=None; If not, then I think no need to judge "if device_target" exist.

kylazhang (commented):
Please check comment inline, thanks

virttest/utils_libvirt/libvirt_disk.py
chunfuwen: @kylazhang 
For snapshotdisk, device_target is not mandatory
For specific device, such as cdrom , inner part in <source  /source> can be  empty

kylazhang: Then if it is not necessary, please add "disk_path=None, device_target=None" in function params

kylazhang (commented):
Please check comments inline, thanks

dzhengfy (approved):

virttest/utils_libvirt/libvirt_disk.py
kylazhang: And also I think it is a better way but your way is not unacceptable after we discuss and think more. It depends on you finally.

kylazhang (approved):

PR #3004, opened 2021-04-15
libvirt_cgroup: Add a method to get output as a dict

Yingshun:
Some cases check the result's exist status of virsh command,
so split get_virsh_output_dict() into 2 methods to get the dict
type output from the command object.

BTW, update the values of "quota" related keys to 'max'.

test results:
https://github.com/autotest/tp-libvirt/pull/3437#issue-615842628

### DIFF
--- a/virttest/libvirt_cgroup.py
+++ b/virttest/libvirt_cgroup.py
@@ -195,6 +195,11 @@ class CgroupTest(object):
                     cg_file_value = cg_file.read().strip()
                     if cg_file_value == max_cpu_value:
                         cg_file_value = "max"
+                    if "quota" in cg_key:
+                        if cg_file_value in ["-1", "18446744073709551",
+                                             "17592186044415"]:
+                            standardized_cgroup_info[cg_key] = "max"
+                            continue
                     standardized_cgroup_info[cg_key] = cg_file_value
         else:
             logging.error("You've provided a wrong virsh cmd: %s", virsh_cmd)
@@ -291,6 +296,19 @@ class CgroupTest(object):
             logging.error("There is no virsh cmd '%s'", virsh_cmd)
             return None
         result = func(vm_name, ignore_status=True)
+        return self.convert_virsh_output_to_dict(result)
+
+    def convert_virsh_output_to_dict(self, result):
+        """
+        Convert virsh command output to a dict.
+        This method is only applicable to the virsh command
+        with ':' in the output for now.
+
+        :param result: Virsh command's result.
+            For example, the result of virsh schedinfo, blkiotune, memtune.
+
+        :return: The virsh cmd output, as a dict.
+        """
         output = result.stdout_text.strip()
         output_list = output.splitlines()
         output_dict = {}


virttest/libvirt_cgroup.py
```diff
@@ -275,6 +280,16 @@ def get_virsh_output_dict(self, vm_name=None, virsh_cmd=None):
             logging.error("There is no virsh cmd '%s'", virsh_cmd)
             return None
         result = func(vm_name, ignore_status=True)
+        return self.get_dict_output(result)
+
+    def get_dict_output(self, result):
```
chunfuwen, line 285: whether consider change get_dict_output to __get_dict_output since below there is another  example method: __get_dev_major_minor

Yingshun: Thank you for your comment! I'd like to keep it because it'll be called in other tests in tp-libvirt directly.

dzhengfy, line 285: This function is to parse the virsh command result into a dict. It is so common usage. So I suggest it should be put into other place, maybe virttest/virsh.py as it is much close to virsh cmd implementation?

Yingshun: makes sense, updated.

chloerh: @dzhengfy @Yingshun I have a little concern about moving this function to `virsh`. Those functions in virsh are all about virsh commands which means when we call a virsh.xxx(), xxx is supposed to be an existing virsh command. I guess it feels a little bit weird when a function not including virsh command comes into 'virsh'.

virttest/virsh.py
```diff
@@ -714,6 +714,28 @@ def command(cmd, **dargs):
     return ret
 
 
+def convert_to_dict(result):
+    """
+    Convert virsh command output to a dict.
+
```
chunfuwen, line 720: Can any virsh command output be converted dict?
Or add one reminder that this is specific to virsh command output?

Yingshun: Aha, nice catch, I think no. Converting the output of 'virsh dumpxml' to a dict is meaningless I think. I may move it back to libvirt_cgroup later if there is no better choice.

virttest/libvirt_cgroup.py
Yingshun: @chloerh yeah... I've rechecked the virsh.py, all of functions are related to command(). Thanks for your reminder!
@dzhengfy The module libvirt_cgroup belongs to virttest package, so I wonder if I can keep it in libvirt_cgroup. If we have other requirements in the future, I think we can create a new function under utils_libvirt. How do you think?

virttest/virsh.py
Yingshun: BTW, I will give an example in next push.

virttest/libvirt_cgroup.py
dzhengfy: Accepted.

virttest/virsh.py
Yingshun: docstring updated.

Yingshun:
@dzhengfy @chloerh @chunfuwen I've updated the pr, would you please help re-review? Thanks!

dzhengfy (approved):

chloerh (approved):

chunfuwen (approved):
lgtm

PR #3007, opened 2021-04-19
Add primitive create disk

chunfuwen:


chunfuwen:
PASS 1-type_specific.io-github-autotest-libvirt.virtual_disks.luks.positive_test.coldplug.device_disk.ceph_test.enable_auth.auth_in_source.encryption_in_source

chunfuwen:
this aims to avoid pass in whole dict as the only parameter, where create_vm_disk currently does.

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -40,3 +43,32 @@ def create_disk(disk_type, path=None, size="500M", disk_format="raw", extra='',
     else:
         return libvirt.create_local_disk(disk_type, path=path, size=size,
                                          disk_format=disk_format, extra=extra)
+
+
+def create_primitive_disk(type_name, disk_device, device_target, device_bus,
+                          device_format, disk_src_dict, disk_auth_dict):
+    """
+    Creates primitive disk xml
```
kylazhang, line 51: 1 more blank line needed here

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -40,3 +43,32 @@ def create_disk(disk_type, path=None, size="500M", disk_format="raw", extra='',
...
+    :param type_name: disk type
+    :param disk_device: disk device
+    :param device_target: target device
+    :param device_bus: device bus
+    :param device_format: device format
+    :param disk_src_dict: disk source
+    :param disk_auth_dict: disk auth information
```
kylazhang, line 58: I think above 2 need to give explanation on what type of dict it should give, but not the normal dict

kylazhang (commented):
Others are OK for me

virttest/utils_libvirt/libvirt_disk.py
chunfuwen: updated

virttest/utils_libvirt/libvirt_disk.py
chunfuwen: I add one example in doc string

kylazhang (approved):

dzhengfy (approved):
LGTM

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -40,3 +43,43 @@ def create_disk(disk_type, path=None, size="500M", disk_format="raw", extra='',
...
+
+
+def create_primitive_disk(type_name, disk_device, device_target, device_bus,
+                          device_format, disk_src_dict, disk_auth_dict):
+    """
+    Creates primitive disk xml
+
+    usage: disk_auth_dict = {"auth_user": ceph_auth_user,
+                              "secret_type": auth_sec_usage_type,
+                              "secret_uuid": auth_sec_uuid}
```
kylazhang, line 55: @chunfuwen Please check if the usage is copy error here, thanks

chunfuwen: updated

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -40,3 +43,40 @@ def create_disk(disk_type, path=None, size="500M", disk_format="raw", extra='',
     else:
         return libvirt.create_local_disk(disk_type, path=path, size=size,
                                          disk_format=disk_format, extra=extra)
+
+
+def create_primitive_disk(type_name, disk_device, device_target, device_bus,
```
smitterl, line 48: This signature doesn't indicate that only the xml is created as opposed to e.g. `create_disk` in the same file. Could you change the name please to `create_primitive_disk_xml`?

Why is this function "primitive"? I'm wondering if there's a way to name or document this function to justify its existence https://github.com/avocado-framework/avocado-vt/blob/7f273c05d787908b5f7cef415a6ef9fee9ea1a03/virttest/utils_test/libvirt.py#L1611 ?

Could the comment maybe indicate which use cases are covered by `create_primitive_disk` as opposed to `create_disk_xml`? Is this a `ceph` specific function as suggested by the parameter doc (`ceph_disk_name`)? Should this be reflected in its name?

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -40,3 +43,40 @@ def create_disk(disk_type, path=None, size="500M", disk_format="raw", extra='',
...
+    :param device_format: device format
+    :param disk_src_dict: disk source, dict format like below
+           disk_src_dict = {"attrs": {"protocol": "rbd",
+                                   "name": ceph_disk_name},
+                            "hosts":  [{"name": ceph_mon_ip,
+                                     "port": ceph_host_port}]}
+    :param disk_auth_dict: disk auth information, dict format like below
+           disk_auth_dict = {"auth_user": ceph_auth_user,
+                             "secret_type": auth_sec_usage_type,
+                             "secret_uuid": auth_sec_uuid}
```
smitterl, line 66: `disk_auth_dict` and `disk_src_dict` have `ceph_` parameter names. Is this a ceph specific function? Should the function maybe better be called `create_basic_ceph_disk_xml`?

If not, please change the names.

smitterl (changes requested):
Hi @chunfuwen Thanks for this. I'm a bit worried that we create this function with some unclear parameter and function names. Could you please have a look?

virttest/utils_libvirt/libvirt_disk.py
chunfuwen: update name to create_primitive_disk_xml

virttest/utils_libvirt/libvirt_disk.py
chunfuwen: function has been changed to create_primitive_disk_xml.
This is more of general disk generation, disk_auth_dict given here is just one demo how it looks like

chunfuwen:
@smitterl 

Thank you for your review time.
Your commented are addressed accordingly

smitterl (approved):

PR #3008, opened 2021-04-20
qemu_vm: Replace password with password-secret for spice

vivianQizhu:
id: 1949060

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -3201,3 +3201,28 @@ class DevContainer(object):
                     aobject=aobject, backend='hmat_type'))
 
         return hmat_cache_devs
+
+    def secret_object_define_by_varibles(self, secret_id, data,
+                                         secret_format=None, inline=True):
+        """
+        Generate secret object device with given data, will create a new file
+        when inline is False
+
+        :param secret_id: the id of the secret object
+        :param data: password, in raw or base64 format
+        :param secret_format: raw or base64, by default not specify it
+        :param inline: append data=xxx if True and file=xxx if False
+        :return: the secret object device QObject
+        """
+        params = {'id': secret_id}
+        if secret_format:
+            params.update({'format': secret_format})
+        if not inline:
+            secret_file = os.path.join(
+                data_dir.get_tmp_dir(), '%s.secret' % secret_id)
+            with open(secret_file, 'w') as f:
+                f.write(data)
+            params.update({'file': secret_file})
+        else:
+            params.update({'data': data})
+        return qdevices.QObject('secret', params=params)
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1043,7 +1043,19 @@ class VM(virt_vm.BaseVM):
             elif optget("spice_port") != "no":
                 set_value("port=%s", "spice_port")
 
-            set_value("password=%s", "spice_password", "disable-ticketing=on")
+            password = optget("spice_password")
+            secret_cmdline = ''
+            if password:
+                if ("password-secret" not in
+                        devices.execute_qemu("-spice help")):
+                    spice_opts.append('password=%s' % password)
+                else:
+                    secret_id = 'spice_sec0'
+                    secret_cmdline = devices.secret_object_define_by_varibles(
+                        secret_id, data=password).cmdline()
+                    spice_opts.append('password-secret=%s' % secret_id)
+            else:
+                spice_opts.append("disable-ticketing=on")
             ip_ver = optget("listening_addr")
             if ip_ver:
                 host_ip = utils_net.get_host_ip_address(self.params, ip_ver)
@@ -1125,7 +1137,7 @@ class VM(virt_vm.BaseVM):
             set_yes_no_value("spice_ipv6",
                              yes_value="ipv6=on", no_value="ipv6=off")
 
-            return " -spice %s" % (",".join(spice_opts))
+            return secret_cmdline + " -spice %s" % (",".join(spice_opts))
 
         def add_qxl(qxl_nr, base_addr=29):
             """


vivianQizhu:
Get expected command line with qemu6.0:
```
    -object secret,id=spice_sec0,format=base64,file=/tmp/avocado_k32ftdie/spice_sec0.secret \
    -spice port=3180,password-secret=spice_sec0,addr=0,image-compression=auto_glz,zlib-glz-wan-compression=auto,streaming-video=all,agent-mouse=on,playback-compression=on,ipv4=on,ipv6=off  \
```

vivianQizhu:
@yanan-fu Would you help review? Thanks.

yanan-fu:
@vivianQizhu  Could you check the issue id, i think it should be 1949060.

virttest/qemu_vm.py
```diff
@@ -1043,7 +1043,20 @@ def set_value(opt_string, key, fallback=None):
             elif optget("spice_port") != "no":
                 set_value("port=%s", "spice_port")
 
-            set_value("password=%s", "spice_password", "disable-ticketing=on")
+            password = optget("spice_password")
+            secret_cmdline = ''
+            if password:
+                if process.run(
+                        '%s -spice help | grep password-secret' % qemu_binary,
+                        ignore_status=True, shell=True).exit_status:
```
yanan-fu, line 1051: Suggest `if "password-secret" not in devices.execute_qemu("-spice help"):`.

virttest/qemu_devices/qcontainer.py
```diff
@@ -3201,3 +3202,41 @@ def numa_hmat_cache_define_by_params(self, nodeid, params):
...
+                      encrypt=True, inline=True, secret_file=None):
+        """
+        Generate secret object device with given data, will create a new file
+        when inline is False and file is None
+
+        :param secret_id: the id of the secret object
+        :param data: password, could be None when file is provided and
+                     inline is True
+        :param secret_format: raw or base64
+        :param encrypt: encrypt data when True and format is not raw
```
yanan-fu, line 3216: Does the `secret_format` indicates whether it is encrypted or not  already ?
I mean, when using `format=base64` in qemu command, the `data` is encrypted already, is it ?

vivianQizhu: @yanan-fu I suppose the user could give the plain text in data, and `encrypt=True` to let the function help encrypt it to `base64`, or it could give directly `base64` format data and `encrypt=False`, so the usage could be:
`secret_object(secret_id, data='123456', format='base64', encrypt=True)` or
`secret_object(secret_id, data='MTIzNDU2', format='base64', encrypt=False)`
What do you think?

virttest/qemu_vm.py
vivianQizhu: @yanan-fu Updated this one as you recommended, thanks.

virttest/qemu_devices/qcontainer.py
yanan-fu: Ah~ i only suppose user define the plain text data, and qemu command line use encrypted data that handled by this function, the scenario one indeed.
No storage idea with the scenario two that user define a encrypted data. Call for more idea with other uses that use secret object too  @zhencliu could you have a look ?  Thanks!

zhencliu: The secret object used for storage access fully depends on user input, i.e.  user can represent secret data in two ways: 'plain text data(data) + raw(format)' or 'data in base64(data) + base64(format)', we never encrypt/decrypt the sensitive data in VT.  Simply, show me the data and its format, I use it.

Also, for storage access, usually we have no choice for using key-secret(inline passwd) or password-secret(a secret file should exist, containing the passwd), because  only one is supported. 

For this patch, it seems raw/base64 password and key-secret/password-secret are all OK, we can even allow user to pass a file containing password, that's very flexible, but consider devices.secret_object('id'), this is correct from grammer perspective, but it might not make sense. 
What about 'def secret_object(self, secret_id, data, secret_format='raw', inline=True)', data is required, adding data encryption seems not part of responsibility for this function.

vivianQizhu: > The secret object used for storage access fully depends on user input, i.e. user can represent secret data in two ways: 'plain text data(data) + raw(format)' or 'data in base64(data) + base64(format)', we never encrypt/decrypt the sensitive data in VT. Simply, show me the data and its format, I use it.
> 
> Also, for storage access, usually we have no choice for using key-secret(inline passwd) or password-secret(a secret file should exist, containing the passwd), because only one is supported.
> 
> For this patch, it seems raw/base64 password and key-secret/password-secret are all OK, we can even allow user to pass a file containing password, that's very flexible, but consider devices.secret_object('id'), this is correct from grammer perspective, but it might not make sense.

Thanks @zhencliu Good point.

> What about 'def secret_object(self, secret_id, data, secret_format='raw', inline=True)', data is required, adding data encryption seems not part of responsibility for this function.

I am okay with both ways, even it is not flexible enough but it is still sufficient for any type of users, what's your opinion, @yanan-fu ?

yanan-fu: Agree, I am ok with this too, function is more concise while can meet the different type too.

virttest/qemu_devices/qcontainer.py
```diff
@@ -3201,3 +3202,41 @@ def numa_hmat_cache_define_by_params(self, nodeid, params):
...
+        if not inline:
+            if not secret_file:
+                suffix = '.secret'
+                secret_file = os.path.join(
+                    data_dir.get_tmp_dir(), '%s' % secret_id)
+                i = 1
+                file_dir_tmp = secret_file
+                while os.path.exists(file_dir_tmp + suffix):
+                    file_dir_tmp = secret_file + '_%s' % i
+                    i += 1
```
yanan-fu, line 3235: If run test with a big loop(use inline=False), there will be a lot of secret_files, and no clean up action, it is a problem i think.
Discussed with @zhencliu  about this too, without adding clean up action,  we can use `secret_id` as the file name, and overwrite if the file exist.  How do you think about this ? @vivianQizhu

vivianQizhu: @yanan-fu It is in `data_dir.get_tmp_dir` directory, will be removed somehow I suppose?

yanan-fu: It is the framework's tempoary dir with host, no clean up as far as i know.

vivianQizhu: @yanan-fu I thought avocado will clean it, and normally after the job stop those temporary directory like named avocado_xxx would disappear, correct me if I was wrong.
```
# Destroy the data_dir.get_tmpdir ...
data_dir._tmp_tracker.unittest_refresh_dir_tracker()
```

yanan-fu: @vivianQizhu The `avocado_xxx` is same when launch `avocado run` one time - a test loop include multiple test cases, avocado will not clean it for every case (post process).

vivianQizhu: @yanan-fu Okay, I get your point, it creates one more secret file for each test case(for spice) so you don't like too many files there when the job has many cases, I am okay anyway, will remove it. Thanks.

vivianQizhu:
@yanan-fu @zhencliu Thanks for your review, updated per your comments, would you help re-review it?

virttest/qemu_devices/qcontainer.py
```diff
@@ -3201,3 +3201,27 @@ def numa_hmat_cache_define_by_params(self, nodeid, params):
...
+        when inline is False
+
+        :param secret_id: the id of the secret object
+        :param data: password, in raw or base64 format
+        :param secret_format: raw or base64
+        :param inline: append data=xxx if True and file=xxx if False
+        :return: the secret object device QObject
+        """
+        params = {'id': secret_id}
+        if secret_format != 'raw':
```
yanan-fu, line 3217: How about add `format` in any case?  Does not depends on the QEMU default behavior.

vivianQizhu: @yanan-fu Em...it also make sense, but I hope it could emulate the manual test, and maybe to discover the potential bz like the default one is not take effect.

yanan-fu: To discover the potential bz like the default one does not take effect, i would like to use:
`def secret_object(self, secret_id, data, secret_format, inline=True):`

Do not update format if no `secret_format` was specified in the calling. How do you think ?

vivianQizhu: > To discover the potential bz like the default one does not take effect, i would like to use:
> `def secret_object(self, secret_id, data, secret_format, inline=True):`

Do you mean `def secret_object(self, secret_id, data, secret_format=None, inline=True):`? If so, I am okay, will update it.

virttest/qemu_devices/qcontainer.py
```diff
@@ -3201,3 +3201,27 @@ def numa_hmat_cache_define_by_params(self, nodeid, params):
                     aobject=aobject, backend='hmat_type'))
 
         return hmat_cache_devs
+
+    def secret_object(self, secret_id, data, secret_format=None, inline=True):
```
yanan-fu, line 3205: How about  `secret_object_define_by_params` ?  Same format with other fuctions.

Others LGTM

vivianQizhu: > How about `secret_object_define_by_params` ? Same format with other fuctions.

Thanks @yanan-fu Actually from the meaning it should be `secret_object_define_by_variables`. If you check the above definitions, some devices have `xxx_define_by_variables` and `xxx_define_by_params`, `xxx_define_by_params` take params object(dictionary) and convert params to some variables then calling `xxx_define_by_variables`.
At the beginning I use `xxx_define_by_variables` but finally I think not necessary since only this one function so do not need to distinguish, if the `secret_object` seems not very objective, what do you think about `generate_secret_object`, or you prefer `xxx_define_by_variables`? I can make the change accordingly.

zhencliu:
LGTM

virttest/qemu_devices/qcontainer.py
yanan-fu: I prefer `xxx_define_by_variables`  a bit more ~ Thanks!

virttest/qemu_vm.py
```diff
@@ -1043,7 +1043,19 @@ def set_value(opt_string, key, fallback=None):
...
-            set_value("password=%s", "spice_password", "disable-ticketing=on")
+            password = optget("spice_password")
+            secret_cmdline = ''
+            if password:
+                if ("password-secret" not in
+                        devices.execute_qemu("-spice help")):
+                    spice_opts.append('password=%s' % password)
+                else:
+                    secret_id = 'spice_sec0'
+                    secret_cmdline = devices.secret_object(
```
yanan-fu, line 1054: Missed,  `secret_object_define_by_varibles`

vivianQizhu: @yanan-fu Ops, failed to add the change, just updated, thanks.

yanan-fu:
Call for latest test result with this patch, thanks @vivianQizhu 

Others LGTM

vivianQizhu:
Thanks @yanan-fu Test on both old and new versions, all passed.
 (1/1) Host_RHEL.m8.u5.product_av.spice.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.4.0.x86_64.io-github-autotest-qemu.boot: PASS (44.27 s)
```
    -object secret,id=spice_sec0,data=123456 \
    -spice port=3115,password-secret=spice_sec0,addr=0,image-compression=auto_glz,zlib-glz-wan-compression=auto,streaming-video=all,agent-mouse=on,playback-compression=on,ipv4=on,ipv6=off
```

yanan-fu (approved):
Ack

vivianQizhu:
Thanks @yanan-fu @zhencliu

PR #3010, opened 2021-04-21
Add func to get first disk

dzhengfy:
This function is to get the first disk on the host or guest, especially
to detect the disk is 'vda' or 'sda'.

### DIFF
--- a/virttest/utils_disk.py
+++ b/virttest/utils_disk.py
@@ -1102,6 +1102,23 @@ def get_parts_list(session=None):
     return parts
 
 
+def get_first_disk(session=None):
+    """
+    Get the first disk device on host or guest
+
+    :param session: the vm session
+    :return: str, the disk device, like 'vda' or 'sda'
+    """
+    first_disk = ""
+    disks = get_parts_list(session=session)
+    for disk in disks:
+        pattern = re.compile('[0-9]+')
+        if not pattern.findall(disk):
+            first_disk = disk
+            break
+    return first_disk
+
+
 def get_disk_by_serial(serial_str, session=None):
     """
     Get disk by serial in VM or host


chunfuwen:
There is similar method in virttest/libvirt_vm.py
```
    def get_first_disk_devices(self):
        """
        Get vm's first disk type block devices.
        """
        disk = {}
        options = "--details"
        result = virsh.domblklist(self.name, options, ignore_status=True,
                                  uri=self.connect_uri)
        blklist = result.stdout_text.strip().splitlines()
        if result.exit_status != 0:
            logging.info("Get vm devices failed.")
        else:
            blklist = blklist[2:]
            linesplit = blklist[0].split(None, 4)
            disk = {'type': linesplit[0],
                    'device': linesplit[1],
                    'target': linesplit[2],
                    'source': linesplit[3]}
        return disk
```

virttest/utils_disk.py
```diff
@@ -1102,6 +1102,23 @@ def get_parts_list(session=None):
     return parts
 
 
+def get_first_disk(session=None):
+    """
+    Get the first disk device on host or guest
+
+    :param session: the vm session
```
Yingshun, line 1109: only support vm session? what about remote host session?

dzhengfy: Currently I have not test case to test it if I support remote host. So I would like to enhance it in future if we have the requirement. Thanks.

dzhengfy:
> There is similar method in virttest/libvirt_vm.py
> 
> ```
>     def get_first_disk_devices(self):
>         """
>         Get vm's first disk type block devices.
>         """
>         disk = {}
>         options = "--details"
>         result = virsh.domblklist(self.name, options, ignore_status=True,
>                                   uri=self.connect_uri)
>         blklist = result.stdout_text.strip().splitlines()
>         if result.exit_status != 0:
>             logging.info("Get vm devices failed.")
>         else:
>             blklist = blklist[2:]
>             linesplit = blklist[0].split(None, 4)
>             disk = {'type': linesplit[0],
>                     'device': linesplit[1],
>                     'target': linesplit[2],
>                     'source': linesplit[3]}
>         return disk
> ```

Yes, I notice this function. But sometimes we also need to support both host and guest. That function does not meet the needs. What do you think?

Yingshun (approved):
LGTM.

virttest/utils_disk.py
Yingshun: Ok

Yingshun: BTW, disks can be listed by avocado.utils.disk.get_disks() on localhost. So I wonder if we can get disks from lsblk command directly: 
for example:
```
json_result = process.run('lsblk --json')
json_data = json.loads(json_result.stdout_text)
[str(disk['name']) for disk in json_data['blockdevices'] if disk['type']=="disk"]
```

chloerh (approved):

chunfuwen (approved):
lgtm

kamilvarga (commented):
LGTM.

PR #3011, opened 2021-04-21
do not use /tmp

dzhengfy:
For gluster volume setup, /tmp sometimes causes unexpected error because
tmpfs (/tmp sometimes use) filesystem can not support the function. So
we decide to use other directory instead.
For iscsi block,  dd 10G sometimes will fail due to no enough space on /tmp. 
So it is also changed.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -534,7 +534,7 @@ def setup_or_cleanup_iscsi(is_setup, is_login=True,
     :param chap_passwd: CHAP authentication password
     :return: iscsi device name or iscsi target
     """
-    tmpdir = data_dir.get_tmp_dir()
+    tmpdir = data_dir.get_data_dir()
     emulated_path = os.path.join(tmpdir, emulated_image)
     emulated_target = ("iqn.%s.com.virttest:%s.target" %
                        (time.strftime("%Y-%m"), emulated_image))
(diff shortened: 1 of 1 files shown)


chunfuwen:
Make sense, Better have a simple test result

dzhengfy:
Test result see https://github.com/autotest/tp-libvirt/pull/3455#issuecomment-825477905

virttest/utils_test/libvirt.py
```diff
@@ -2567,9 +2567,8 @@ def set_vm_disk(vm, params, tmp_dir=None, test=None):
                         (src_disk_format, disk_format, blk_source, dist_img))
         else:
             # create another disk without convert
-            disk_cmd = "qemu-img create -f %s /mnt/%s 10M" % (src_disk_format,
-                                                              dist_img)
-
+            disk_cmd = "qemu-img create -f %s -o preallocation=full /mnt/%s 10M" % (src_disk_format,
```
chunfuwen, line 2570: is there is special case that need this option : preallocation=full?
Better we keep it as pass in parameter,rather than hardcode his  option in  since there are quite numbers of cases dependent on this method.

kylazhang (approved):

virttest/utils_test/libvirt.py
dzhengfy: @chunfuwen this comes from https://github.com/avocado-framework/avocado-vt/pull/2914. The feature owner explained to me that this is to make sure a 'no space' error message is able to be displayed in logging file when 'qemu-img create'. Without it,  'qemu-img create' will not report an error and then vm starting will hang and no error message is displayed.

chunfuwen (approved):
lgtm

PR #3014, opened 2021-04-22
Add support to check if backup job canceled

lento-sun:
There are 3 parts in this commit:
1. Add a new BackupCanceledError exception.
2. Add 'extra' in virsh.domjobinfo to allow it take more params
such as '--complete' when backup job canceled.
3. Add a 'is_backup_canceled' method to check if backup job
just canceled.

### DIFF
--- a/virttest/utils_backup.py
+++ b/virttest/utils_backup.py
@@ -55,6 +55,19 @@ class BackupTLSError(BackupError):
         return "Backup TLS failure: %s" % self.error_info
 
 
+class BackupCanceledError(BackupError):
+
+    """
+    Class of backup canceled exception
+    """
+
+    def __init__(self):
+        BackupError.__init__(self)
+
+    def __str__(self):
+        return "Backup job canceled"
+
+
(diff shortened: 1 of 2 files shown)


virttest/utils_backup.py
```diff
@@ -55,6 +55,19 @@ def __str__(self):
         return "Backup TLS failure: %s" % self.error_info
 
 
+class BackupCanceledError(BackupError):
+
+    """
+    Class of backup canceled exception
+    """
+
+    def __init__(self):
```
chunfuwen, line 64: whether consider add parameter e.g message to allow customized exception message

kamilvarga (approved):
LGTM

virttest/utils_backup.py
lento-sun: It's NOT always an error when backup canceled, for examlpe, we can use 'domjobabort' to cancel the backup. So there is no error message produced by libvirt or kernel that need to be captured. And other user message is not worth to be put here, but should be put in somewhere such as logging.debug.

chunfuwen (approved):
lgtm

PR #3016, opened 2021-04-23
Add the multiqueue check and set functions

yalzhang:
To check if the multiqueue setting is effective, run "ethtool -l" on
the vm. And set it by "ethtool -L". Add these 2 functions to check and
set the multiqueue in the vm.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -4351,3 +4351,60 @@ def delete_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
     return utils_misc.cmd_status_output(tmux_cmd, shell=True, verbose=True,
                                         ignore_status=ignore_status,
                                         session=session)
+
+
+def get_channel_info(session, interface):
+    """
+    Get channel parameters of interface in vm.
+
+    :param session: the guest session
+    :param interface: the interface name shows on the guest, for example: eth0
+    :return: 2 dictionaries, one is the maximum values for hardware, the other is current value,
+    in format as {"RX": "n/a", "TX": "n/a", "Other": "n/a", "Combined": "1"}
+    """
+    cmd = "ethtool -l %s" % interface
+    s, o = session.cmd_status_output(cmd)
+    if s:
+        logging.error("Get channel parameters for vm failed:%s" % o)
+        return {}, {}
+    maximum = {}
+    current = {}
+    # Just for temp
+    settings = {}
+    for line in o.splitlines():
+        if line.count("maximums"):
+            settings = maximum
+        elif line.count("Current"):
+            settings = current
+        parameter = line.split(':')[0].strip()
+        if parameter in ["RX", "TX", "Other", "Combined"]:
+            settings[parameter] = line.split(':')[1].strip()
+    return maximum, current
+
+
+def set_channel(session, interface, parameter, value):
+    """
+    Set channel parameters for interface, then check if it is set successfully
+
+    :param session: the session for the vm
+    :param interface: the interface name shows on the guest, for example: eth0
+    :param parameter: the parameter to be set, it can be: "rx", "tx", "other" or "combined"
+    :param value: the value will be set for the parameter
+    :return: False or True to indicate if the the parameter is set successfully
+    """
+    cmd = "ethtool -L %s %s %s" % (interface, parameter, value)
+    logging.debug("The cmd to set the channel info '%s'" % cmd)
+    s, o = session.cmd_status_output(cmd)
+    if s:
+        logging.error("Setting %s to %s failed:%s", parameter, value, o)
+        return False
+    _, current = get_channel_info(session, interface)
+    logging.debug("After set, the current value is %s" % current)
+    try:
+        if int(current['Combined']) == int(value):
+            return True
+        else:
+            logging.error("Setting passed, but checking failed:%s", current)
+    except KeyError:
+        logging.error("No 'Combined' found in the channel info")
+    return False


virttest/utils_test/libvirt.py
```diff
@@ -2092,6 +2092,51 @@ def create_nwfilter_xml(params):
         raise exceptions.TestFail("Fail to create nwfilter XML: %s" % detail)
 
 
+def get_channel_info(session, interface):
+    """
+    Get channel parameters of interface in vm.
+    """
```
dzhengfy, line 2098: Complete docstring for params and return

virttest/utils_test/libvirt.py
```diff
@@ -2092,6 +2092,51 @@ def create_nwfilter_xml(params):
...
+    cmd = "ethtool -l %s" % interface
+    s, o = session.cmd_status_output(cmd)
+    if s:
+        logging.info("Get channel parameters for vm failed:%s" % o)
```
dzhengfy, line 2102: logging.error() is better because this means the function return errors.

virttest/utils_test/libvirt.py
```diff
@@ -2092,6 +2092,51 @@ def create_nwfilter_xml(params):
...
+        parameter = line.split(':')[0].strip()
+        if parameter in ["RX", "TX", "Other", "Combined"]:
+            settings[parameter] = line.split(':')[1].strip()
+    return maximum, current
+
+
+def setting_channel(session, interface, parameter, value):
+    """
+    Setting channel parameters for interface.
+    """
```
dzhengfy, line 2122: Complete the docstring for params and return

virttest/utils_test/libvirt.py
```diff
@@ -2092,6 +2092,51 @@ def create_nwfilter_xml(params):
...
+    cmd = "ethtool -L %s %s %s" % (interface, parameter, value)
+    logging.debug("The cmd to set the channel info %s" % cmd)
```
dzhengfy, line 2124: Quoting the '%s' will be better to distinguish the command to be executed.

virttest/utils_test/libvirt.py
```diff
@@ -2092,6 +2092,51 @@ def create_nwfilter_xml(params):
...
+    s, o = session.cmd_status_output(cmd)
+    if s:
+        logging.debug("Setting %s to %s failed:%s", parameter, value, o)
```
dzhengfy, line 2127: logging.error

virttest/utils_test/libvirt.py
```diff
@@ -2092,6 +2092,51 @@ def create_nwfilter_xml(params):
...
+        return False
+    maximum, current = get_channel_info(session, interface)
```
dzhengfy, line 2129: If you do not need 'maximum', it is better to use '_, current = xxx'

virttest/utils_test/libvirt.py
```diff
@@ -2092,6 +2092,51 @@ def create_nwfilter_xml(params):
...
+    logging.debug("After set, the current value is %s" % current)
+    try:
+        if int(current['Combined']) == int(value):
```
dzhengfy, line 2132: 'Combined' should be 'parameter' instead?

virttest/utils_test/libvirt.py
```diff
@@ -2092,6 +2092,51 @@ def create_nwfilter_xml(params):
...
+            return True
+    except KeyError:
+        pass
```
dzhengfy, line 2135: It is better to logging.error() here instead of simply ignoring the exception.

dzhengfy (commented):
Please do not add new functions in virttest/utils_test/libvirt.py  any more because it is too big and hard to maintain. 
If your new functions are using virsh command or vm xml related, please add them to utils_libvirt/libvirt_xxx.py.
Otherwise, please add them to virttest/utils_xxx.py. 
It seems you belong to the later case. So we suggest you add new functions to utils_net.py. But utils_net.py is also too big, so you are free to create a new utils_xxx.py for your new functions.

virttest/utils_test/libvirt.py
yalzhang: updated, Thank you!

virttest/utils_test/libvirt.py
yalzhang: updated, Thank you!

virttest/utils_test/libvirt.py
yalzhang: updated, Thank you!

virttest/utils_test/libvirt.py
yalzhang: updated, Thank you!

virttest/utils_test/libvirt.py
yalzhang: updated, Thank you!

virttest/utils_test/libvirt.py
yalzhang: updated

virttest/utils_test/libvirt.py
yalzhang: no, 'Combined' is the key of the dict "current".

virttest/utils_test/libvirt.py
yalzhang: updated, Thank you!

virttest/network.py
```diff
@@ -0,0 +1,57 @@
...
+
+    :param session: the guest session
+    :param interface: the interface name shows on the guest, for example: eth0
+    :return: 2 dictionaries, one is the maximum values for hardware, the other is current value,
+    in format as {"RX": "n/a", "TX": "n/a", "Other": "n/a", "Combined": "1"}
+    """
+    cmd = "ethtool -l %s" % interface
+    s, o = session.cmd_status_output(cmd)
+    if s:
+        logging.info("Get channel parameters for vm failed:%s" % o)
```
dzhengfy, line 16: It seems you forgot to update it to 'error' :D

virttest/network.py
```diff
@@ -0,0 +1,57 @@
...
+    :param interface: the interface name shows on the guest, for example: eth0
+    :param parameter: the parameter to be set, it can be: "rx", "tx", "other" or "combined"
+    :param value: the value will be set for the parameter
+    :return: False or True to indicate if the the parameter is set successfully
+    """
+    cmd = "ethtool -L %s %s %s" % (interface, parameter, value)
+    logging.debug("The cmd to set the channel info '%s'" % cmd)
+    s, o = session.cmd_status_output(cmd)
+    if s:
+        logging.debug("Setting %s to %s failed:%s", parameter, value, o)
```
dzhengfy, line 47: It is accurate to use logging.error()

virttest/network.py
```diff
@@ -0,0 +1,57 @@
+import logging
```
dzhengfy, line 1: Could you change the file name to virttest/utils_network.py to align the existing naming style?

dzhengfy (commented):
Others LGTM

virttest/network.py
yalzhang: Hi @dzhengfy , I'm thinking that even the utils_net.py is very big, but the functions in the file are independent from each other, so it may be better to add these 2 functions into that file, it will be more convenient as they both are related with network. What do you think? Or I should create a new file?

virttest/network.py
```diff
@@ -0,0 +1,57 @@
...
+        logging.debug("Setting %s to %s failed:%s", parameter, value, o)
+        return False
+    _, current = get_channel_info(session, interface)
+    logging.debug("After set, the current value is %s" % current)
+    try:
+        if int(current['Combined']) == int(value):
+            return True
+    except KeyError:
+        logging.error("No 'Combined' found in the channel info")
+    logging.error("Setting passed, but checking failed:%s", current)
```
kamilvarga, line 56: The last two lines (56, 57) seem to be applicable only when 'Combined' is not found in the channel info. Wouldn't be better to move them to the same exception? If any other exception happens the last two lines will not be applicable anyway.

yalzhang: Yes, you are right, update to be as below:   
   try:
        if int(current['Combined']) == int(value):
            return True
        else:
            logging.error("Setting passed, but checking failed:%s", current)
    except KeyError:
        logging.error("No 'Combined' found in the channel info")
    return False

virttest/network.py
yalzhang: updated

virttest/network.py
yalzhang: updated

dzhengfy (approved):
LGTM

kamilvarga (approved):
LGTM.

dzhengfy:
See the test result  https://github.com/autotest/tp-libvirt/pull/3454#issuecomment-826283463

PR #3022, opened 2021-04-25
Add function check_dmesg_output

chloerh:
Allow users to check whether dmesg output contains certain pattern.

### DIFF
--- a/virttest/utils_sys.py
+++ b/virttest/utils_sys.py
@@ -1,0 +1,39 @@
+"""
+Virtualization test utility functions.
+
+:copyright: 2021 Red Hat Inc.
+"""
+
+import logging
+import re
+
+from avocado.utils import process
+
+
+def check_dmesg_output(pattern, expect=True, session=None):
+    """
+    Check whether certain pattern exists in dmesg.
+
+    :param pattern: pattern to search in dmesg
+    :param expect: True if expect to exist, False if not
+    :param session: session of vm to be checked
+    :return: True if result met expectation, False if not met
+    """
+    dmesg_cmd = 'dmesg'
+    func_get_dmesg = session.cmd if session else process.run
+    dmesg = func_get_dmesg(dmesg_cmd)
+
+    prefix = '' if expect else 'Not '
+    logging.info('%sExpecting pattern: "%s".', prefix, pattern)
+
+    # Search for pattern
+    found = bool(re.search(pattern, dmesg))
+    log_content = ('' if found else 'Not') + 'Found "%s"' % pattern
+    logging.debug(log_content)
+
+    if found ^ expect:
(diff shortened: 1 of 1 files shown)


chloerh:
JOB LOG    : /root/avocado/job-results/job-2021-04-25T16.22-abf8fce/job.log
 (1/1) type_specific.local.memory_hotplug.report_failure: PASS (27.41 s)
JOB TIME   : 27.69 s

chloerh:
@Yingshun Help help~

virttest/utils_sys.py
```diff
@@ -0,0 +1,39 @@
+"""
+Virtualization test utility functions.
+
+:copyright: 2008-2009 Red Hat Inc.
```
Yingshun, line 4: Can you please update it to the latest? Others look good to me.

chloerh: Sure

Yingshun (approved):

dzhengfy (approved):
LGTM

pevogam:
Why isn't this placed in utils_misc next to `utils_misc.verify_dmesg()`? It could even be integrated with the functionality there which could reuse some of the pattern matching.

pevogam:
Also, why can't it immediately reuse or extend `avocado.utils.dmesg`? To me it seems like yet another module raising complexity.

chloerh:
@pevogam Thank u for bringing this useful module to us. But sadly we're still running our test with lts branch, and I have checked that the commit of dmesg module is not included in 82lts. But it still could be very useful to us after we switch to the next lts avocado. Thanks again.

pevogam:
Yes that's understandable, we also use the Avocado 82 LTS because of the tii many changes happening there all the time. Then would you be willing to commit to later migrating the dmesg functionality or dropping it in favor of the upstream one?

chloerh:
@pevogam For future tests, I would recommend dmesg utils of avocado. 
Sorry but I don't quite understand what you meant by 'commit to later migrating the dmesg functionality' :sweat_smile:  Did you mean migrate from avocado master to 82lts?

pevogam:
No I mean would you want to migrate the functionality to Avocado if it extends what is already available there or if not add a TODO in the current function's docstring that is should be dropped in the future with the next LTS and a function pointer there?

chloerh:
> No I mean would you want to migrate the functionality to Avocado if it extends what is already available there or if not add a TODO in the current function's docstring that is should be dropped in the future with the next LTS and a function pointer there?

@pevogam I'd prefer the second one, since we're currently running with 82lts and the function is required by one of our tests.

pevogam:
Sounds good enough, a TODO will be good then so that we keep track of such reimplementations.

PR #3023, opened 2021-04-25
migrate_template: Support more migration options

fangge1212:

### DIFF
--- a/virttest/migration_template.py
+++ b/virttest/migration_template.py
@@ -1,5 +1,6 @@
 import logging
 import functools
+import os
 
 from enum import Enum
 
@@ -9,7 +10,7 @@ from avocado.utils import distro
 from avocado.utils import path as utils_path
 
 from virttest import virsh, migration, remote
-from virttest import utils_iptables, utils_selinux
+from virttest import utils_iptables, utils_selinux, utils_misc
 
 from virttest.libvirt_xml import vm_xml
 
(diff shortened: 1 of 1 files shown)


virttest/migration_template.py
```diff
@@ -598,6 +736,30 @@ def _set_vm_disk(self, cache="none"):
...
+
+        """
+        logging.debug("Create disk image on dest host before migration")
+        if self.migrate_flags & (VIR_MIGRATE_NON_SHARED_DISK|VIR_MIGRATE_NON_SHARED_INC):
+            all_vm_disks = self.main_vm.get_blk_devices()
+            for disk in list(itervalues(all_vm_disks)):
+                disk_type = disk.get("type")
+                disk_path = disk.get("source")
+                disk_size = utils_misc.get_image_info(disk_path).get("vsize")
+                disk_format = utils_misc.get_image_info(disk_path).get("format")
```
dzhengfy, line 756: We might invoke get_image_info() once and then read the returned dict for 'vsize' and 'format'. That will be better.

fangge1212: Updated

dzhengfy (approved):
LGTM

virttest/migration_template.py
```diff
@@ -365,16 +431,25 @@ def _post_start_vm(self):
...
-        4.Do live migration
-        5.Check migration result: succeed or fail with expected error
-        6.For postcopy migration: check postcopy event
-        7.Do post migration check: check vm state, uptime, network
+        5.Do live migration
+        6.Check migration result: succeed or fail with expected error
+        7.For postcopy migration: check postcopy event
+        8.Do post migration check: check vm state, uptime, network
         """
+        # Create disk image on dest host if --copy-storage-all/inc is used
```
lento-sun, line 446: Here why we need to create some disk images on target host?
In my previous test, only create a dir on target host is enough and the source image will be copied to target dir, isn't it?

fangge1212: 1. Yes, if we don't create an image on dest host manually, libvirt will create it automatically. But there was a bug in this case: [Bug 1248996 ](https://bugzilla.redhat.com/show_bug.cgi?id=1248996)- Larger size of qcow2 image file on target host after migration with non-shared storage with full disk copy. The consequence of the bug is that full image is copied to dest host even if only part of the disk is allocated, which waste a lot of time during test.

2. In layered product, it pre-creates image on dest host by itself, not by libvirt.

Based on above two reasons, I add "create image" step in automation script.

lento-sun: thx for the explanation, LGTM

lento-sun (approved):

virttest/migration_template.py
```diff
@@ -586,17 +727,40 @@ def _set_vm_disk(self, cache="none"):
...
-                logging.debug("Prepare nfs backed disk in vm xml")
-                for vm in self.vms:
-                    libvirt.update_vm_disk_source(vm.name,
-                                                  self.nfs_mount_dir)
-                    libvirt.update_vm_disk_driver_cache(vm.name,
-                                                        driver_cache=cache)
-            else:
-                # TODO:other storage types
-                self.test.cancel("Only nfs storage is supported for now")
+        if self.storage_type == 'nfs':
```
chunfuwen, line 730: I observed more and more methods are introduced to this template class as more requirements is coming. Eventually it will becomes one big bang class, which will make it harder to be readable,and maintainable.
I think whether we can split them into a few class,such as storagetemplate.class, networktemplate.class, then use migration_template inherit them.

fangge1212: The methods in migration_template.py are all well designed for good readability and easy maintenance. For migration feature itself, there are only two categories in this aspect: migration with storage copy, migration without storage copy. Other features like network/cpu/etc won't be handled in migration_template.py.

chunfuwen (approved):
lgtm

PR #3025, opened 2021-04-27
modify_vm_iface:delete iface source before assigning new value

chloerh:
The old value could still exist after assigning new value, therefore
remove it before assigning new value.
Move log of updated iface xml for all types of operations.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -3543,6 +3543,7 @@ def modify_vm_iface(vm_name, oper, iface_dict, index=0, virsh_instance=virsh):
     if iface_link:
         iface.link_state = iface_link
     if iface_source:
+        iface.del_source()
         iface.source = eval(iface_source)
     if iface_target:
         iface.target = eval(iface_target)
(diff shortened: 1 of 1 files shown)


chloerh:
Before:
```
<interface type="direct">
      <mac address="52:54:00:6a:3b:13" />
      <source dev="eno1" mode="bridge" network="default" />
      <model type="virtio" />
      <address bus="0x01" domain="0x0000" function="0x0" slot="0x00" type="pci" />
    </interface>

```
After
```
<interface type="direct">
      <mac address="52:54:00:6a:3b:13" />
      <model type="virtio" />
      <address bus="0x01" domain="0x0000" function="0x0" slot="0x00" type="pci" />
    <source dev="eno1" mode="bridge" /></interface>

```

chloerh:
@yalzhang

yalzhang (approved):

virttest/utils_test/libvirt.py
```diff
@@ -3543,6 +3543,7 @@ def modify_vm_iface(vm_name, oper, iface_dict, index=0, virsh_instance=virsh):
     if iface_link:
         iface.link_state = iface_link
     if iface_source:
+        iface.del_source()
```
kylazhang, line 3546: I suggest to have separate param like del_source with boolean to check if need to del, then using
```
if del_source:
    iface.del_source()
```

chloerh: Actually, after second thought, I decide to close this pr. Thank u for the review ~

PR #3026, opened 2021-04-27
tmp_fix: Enable incremental backup in RHEL9

lento-sun:
Incremental backup not enabled in rhel9 by default. But it's
already enabled in rhel8. So we need this tmp fix to enable it.
And this fix won't effect rhel8 since the extra xml has no
side effect.


lento-sun:
when bz1953939 fixed, this draft pr can be closed.

lento-sun:
bz fixed, no need to apply this tmp patch

lento-sun:
bz fixed, no need to apply this tmp patch

PR #3027, opened 2021-04-27
libvirt_misc: new function to parse result

dzhengfy:
Some virsh commands have output in same or similar format. So this
function is to used as an unified one to run and parse the result
to a dict. For example, the commands are virsh.vcpupin, virsh.emulatorpin,
virsh.iothreadinfo and so on.

### DIFF
--- a/virttest/utils_libvirt/libvirt_misc.py
+++ b/virttest/utils_libvirt/libvirt_misc.py
@@ -1,0 +1,25 @@
+"""
+Virtualization test - utility functions for libvirt
+
+:copyright: 2021 Red Hat Inc.
+"""
+
+import re
+import logging
+
+
+def convert_to_dict(content, pattern=r'(\d+) +(\S+)'):
+    """
+    Put the content into a dict according to the pattern.
+
+    :param content: str, the string to be parsed
+    :param pattern: str, regex for parsing the command output
+    :return: dict, the dict contains matched result
+    """
+
+    info_dict = {}
+    info_list = re.findall(pattern, content, re.M)
+    for info in info_list:
+        info_dict[info[0]] = info[1]
+    logging.debug("The dict converted is:\n%s", info_dict)
+    return info_dict


chunfuwen:
@dzhengfy 
Better not use 'misc' as filename, where people tends to put everything into this file.
How about libvirt_second.py ? since it is more close libvirt.py, which is freezed.

virttest/utils_libvirt/libvirt_misc.py
```diff
@@ -0,0 +1,28 @@
+"""
+Virsh command related utility functions which do not belong to special
```
Yingshun, line 2: I wonder if we can remove the limitation of 'virsh command' here? If we decide to create a libvirt_misc module, I think it can be used  more widely. How do you think?

virttest/utils_libvirt/libvirt_misc.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    Execute specified virsh command and return parsed result
+
+    :param func_name: The virsh command function to be executed
+    :param vm_name: str, The vm name
+    :param pattern: str, regex for parsing the command output
+    :return: dict, the command result
+    """
+
+    ret = func_name(vm_name, ignore_status=False, debug=True)
```
Yingshun, line 21: In some cases, we may need some special options for the command and need to check the exist status of this command, so how about just parsing the command's output? I mean, the command is executed in other places, and the stdout will be parsed by this function. What do you think?

virttest/utils_libvirt/libvirt_misc.py
chunfuwen, line 2: Better add copyright declare for newly created file

dzhengfy: Currently the parsing logic can only meet certain output formats. I am ok to remove the words 'virsh command' here, but for the function created below, I need to add docstring to say currently this function might only support certain output formats, for example, some virsh command results. What do you think?

virttest/utils_libvirt/libvirt_misc.py
dzhengfy: 1. Originally I want this function can provide a single entry to execute some virsh functions, and the user only needs to provide a virsh function name to invoke it. That is a simple usage.
2. This function is not designed for those commands which the user needs to manually check their execution results, so `ignore_status=False,` is explicitly defined. 
3. This function is not designed for those commands which need more options. So yes, it has limitations.
4. This is only designed for certain virsh commands which have similar output formats. If it is extended to bigger scope, I am not sure how to write the following list parsing codes, like     `  info_dict[info[0]] = info[1]` is hard to meet other scenarios.
@Yingshun

dzhengfy: updated

virttest/utils_libvirt/libvirt_misc.py
dzhengfy: updated

Yingshun (approved):

chunfuwen (approved):
lgtm
If one example output is given, that's better

PR #3034, opened 2021-04-28
postprocess: Don't interrupt the procedure by the failures of single vm

luckyh:


yanglei-rh:
LGTM
Using the current code for testing can effectively solve the problem: Can't get the fd that qemu process opened!

yanglei-rh (approved):

PR #3035, opened 2021-04-28
Add 2 functions to create/delete linux bridge

yalzhang:
Add 2 functions to create/delete linux bridge on the host. And it will
also add or delete a physical interface to the bridge. Using the tmux to
execute the ip commands to avoid the broken of the host network.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -4408,3 +4408,61 @@ def set_channel(session, interface, parameter, value):
     except KeyError:
         logging.error("No 'Combined' found in the channel info")
     return False
+
+
+def create_linux_bridge_tmux(linux_bridge_name, iface_name=None, ignore_status=False):
+    """
+    Create linux bridge and connect a physical interface to the bridge via tmux command on local host.
+    Note: If iface_name is specified, it should be the one in current connection. Or you will break current
+    connection as this function will bring up the bridge and switch the network traffic to this bridge which
+    share the same mac with iface_name.
+    1. use tmux to avoid the network broken;
+    2. add a physical interface to the bridge, and set the bridge up;
+
+    :param linux_bridge_name: The linux bridge
+    :param iface_name: the physical interface to be attached to the bridge, if None, no need to add physical
+    interface to the bridge
+    :param ignore_status: Whether to raise an exception when command fails
+    :return: bridge created or raise exception
+    """
+    # Create bridge
+    br_path = "/sys/class/net/%s" % linux_bridge_name
+    if not utils_package.package_install(['tmux', 'dhcp-client', 'net-tools']):
+        raise exceptions.TestError("Failed to install the required packages.")
+    if os.path.exists(br_path):
+        s, o = delete_linux_bridge_tmux(linux_bridge_name, iface_name)
+        if s:
+            raise exceptions.TestError("Create bridge fail as there is already interface named '%s' on the host "
+                                       "and can not delete with error: %s" % (linux_bridge_name, o))
+    if iface_name:
+        cmd = 'tmux -c "ip link add name {0} type bridge; ip link set {1} up;' \
+              ' ip link set {1} master {0}; ip link set {0} up;' \
+              ' pkill dhclient; sleep 6; dhclient {0}; ifconfig {1} 0"'.format(linux_bridge_name, iface_name)
+    else:
+        cmd = 'ip link add %s type bridge' % linux_bridge_name
+    return utils_misc.cmd_status_output(cmd, shell=True, verbose=True, ignore_status=ignore_status)
+
+
+def delete_linux_bridge_tmux(linux_bridge_name, iface_name=None, ignore_status=False):
+    """
+    Delete the linux bridge on the host, and recover the network on the physical interface
+
+    :param linux_bridge_name: The linux bridge
+    :param iface_name: the physical interface has been attached to the bridge, if None, it means no interface
+    is attached to the bridge
+    :param ignore_status: Whether to raise an exception when command fails
+    :return: bridge deleted or raise exception
+    """
+    # Delete the linux bridge
+    br_path = "/sys/class/net/%s" % linux_bridge_name
+    if not utils_package.package_install(['tmux', 'dhcp-client', 'procps-ng', 'net-tools']):
+        raise exceptions.TestError("Failed to install the required packages.")
+    if not os.path.exists(br_path):
+        logging.info("There is no bridge named '%s' on the host" % linux_bridge_name)
+        return
+    if iface_name:
+        cmd = 'tmux -c "ip link set {1} nomaster; ip link delete {0}; pkill dhclient; ' \
+              'sleep 5; dhclient {1}"'.format(linux_bridge_name, iface_name)
+    else:
+        cmd = 'ip link delete %s' % linux_bridge_name
+    return utils_misc.cmd_status_output(cmd, shell=True, verbose=True, ignore_status=ignore_status)


virttest/utils_net.py
```diff
@@ -4351,3 +4351,46 @@ def delete_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
...
+
+    :param linux_bridge_name: The linux bridge
+    :param iface_name: the physical interface to be attached to the bridge
+    :param ignore_status: Whether to raise an exception when command fails
+    :return: bridge created or raise exception
+    """
+    # Create bridge
+    if not utils_package.package_install(['tmux', 'dhcp-client']):
+        raise exceptions.TestError("Failed to install the required packages.")
+    current_iface = process.run("ls /sys/class/net", shell=True, verbose=True, ignore_status=True).stdout_text.split()
```
chunfuwen, line 4370: ignore_status=True or should be ignore_status=False ?

virttest/utils_net.py
```diff
@@ -4351,3 +4351,46 @@ def delete_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
...
+    :param iface_name: the physical interface to be attached to the bridge
+    :param ignore_status: Whether to raise an exception when command fails
+    :return: bridge deleted or raise exception
+    """
+    # Delete the linux bridge
+    if not utils_package.package_install(['tmux', 'dhcp-client', 'procps-ng']):
+        raise exceptions.TestError("Failed to install the required packages.")
+    cmd = 'tmux -c "ip link set {1} nomaster; ip link delete {0}; pkill dhclient; ' \
+          'sleep 5; dhclient {1}"'.format(linux_bridge_name, iface_name)
+    return utils_misc.cmd_status_output(cmd, shell=True, verbose=True,
```
chunfuwen, line 4395: Better return True or False ,rather than standard output

virttest/utils_net.py
yalzhang: have updated to:
"br_path = "/sys/class/net/%s" % linux_bridge_name
if os.path.exists(br_path):
    raise exceptions.TestError(xxx)"
So the process.run() is not needed now.

virttest/utils_net.py
yalzhang: Thank you for the suggestion.
update to:
s, o = utils_misc.cmd_status_output(cmd, shell=True, verbose=True, ignore_status=ignore_status)
return s

virttest/utils_net.py
```diff
@@ -4351,3 +4351,50 @@ def delete_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
...
+    br_path = "/sys/class/net/%s" % linux_bridge_name
+    if not utils_package.package_install(['tmux', 'dhcp-client']):
+        raise exceptions.TestError("Failed to install the required packages.")
+    if os.path.exists(br_path):
+        raise exceptions.TestError("Create bridge error as there is already an interface named '%s' on the host"
+                                   % linux_bridge_name)
+    cmd = 'tmux -c "ip link add name {0} type bridge; ip link set {1} up;' \
+          ' ip link set {1} master {0}; ip link set {0} up;' \
+          ' pkill dhclient; sleep 6; dhclient {0}; ifconfig {1} 0"'.format(linux_bridge_name, iface_name)
+    s, o = utils_misc.cmd_status_output(cmd, shell=True, verbose=True, ignore_status=ignore_status)
```
chunfuwen, line 4377: since o is not used, better change to s,_ = utils_misc()

virttest/utils_net.py
```diff
@@ -4351,3 +4351,50 @@ def delete_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
...
+    :param iface_name: the physical interface to be attached to the bridge
+    :param ignore_status: Whether to raise an exception when command fails
+    :return: bridge deleted or raise exception
+    """
+    # Delete the linux bridge
+    br_path = "/sys/class/net/%s" % linux_bridge_name
+    if not utils_package.package_install(['tmux', 'dhcp-client', 'procps-ng']):
+        raise exceptions.TestError("Failed to install the required packages.")
+    if not os.path.exists(br_path):
+        raise exceptions.TestError("There is no bridge named %s on the host to delete" % linux_bridge_name)
```
Yingshun, line 4395: if there is no br_path,I think we'd better log a message and 'return' directly, rather than raising an exception. How do you think?

virttest/utils_net.py
```diff
@@ -4351,3 +4351,50 @@ def delete_ovs_bridge(ovs_bridge_name, session=None, ignore_status=False):
...
+    :param iface_name: the physical interface to be attached to the bridge
+    :param ignore_status: Whether to raise an exception when command fails
+    :return: bridge created or raise exception
+    """
+    # Create bridge
+    br_path = "/sys/class/net/%s" % linux_bridge_name
+    if not utils_package.package_install(['tmux', 'dhcp-client']):
+        raise exceptions.TestError("Failed to install the required packages.")
+    if os.path.exists(br_path):
+        raise exceptions.TestError("Create bridge error as there is already an interface named '%s' on the host"
```
Yingshun, line 4372: Is it possible to delete the existing bridge by calling delete_linux_bridge?

virttest/utils_net.py
Yingshun: utils_misc.cmd_status_output will handle the exception and return status and stdout. So I think the previous code is correct. It should return utils_misc.cmd_status_output. How do you think? @yalzhang @chunfuwen

yalzhang: I think chunfu means  "s,_ = utils_misc.cmd_status_output"  instead of "s,o=utils_misc.cmd_status_output". So I think it is ok to use "_" to replace "o"

virttest/utils_net.py
yalzhang: I'm thinking of it, too. I have updated it now, Thank you!

virttest/utils_net.py
yalzhang: yes, I have updated the function to be like delete_linux_bridge(linux_bridge_name, iface_name=None, ignore_status=False)
If there is physical interface attached to the bridge, the bridge name and interface name both need to be specified. The interface name is needed when we try to recover the connection.
If there is no physical interface attached to the bridge, only the bridge name is needed.

yalzhang:
@Yingshun  Please help to review it again, Thank you!

virttest/utils_net.py
Yingshun: Ok. You can call delete_linux_bridge🙂

virttest/utils_net.py
yalzhang: > Better return True or False ,rather than standard output
As the utils_misc.cmd_status_output() with return "status" and "outputs", if we only return True or False, it will loss some information in the outputs like error message if create failure.

Yingshun (approved):

yalzhang:
test result refer to:
https://github.com/autotest/tp-libvirt/pull/3473#issuecomment-839432750

pevogam:
What is the difference between this proposal and `Bridge.add_bridge` in the `utils_net`?

yalzhang:
> What is the difference between this proposal and `Bridge.add_bridge` in the `utils_net`?
Thank you for pointing that out. The Bridge.add_bridge works well to add a Linux bridge, but when I tried to add an active interface to this bridge, the system network broken. It is because Bridge.add_port did not set the bridge up, even it do, it will also cause network broken. If we use tmux to execute the commands, we can avoid it.

pevogam:
Then can this be documented? Or better yet, can the `Bridge.add_bridge` gain a `use_tmux` argument to be made to work again for everyone but not break anything for people that won't enable a `use_tmux` flag?

chunfuwen (approved):
lgtm
Please resolve code conflicts

dzhengfy:
@yalzhang please fix the conflicts.

pevogam (changes requested):
@yalzhang Ok, in that case please rename the `create_linux_bridge` to something like `create_linux_bridge_tmux` as the name you use seems way more general than it actually is. I am still wondering if there is another reason to have broken networking and the need of this complexity can be avoided but at the very least the functions should be renamed to mean what they actually mean to do.

yalzhang:
> @yalzhang Ok, in that case please rename the `create_linux_bridge` to something like `create_linux_bridge_tmux` as the name you use seems way more general than it actually is. I am still wondering if there is another reason to have broken networking and the need of this complexity can be avoided but at the very least the functions should be renamed to mean what they actually mean to do.
@pevogam  Thank you for the suggestion. I have updated the name for the functions. I have tried to search the document about "why it will break the networking" but could not find one. In our testing, we execute the ip commands or the Bridge.addport() on a remote host, and we connect the remote host by ssh.  If there is only 1 active interface on the host, see 'eth0', after we execute the "ip link set eth0 master br0", the ssh connection will be broken at this point. And the network of this host will broken, until we connect the host locally to set the br0 up successfully, the network of this system will recover. By using 'tmux' or executing all the commands in a shell script will ensure all the commands including setting bridge up and recovering the network to be executed. And it avoid the broken of current ssh connection.

pevogam:
> > @yalzhang Ok, in that case please rename the `create_linux_bridge` to something like `create_linux_bridge_tmux` as the name you use seems way more general than it actually is. I am still wondering if there is another reason to have broken networking and the need of this complexity can be avoided but at the very least the functions should be renamed to mean what they actually mean to do.
> > @pevogam  Thank you for the suggestion. I have updated the name for the functions. I have tried to search the document about "why it will break the networking" but could not find one. In our testing, we execute the ip commands or the Bridge.addport() on a remote host, and we connect the remote host by ssh.  If there is only 1 active interface on the host, see 'eth0', after we execute the "ip link set eth0 master br0", the ssh connection will be broken at this point. And the network of this host will broken, until we connect the host locally to set the br0 up successfully, the network of this system will recover. By using 'tmux' or executing all the commands in a shell script will ensure all the commands including setting bridge up and recovering the network to be executed. And it avoid the broken of current ssh connection.

Hi @yalzhang, to me this seems like something that could be solved in a much simpler way by simply creating a custom bridge for the tests and not using the main bridge for the host machine. This is also more natural solution and helps for better isolated testing. But this is your choice, I am just giving my 2 cents here.

pevogam (approved):

PR #3036, opened 2021-04-30
Add vhostuser disk support in libvirt.create_disk_xml

chunfuwen:
vhostuser disk is newly introduced backend storage
```
vhostuser disk looks like as below:
    <disk type='vhostuser' device='disk' snapshot='no'>
      <driver name='qemu' type='raw'/>
      <source type='unix' path='/tmp/vhost.sock'/>
      <target dev='vdb' bus='virtio'/>
    </disk>
```

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1636,6 +1636,9 @@ def create_disk_xml(params):
             source_attrs = {'file': source_file}
             if slice_in_source:
                 source_attrs = {'file': input_source_file}
+        elif type_name == "vhostuser":
+            source_file = params.get("source_file", "")
+            source_attrs = {'type': 'unix', 'path': source_file}
         elif type_name == "block":
             source_file = params.get("source_file", "")
             source_attrs = {'dev': source_file}


chunfuwen:
>>> from virttest.utils_test import libvirt
>>> from avocado.utils import process
>>> params = {'type_name': 'vhostuser', 'disk_snapshot_attr': 'no', 'driver_type': 'raw', 'source_file': '/var/lib/image.qcow2'}
>>> disk = libvirt.create_disk_xml(params)
>>> output = process.run('cat %s' % disk, ignore_status=False, shell=True).stdout_text
>>> print(output)
<disk device="disk" snapshot="no" type="vhostuser"><source path="/var/lib/image.qcow2" type="unix" /><driver name="qemu" type="raw" /><target bus="virtio" dev="vdb" /></disk>
>>>

kamilvarga (approved):
LGTM.

virttest/utils_test/libvirt.py
```diff
@@ -1636,6 +1636,9 @@ def create_disk_xml(params):
             source_attrs = {'file': source_file}
             if slice_in_source:
                 source_attrs = {'file': input_source_file}
+        if type_name == "vhostuser":
```
Yingshun, line 1639: 'if' should be 'elif'? I think we'd better keep original style of code, how do you think?

chunfuwen: @Yingshun 
updated accordingly.

chunfuwen:
@kylazhang 

please have a look

Yingshun (approved):

dzhengfy (approved):
LGTM

PR #3039, opened 2021-05-08
Support image convert disable

dzhengfy:
This is to make the command 'qemu-img convert' optional for netfs.
Because image_convert is True as default, this change will not
impact the existing invoke codes. The reason to change this is
because in most cases there is no need to do the image convert
which increases the migration duration. After this, the user
invokes set_vm_disk() with image_convert='no' to avoid converting
the image.


Yingshun (approved):

dzhengfy:
See test result https://github.com/autotest/tp-libvirt/pull/3481#issuecomment-847494726

chloerh (approved):

PR #3047, opened 2021-05-11
migration: control migration speed

dzhengfy:
Add a function to set the migration speed for precopy and postcopy.
This is usually used to decrease the speed in order to give the
opportunity to run other functions during migration process.

### DIFF
--- a/virttest/migration.py
+++ b/virttest/migration.py
@@ -9,6 +9,7 @@ from avocado.utils import path as utils_path
 from avocado.utils import distro
 from avocado.core import exceptions
 
+from virttest import libvirt_version
 from virttest import remote
 from virttest import virsh
 from virttest import utils_disk
@@ -577,6 +578,53 @@ class MigrationTest(object):
                               virsh_event_session.get_stripped_output()), 30):
             raise exceptions.TestError("Unalbe to find event {}"
                                        .format(exp_str))
+
+    def control_migrate_speed(self, vm_name, to_speed=1, mode="precopy"):
+        """
+        Set migration speed to control the migration duration
+
+        :param vm_name: vm name
+        :param to_speed: the speed value in Mbps to be set for migration
+        :param mode: one of ['postcopy', 'precopy', 'both']
+        """
+        def _set_speed(extra_option=''):
+            """
+            Inner function to set migration speed
+
+            :param extra_option: str, it might include '--postcopy' or not
+            """
+            virsh_args = {"ignore_status": False}
+            old_speed = virsh.migrate_getspeed(vm_name,
+                                               extra=extra_option,
+                                               **virsh_args)
+            logging.debug("Current %s migration speed is %s "
+                          "MiB/s\n", extra_option, old_speed.stdout_text.strip())
+            logging.debug("Set %s migration speed to %d "
+                          "MiB/s\n", extra_option, to_speed)
+            virsh.migrate_setspeed(vm_name, to_speed,
+                                   extra=extra_option,
+                                   **virsh_args)
+
+        if mode not in ['postcopy', 'precopy', 'both']:
+            raise exceptions.TestError("'mode' only supports "
+                                       "'postcopy', 'precopy', 'both'")
+        warning_msg = "libvirt version should be larger than or equal to " \
+                      "'5.0.0' when setting postcopy migration speed."
+
+        if not libvirt_version.version_compare(5, 0, 0):
+            if mode == 'both':
+                logging.warning("%s Only precopy speed is set.", warning_msg)
+                mode = 'precopy'
+            if mode == 'postcopy':
+                logging.warning("%s Skipping", warning_msg)
+                return
+        if mode == 'both':
+            _set_speed()
+            _set_speed(extra_option='--postcopy')
+        elif mode == 'postcopy':
+            _set_speed(extra_option='--postcopy')
+        else:
+            _set_speed()
 
     def run_stress_in_vm(self, vm, params):
         """


virttest/migration.py
```diff
@@ -578,6 +579,29 @@ def set_migratepostcopy(self, vm_name, uri=None):
...
+    def control_migrate_speed(self, vm_name, to_speed=1, opts=""):
+        """
+        Set migration speed to control the migration duration
+
+        :param vm_name: vm name
+        :param to_speed: the speed value in Mbps to be set for migration
+        :param opts: the options for virsh command
+        """
+
+        def _set_speed(vm_name, to_speed, opts=''):
```
chunfuwen, line 591: I assume those parameters can be referred directly from control_migrate_speed since _set_speed is inner method of control_migrate_speed

virttest/migration.py
```diff
@@ -578,6 +579,29 @@ def set_migratepostcopy(self, vm_name, uri=None):
...
+            virsh_args = {}
+            virsh_args.update({"ignore_status": False})
+            old_speed = virsh.migrate_getspeed(vm_name, extra=opts, **virsh_args)
+            logging.debug("Current %s migration speed is %s MiB/s\n", opts, old_speed.stdout_text.strip())
+            logging.debug("Set %s migration speed to %d MiB/s\n", opts, to_speed)
+            virsh.migrate_setspeed(vm_name, to_speed, extra=opts, **virsh_args)
+            actual_speed = virsh.migrate_getspeed(vm_name, extra=opts, **virsh_args)
```
Yingshun, line 598: Does it necessary to check if the speed is set as expected?

virttest/migration.py
```diff
@@ -578,6 +579,29 @@ def set_migratepostcopy(self, vm_name, uri=None):
...
+            logging.debug("New %s migration speed is %s MiB/s\n", opts, actual_speed.stdout_text.strip())
+
+        _set_speed(vm_name, to_speed)
+        if '--postcopy' in opts and libvirt_version.version_compare(5, 0, 0):
+            _set_speed(vm_name, to_speed, opts=opts)
```
Yingshun, line 603: I can imagine why you set the speed for both precopy and postcopy here, we may need to slowdown the whole process. But I'm still feel a little bit confused that when **opts=='--postcopy'**, it setsspeed for precopy as well. is it expected? I'm wondering if we need a parameter(eg, mode) to specify whether to set pre-cpy, postcopy or both precopy-cpy and postcopy migration speed. What do you think?

virttest/migration.py
dzhengfy: updated

virttest/migration.py
dzhengfy: When opts=='--postcopy', it setsspeed for precopy as well. This is usually what feature owner does in their test. So I would like to mock up this step to ensure the migration has enough time to shift to postcopy from precopy.

virttest/migration.py
dzhengfy: updated

virttest/migration.py
```diff
@@ -578,6 +579,45 @@ def set_migratepostcopy(self, vm_name, uri=None):
...
+            Inner function to set migration speed
+
+            :param extra_option: str, it might include '--postcopy' or not
+            :raises: exceptions.TestError when not support
+            """
+            if ('--postcopy' in extra_option and
+                    not libvirt_version.version_compare(5, 0, 0)):
+                raise exceptions.TestError("libvirt version should be larger "
+                                           "than or equal to '5.0.0' when "
+                                           "setting postcopy migration speed")
```
Yingshun, line 602: Is 'speed set' a 'must have' step? If no, I think we'd better log an error/return directly rather than raising TestError to stop testing. What do you think?

virttest/migration.py
Yingshun: Ok, I see. But I think we'd better set a switch here... or move _set_speed to outside of the control_migrate_speed, in case we only need to set postcopy speed.

dzhengfy: Thanks for your suggestion. I think it is a good idea. Updated.

virttest/migration.py
dzhengfy: updated

Yingshun (approved):
LGTM

virttest/migration.py
```diff
@@ -578,6 +579,54 @@ def set_migratepostcopy(self, vm_name, uri=None):
...
+            """
+            virsh_args = {}
+            virsh_args.update({"ignore_status": False})
+            old_speed = virsh.migrate_getspeed(vm_name,
+                                               extra=extra_option,
+                                               **virsh_args)
+            logging.debug("Current %s migration speed is %s "
+                          "MiB/s\n", opts, old_speed.stdout_text.strip())
+            logging.debug("Set %s migration speed to %d "
+                          "MiB/s\n", opts, to_speed)
```
Yingshun, line 604: s/opts/extra_option

dzhengfy: fixed. thanks.

dzhengfy:
Test for latest change:
 (1/1) type_specific.io-github-autotest-libvirt.virsh.live_migration.domjobabort.p2p_live.without_postcopy: PASS (217.88 s)
JOB TIME   : 219.05 s

virttest/migration.py
```diff
@@ -578,6 +579,54 @@ def set_migratepostcopy(self, vm_name, uri=None):
...
+        :param mode: one of ['postcopy', 'precopy', 'both']
+        """
+        def _set_speed(extra_option=''):
+            """
+            Inner function to set migration speed
+
+            :param extra_option: str, it might include '--postcopy' or not
+            """
+            virsh_args = {}
+            virsh_args.update({"ignore_status": False})
```
chloerh, line 597:
```suggestion
            virsh_args = {"ignore_status": False}
```

kylazhang (approved):

virttest/migration.py
dzhengfy: update

chloerh (approved):

virttest/migration.py
dzhengfy: @chunfuwen I have solved your comments. Because I have an urgent requirement for this pr and another pr depending on it to be merged asap, I requested to merge it now without your approval. If you have further comments, let me know, then I 'd like to solve them in new PR. Thanks.

kylazhang:
As chunfuwen's comments already resolved so merge it.

PR #3058, opened 2021-05-14
Add set-user-sshkeys and get-user-sshkeys cmd to virsh.py

iccaszhulili:

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -4857,3 +4857,28 @@ def guestinfo(name, options="", **dargs):
     :return: CmdResult object.
     """
     return command("guestinfo %s %s" % (name, options), **dargs)
+
+
+def get_user_sshkeys(name, user, **dargs):
+    """
+    list authorized SSH keys for given user (via agent)
+
+    :param name: VM name
+    :param user: user to list authorized keys for
+    :param dargs: standardized virsh function API keywords
+    :return: CmdResult object.
+    """
+    return command("get-user-sshkeys %s %s" % (name, user), **dargs)
+
+
+def set_user_sshkeys(name, user, options="", **dargs):
+    """
+    manipulate authorized SSH keys file for given user (via agent)
+
+    :param name: VM name
+    :param user: user to set authorized keys for
+    :param options: options of set-user-sshkeys command
+    :param dargs: standardized virsh function API keywords
+    :return: CmdResult object.
+    """
+    return command("set-user-sshkeys %s %s %s" % (name, user, options), **dargs)


virttest/virsh.py
```diff
@@ -4857,3 +4857,37 @@ def guestinfo(name, options="", **dargs):
...
+    :param options: options of get-user-sshkeys command
+    :param key_file: optional file to read keys from
+    :param dargs: standardized virsh function API keywords
+    :return: CmdResult object.
+
+    """
+    cmd = "set-user-sshkeys %s %s " % (name, user)
+    if options == "--reset":
+        cmd += "--reset "
+    if options == "--remove":
```
chunfuwen, line 4889: If options is --remove, is key_file necessary to be provided
If this is case, better change to cmd += "--remove  --file %s" % key_file

iccaszhulili: --file can be omitted

virttest/virsh.py
```diff
@@ -4857,3 +4857,37 @@ def guestinfo(name, options="", **dargs):
...
+    :param name: VM name
+    :param user: user to set authorized keys for
+    :param options: options of get-user-sshkeys command
+    :param key_file: optional file to read keys from
+    :param dargs: standardized virsh function API keywords
+    :return: CmdResult object.
+
+    """
+    cmd = "set-user-sshkeys %s %s " % (name, user)
+    if options == "--reset":
```
Yingshun, line 4887: If options is not '--reset' or '--remove', it will be ignored. Is this the expected result?
Or Line4886 should be `cmd = "set-user-sshkeys %s %s  %s" % (name, user, options)`?

iccaszhulili: if options is "--file", then there will be 2 --file option, which will cause error. 
Now in the code, if "--file" option is passed in，it will be ignored directly.

virttest/virsh.py
```diff
@@ -4857,3 +4857,37 @@ def guestinfo(name, options="", **dargs):
...
+    return command("get-user-sshkeys %s %s" % (name, user), **dargs)
+
+
+def set_user_sshkeys(name, user, options="", key_file=None, **dargs):
+    """
+    manipulate authorized SSH keys file for given user (via agent)
+
+    :param name: VM name
+    :param user: user to set authorized keys for
+    :param options: options of get-user-sshkeys command
```
Yingshun, line 4880: :%s/get-user-sshkeys/set-user-sshkeys/

chunfuwen (approved):
lgtm

virttest/virsh.py
Yingshun: Ok, I see. To be honest, I think we should handle '--file' option for this situation rather than others. How about updating it as below? 
cmd = "set-user-sshkeys %s %s %s" % (name, user, options)
if key_file:
       cmd = cmd.replace("--file", "")
        cmd += "--file %s" % key_file

iccaszhulili: updated

virttest/virsh.py
iccaszhulili: @Yingshun updated

Yingshun (approved):

PR #3063, opened 2021-05-18
Add todo info for function check_dmesg_output

chloerh:
This function could be replaced with another function of master
branch of avocado. Add a todo here as a reminder.

### DIFF
--- a/virttest/utils_sys.py
+++ b/virttest/utils_sys.py
@@ -10,6 +10,7 @@ import re
 from avocado.utils import process
 
 
+# TODO: check function in avocado.utils after the next LTS
 def check_dmesg_output(pattern, expect=True, session=None):
     """
     Check whether certain pattern exists in dmesg.


chloerh:
@pevogam Would you like to take a look?

pevogam (approved):
LGTM, if you want to keep it easier in the future registering an issue at github is another good way to do the same.

virttest/utils_sys.py
```diff
@@ -10,6 +10,7 @@
 from avocado.utils import process
 
 
+# todo: check function in avocado.utils after the next LTS
```
Yingshun, line 13: I'm not quite sure but I think it should be in upper case. TODO Tree(extension of vscode) cannot search for comment tag todo by default.

pevogam: Alternatively we could also go for sphinx compatible `..todo:: ` but I think our approach shouldn't depend on the choice of IDE-s.

chloerh: @Yingshun @pevogam Thanks for the suggestions. I'm going to follow the tradition of 'todo's of avocado-vt which is to replace it with 'TODO'.

Yingshun (approved):

PR #3064, opened 2021-05-18
[utils_zchannel] Define helpers for handling (sub)channel paths on s390x

smitterl:
Add two new classes to handle subchannel and channel paths using
s390-tools commands lcss, chchp.

### DIFF
--- a/virttest/utils_zchannels.py
+++ b/virttest/utils_zchannels.py
@@ -1,0 +1,169 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Red Hat Inc. 2020
+# Author: Sebastian Mitterle <smitterl@redhat.com>
+
+"""
+Module for channel device management on s390x aka IBM Z
+
+Example:
+    # Retrieve device to remove it from host
+
+    subchannel_paths = SubchannelPaths()
+    subchannel_paths.get_info()
+
+    device = subchannel_paths.get_first_unused_and_safely_removable()
+    chpids = device[subchannel_paths.header["CHPIDs"]]
+
+    channel_paths = ChannelPaths()
+    channel_paths.set_standby(chipds)
+"""
+from virttest.utils_misc import cmd_status_output
+
+# timeout value in seconds for any command run in this module
+CMD_TIMEOUT = 60
+# for debugging, set True to get verbose command run output
+VERBOSE = False
+
+
+class SubchannelPaths(object):
+    """
+    Represents channel subsystem info
+    """
+
+    HEADER = {
+            "Device": 0,
+            "Subchan.": 1,
+            "DevType": 2,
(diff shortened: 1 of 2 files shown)


smitterl:
Unit test run:
```bash
# python unittests/test_utils_zchannels.py --verbose
test__split (__main__.TestChannelPaths) ... ok
test_get_first_unused_and_safely_removable (__main__.TestSubchannelPaths) ... ok
test_get_first_unused_and_safely_removable_not_first (__main__.TestSubchannelPaths) ... ok
test_get_first_unused_and_safely_removable_not_safe (__main__.TestSubchannelPaths) ... ok
test_get_info (__main__.TestSubchannelPaths) ... ok

Ran 5 tests in 0.001s

OK
``

**Comment by smitterl** [comment]
Used in new test cases in https://github.com/autotest/tp-libvirt/pull/3493

File: virttest/utils_zchannels.py [hunk 1]
```diff
@@ -0,0 +1,168 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
```
chunfuwen, line 3: I observe that avocado-vt is declared license as GPL-2.0 License. 
I wonder whether we need explicitly declare it again in newly created file.

virttest/utils_zchannels.py
```diff
@@ -0,0 +1,168 @@
...
+        return None
+
+
+class ChannelPaths(object):
+    """
+    Represents channel path info
+    """
+
+    @staticmethod
+    def _split(chpids):
```
chunfuwen, line 112: Could this method level up to independent method in utils_zchannels.py  since it is not relevant to class ChannelPaths?

virttest/utils_zchannels.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+
+    def __init__(self, session=None):
+        """
+        Initializes instance
+
+        :param session: guest session; if None, host info is handled
+        """
+        self.session = session
+        self.header = {
```
chunfuwen, line 50: Could header is  Class-wide  constant since it looks like its values are hardcode?

virttest/utils_zchannels.py
smitterl: @chunfuwen No, I prefer not to. According to the class design, ChannelPaths should be used for any channel path handling. By pushing it to top, we'd break encapsulation I believe. Quoting [1]:

"To me, most fundamentally, encapsulation means hiding implementation details. Instead of expressing how to do something, we express what to do. I like to say that encapsulation is hiding the “how” with the “what.” Encapsulation is an important concept and an important code quality because it gives us the ability to change the implementation details without affecting other parts of the system."

[1] https://dzone.com/articles/quality-code-is-encapsulated

virttest/utils_zchannels.py
smitterl: @chunfuwen Libvirt adds license to file header of each file, I took that and that we did include license declarations in other code files, too, as authorative. Please let me know if you want me to remove it.

virttest/utils_zchannels.py
smitterl: @chunfuwen It could. I updated it. But you'll see that in this case, IMO, it's only syntactic sugar, we still need to use `self.HEADER` or the longer `SubchannelPaths.HEADER` to use it; using only `HEADER` will lead to an error. However, as it's conceptually true - the header can be expected to be constant - I updated it and re-ran the unittests.

kamilvarga (approved):
LGTM.

chunfuwen (approved):
lgtm

PR #3067, opened 2021-05-21
[virttest/migration] Add default value to avoid TypeError

smitterl:
`migrate_options` might not be set. Use empty string to avoid tests
failing with
"TypeError: argument of type 'NoneType' is not iterable"


Yingshun (approved):

chunfuwen (approved):
lgtm

PR #3069, opened 2021-05-24
qemu_vm: Fix QMPCmdError

huangyum:
ID: 1961003

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -280,18 +280,6 @@ class VM(virt_vm.BaseVM):
         except virt_vm.VMStatusError:
             return False
 
-    def is_preconfig(self):
-        """
-        Return True if the qemu process is in preconfig status
-        """
-        if self.is_dead():
-            return False
-        try:
-            self.verify_status("preconfig")
-            return True
-        except virt_vm.VMStatusError:
-            return False
-
     def is_panicked(self):
         """
         Return True if the qemu process is panicked
@@ -3147,7 +3135,8 @@ class VM(virt_vm.BaseVM):
 
             logging.debug("VM appears to be alive with PID %s", self.get_pid())
             # Record vcpu infos in debug log
-            if not self.is_preconfig():
+            is_preconfig = params.get_boolean("qemu_preconfig")
+            if not is_preconfig:
                 self.get_vcpu_pids(debug=True)
             vhost_thread_pattern = params.get("vhost_thread_pattern",
                                               r"\w+\s+(\d+)\s.*\[vhost-%s\]")
@@ -3168,7 +3157,7 @@ class VM(virt_vm.BaseVM):
             # such as serial console.
             time.sleep(1)
 
-            if self.is_preconfig():
+            if is_preconfig:
                 return
 
             if params.get("paused_after_start_vm") != "yes":


virttest/qemu_vm.py
```diff
@@ -3147,7 +3159,7 @@ def create(self, name=None, params=None, root_dir=None,
 
             logging.debug("VM appears to be alive with PID %s", self.get_pid())
             # Record vcpu infos in debug log
-            if not self.is_preconfig():
+            if not self.is_preconfig() and not self.is_prelaunch():
```
vivianQizhu, line 3162: If they are the same thing with different expression in different versions, I would prefer to implement it in one same function, not only because of the meaning, also for the reference of `is_preconfig` at other places.

yanan-fu: I have the same idea with @vivianQizhu .

The change is only about the qemu monitor output for different qemu version, no change with qemu command:
https://github.com/avocado-framework/avocado-vt/blob/b855e7ec9b2c1f46593d79199668e71f7b92ff3b/virttest/qemu_vm.py#L1514~L1516

I think, `is_preconfig` is still reasonable even if the status string in qemu monitor output is `prelaunch`, so suggest update `is_preconfig` to work with different qemu version.

yanan-fu: I have the same idea with @vivianQizhu .

The change is only about the qemu monitor output for different qemu version, no change with qemu command:
https://github.com/avocado-framework/avocado-vt/blob/b855e7ec9b2c1f46593d79199668e71f7b92ff3b/virttest/qemu_vm.py#L1514~L1516

I think, `is_preconfig` is still reasonable even if the status string in qemu monitor output is `prelaunch`, so suggest update `is_preconfig` to work with different qemu version.

huangyum:
@yanan-fu @vivianQizhu Updated, pls help review again.
 (1/1) Host_RHEL.m8.u5.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.numa_cpu.with_preconfig: PASS (43.54 s)

virttest/qemu_vm.py
```diff
@@ -284,13 +284,9 @@ def is_preconfig(self):
...
         Return True if the qemu process is in preconfig status
         """
-        if self.is_dead():
-            return False
-        try:
-            self.verify_status("preconfig")
-            return True
-        except virt_vm.VMStatusError:
-            return False
+        if self.params.get_boolean("qemu_preconfig"):
```
yanan-fu, line 287: IMO, the value of the param and the vm status are two different thing, although the result is same sometimes.
I still suggest to checking vm status, and use different status name for different qemu version.

huangyum: It won't work for other cases. In all the cases, vm status is prelaunch at first. As a result, is_preconfig() returns true and jump to L3172, which is not what we expect.

vivianQizhu: @huangyum If modified it this way, this whole function will be meaningless, why not calling `is_paused` directly instead of `is_preconfig` then?

huangyum: Because VM is always paused at this moment, and will return in L3186.

vivianQizhu: @huangyum `if self.is_preconfig()` just now equals to `if self.params.get_boolean("qemu_preconfig") and self.is_paused()`, this function still seems meaningless to me. I mean, then this function should not be implemented this way.

huangyum: Do you mean to replace `if is_preconfig()` with` if params.get_boolean("qemu_preconfig") and is_paused()` ?

xutian: agree with @vivianQizhu .

huangyum:
(1/1) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.numa_cpu.with_preconfig: PASS (164.97 s)

@yanan-fu @vivianQizhu Removed the function, pls help review, thanks.

virttest/qemu_vm.py
vivianQizhu: > Do you mean to replace `if is_preconfig()` with` if params.get_boolean("qemu_preconfig") and is_paused()` ?

@huangyum That's not what I mean, I still not quite understand, the original issue seems to be the name of the status `preconfig` is modified in new version qemu, why we now just simply remove the checking? So previously the checking is totally meaningless from the beginning?

huangyum: With qemu<6.0, the status is preconfig at this point if preconfig is set, and prelaunch if not. So it's ok to check if is_preconfig by checking status. But with qemu6.0, the status is always prelaunch no matter it's preconfig or not. Seems checking the param is the only way to find out if it's preconfig.

vivianQizhu:
@huangyum Thanks for the clarification, now it seems clearer for me. However, is there any other difference when you append `--preconfig` on the command line? If no any difference, what's the point of this command line? From the doc it says:
```
The experimental --preconfig command line option allows pausing QEMU before the initial VM creation, in a “preconfig” state, where additional queries and configuration can be performed via QMP before moving on to the resulting configuration startup. In the preconfig state, QEMU only allows a limited set of commands over the QMP monitor, where the commands do not depend on an initialized machine, including but not limited to:
qmp_capabilities
query-qmp-schema
query-commands
query-status
x-exit-preconfig
```
 If the commands allowed also changed? Can we check the status via the above commands?

huangyum:
> @huangyum Thanks for the clarification, now it seems clearer for me. However, is there any other difference when you append `--preconfig` on the command line? If no any difference, what's the point of this command line? From the doc it says:
> '''
> The experimental --preconfig command line option allows pausing QEMU before the initial VM creation, in a “preconfig” state, where additional queries and configuration can be performed via QMP before moving on to the resulting configuration startup. In the preconfig state, QEMU only allows a limited set of commands over the QMP monitor, where the commands do not depend on an initialized machine, including but not limited to:
> qmp_capabilities
> query-qmp-schema
> query-commands
> query-status
> x-exit-preconfig
> '''
> If the commands allowed also changed? Can we check the status via the above commands?

No as my knowledge. There are no difference in all query related cmds output between preconfig and non-preconfig.

vivianQizhu:
@huangyum Emm...then, before we modify this, I think we should ask devel why there is still `--preconfig` if nothing different.

vivianQizhu:
From the doc it says `--preconfig` is `experimental`, if it is supposed to be removed in qemu6.0(if no difference then that should be a reasonable speculation), maybe we should remove all `preconfig` related operations, including the parameter `qemu_preconfig`. Before that, should confirm with developer.

huangyum:
> @huangyum Emm...then, before we modify this, I think we should ask devel why there is still `--preconfig` if nothing different.

I may not make myself clear. It's not nothing different. It only makes no difference when execute these allowed query command. The '--preconfig' option allows to configure guest before start, which is the test step in tp-qemu scripts. We check if it's preconfig here because at this state, some qmp commands are not allowed and will fail the case directly. So we wanna skip these cmds if it's preconfig.

vivianQizhu:
@huangyum So we can keep the `is_preconfig` function and check it via those unallowed command(allowed when --preconfig, if I understand correctly), right?

huangyum:
> @huangyum So we can keep the `is_preconfig` function and check it via those unallowed command(allowed when --preconfig, if I understand correctly), right?

Could you please tell the reason to keep the function? It's intended to skip the unallowed cmds, now you are suggesting to run them inside the function. Besides, if the cmds go wrong with some issues, all cases end up in preconfig status.

yanan-fu:
> @huangyum So we can keep the `is_preconfig` function and check it via those unallowed command(allowed when --preconfig, if I understand correctly), right?

@vivianQizhu I have different opinion about this.
1. I think `query-status` is the appropriate way to indicate the VM's status.
2. In prefonfig status, checking supported command is not supported, refer d5d98723a7ddda69a1b0a5a3a4bff273db0e3464,  to determin  `is_preconfig` with unallowed command is an negative test.
3. Which unallowd command should be used is a problem too i think. It is a risk that the unallowed command are different with different version.

yanan-fu:
I would like to explain my thought and some backgroud:
1. Old behaviour
With `--preconfig`,  the VM status value is 'preconfig', and vm can be this status only when use `--preconfig`. We use this status to determine `is_preconfig`, and define some `preconfig` specific behaviours.

2. New behaviour
With `--preconfig`, the VM status value is `prelaunch`, this status is not for `--preconfig` only,  boot with "-S" is in this status too.  So we can not use `prelaunch`(status) to determind  `is_preconfig` now.

Here is some of my thought:
1.  `preconfig` and `prelaunch` is irreversible
2.  The current change is ok in vt level, just not very friendly for case level ?(no case use it at current stage)
3. How do you think about use both `qemu_preconfig` and the status to define `is_preconfig`
```
if params.get_boolean("qemu_preconfig"):
    try:
        self.verify_status("***")
   ...
```

vivianQizhu:
@huangyum @yanan-fu I am convinced, in case no other usage of this function, instead of define a `is_preconfig`, I prefer the current code to only check it by parameter `qemu_preconfig`. Thanks.

vivianQizhu (approved):

yanan-fu:
>     2. The current change is ok in vt level, just not very friendly for case level ?(no case use it at current stage)
>     3. How do you think about use both `qemu_preconfig` and the status to define `is_preconfig`
> ```
> if params.get_boolean("qemu_preconfig"):
>     try:
>         self.verify_status("***")
>    ...
> ```


@huangyum Do you have any comments about the info above ?

The status need be verified is `preconfig` and `prelaunch`.   
The solution in the pre version which use `is_paused` is not correct, since pause vm after exit_preconfig can match it by mistake.

huangyum:
> > 1. The current change is ok in vt level, just not very friendly for case level ?(no case use it at current stage)
> > 2. How do you think about use both `qemu_preconfig` and the status to define `is_preconfig`
> > 
> > ```
> > if params.get_boolean("qemu_preconfig"):
> >     try:
> >         self.verify_status("***")
> >    ...
> > ```
> 
> @huangyum Do you have any comments about the info above ?
> 
> The status need be verified is `preconfig` and `prelaunch`.
> The solution in the pre version which use `is_paused` is not correct, since pause vm after exit_preconfig can match it by mistake.

First of all, this such kind of function was rejected before regardless of correctness, pls see the review comments above. Second, I don's see anywhere else need to use it. Under current circumstance(in vm.create()), I think just checking the param is fine enough. We can update later if somewhere else need to use it. What do you think?

yanan-fu:
> First of all, this such kind of function was rejected before regardless of correctness, pls see the review comments above. Second, I don's see anywhere else need to use it. Under current circumstance(in vm.create()), I think just checking the param is fine enough. We can update later if somewhere else need to use it. What do you think?

1. I think you mean the comment from @vivianQizhu : `@huangyum if self.is_preconfig() just now equals to if self.params.get_boolean("qemu_preconfig") and self.is_paused(), this function still seems meaningless to me. I mean, then this function should not be implemented this way.`  ?   But after that, @vivianQizhu have abandon the original thought i think.

Let's ignore the history and discuss the problem itself now.

2.  My thought is: perfect the existing function `is_preconfig` is useful for both vt and test case.  If update it later when somewhere need use it in the future,  the change in this patch need be revert then.  So i prefer to update it now.

huangyum:
> > First of all, this such kind of function was rejected before regardless of correctness, pls see the review comments above. Second, I don's see anywhere else need to use it. Under current circumstance(in vm.create()), I think just checking the param is fine enough. We can update later if somewhere else need to use it. What do you think?
> 
>     1. I think you mean the comment from @vivianQizhu : `@huangyum if self.is_preconfig() just now equals to if self.params.get_boolean("qemu_preconfig") and self.is_paused(), this function still seems meaningless to me. I mean, then this function should not be implemented this way.`  ?   But after that, @vivianQizhu have abandon the original thought i think.
> 

I'm not sure about that. She approved the current change. @vivianQizhu Would you please help confirm? What do you think about  the design to put both param check and status check into the function?

> 
> Let's ignore the history and discuss the problem itself now.
> 
>     1. My thought is: perfect the existing function `is_preconfig` is useful for both vt and test case.  If update it later when somewhere need use it in the future,  the change in this patch need be revert then.  So i prefer to update it now.

Why do we need revert this patch? Checking param in vm.create() still works. The case that pause vm after exit_preconfig can never happen here.

vivianQizhu:
> > First of all, this such kind of function was rejected before regardless of correctness, pls see the review comments above. Second, I don's see anywhere else need to use it. Under current circumstance(in vm.create()), I think just checking the param is fine enough. We can update later if somewhere else need to use it. What do you think?
> 
> 1. I think you mean the comment from @vivianQizhu : `@huangyum if self.is_preconfig() just now equals to if self.params.get_boolean("qemu_preconfig") and self.is_paused(), this function still seems meaningless to me. I mean, then this function should not be implemented this way.`  ?   But after that, @vivianQizhu have abandon the original thought i think.
> 
> Let's ignore the history and discuss the problem itself now.
> 
> 1. My thought is: perfect the existing function `is_preconfig` is useful for both vt and test case.  If update it later when somewhere need use it in the future,  the change in this patch need be revert then.  So i prefer to update it now.

Yes, if we have proper way to judge the condition of `preconfig` that would be better from my view, that's exactly why I ask so many questions previously, I also hope to figure out a better `is_preconfig()`, but if it is always `prelaunch` in new version then that seems meaningless to check the status.
If leave the context of the current usage,  Obviously `self.params.get_boolean("qemu_preconfig")` and `self.is_paused()` can not equal to `is_preconfig()` cause what if the guest only pauses but not at prelaunch phase? So I don't accept this one.
While if organize `self.params.get_boolean("qemu_preconfig")` and `prelaunch`/`preconfig`(judge version first) together(so come back to the original recommendation), although not very meaningful for current usage, but if you suppose there will be further usage on test case, I am okay with it.

yanan-fu:
> > 1. My thought is: perfect the existing function `is_preconfig` is useful for both vt and test case.  If update it later when somewhere need use it in the future,  the change in this patch need be revert then.  So i prefer to update it now.
> 
> Yes, if we have proper way to judge the condition of `preconfig` that would be better from my view, that's exactly why I ask so many questions previously, I also hope to figure out a better `is_preconfig()`, but if it is always `prelaunch` in new version then that seems meaningless to check the status.
> If leave the context of the current usage, Obviously `self.params.get_boolean("qemu_preconfig")` and `self.is_paused()` can not equal to `is_preconfig()` cause what if the guest only pauses but not at prelaunch phase? So I don't accept this one.

Yeap, `is_paused` is incorrect, can not indicate the preconfig status correctly.

> While if organize `self.params.get_boolean("qemu_preconfig")` and `prelaunch`/`preconfig`(judge version first) together(so come back to the original recommendation), although not very meaningful for current usage, but if you suppose there will be further usage on test case, I am okay with it.

Yes, this is my proposal before, use `self.params.get_boolean("qemu_preconfig")` and `prelaunch`/`preconfig`(jdependens on version).   But after much reasoning and testing, it is still incorrect :(

Boot vm with `--preconfig` and `-S`, after `(x-)exit-preconfig`, the vm status is still in `prelaunch`, but it is not the preconfig, vm state already changed, supported qmp command changed too.

So,  i negate myself again,  no effective method to indicate vm is in preconfig status  by now :(

From the source code, `preconfig` state was removed, and introduce MachineInitPhase to indicate the phase. 
https://gitlab.com/redhat/rhel/src/qemu-kvm/qemu-kvm/-/commit/164dafd1744c69d268b89015977e19d8a9617fdf
https://gitlab.com/redhat/rhel/src/qemu-kvm/qemu-kvm/-/commit/2f181fbd5a9d456d1da291bea61d7e3ad10ec7d1

But i have no idea about how to check it from monitor or we can say from user's perspective :(

yanan-fu:
Ack with the code change.

The QMPCmdError looks not clear enough to explain this issue, could you explain with more details ? @huangyum

huangyum:
> > > 1. My thought is: perfect the existing function `is_preconfig` is useful for both vt and test case.  If update it later when somewhere need use it in the future,  the change in this patch need be revert then.  So i prefer to update it now.
> > 
> > 
> > Yes, if we have proper way to judge the condition of `preconfig` that would be better from my view, that's exactly why I ask so many questions previously, I also hope to figure out a better `is_preconfig()`, but if it is always `prelaunch` in new version then that seems meaningless to check the status.
> > If leave the context of the current usage, Obviously `self.params.get_boolean("qemu_preconfig")` and `self.is_paused()` can not equal to `is_preconfig()` cause what if the guest only pauses but not at prelaunch phase? So I don't accept this one.
> 
> Yeap, `is_paused` is incorrect, can not indicate the preconfig status correctly.
> 
> > While if organize `self.params.get_boolean("qemu_preconfig")` and `prelaunch`/`preconfig`(judge version first) together(so come back to the original recommendation), although not very meaningful for current usage, but if you suppose there will be further usage on test case, I am okay with it.
> 
> Yes, this is my proposal before, use `self.params.get_boolean("qemu_preconfig")` and `prelaunch`/`preconfig`(dependens on version). But after much reasoning and testing, it is still incorrect :(
> 
> Boot vm with `--preconfig` and `-S`, after `(x-)exit-preconfig`, the vm status is still in `prelaunch`, but it is not the preconfig, vm state already changed, supported qmp command changed too.
> 
> So, i negate myself again, no effective method to indicate vm is in preconfig status by now :(
> 
> From the source code, `preconfig` state was removed, and introduce MachineInitPhase to indicate the phase.
> https://gitlab.com/redhat/rhel/src/qemu-kvm/qemu-kvm/-/commit/164dafd1744c69d268b89015977e19d8a9617fdf
> https://gitlab.com/redhat/rhel/src/qemu-kvm/qemu-kvm/-/commit/2f181fbd5a9d456d1da291bea61d7e3ad10ec7d1
> 

Thanks for all the investigation. Seems there's no appropriate way to decide if's in preconfig phase for sure. Luckily, usually we only check is_preconfig at very beginning, e.g. vm.create(), so I think we can remove the function and only check param qemu_preconfig here. Thanks.

> But i have no idea about how to check it from monitor or we can say from user's perspective :(

yanan-fu (approved):

vivianQizhu:
Test passed with `qemu_preconfig = off`, thanks @huangyum and @yanan-fu

PR #3072, opened 2021-05-26
Customize tls setup

dzhengfy:
Before this PR, we always scp all required files like cakey, cacert,
severkey ... to the TLS server host and client host which are normal
way for TLS connection setup. But sometimes we need partial setup.
So this PR is to enable the capability to allow the user to define
which files will be scped to the TLS server/client host.

### DIFF
--- a/virttest/utils_conn.py
+++ b/virttest/utils_conn.py
@@ -10,6 +10,7 @@ import tempfile
 import aexpect
 from aexpect import remote
 
+from avocado.core import exceptions
 from avocado.utils import path
 from avocado.utils import process
 
@@ -925,6 +926,8 @@ class TLSConnection(ConnectionBase):
     special_cn: Use special cn in /etc/hosts, default don't use
     server_setup_local: Whether to setup tls server on local host
     server_info_ip: Use a specific IP address in server.info
+    scp_list_server: the file list to be scped to the server
+    scp_list_client: the file list to be scped to the client
     """
     __slots__ = ('server_cn', 'client_cn', 'ca_cn', 'CERTTOOL', 'pki_CA_dir',
                  'libvirt_pki_dir', 'libvirt_pki_private_dir', 'client_hosts',
@@ -937,7 +940,8 @@ class TLSConnection(ConnectionBase):
                  'server_saslconf', 'server_qemuconf', 'client_qemuconf',
                  'server_libvirtd_tls_socket', 'client_libvirtd_tls_socket',
                  'special_cn', 'server_setup_local', 'server_info_ip',
-                 'daemon_conf', 'daemon_socket_conf')
+                 'daemon_conf', 'daemon_socket_conf', 'scp_list_server',
+                 'scp_list_client')
 
     def __init__(self, *args, **dargs):
         """
@@ -967,6 +971,12 @@ class TLSConnection(ConnectionBase):
         init_dict['special_cn'] = init_dict.get('special_cn', 'no')
         init_dict['server_setup_local'] = init_dict.get('server_setup_local', False)
         init_dict['server_info_ip'] = init_dict.get("server_info_ip")
+        init_dict['scp_list_server'] = init_dict.get("scp_list_server",
+                                                     'cacert,cakey,'
+                                                     'servercert,serverkey')
+        init_dict['scp_list_client'] = init_dict.get("scp_list_client",
+                                                     'cacert,cakey,'
+                                                     'clientcert,clientkey')
 
         super(TLSConnection, self).__init__(init_dict)
         # check and set CERTTOOL in slots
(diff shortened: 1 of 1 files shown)


dzhengfy:
@Yingshun Please help review.

virttest/utils_conn.py
```diff
@@ -1292,17 +1300,18 @@ def server_setup(self, on_local=False):
...
-                        serverkey_path: self.libvirt_pki_private_dir}
             cmd = "mkdir -p {0}".format(self.pki_CA_dir)
             status, output = server_session.cmd_status_output(cmd)
             if status:
                 raise ConnMkdirError(self.pki_CA_dir, output)
+            params_scp = {'cacert': cacert_path,
+                          'cakey': cakey_path,
+                          'serverkey': serverkey_path,
+                          'servercert': servercert_path}
+        scp_dict = self._update_scp_list(self.scp_list_server, params_scp)
```
chunfuwen, line 1314: I notice that previous implementation don't this step. Is the step conditionally required or necessary required now?

virttest/utils_conn.py
```diff
@@ -1458,6 +1467,42 @@ def server_setup(self, on_local=False):
                                "%s %s" % (self.client_ip, hostname)}
             self.server_hosts.sub_else_add(pattern_to_repl)
 
+    def _update_scp_list(self, scp_list, params):
+        """
+        Generate the scp file list to remote host
+
+        :param params: dict, the required parameters are like:
```
chunfuwen, line 1474: mismatch doc string order

dzhengfy:
@Yingshun please help have a look. thanks.

virttest/utils_conn.py
```diff
@@ -1292,17 +1300,18 @@ def server_setup(self, on_local=False):
             raise ConnMkdirError(self.libvirt_pki_private_dir, output)
 
         if scp_new_cacert == 'no':
-            scp_dict = {servercert_path: self.libvirt_pki_dir,
-                        serverkey_path: self.libvirt_pki_private_dir}
+            params_scp = {'serverkey': serverkey_path,
+                          'servercert': servercert_path}
```
Yingshun, line 1304: Is scp_new_cecert still useful? can we set to scp_list_xxx directly?

dzhengfy: scp_new_cecert is the indicator to decide what files will be scped. So I think it is still useful.

virttest/utils_conn.py
dzhengfy: If you are asking to invoke for _update_scp_list(), then this step is necessary for all. It is a new created function which applies to all. If the user does not provide scp_list_client/scp_list_server, then invoking  this function makes no effect on them. It is just unified entry.

virttest/utils_conn.py
dzhengfy: updated, thanks

virttest/utils_conn.py
```diff
@@ -937,7 +938,8 @@ class TLSConnection(ConnectionBase):
                  'server_saslconf', 'server_qemuconf', 'client_qemuconf',
                  'server_libvirtd_tls_socket', 'client_libvirtd_tls_socket',
                  'special_cn', 'server_setup_local', 'server_info_ip',
-                 'daemon_conf', 'daemon_socket_conf')
+                 'daemon_conf', 'daemon_socket_conf', 'scp_list_server',
```
chunfuwen, line 941: please add those parameters description after line 928

virttest/utils_conn.py
```diff
@@ -1292,17 +1300,18 @@ def server_setup(self, on_local=False):
...
         else:
-            scp_dict = {cacert_path: self.pki_CA_dir,
-                        cakey_path: self.pki_CA_dir,
-                        servercert_path: self.libvirt_pki_dir,
-                        serverkey_path: self.libvirt_pki_private_dir}
             cmd = "mkdir -p {0}".format(self.pki_CA_dir)
             status, output = server_session.cmd_status_output(cmd)
             if status:
                 raise ConnMkdirError(self.pki_CA_dir, output)
+            params_scp = {'cacert': cacert_path,
```
chunfuwen, line 1310: Here what params_scp makes difference with params_scp in line 1303  are newly added two keys:cacert,cakey. 
So the proper way could be using params_scp.update(xx)

virttest/utils_conn.py
```diff
@@ -1458,6 +1467,42 @@ def server_setup(self, on_local=False):
                                "%s %s" % (self.client_ip, hostname)}
             self.server_hosts.sub_else_add(pattern_to_repl)
 
+    def _update_scp_list(self, scp_list, params):
+        """
+        Generate the scp file list to remote host
```
chunfuwen, line 1472: Generate -- > update ?

chunfuwen (commented):
see inline comments

virttest/utils_conn.py
dzhengfy: update

virttest/utils_conn.py
dzhengfy: This is just to define a new dict variable (params_scp) with different initial values. params_scp is not defined before this statement. So I think params_scp.update() is not proper.

virttest/utils_conn.py
dzhengfy: Update

chunfuwen (approved):
LGTM

Yingshun (approved):

kamilvarga (approved):
LGTM.

PR #3073, opened 2021-05-27
Add machine type option memory-backend

huangyum:
ID: 1926689

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -1438,6 +1438,12 @@ class DevContainer(object):
                         cmd += ",accel=%s" % machine_accel
                     if machine_type_extra_params:
                         cmd += ",%s" % machine_type_extra_params.strip(',')
+                    machine_help = self.execute_qemu("-M %s,?" % machine_type)
+                    opt = "memory-backend"
+                    backend_id = "machine_mem"
+                    if re.search(r"%s=" % opt, machine_help, re.MULTILINE) \
+                            and not params.get("guest_numa_nodes"):
+                        cmd += ",memory-backend=mem-%s" % backend_id
                 else:
                     cmd = ""
                 if 'q35' in machine_type:   # Q35 + ICH9
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -48,6 +48,7 @@ from virttest import error_event
 from virttest.qemu_devices import qdevices, qcontainer
 from virttest.qemu_devices.utils import DeviceError
 from virttest.qemu_capabilities import Flags
+from virttest.utils_params import Params
 
 
 class QemuSegFaultError(virt_vm.VMError):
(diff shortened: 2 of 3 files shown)


huangyum:
Ran huge_page test loop with both qemu6.0 and qemu4.2, all cases passed with correct qemu cli.
@yfu Pls help review, thanks.

virttest/qemu_devices/qcontainer.py
```diff
@@ -1438,6 +1438,15 @@ def machine_other(cmd=False):
                         cmd += ",accel=%s" % machine_accel
                     if machine_type_extra_params:
                         cmd += ",%s" % machine_type_extra_params.strip(',')
+                    machine_help = process.run("%s -M %s,\?" %
+                                               (self.__qemu_binary,
+                                                machine_type),
+                                               shell=True).stdout_text
+                    opt = "memory-backend"
+                    backend_id = "machine_backend"
```
yanan-fu, line 1446: The id is fixed, how do you think about move it to base.cfg, make it clear for user to set params for the memory object ?

virttest/qemu_vm.py
```diff
@@ -996,6 +996,22 @@ def add_memorys(devices, params):
                 for name in params.objects("mem_devs"):
                     dev = devices.memory_define_by_params(params, name)
                     devs.extend(dev)
+            machine_dev = devices.get_by_properties({"type": "machine"})[0]
+            machine_cmd = machine_dev.cmdline_nd()
+            output = re.findall(r",memory-backend=mem-([\w|-]+)", machine_cmd)
+            if output:
+                name = output[0]
+                params["size_%s" % name] = "%sM" % params["mem"]
+                params["use_mem_%s" % name] = "no"
```
yanan-fu, line 1005: For this memory object for machine, can `use_mem` only be `no` ?

huangyum: Yes, it must be 'no', otherwise, a dimm device will generate.

virttest/qemu_devices/qcontainer.py
huangyum: Make sense. Added it to base.cfg, pls help review again.

virttest/qemu_vm.py
yanan-fu: ok, thanks!

yanan-fu:
@luckyh  Could you help review ? Thanks!
1. Check if use the new command in `machine_by_params`, add `memory-backend` option accordingly
2.  `add_memory` use new command or not depends on if `memory-backend` was added to machine.

For hugepage, some new change with the old version: remove the `extra_params += " -mem-path /mnt/kvm_hugepage"` in case cfg, add the string dev based on the hugepage_path in VT. (before, It is tricky, hardcode the mem-patch path in test case to use the same value with the hugepage_path in test_setup.HugePageConfig).

virttest/qemu_devices/qcontainer.py
luckyh: @yanan-fu what is the use case for user specified id?

virttest/qemu_devices/qcontainer.py
```diff
@@ -1438,6 +1438,12 @@ def machine_other(cmd=False):
                         cmd += ",accel=%s" % machine_accel
                     if machine_type_extra_params:
                         cmd += ",%s" % machine_type_extra_params.strip(',')
+                    machine_help = self.execute_qemu("-M %s,?" %machine_type)
```
luckyh, line 1441: @huangyum does it work well with non-x86 platforms?

virttest/qemu_devices/qcontainer.py
yanan-fu: The id is expected to bz fixed, and distinguish with other memory object. Do you have any other suggestion ?

virttest/qemu_devices/qcontainer.py
huangyum: Yes, test passed on ppc host.

huangyum: It works on arm and s390 platforms as well.

luckyh: Get it, thanks!

virttest/qemu_vm.py
```diff
@@ -996,6 +997,24 @@ def add_memorys(devices, params):
...
+                if params.get("hugepage_path"):
+                    backend_options["backend"] = "memory-backend-file"
+                    backend_options["mem-path"] = params["hugepage_path"]
+                backend_param = Params(backend_options)
+                dev = devices.memory_object_define_by_params(backend_param,
+                                                             name)
+                devs.append(dev)
+            elif params.get("hugepage_path"):
+                cmd = "-mem-path %s" % params["hugepage_path"]
+                devs.append(StrDev('extra', cmdline=cmd))
```
luckyh, line 1016: Suggest the name `mem-path` instead of `extra`

huangyum: Thanks, updated.  
Test passed on s390 host. 
@luckyh @yanan-fu Pls help review, thanks.

luckyh (approved):
LGTM

yanan-fu (approved):
LGTM

luckyh:
Thanks.

PR #3074, opened 2021-05-28
Enable multi errors

dzhengfy:
This is to use a seperator to support multiple error messages
detecting.

### DIFF
--- a/virttest/migration.py
+++ b/virttest/migration.py
@@ -700,11 +700,12 @@ class MigrationTest(object):
 
         if status_error:  # Migration should fail
             if err_msg:   # Special error messages are expected
-                if not re.search(err_msg, result.stderr_text.strip()):
-                    raise exceptions.TestFail("Can not find the expected "
-                                              "patterns '%s' in output '%s'"
-                                              % (err_msg,
-                                                 result.stderr_text.strip()))
+                err_msg_list = err_msg.split('|')
+                if not any(re.search(patt.strip(), result.stderr_text.strip())
+                           for patt in err_msg_list):
+                    raise exceptions.TestFail("Expect to match with one of %s,"
+                                              "but failed with output: %s" %
+                                              (err_msg_list, result.stderr_text.strip()))
                 else:
                     logging.debug("It is the expected error message")
             else:


dzhengfy:
The case using single error message:
 (1/1) type_specific.io-github-autotest-libvirt.virsh.live_migration.domjobabort.p2p_live.without_postcopy: PASS (225.27 s)

The case using multiple error messages:
(1/1) type_specific.io-github-autotest-libvirt.virsh.live_migration.native_tls.migrate_x509_no.default_x509_no.by_storage.p2p_live.without_postcopy: PASS (304.30 s)

virttest/migration.py
```diff
@@ -700,11 +700,12 @@ def check_result(self, result, params):
 
         if status_error:  # Migration should fail
             if err_msg:   # Special error messages are expected
-                if not re.search(err_msg, result.stderr_text.strip()):
-                    raise exceptions.TestFail("Can not find the expected "
-                                              "patterns '%s' in output '%s'"
-                                              % (err_msg,
-                                                 result.stderr_text.strip()))
+                err_msg_list = err_msg.split('|')
```
Yingshun, line 703: '|' means 'OR' for a regular expression, why do we need split the string to list here? 
We already have some cases which have multiple error messages, please check them again. Thanks!
```
migration/migrate_vm.cfg:                                    err_msg = "error: internal error: unable to execute QEMU command '(nbd-server-add|block-export-add)': Block node is read-only"
migration/migrate_options_shared.cfg:                                    err_msg = "error: operation failed: migration out job:|client socket is closed"
```

virttest/migration.py
```diff
@@ -700,11 +700,12 @@ def check_result(self, result, params):
...
+                if not any(re.search(patt.strip(), result.stderr_text.strip())
```
chunfuwen, line 704: there is one method named: check_status_output in virttest/utils_test/libvirt.py, not sure whether it can satisfy your request.

virttest/migration.py
dzhengfy: You are right.

dzhengfy:
This seems necessary because '|' can be automatically recognized in regular expression.

PR #3086, opened 2021-06-04
utils_sriov: Add a method to get vf's mac

Yingshun:
This PR adds a method to get the mac address of VF.

`type_specific.io-github-autotest-libvirt.sriov_net_failover.hotplug_hostdev_device_with_teaming: PASS (97.59 s)`

### DIFF
--- a/virttest/utils_sriov.py
+++ b/virttest/utils_sriov.py
@@ -77,6 +77,25 @@ def set_vf_mac(ethname, mac_addr, vf_idx=0, session=None):
             cmd, shell=True, verbose=True, session=session)
 
 
+def get_vf_mac(ethname, vf_idx=0, session=None):
+    """
+    Get admin mac address for VF via 'ip' command.
+
+    :param ethname: The name of the network interface
+    :param vf_idx: The index of VF
+    :param session: The session object to the host
+    :return: VF's admin mac
+    """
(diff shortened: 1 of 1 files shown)


Yingshun:
@yalzhang  could you please help review? Thanks!

yalzhang (approved):

virttest/utils_sriov.py
```diff
@@ -77,6 +77,25 @@ def set_vf_mac(ethname, mac_addr, vf_idx=0, session=None):
             cmd, shell=True, verbose=True, session=session)
 
 
+def get_vf_mac(ethname, vf_idx=0, session=None):
+    """
+    Get mac address for VF
```
yalzhang, line 82: Suggest to add more information here, it may make others confused as vf mac can be the mac we set in the vm, or the output in "ifconfig $vf"(they may be different in some cass). In this function, we get the VF's original admin mac.

Yingshun: Thank you @yalzhang for your comment! I think we should not limit this function to get the VF's **original** admin mac. It can also use to get the mac after setting it to other address.  I did a minor update in docstring, could you please re-review?

yalzhang: You are right. It looks better for me now! :-)

PR #3090, opened 2021-06-07
Add new option

dzhengfy:
This is to provide a new option to customize the removing that is to
only local, only remote, or both local and remote.

### DIFF
--- a/virttest/utils_libvirt/libvirt_config.py
+++ b/virttest/utils_libvirt/libvirt_config.py
@@ -64,6 +64,42 @@ def remove_key_in_conf(value_list, conf_type="libvirtd",
         return target_conf
 
 
+def remove_key_by_mode(value_list, conf_type="libvirtd",
+                       remote_params=None, restart_libvirt=False,
+                       mode='local'):
+    """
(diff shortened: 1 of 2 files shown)


Yingshun (changes requested):
This function is called by remove_key_for_modular_daemon(). If you add this option, we have to think about the testing when the split daemons are enabled.

virttest/utils_libvirt/libvirt_config.py
```diff
@@ -15,7 +15,8 @@
 
 
 def remove_key_in_conf(value_list, conf_type="libvirtd",
-                       remote_params=None, restart_libvirt=False):
+                       remote_params=None, restart_libvirt=False,
+                       mode='local'):
```
chunfuwen, line 19: From the function itself, 'local' removal happen no matter what situation it is.
so remote_params can be used to judge whether it should happen on remote host.
Not sure why we need introduce another parameter: 'mode'

dzhengfy:
I suddenly realize this enhancement's value is not big so far. So I decide to close it for now. I will open it if we have this need in future.

PR #3091, opened 2021-06-10
tpm dev: new attribute persistent_state

yanqzhan:
Add new attribute persistent_state for tpm device, supported from
libvirt 7.0.0.

Signed-off-by: Yanqiu Zhang <yanqzhan@redhat.com>

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -2317,6 +2317,7 @@ def create_tpm_dev(params):
     tpm_model = params.get("tpm_model", 'tpm-crb')
     backend_type = params.get("backend_type")
     backend_version = params.get("backend_version")
+    persistent_state = params.get("persistent_state")
     encryption_secret = params.get("encryption_secret")
     device_path = params.get("device_path")
 
@@ -2327,6 +2328,8 @@ def create_tpm_dev(params):
         tpm_backend.backend_type = backend_type
         if backend_version:
             tpm_backend.backend_version = backend_version
+        if persistent_state:
+            tpm_backend.persistent_state = persistent_state
         if encryption_secret:
             tpm_backend.encryption_secret = encryption_secret
         if device_path:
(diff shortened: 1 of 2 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -2327,6 +2328,8 @@ def create_tpm_dev(params):
         tpm_backend.backend_type = backend_type
         if backend_version:
             tpm_backend.backend_version = backend_version
+        if persistent_state:
```
chloerh, line 2331: According to line 2320, persistent_state won't be 'False' unless being assigned with 'False', empty str/list/dict....  
So basically this line is always True, right?

virttest/utils_test/libvirt.py
```diff
@@ -2317,6 +2317,7 @@ def create_tpm_dev(params):
     tpm_model = params.get("tpm_model", 'tpm-crb')
     backend_type = params.get("backend_type")
     backend_version = params.get("backend_version")
+    persistent_state = params.get("persistent_state", "no")
```
chunfuwen, line 2320: persistent_state may not be set one default value if you want to use it as judgement point in line 2331.

yanqzhan: Done. Thanks. Actually it's not used by tpm_device.py, just update it for other cases using.

virttest/utils_test/libvirt.py
yanqzhan: Modified. Thanks.

dzhengfy (approved):
LGTM

chunfuwen (approved):
LGTM

dzhengfy:
@chloerh , I saw your comment had been solved. The https://github.com/autotest/tp-libvirt/pull/3543 was already merged by accident, so I prefer to merge it now. If you have other questions, let us file new PR to address. Thanks.

PR #3099, opened 2021-06-16
libvirt_disk: create remote disk

dzhengfy:
This is to create a same virtual size empty disk image on remote
host as the first disk of the vm on local host. This is common
used for storage migration test.

### DIFF
--- a/virttest/utils_libvirt/libvirt_disk.py
+++ b/virttest/utils_libvirt/libvirt_disk.py
@@ -10,6 +10,7 @@ from avocado.core import exceptions
 from avocado.utils import process
 
 from virttest import libvirt_storage
+from virttest import remote as remote_old
 from virttest import utils_misc
 from virttest import utils_disk
 from virttest import virsh
@@ -34,7 +35,6 @@ def create_disk(disk_type, path=None, size="500M", disk_format="raw", extra='',
     :return: The path of disk
     :raise: TestError if the disk can't be created
     """
-
     if session:
         if disk_type == "file":
             disk_cmd = ("qemu-img create -f %s %s %s %s"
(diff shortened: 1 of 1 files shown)


virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -510,3 +510,30 @@ def check_in_vm(vm, target, old_parts, is_equal=False):
     except Exception as e:
         logging.error(str(e))
         return False
+
+
+def create_remote_disk_by_first(vm, params):
```
chunfuwen, line 515: If this function aims to create empty image on remote host, I assume it is not related to specific VM.
So vm parameter may not be used

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -510,3 +510,30 @@ def check_in_vm(vm, target, old_parts, is_equal=False):
...
+    """
+    Create an empty image on remote host
+
+    :param vm:  the VM object migrated
+    :param params:  dict, parameters used for migration
```
Yingshun, line 520: Should we remove 'migrated' and 'migration'?

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -510,3 +510,30 @@ def check_in_vm(vm, target, old_parts, is_equal=False):
     except Exception as e:
         logging.error(str(e))
         return False
+
+
+def create_remote_disk_by_first(vm, params):
+    """
+    Create an empty image on remote host
```
Yingshun, line 517: How about making the name and the docstring more specific?
This function is to create a disk with the same name, size, type and path as the VM's first disk on the remote host. It seems that the current ones are too simple.. What do you think?

virttest/utils_libvirt/libvirt_disk.py
dzhengfy: Unfortunately, I can not remove vm parameter because vm is used in the codes. See L528
` blk_source = get_first_disk_source(vm)`

virttest/utils_libvirt/libvirt_disk.py
dzhengfy: updated

virttest/utils_libvirt/libvirt_disk.py
dzhengfy: Change to create_remote_disk_by_same_metadata()

kamilvarga (approved):
LGTM.

Yingshun (approved):

chunfuwen (approved):
LGTM

PR #3105, opened 2021-06-21
add a parameter to set extra verifier 

leidwang:
ID:1926551

### DIFF
--- a/virttest/utils_test/qemu/__init__.py
+++ b/virttest/utils_test/qemu/__init__.py
@@ -137,8 +137,12 @@ def setup_win_driver_verifier(session, driver, vm, timeout=300):
     """
 
     win_verifier_flags = vm.params.get("windows_verifier_flags")
+    extra_verifier = vm.params.get("extra_verifier")
     verifier_status = _check_driver_verifier(session, driver,
                                              win_verifier_flags)[0]
+
+    if extra_verifier:
+        driver = driver + " " + extra_verifier
     if not verifier_status:
         error_context.context("Enable %s driver verifier" % driver,
                               logging.info)


leidwang:
@PaulYuuu @yanglei-rh  Could you help to review this patch? Thanks a lot.

virttest/shared/cfg/guest-os/Windows.cfg
```diff
@@ -41,9 +41,11 @@
     # Set verifier flags for windows guests
     Win2016, Win2019:
         windows_verifier_flags = "0x009209bb"
+        windows_verifier_sys = "netkvm ndis"
 
     Win8..1, Win2012..r2:
         windows_verifier_flags = "0x008209bb"
+        windows_verifier_sys = "netkvm ndis"
```
lijinlijin, line 48: We only want to enable netkvm and ndis for netkvm cases, other features' case need to enable their own features' files.
For your changes, all feature will enable netkvm ndis, that's not we expect.

leidwang: Okay, will update the code.

yanglei-rh:
@leidwang Please carefully check the bug's comments 10 and 13, your current usage is inconsistent with what the development expressed.

leidwang:
> @leidwang Please carefully check the bug's comments 10 and 13, your current usage is inconsistent with what the development expressed.

Okay, I think you may have misunderstood something, we can discuss it later.

leidwang:
> I don’t know much about windows guests testing, but seems this should not an enhancement of VT, only some specific test cases need it as @lijinlijin said.
> 
> So, is better to modify the `driver_name` of related netkvm test cases in tp-qemu?
> 
> Or append `ndis` when driver contains `netkvm`?
Do you mean to modify the code like this？

`extra_verifier_sys = "ndis"`

`if extra_verifier_sys and "netkvm" in driver:
        driver += extra_verifier_sys`
> https://github.com/avocado-framework/avocado-vt/blob/78dc750968b009ebfe42ab34c5b403e80d35bc9a/virttest/utils_test/qemu/__init__.py#L129



> I don’t know much about windows guests testing, but seems this should not an enhancement of VT, only some specific test cases need it as @lijinlijin said.
> 
> So, is better to modify the `driver_name` of related netkvm test cases in tp-qemu?
> 
> Or append `ndis` when driver contains `netkvm`?
Do you mean to modify the code like this？

`extra_verifier_sys = "ndis"`

`if extra_verifier_sys and "netkvm" in driver:
        driver += extra_verifier_sys`

> https://github.com/avocado-framework/avocado-vt/blob/78dc750968b009ebfe42ab34c5b403e80d35bc9a/virttest/utils_test/qemu/__init__.py#L129

PaulYuuu:
> > I don’t know much about windows guests testing, but seems this should not an enhancement of VT, only some specific test cases need it as @lijinlijin said.
> > So, is better to modify the `driver_name` of related netkvm test cases in tp-qemu?
> > Or append `ndis` when driver contains `netkvm`?
> > Do you mean to modify the code like this？
> 
> `extra_verifier_sys = "ndis"`
> 
> `if extra_verifier_sys and "netkvm" in driver: driver += extra_verifier_sys`
> 
> > https://github.com/avocado-framework/avocado-vt/blob/78dc750968b009ebfe42ab34c5b403e80d35bc9a/virttest/utils_test/qemu/__init__.py#L129

Probably like this. For this change, we need to bin `netkvm` and `ndis` as a group.

leidwang:
> > > I don’t know much about windows guests testing, but seems this should not an enhancement of VT, only some specific test cases need it as @lijinlijin said.
> > > So, is better to modify the `driver_name` of related netkvm test cases in tp-qemu?
> > > Or append `ndis` when driver contains `netkvm`?
> > > Do you mean to modify the code like this？
> > 
> > 
> > `extra_verifier_sys = "ndis"`
> > `if extra_verifier_sys and "netkvm" in driver: driver += extra_verifier_sys`
> > > https://github.com/avocado-framework/avocado-vt/blob/78dc750968b009ebfe42ab34c5b403e80d35bc9a/virttest/utils_test/qemu/__init__.py#L129
> 
> Probably like this. For this change, we need to bin `netkvm` and `ndis` as a group.
> If `'netkvm' in driver and 'ndis' not in driver`, append ndis, and vice versa.

Let me explain, we only need to add "ndis" to some windows guests, not all windows guests. So I added a parameter(extra_verifier_sys) to cfg

PaulYuuu:
With your patch, `driver` will no longer be fixed, when `driver` includes `netkvm`, it will become `win_verifier_sys`, no matter what the content of the driver is. This should not a good idea.

Hello @vivianQizhu , would you possibly help review this patch as well? Since I am not an expert in windows testing, I think the fundamental solution needs to be determined together.

vivianQizhu:
I agree with @PaulYuuu that this should not be implemented at avocado-vt, since the module actually fully support multiple driver with `verifier` related command lines: `verifier_setup_cmd = "verifier /standard /driver %s.sys" % (".sys ".join(driver.split()))`, it is not proper to specifically make up `netkvm` with `netkvm ndis` within the framework module. I understand you might need to change a lot files if you replace all the `netkvm` with `netkvm ndis` in tp-qemu, but that's the proper way from my perspective.

leidwang:
> I agree with @PaulYuuu that this should not be implemented at avocado-vt, since the module actually fully support multiple driver with `verifier` related command lines: `verifier_setup_cmd = "verifier /standard /driver %s.sys" % (".sys ".join(driver.split()))`, it is not proper to specifically make up `netkvm` with `netkvm ndis` within the framework module. I understand you might need to change a lot files if you replace all the `netkvm` with `netkvm ndis` in tp-qemu, but that's the proper way from my perspective.

Okay,agree with you,will fix this issue in tp-qemu.
Thanks all.

PR #3112, opened 2021-06-28
Add sharebacking attribute to disk xml

MaXinjian:
Add support for the following libvirt new features:
```
v7.4.0

qemu: Add support for sharing base image of <transient/> disks

Users can use <transient shareBacking='yes'/> to tell the qemu driver
 to never open the base image in write mode thus multiple VMs can share
the same image. Note that the disk will be hotplugged during startup.
```

Signed-off-by: Ma Xinjian <maxj.fnst@fujitsu.com>

### DIFF
--- a/virttest/libvirt_xml/devices/disk.py
+++ b/virttest/libvirt_xml/devices/disk.py
@@ -18,6 +18,8 @@ class Disk(base.TypedDeviceBase):
             string, how exposted to guest
         rawio:
             string (yes/no), disk needs rawio capability
+        sharebacking:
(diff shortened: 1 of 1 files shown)


MaXinjian:
@chunfuwen
This PR is related to new transient sharebacking test cases I have already send [PR](https://github.com/autotest/tp-libvirt/pull/3587) in tp-libvirt.

chunfuwen (approved):
LGTM

dzhengfy (approved):
LGTM

PR #3120, opened 2021-07-01
virt_vm: allow use a serial connection when remote login failed

yanan-fu:
Some cases test with vm without nic need a serial session, make
it is available.

id: 1968402

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -521,7 +521,7 @@ def session_handler(func):
                 vm.connect_uri = uri
                 vm.session = vm.wait_for_serial_login()
             else:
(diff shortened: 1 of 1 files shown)


yanan-fu:
@vivianQizhu  @PaulYuuu  Could you help review ? Thanks!

vivianQizhu (approved):
Checked all usage of `session_handler()`, for all those functions it should be okay to login with serial when nic is not available, and I do not see any risk to escape bug for missing nic.  So ACK.

PaulYuuu:
Yep, the serial session can be used as the alternate session. ACK

PR #3122, opened 2021-07-02
utils_package: print error log

dzhengfy:
This is to print the error message when installation fails.

### DIFF
--- a/virttest/utils_package.py
+++ b/virttest/utils_package.py
@@ -106,14 +106,21 @@ class RemotePackageMgr(object):
                 need = True
             if need:
                 cmd = self.cmd + pkg
-                if self.session.cmd_status(cmd, timeout, internal_timeout):
+                status, output = self.session.cmd_status_output(cmd,
+                                                                timeout,
+                                                                internal_timeout)
+                if status:
+                    logging.error("'%s' execution failed "
+                                  "with %s", cmd, output)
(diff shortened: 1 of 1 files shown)


Yingshun (approved):

virttest/utils_package.py
```diff
@@ -106,14 +106,21 @@ def operate(self, timeout, default_status, internal_timeout=2):
                 need = True
             if need:
                 cmd = self.cmd + pkg
-                if self.session.cmd_status(cmd, timeout, internal_timeout):
+                status, output = self.session.cmd_status_output(cmd,
+                                                                timeout,
+                                                                internal_timeout)
+                if status:
+                    logging.error("Operate %s with %s "
+                                  "failed:%s", pkg, cmd, output)
```
kamilvarga, line 114: pkg is already part of cmd from line 108, do we really want to output it twice?

dzhengfy: @kamilvarga , ok, I just copied from original codes. Now I have refined it.

chloerh (approved):

kamilvarga (approved):
LGTM.

dzhengfy:
As this fix is only for the installation failure which is not easy to reproduce, I will not paste test result. Anyway this fix is simple and low risk. Merge it now.

PR #3128, opened 2021-07-08
virsh: set default timeout for snapshot_create_as

liuyd96:
Signed-off-by: Liu Yiding <liuyd.fnst@fujitsu.com>

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -3052,7 +3052,10 @@ def snapshot_create_as(name, options="", **dargs):
     if options is not None:
         cmd += " %s" % options
 
-    return command(cmd, **dargs)
+    # Avoid snapshot-create-as cmd hang on
+    timeout = dargs.get('timeout', 300)
+
+    return command(cmd, timeout=timeout, **dargs)
 
 
 def snapshot_parent(name, options, **dargs):


liuyd96:
Hi, @kylazhang @chloerh 

Particular snapshot cases sometimes TIMEOUT in CI env and I can't reproduce TIMEOUT issue mannually.

Our CI interupt TIMEOUT job directly so cleanup steps don't work. That block other snapshot jobs.
 
I checked the log and found them stuck on virsh.snapshot_create_as.

virttest/virsh.py
```diff
@@ -3052,7 +3052,10 @@ def snapshot_create_as(name, options="", **dargs):
     if options is not None:
         cmd += " %s" % options
 
-    return command(cmd, **dargs)
+    # Avoid snapshot-create-as cmd hang on
+    timeout = dargs.get('timeout', 300)
```
chloerh, line 3056: I think we could just run with 
```
virsh.snapshot_create(vm_name, options, timeout=300)
```
in tp-libvirt, it should work.

liuyd96: You mean virsh.snapshot_create_as(vm_name, options, timeout=300) right?
 
Of course that should work. 
But i think set default timeout inside func snapshot_create_as is much better,
Reason
1. Many jobs have already introduced snapshot_create_as with setting timeout option
2. When devlopers don't set timeout option these cases which call snapshot_create_as may be TIMEOUT in CI env.
 
To avoid this situation, i think a default timeout inside snap_create_as is needed or make timeout option a must.

chloerh: I'm not sure about that. My opinion is we shouldn't set a fixed timeout just because our ci is limiting time of a case. Sometimes the image could be big, or the host performance is not so good, that could cause the execution time longer than usual. 
Anyway, I'll ask @kylazhang 's opinion too.

Yingshun: @liuyd96 I agree with @chloerh . I think timed out issue in CI may caused by several reasons, ci timeout setting, case setting or product issue or a performance issue as @chloerh mentioned. As a common func, I think we shouldn't set it.

liuyd96: Ok, i will close this PR and try other solutions.
Thanks.

liuyd96:
Reopen this PR. I will specify this PR in CI env to debug the root cause.
Please ignore this PR. Thanks.

PR #3132, opened 2021-07-09
libvirt_disk: update vm disk target

dzhengfy:
This is to provide an utility function to update the vm disk's
target attributes.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1619,6 +1619,7 @@ def create_disk_xml(params):
     type_name = params.get("type_name", "file")
     target_dev = params.get("target_dev", "vdb")
     target_bus = params.get("target_bus", "virtio")
+    target_rotation = params.get("target_rotation")
     diskxml = disk.Disk(type_name)
     diskxml.device = params.get("device_type", "disk")
     snapshot_attr = params.get('disk_snapshot_attr')
@@ -1757,6 +1758,8 @@ def create_disk_xml(params):
         diskxml.readonly = "yes" == params.get("readonly", "no")
         diskxml.share = "yes" == params.get("shareable", "no")
         diskxml.target = {'dev': target_dev, 'bus': target_bus}
+        if target_rotation:
+            diskxml.target = dict(diskxml.target, **{'rotation_rate': target_rotation})
         alias = params.get('alias')
         if alias:
             diskxml.alias = {'name': alias}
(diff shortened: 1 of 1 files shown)


virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -538,3 +538,38 @@ def create_remote_disk_by_same_metadata(vm, params):
 
     remote_session.close()
     return blk_source
+
+
+def update_vm_disk_target(vm_name, target_dict, disk_index=0):
```
chunfuwen, line 543: @dzhengfy 
There is existed method libvirt.set_vm_disk(vm, params, tmp_dir) to achieve update disk.
So is there other requirements that set_vm_disk can not satisfy ,and therefore create this new one?

dzhengfy: @chunfuwen The main reasons are below:
1. set_vm_disk() is too complex and readability is not good any more. 
2. I need to support new attribute in <target rotation_rate='' >, but set_vm_disk() does not support. I would not like to add new support to that function.
So I need a new style function to update disk target attributes, instead of hard coding some attributes in the function

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -538,3 +538,44 @@ def create_remote_disk_by_same_metadata(vm, params):
...
+        devices = vmxml.devices
+        device_index = devices.index(devices.by_device_tag('disk')[disk_index])
+        disk = devices[device_index]
+
+        # Update disk target attributes
+        current_target_dict = disk.target
+        for key, value in target_dict.items():
+            current_target_dict[key] = value
+
+        disk.target = current_target_dict
```
Yingshun, line 566:
```suggestion
        disk.target = dict(disk.target, **target_dict)
```

virttest/utils_libvirt/libvirt_disk.py
chunfuwen: @dzhengfy, here is what we need do if update existed disk.
And you need update target, it also allow you update disk_target and disk_target_bus accordingly.

    disk_target = params.get("disk_target", 'vda')
    disk_target_bus = params.get("disk_target_bus", "virtio")
```
            params.update({'disk_source_name': replace_disk_image,
                           'disk_type': 'file',
                           'disk_format': 'qcow2',
                           'disk_source_protocol': 'file'})
            libvirt.set_vm_disk(vm, params, tmp_dir)
```

virttest/utils_libvirt/libvirt_disk.py
dzhengfy: @Yingshun Thanks for your comments and I learned much.

virttest/utils_libvirt/libvirt_disk.py
dzhengfy: @chunfuwen  Update to use set_vm_disk(). Please review again. Thanks.

virttest/utils_libvirt/libvirt_disk.py
dzhengfy: @Yingshun I used it in new place, Please help review. Thanks.

Yingshun (approved):
LGTM

chunfuwen (approved):
LGTM

PR #3135, opened 2021-07-12
libvirt_vmxml: add func to check dumpxml

dzhengfy:

### DIFF
--- a/virttest/utils_libvirt/libvirt_vmxml.py
+++ b/virttest/utils_libvirt/libvirt_vmxml.py
@@ -5,7 +5,11 @@ http://libvirt.org/formatdomain.html
 
 
 import logging
+import re
 
+from avocado.core import exceptions
+
+from virttest import virsh
 from virttest.libvirt_xml import vm_xml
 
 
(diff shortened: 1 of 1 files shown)


Yingshun (approved):
LGTM

virttest/utils_libvirt/libvirt_vmxml.py
```diff
@@ -25,6 +29,26 @@ def set_vm_attrs(vmxml, vm_attrs):
...
+def check_guest_xml(vm_name, pat_in_dumpxml, option='', status_error=False):
+    """
+    Check the given pattern in the vm dumpxml
+
+    :param vm_name: vm name
+    :param pat_in_dumpxml:  str, the pattern to search in dumpxml
+    :param status_error: True if expect not existing, otherwise False
+    :raises: TestFail if the result is not expected
+    """
+    ret_stdout = virsh.dumpxml(vm_name, extra=option).stdout.strip()
```
chunfuwen, line 41: change 
stdout -->stdout_text ?

dzhengfy: @chunfuwen , when I check def command() in virsh.py, I find below in Line 812,813, so I think here 'stdout' is ok to use. How do you think of?
```
        ret.stdout = ret.stdout_text
        ret.stderr = ret.stderr_text
```

chunfuwen (approved):
LGTM

PR #3141, opened 2021-07-14
update iscsi setup and cleanup in preprocess and postprocess functions

chunfuwen:
In libvirt, it will call libvirt.setup_or_cleanup_iscsi to setup or cleanup iscsi storage
Therefore, update the code to avoid bringing up some noises here

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1037,7 +1037,7 @@ def preprocess(test, params, env):
         brcfg.setup()
 
     base_dir = data_dir.get_data_dir()
-    if params.get("storage_type") == "iscsi":
+    if params.get("storage_type") == "iscsi" and vm_type != "libvirt":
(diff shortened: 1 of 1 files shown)


virttest/env_process.py
```diff
@@ -1037,7 +1037,7 @@ def preprocess(test, params, env):
         brcfg.setup()
 
     base_dir = data_dir.get_data_dir()
-    if params.get("storage_type") == "iscsi":
+    if params.get("storage_type") == "iscsi" and vm_type != "libvirt":
```
kylazhang, line 1040: I think "lvm" type prepare can also be ignored for libvirt as I saw it also use qemu_storage function. Could you please deal with it also?

dzhengfy: Agree. We can also fix lvm part which is definitely not used by libvirt cases.

chunfuwen:
close it firstly after it need more validation

PR #3146, opened 2021-07-21
utils_sriov: Add a function to get pf info by the given pci id

Yingshun:
Some tests require specific pf details, so add a function to get
pf information by the given pci id.

Test results:
https://github.com/autotest/tp-libvirt/pull/3658#issue-694170220

### DIFF
--- a/virttest/utils_sriov.py
+++ b/virttest/utils_sriov.py
@@ -82,7 +82,7 @@ def get_pf_info(session=None):
                              'status': utils_net.get_net_if_operstate(
                                           iface_name.strip(), runner=runner)})
             pf_info.update({pci: tmp_info})
-
+    logging.debug("PF info: %s.", pf_info)
     return pf_info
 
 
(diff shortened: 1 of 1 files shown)


virttest/utils_sriov.py
```diff
@@ -99,6 +99,23 @@ def get_pf_pci(session=None):
...
+    :return: Dict, pf's info,
+        eg. {'driver': 'ixgbe', 'pci_id': '0000:3b:00.0',
+             'iface': 'ens1f0', 'status': 'up'}
+    """
+    pf_info = get_pf_info(session=session)
+    for pf in pf_info.values():
+        if pf.get('pci_id') == pci_id:
+            logging.debug("PF %s details: %s.", pci_id, pf)
+            return pf
+
```
chunfuwen, line 117: Better explicitly return None if pci_id is not anticipated one

Yingshun: I don't think so. It's redundant. May I ask why you want that?

Yingshun:
@yalzhang @chloerh would you please help review? Thanks!

chloerh (approved):

chloerh:
@yalzhang we're waiting for your review, thanks. :handshake:

virttest/utils_sriov.py
```diff
@@ -99,6 +99,23 @@ def get_pf_pci(session=None):
...
+    """
+    Get the pci info by the given pci id
+
+    :param pci_id: PF's pci id, eg. 0000:3b:00.0
+    :param session: The session object to the host
+    :return: Dict, pf's info,
+        eg. {'driver': 'ixgbe', 'pci_id': '0000:3b:00.0',
+             'iface': 'ens1f0', 'status': 'up'}
+    """
+    pf_info = get_pf_info(session=session)
```
yalzhang, line 112: how about the get_pf_info() return None? may need add an judgement here.

Yingshun: If there is no pfs, it'll return '{}' and the following for loop will do nothing.

yalzhang (approved):

virttest/utils_sriov.py
chunfuwen: @Yingshun , perhaps it is my personal practice that what if you return sth in some function , we expect have consistent return value whatever condition  it meets. Maybe bring it to PR discussion for other opinoins.

chunfuwen (commented):
see inline comments

virttest/utils_sriov.py
Yingshun: How about we agree to disagree? I'd prefer to keep the current way, but I wouldn't disagree with others if they explicitly return 'None' at the end of the function. 
BTW, is this a blocker for this function? If not, I think maybe we could continue to review. What do you think?

chunfuwen (approved):
LGTM

PR #3154, opened 2021-07-28
Fix the issue in modify_vm_iface()

yalzhang:
When there is no interface device in the vm xml, it will create a new one.
But in previous code, it forget to add it into the vm xml. If the oper is
"update_iface", the function will not work as expected.

Signed-off-by: Yalan <yalzhang@redhat.com>


Yingshun (approved):
LGTM

yalzhang:
@chloerh  Please help to reivew, Thank you!

kylazhang (approved):

PR #3155, opened 2021-07-28
Add one function to ensure default network exists

yalzhang:
Ensure the default network exists and in the expected status.

### DIFF
--- a/virttest/utils_libvirt/libvirt_network.py
+++ b/virttest/utils_libvirt/libvirt_network.py
@@ -163,3 +163,18 @@ def modify_network_xml(net_dict, testnet_xml):
         net_outbound = ast.literal_eval(net_bandwidth_outbound)
         testnet_xml.bandwidth_outbound = net_outbound
     return testnet_xml
+
+
+def ensure_default_network():
+    """
+    Ensure the default network exists on the host and in active status
+    :return: None
+    """
+    net_state = virsh.net_state_dict()
+    if 'default' not in net_state:
+        # define and start the default network
+        virsh.net_define("/usr/share/libvirt/networks/default.xml",
+                         debug=True, ignore_status=False)
+    if not net_state["default"].get("active"):
+        virsh.net_start("default", debug=True, ignore_status=False)
+        virsh.net_autostart("default", debug=True, ignore_status=False)


virttest/utils_net.py
```diff
@@ -4467,3 +4468,19 @@ def delete_linux_bridge_tmux(linux_bridge_name, iface_name=None, ignore_status=F
     else:
         cmd = 'ip link delete %s' % linux_bridge_name
     return utils_misc.cmd_status_output(cmd, shell=True, verbose=True, ignore_status=ignore_status)
+
+
+def ensure_default_network():
```
Yingshun, line 4473: I think utils_net module should be focus on system network, so how about adding it to utils_libvirt.libvirt_network?

yalzhang: Updated, Thank you!

virttest/utils_libvirt/libvirt_network.py
```diff
@@ -163,3 +163,19 @@ def modify_network_xml(net_dict, testnet_xml):
...
+    """
+    net_state = virsh.net_state_dict()
+    if 'default' not in net_state:
+        # define and start the default network
+        virsh.net_define("/usr/share/libvirt/networks/default.xml",
+                         debug=True, ignore_status=False)
+        virsh.net_start("default", debug=True, ignore_status=False)
+        virsh.net_autostart("default", debug=True, ignore_status=False)
+    elif not net_state["default"].get("active"):
+        virsh.net_start("default", debug=True, ignore_status=False)
```
chunfuwen, line 181: Do we need also apply virsh.net_autostart("default", debug=True, ignore_status=False) after this?

virttest/utils_libvirt/libvirt_network.py
```diff
@@ -163,3 +163,20 @@ def modify_network_xml(net_dict, testnet_xml):
...
+        virsh.net_autostart("default", debug=True, ignore_status=False)
```
Yingshun, line 182: How about changing it like:
```suggestion
    if 'default' not in net_state:
        # define and start the default network
        virsh.net_define("/usr/share/libvirt/networks/default.xml",
                         debug=True, ignore_status=False)
    if not net_state["default"].get("active"):
        virsh.net_start("default", debug=True, ignore_status=False)
        virsh.net_autostart("default", debug=True, ignore_status=False)
```

virttest/utils_libvirt/libvirt_network.py
yalzhang: you are right. Updated.

virttest/utils_libvirt/libvirt_network.py
yalzhang: Updated, Thank you!

Yingshun (approved):

chunfuwen (approved):
LGTM

PR #3162, opened 2021-08-02
Add support to get cpu model and flags from virsh

huangyum:
ID: 1989001


huangyum:
@yanan-fu @luckyh Would you pls help review? 
Tested with x86_cpu_flags and cpu hotplug test cases with libvirt policy, test passed, and the generated qemu cli is as expected. Thanks.

virttest/env_process.py
```diff
@@ -1167,8 +1167,25 @@ def preprocess(test, params, env):
...
-            env["cpu_model"] = cpu.get_qemu_best_cpu_model(params)
-        params["cpu_model"] = env.get("cpu_model")
+            policy_map = {"libvirt": "get_cpu_info_from_virsh",
+                          "virttest": "get_qemu_best_cpu_info"}
+            auto_cpu_policy = params.get("auto_cpu_policy", "virttest").split()
+            for policy in auto_cpu_policy:
+                func = getattr(cpu, policy_map[policy])
+                cpu_info = func(params)
+                if cpu_info:
+                    break
```
luckyh, line 1177:
```python
try:
    cpu_info = func(params)
    if cpu_info:
        break
except:
    logging.error(...)
    continue
```

virttest/env_process.py
```diff
@@ -1167,8 +1167,25 @@ def preprocess(test, params, env):
...
+            else:
+                test.cancel("Failed to get cpu model with policy %s"
+                            % auto_cpu_policy)
+
+            env["cpu_model"] = cpu_info["cpu_model"]
+            env["cpu_flags"] = cpu_info["cpu_flags"]
+            params["cpu_model"] = env["cpu_model"]
```
luckyh, line 1184: This line has to be out of the `if` clause, right?

virttest/env_process.py
```diff
@@ -1167,8 +1167,25 @@ def preprocess(test, params, env):
...
+            if env["cpu_flags"]:
+                cpu_flags = params.get("cpu_model_flags")
+                params["cpu_model_flags"] = cpu.combine_flags(env["cpu_flags"],
+                                                              cpu_flags)
```
luckyh, line 1188: And also, this block.

luckyh: And we'd better leave a comment like:
```python
# we assumed that `env["cpu_flags"]` would always be existing if there is `env["cpu_model"]`
# hence we do not need to call the above routine for 'cpu_flags' only
if env["cpu_flags"]:
    ...
```

virttest/cpu.py
```diff
@@ -1258,3 +1260,108 @@ def get_qemu_best_cpu_model(params):
...
+                </os>
+                <cpu mode='host-model'/>
+            </domain>
+          """ % (name, arch, machine)
+    xml_file = os.path.join(data_dir.get_tmp_dir(), "xml")
+    with open(xml_file, "w") as f:
+        f.write(xml)
+    try:
+        logging.info("Get cpu model and flags from virsh")
+        virsh.define(xml_file)
```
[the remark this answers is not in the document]
huangyum: I guess there might be error when virsh.define ?

virttest/cpu.py
```diff
@@ -1258,3 +1260,108 @@ def get_qemu_best_cpu_model(params):
...
+    cpu_info["cpu_model"] = cpu_model
+    cpu_info["cpu_flags"] = cpu_flags
+    return cpu_info
+
+
+def get_cpu_info_from_virsh(params):
+    """
+    Try to get cpu model and flags from virsh with 'host-model'
+    """
+    arch = params["vm_arch_name"]
```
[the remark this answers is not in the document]
huangyum: Confirmed with @nanliu-r, it's host arch. Have updated accordingly, thanks.

virttest/env_process.py
```diff
@@ -1167,8 +1167,32 @@ def preprocess(test, params, env):
     if (params.get("auto_cpu_model") == "yes" and
             vm_type == "qemu"):
         if not env.get("cpu_model"):
```
huangyum, line 1169: Seems env is not cleared when case finishes. If the next case runs with different machine type, the cpu info stored in env might be incorrect. Shall we remove this if condition here?

huangyum: Not only when machine type change, the next case might use different policies, then the env['cpu_*'] is definitely not what it needs. So I remove it. Thanks.

virttest/cpu.py
```diff
@@ -1258,3 +1261,132 @@ def get_qemu_best_cpu_model(params):
...
+            <domain type='kvm'>
+                <name>%s</name>
+                <memory>1</memory>
+                <os>
+                    <type arch='%s' machine='%s'>hvm</type>
+                </os>
+                <cpu mode='host-model'/>
+            </domain>
+          """ % (name, vm_arch, machine)
+    xml_file = os.path.join(data_dir.get_tmp_dir(), "temp_xml_for_cpu")
```
yanan-fu, line 1334: The `xml_file` is not unique if machine changed ?  Suggest use the same strategy for cache_file.

huangyum: It's just a temporary file. We will never reuse it. So I think the file name doesn't matter.

virttest/cpu.py
```diff
@@ -1258,3 +1261,132 @@ def get_qemu_best_cpu_model(params):
...
+        return cpu_info
+    finally:
+        if virsh.is_alive(name):
+            virsh.destroy(name, ignore_status=True)
+        virsh.undefine(name, ignore_status=True)
+
+
+def combine_flags(flags_dict, flags_string):
+    """
+    Combine the flags in flags_dict and flags_string
```
yanan-fu, line 1370: A bit suggestion about the parameter name,  how do you think about indicate the source of the flags, and the format of the flags can be explained with the doc:
```
def combine_flags(auto_cpu_flags, custom_cpu_flag):
```

virttest/cpu.py
yanan-fu: Yes, it is true, since use `w` every time. i wonder if this step can be optimized to use an unique xml and reuse it ?
@luckyh What's your opinion about this ?

virttest/cpu.py
```diff
@@ -1258,3 +1261,132 @@ def get_qemu_best_cpu_model(params):
...
+    """
+    Get the best cpu model and features available for qemu
+    """
+    cache_file = "qemu_best_cpu_cache"
+    cache_path = os.path.join(data_dir.get_tmp_dir(), cache_file)
+    cpu_info = get_cpu_info_from_cache(cache_path)
+    if not cpu_info:
+        cpu_info = {}
+        cpu_model = get_qemu_best_cpu_model(params)
+        cpu_features = ""
```
yanan-fu, line 1297: How about add a short comment for this line to explain it is for better expansion in the future ? Someone may wonder why not combine it to L1299 directly.

virttest/cpu.py
huangyum: I guess the function itself doesn't care if the flags are auto generated or defined by user. It cares more about the param type. I think the variable names are more useful when the function is called other than defined. Thanks.

yanan-fu: It depends on the purpose of this function, whether the implemention of this function is available for other dict + string combination, not only cpu flags. IMO, it is for cpu flags only in cpu.py,  correct me if my understanding is wrong.

virttest/cpu.py
huangyum: Good idea. Do you have any suggestion on the comment?

virttest/cpu.py
huangyum: I'm not quite sure about the dependency.  Yes, it's only for cpu flags now. But as to function definition has to indicate the param source, I doubt it.

virttest/env_process.py
```diff
@@ -1166,9 +1166,29 @@ def preprocess(test, params, env):
...
+            test.cancel("Failed to get cpu info with policy %s"
+                        % auto_cpu_policy)
+
+        env["cpu_model"] = cpu_info["model"]
+        env["cpu_flags"] = cpu_info["features"]
+        params["cpu_model"] = env["cpu_model"]
+        if env["cpu_flags"]:
+            cpu_flags = params.get("cpu_model_flags")
+            params["cpu_model_flags"] = cpu.combine_flags(env["cpu_flags"],
+                                                          cpu_flags)
```
yanan-fu, line 1191: `env['cpu_flags']`  is a new env params,  what exactly does it mean ?
I wonder do we need to update `env['cpu_flags']`  after combine ?  Indicate the cpu flags that be used in vm finally.

yanan-fu: Or, no need to save cpu_flags with env, the new design use the cache file strategy, and keep `env['cpu_model']` just to compatible with the older code.

huangyum:
@luckyh @yanan-fu @nanliu-r There is a small change that we decided to get the cpu info from qemu cli instead of dumpxml now. In case of libvirt bug, we also analyze dumpxml to check if the flags are same with qemu cli. Test passed on my side. Please help review, thanks.

virttest/cpu.py
yanan-fu: No cache file, no env,  so use fresh xml file each time, ack for this.

virttest/cpu.py
```diff
@@ -1258,3 +1261,111 @@ def get_qemu_best_cpu_model(params):
...
+    except Exception as err:
+        logging.error(err)
+        return
+    else:
+        cpu_info = get_cpu_info_from_qemu_cli(name)
+        return cpu_info
+    finally:
+        if virsh.is_alive(name):
+            virsh.destroy(name, ignore_status=True)
+        virsh.undefine(name, ignore_status=True)
```
yanan-fu, line 1355: With the latest version,  use virsh start vm for every time, i think `ignore_status=True` is not suitable.

huangyum: Could you explain more? I don't get it.

yanan-fu: Define multiple vms with different name is ok, so after thinking about this more, i am ok with `ignore_status=True`.

One more questions, will these vms be cleaned at the end of test by force ?

virttest/cpu.py
yanan-fu: `# Expend if needed`

virttest/cpu.py
```diff
@@ -1258,3 +1261,111 @@ def get_qemu_best_cpu_model(params):
...
+def get_cpu_info_from_qemu_cli(name):
+    """
+    Get cpu info from qemu cmdline
+    """
+    cmd = "ps aux | grep qemu | grep -e '\s-name\s\+guest=%s\(\,\|\s\)'" % name
+    output = process.getoutput(cmd, shell=True)
+    cpu_line = re.findall("-cpu\s+(\S*)\s+", output)[0]
+    sep = cpu_line.index(",")
+    cpu_model = cpu_line[:sep]
+    cpu_flags = cpu_line[sep:]
```
yanan-fu, line 1307:
```
cpu_model, cpu_flags = cpu_line.split(',',1)
```

virttest/cpu.py
huangyum: No. I'm not sure there is an alternative way to undefine. So far, I never hit undefine failure when domain exists.

virttest/cpu.py
huangyum: Thanks, updated, pls help review again.

virttest/cpu.py
yanan-fu: Yes, ignore_status = True is used to ignore the fails here. As my comment above, considering define multiple vms with different name is ok, so i am ok with ignore the fails in this step.  we are on the same page here.

My concern is about the failed situation, vm does not be destroyed even if no effect for the test.
But, after checked libvirt side,  i found `virsh destroy $domain` is destroy vm from host side with signal, and i think it is reliable to destroy vm, so no concern now, thanks!

huangyum: > Yes, ignore_status = True is used to ignore the fails here. As my comment above, considering define multiple vms with different name is ok, so i am ok with ignore the fails in this step. we are on the same page here.
> 
Actually I mean it's unlikely to fail undefine if define succeeds at beginning. So even vms have same names, there shouldn't have conflicts.

> My concern is about the failed situation, vm does not be destroyed even if no effect for the test.
> But, after checked libvirt side, i found `virsh destroy $domain` is destroy vm from host side with signal, and i think it is reliable to destroy vm, so no concern now, thanks!

Thanks for the check and review!

virttest/cpu.py
```diff
@@ -1258,3 +1261,110 @@ def get_qemu_best_cpu_model(params):
...
+    """
+    Get cpu info from qemu cmdline
+    """
+    cmd = "ps aux | grep qemu | grep -e '\s-name\s\+guest=%s\(\,\|\s\)'" % name
+    output = process.getoutput(cmd, shell=True)
+    cpu_line = re.findall("-cpu\s+(\S*)\s+", output)[0]
+    cpu_model, cpu_flags = cpu_line.split(',', 1)
+    cpu_info = {}
+    cpu_info["model"] = cpu_model
+    cpu_info["flags"] = cpu_flags
```
luckyh, line 1309:
```suggestion
    cpu_cmd = cpu_line.split(',', 1)
    cpu_info = {}
    cpu_info["model"] = cpu_cmd[0]
    cpu_info["flags"] = "" if len(cpu_cmd) < 2 else cpu_cmd[1]
```

virttest/env_process.py
```diff
@@ -1164,26 +1162,6 @@ def preprocess(test, params, env):
...
-        if not env.get("cpu_model"):
-            env["cpu_model"] = cpu.get_qemu_best_cpu_model(params)
-        params["cpu_model"] = env.get("cpu_model")
-
-    if vm_type == "qemu":
-        qemu_path = utils_misc.get_qemu_binary(params)
-        if (utils_qemu.has_device_category(qemu_path, "CPU")
-                and params.get("cpu_driver") is None):
-            cpu_model = params.get("cpu_model")
-            if cpu_model:
```
luckyh, line 1178: Would you please elaborate on why this check can be removed for the new approach?

virttest/cpu.py
```diff
@@ -1258,3 +1261,110 @@ def get_qemu_best_cpu_model(params):
...
+    else:
+        cpu_info = get_cpu_info_from_qemu_cli(name)
+        return cpu_info
+    finally:
+        if virsh.is_alive(name):
+            virsh.destroy(name, ignore_status=True)
+        virsh.undefine(name, ignore_status=True)
+
+
+def combine_flags(auto_cpu_flags, custom_cpu_flags):
```
luckyh, line 1357:
```suggestion
def recombine_qemu_cpu_flags(base, suggestion):
```

virttest/env_process.py
huangyum: Because cpu_model is always assigned a value via L1626~L1642 of qemu_vm.py.

virttest/cpu.py
```diff
@@ -1258,3 +1261,110 @@ def get_qemu_best_cpu_model(params):
...
+                flags_dict[item[1:]] = "off"
+            elif item.find("=") != -1:
+                item = item.split("=")
+                flags_dict[item[0]] = item[1]
+            else:
+                flags_dict[item] = None
+    return flags_dict
+
+
+def get_cpu_info_from_qemu_cli(name):
```
yanan-fu, line 1299: It is only one step of `get_cpu_info_from_virsh`, they are not same level,  suggest move into `get_cpu_info_from_virsh` or change the function name.

virttest/cpu.py
```diff
@@ -1258,3 +1261,110 @@ def get_qemu_best_cpu_model(params):
...
+        return cpu_info
+    finally:
+        if virsh.is_alive(name):
+            virsh.destroy(name, ignore_status=True)
+        virsh.undefine(name, ignore_status=True)
+
+
+def combine_flags(auto_cpu_flags, custom_cpu_flags):
+    """
+    Combine the flags in flags_str1 and flags_str2
```
yanan-fu, line 1359: doc update

virttest/cpu.py
yanan-fu: @luckyh  How to explain the meaning of `base` and `suggestion` ?   On point i can think of is: the priority of `suggestion` is higher than `base`.

virttest/cpu.py
huangyum: It can be used by tp-qemu for the new case. What do you mean by changing the function name?

virttest/env_process.py
luckyh: What if the configuration would be like this?
```
auto_cpu_model = no
cpu_model =
default_cpu_model =
```

virttest/cpu.py
luckyh: Yes, it is, as we used `dict.update` at here.

virttest/qemu_vm.py
```diff
@@ -1598,6 +1599,29 @@ def sort_key(dev):
         add_memorys(devices, params)
         mem = int(params.get("mem", 0))
 
+        if params.get("auto_cpu_model"):
```
luckyh, line 1602: A regression, btw.
```suggestion
        if params.get("auto_cpu_model") == "yes":
```

virttest/env_process.py
huangyum: That's like an unreasonable misconfiguration to me. Do we need to handle all the user configuration error? It will fail all the time.

virttest/cpu.py
yanan-fu: > It can be used by tp-qemu for the new case. What do you mean by changing the function name?

Does this means tp-qemu skip `get_cpu_info_from_virsh`, use specified xml and define/start/destroy/undefine vm in test case level ?

virttest/env_process.py
luckyh: Well, qemu allows users to launch VMs without setting `-cpu`, so from my point of view it's not a misconfiguration.
```
$ qemu-kvm -m 512 -monitor stdio
QEMU 5.2.0 monitor - type 'help' for more information
(qemu) 
```

huangyum: Added the check back, I guess we better not change the behavior. Thanks.

huangyum: > Well, qemu allows users to launch VMs without setting `-cpu`, so from my point of view it's not a misconfiguration.
> 
> ```
> $ qemu-kvm -m 512 -monitor stdio
> QEMU 5.2.0 monitor - type 'help' for more information
> (qemu) 
> ```

Make sense. I missed the scenario. Pls help review again.

virttest/cpu.py
huangyum: I'm not sure. If use `get_cpu_info_from_virsh` directly, it will need to run virsh define/start twice.  @nanliu-r Could you please share your thoughts?

nanliu-r: > I'm not sure. If use `get_cpu_info_from_virsh` directly, it will need to run virsh define/start twice. @nanliu-r Could you please share your thoughts?

Yes, tp-qemu can use this function directly while developing the cpu_model case, we can keep it and change the name, thanks.

virttest/qemu_vm.py
```diff
@@ -1598,6 +1599,29 @@ def sort_key(dev):
...
+                        break
+                except Exception as err:
+                    logging.error("Failed to get cpu info with policy %s: %s"
+                                  % (policy, err))
+                    continue
+            else:
+                raise exceptions.TestCancel("Failed to get cpu info with "
+                                            "policy %s" % auto_cpu_policy)
+
+            params["cpu_model"] = cpu_info["model"]
```
luckyh, line 1619: Looks here would be a use-before-define error if the procedure being performed at this stage.

https://github.com/avocado-framework/avocado-vt/blob/9c2bc1139c163dd9e78cf4ac51adce47e4da2b07/virttest/qemu_vm.py#L1547
https://github.com/avocado-framework/avocado-vt/blob/9c2bc1139c163dd9e78cf4ac51adce47e4da2b07/virttest/qemu_devices/qcontainer.py#L1149

virttest/cpu.py
yanan-fu: > > I'm not sure. If use `get_cpu_info_from_virsh` directly, it will need to run virsh define/start twice. @nanliu-r Could you please share your thoughts?
> 
> Yes, tp-qemu can use this function directly while developing the cpu_model case, we can keep it and change the name, thanks.

Except ` <cpu mode='host-model'/>`,  to test other cpu model with virsh in tp-qemu.
So from my point of view, it is a new requirement that didn't be raised before,   make cpu model configurable while host-model is the default in avocado-vt.

cc @luckyh

virttest/qemu_vm.py
huangyum: Seems so.  But I didn't run into this error with either q35 or i440fx machine type. I wonder what's the QCPUBUS for, seems it's only used in these machine functions.
BTW, with configuration ''auto_cpu_model = no cpu_model = ", it goes to the same situation.

virttest/cpu.py
huangyum: I don't get your point. I just wrapped some steps into a function, which happens to be able to be used a tp-qemu case. What's the new requirement?

yanan-fu: My understanding for @nanliu-r 's comment is,  there are requirement to use virsh to test other cpu module and get cpu info for qemu cli. 

IMO, to resue `get_cpu_info_from_qemu_cli` in tp-qemu,  `get_cpu_info_from_virsh` need be rewrite in tp-qemu as different requirement. Correct me if my understanding is wrong.

huangyum: I think there is a misunderstanding here. The case needs to dumpxml and parse xml, apart from that, other steps like virsh define/start, get_cpu_info_from_qemu_cli etc, are same.
Let's put aside the tp-qemu case. If we prefer to keep this function, what kind of function name do you want to change to?

nanliu-r: > My understanding for @nanliu-r 's comment is, there are requirement to use virsh to test other cpu module and get cpu info for qemu cli.
> 
> IMO, to resue `get_cpu_info_from_qemu_cli` in tp-qemu, `get_cpu_info_from_virsh` need be rewrite in tp-qemu as different requirement. Correct me if my understanding is wrong.

I want to say that it just is convenient to reuse it but this doesn't mean that the function is designed for tp-qemu reuse.

virttest/cpu.py
```diff
@@ -1258,3 +1261,110 @@ def get_qemu_best_cpu_model(params):
...
+                item = item.split("=")
+                flags_dict[item[0]] = item[1]
+            else:
+                flags_dict[item] = None
+    return flags_dict
+
+
+def get_cpu_info_from_qemu_cli(name):
+    """
+    Get cpu info from qemu cmdline
```
yanan-fu, line 1301: `Get cpu info from qemu cmdline that generated by virsh`

virttest/cpu.py
yanan-fu: Ok for me if no other cpu model be involved.  Test case response for generate xml, define/start/... , use the vt funtion to get cpu info.
How about  `get_cpu_info_from_virsh_qemu_cli` ?  One option for virsh,  (same level with something like `get_cpu_info_from_virsh_xml`)

virttest/qemu_vm.py
huangyum: How about move L1602~L1654 before L1548? That should solve the issue.

virttest/cpu.py
huangyum: Updated, pls help review again.

huangyum:
@luckyh @yanan-fu Do you have more comments?

yanan-fu:
For the latest code, LGTM too.

However, during regression test, hit issue with case `x86_cpu_flags.others.test_dies` when use `auto_cpu_model=libvirt_host_model`, while use `auto_cpu_model = virttest` is ok.

Details be added to the issue, please help check, may be need new issue to fix it, thanks!

huangyum:
> For the latest code, LGTM too.
> 
> However, during regression test, hit issue with case `x86_cpu_flags.others.test_dies` when use `auto_cpu_model=libvirt_host_model`, while use `auto_cpu_model = virttest` is ok.
> 
> Details be added to the issue, please help check, may be need new issue to fix it, thanks!

Actually it's a known product bug.

yanan-fu:
> > For the latest code, LGTM too.
> > However, during regression test, hit issue with case `x86_cpu_flags.others.test_dies` when use `auto_cpu_model=libvirt_host_model`, while use `auto_cpu_model = virttest` is ok.
> > Details be added to the issue, please help check, may be need new issue to fix it, thanks!
> 
> Actually it's a known product bug.

No other comments if it is a known product bug, thanks!

yanan-fu (approved):

virttest/env_process.py
```diff
@@ -1164,26 +1164,6 @@ def preprocess(test, params, env):
                 vm.destroy()
                 del env[key]
 
-    if (params.get("auto_cpu_model") == "yes" and
-            vm_type == "qemu"):
```
luckyh, line 1168: Missing this check for the new approach.

nanliu-r:
Test results all look good, thanks yumei, hanxu and yanan!

Ack.

luckyh (approved):

virttest/cpu.py
```diff
@@ -1258,3 +1261,110 @@ def get_qemu_best_cpu_model(params):
...
+                </os>
+                <cpu mode='host-model'/>
+            </domain>
+          """ % (name, vm_arch, machine)
+    xml_file = os.path.join(data_dir.get_tmp_dir(), "temp_xml_for_cpu")
+    with open(xml_file, "w") as f:
+        f.write(xml)
+    try:
+        logging.info("Get cpu model and features from virsh")
+        virsh.define(xml_file)
```
YongxueHong, line 1343: Hi @huangyum 
Recently, we hit a hanging error when the virsh is defining xml since we didn't add the timeout for this process which will result in the job always hanging.
The gdb info:
```
Traceback (most recent call first):
  <built-in method waitpid of module object at remote 0x7f59bc9a42c0>
  File "/usr/lib64/python3.9/subprocess.py", line 1891, in _try_wait
    (pid, sts) = os.waitpid(self.pid, wait_flags)
  File "/usr/lib64/python3.9/subprocess.py", line 1933, in _wait
    (pid, sts) = self._try_wait(0)
  File "/usr/lib64/python3.9/subprocess.py", line 1189, in wait
    return self._wait(timeout=timeout)
  File "/home/kar/workspace/avocado/avocado/utils/process.py", line 856, in wait
    rc = self._popen.wait()
  File "/home/kar/workspace/avocado/avocado/utils/process.py", line 942, in run
    self.wait(timeout, sig)
  File "/home/kar/workspace/avocado/avocado/utils/process.py", line 1009, in run
    cmd_result = sp.run(timeout=timeout)
  File "/home/kar/workspace/avocado-vt/virttest/virsh.py", line 819, in command
    ret = process.run(cmd, timeout=timeout, verbose=debug,
  File "/home/kar/workspace/avocado-vt/virttest/virsh.py", line 1580, in define
    return command(cmd, **dargs)
  File "/home/kar/workspace/avocado-vt/virttest/cpu.py", line 1351, in get_cpu_info_from_virsh
    virsh.define(xml_file)
--Type <RET> for more, q to quit, c to continue without paging--
  File "/home/kar/workspace/avocado-vt/virttest/env_process.py", line 1351, in preprocess
    cpu_info = policy_map[policy](params)
  File "/home/kar/workspace/avocado-vt/virttest/error_context.py", line 135, in new_fn
    return fn(*args, **kwargs)
  File "/home/kar/workspace/avocado-vt/avocado_vt/test.py", line 252, in _runTest
    params = env_process.preprocess(self, params, env)
....
```

CC @qingwangrh

huangyum: Hi @YongxueHong , just checked all the code calling virsh.define, none of them add timeout, wondering why the issue is never hit before. Anyway, I agree 'timeout' should be set for any process in case of process hanging and never return. Adding timeout when invoking virsh.define looks like a workaround to me, maybe we could move the `if` block in https://github.com/avocado-framework/avocado-vt/blob/master/virttest/virsh.py#L792C7-L792C7 out of `if session:` in L788. What do you think?

YongxueHong: Hi @huangyum 
For your solution, we need to get the session id of an existing `VirshSession` in advance, 
and the way does not support `readonly` mode if we want to use `virsh.define` with `readonly` mode in the future.
and we can not support other options such as `virsh_opt`, `uri`
https://github.com/avocado-framework/avocado-vt/blob/fe0abdc708399b9f2b9e9321b2395b11a48b9047/virttest/virsh.py#L788-L795

And the timeout is default 60 if we don't pass the timeout here.
In my point of view, actually, this solution is the same as virsh.define(xml_file, None, timeout=60)

So I prefer to specify the argument timeout when calling `virsh.define`. Please let me know your opinion.

And we should consider adding the timeout for other places where call the `virsh` method without the limitation of the timeout, like `virsh.start(name)`, `virsh.destroy(name, ignore_status=True)` and `virsh.undefine(name, ignore_status=True)`.

YongxueHong: > Hi @huangyum For your solution, we need to get the session id of an existing `VirshSession` in advance, and the way does not support `readonly` mode if we want to use `virsh.define` with `readonly` mode in the future. and we can not support other options such as `virsh_opt`, `uri`
> 
> https://github.com/avocado-framework/avocado-vt/blob/fe0abdc708399b9f2b9e9321b2395b11a48b9047/virttest/virsh.py#L788-L795
> 
> And the timeout is default 60 if we don't pass the timeout here. In my point of view, actually, this solution is the same as virsh.define(xml_file, None, timeout=60)
> 
> So I prefer to specify the argument timeout when calling `virsh.define`. Please let me know your opinion.
> 
> And we should consider adding the timeout for other places where call the `virsh` method without the limitation of the timeout, like `virsh.start(name)`, `virsh.destroy(name, ignore_status=True)` and `virsh.undefine(name, ignore_status=True)`.

Hi @huangyum 
What's your opinion?

huangyum: Agree, sent out the patch https://github.com/avocado-framework/avocado-vt/pull/3754. Thanks.

PR #3166, opened 2021-08-05
Fix I1101 issue in virttest/lvsbs.py

chunfuwen:
Change column_names from None type to list


smitterl (approved):

smitterl:
IIUC, this is necessary to solve recent build issues where the type is inferred incorrectly, so it looks like we hit https://github.com/PyCQA/pylint/issues/1498

dzhengfy (approved):
LGTM

PR #3174, opened 2021-08-10
BaseVM.wait_for_login: Set dumb terminal per default after login

smitterl:
This PR addresses https://github.com/avocado-framework/avocado-vt/issues/3116

If the `TERM` variable is not `dumb` control characters like
bracketed paste `\x1b[?2004l127\n\x1b[?2004h` might be written
to the session output. This leads to several issues as regular
expressions don't match anymore. E.g. `aexpect.client.cmd_status_output`
matching for exit code via regex `"^-?[0-9]+$"` fails and the
method fails with `ShellStatusError`.

Theoretically, the guest image creation process could make sure
that the variable is set for any console session
(e.g. in the `.bashrc`).

However, this issue is not obvious, so lets set the variable
per default after login for any guest image.

The new parameter in the function allows for test implementers
to explicitly suppress this behavior and instead either use
the default value from inside the guest or set another value
in the test.

The variable is set just before returning a session after
successful login to not interfere with any test commands.

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -1152,9 +1152,9 @@ class BaseVM(object):
         return cmd
 
     def wait_for_login(self, nic_index=0, timeout=LOGIN_WAIT_TIMEOUT,
-                       internal_timeout=LOGIN_TIMEOUT,
-                       serial=False, restart_network=False,
-                       username=None, password=None, status_check=True):
+                       internal_timeout=LOGIN_TIMEOUT, serial=False,
+                       restart_network=False, username=None, password=None,
+                       status_check=True, set_dumb_terminal=True):
         """
         Make multiple attempts to log into the guest via SSH/Telnet/Netcat.
 
@@ -1168,6 +1168,9 @@ class BaseVM(object):
         :param status_check: Whether to call verify_alive to detect bad
             VM state early. Disable this when VM status might be unreliable,
             eg. during reboot or pause)
+        :param set_dumb_terminal: Wether to set TERM=dumb in order to avoid
+            control sequences/unprintable characters being written to the
+            session which can cause failures in regular expression matching
         :return: A ShellSession object.
         """
         def print_guest_network_info():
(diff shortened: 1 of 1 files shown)


virttest/virt_vm.py
```diff
@@ -1231,8 +1239,11 @@ def print_guest_network_info():
 
         print_guest_network_info()
         if serial:
-            return self.wait_for_serial_login(timeout, internal_timeout,
-                                              False, username, password)
+            session = self.wait_for_serial_login(timeout, internal_timeout,
+                                                 False, username, password)
+            if set_dumb_terminal:
+                session.cmd("export TERM=dumb")
```
chunfuwen, line 1245: May I ask whether serial console is impacted by  set TERM=dumb too, or this is limited to login only?

smitterl: @chunfuwen The environment variable can only be set after log in.

virttest/virt_vm.py
```diff
@@ -1209,6 +1212,8 @@ def print_guest_network_info():
                                                  restart_network, username,
                                                  password, False)
             if serial:
+                if set_dumb_terminal:
+                    session.cmd("export TERM=dumb")
```
luckyh, line 1216: VMs can be Windows guests therefore a check of OS type is required before setting the `TERM` variable.
```python
if set_dumb_terminal and self.params.get("os_type") != "windows":
    session.cmd("export TERM=dumb")
```
https://github.com/avocado-framework/avocado-vt/blob/d4c323aac396667cde94a3d22ba419647c8e6751/virttest/shared/cfg/guest-os/Windows.cfg#L6

dzhengfy:
@smitterl, it has been long time since it was updated last time. Could you help have a look?

smitterl:
> @smitterl, it has been long time since it was updated last time. Could you help have a look?

Thanks for the ping. :) We finally decided to update our expectations about the guests, s. https://github.com/avocado-framework/avocado-vt/pull/3217

So I close this.

PR #3178, opened 2021-08-12
Update marshal_from function to support setup_attrs

chloerh:
For XMLElementList accessor that has subclass, we need marshal_from
function to accept dict as arguments to support function setup_attrs


Yingshun (approved):

dzhengfy (approved):
LGTM

Issue #3169, opened 2021-08-06
Reuse the avocado logger


PaulYuuu:
This is a continuation of the discussion on https://github.com/avocado-framework/avocado/issues/4797

After https://github.com/avocado-framework/avocado/pull/4787 has been introduced, the VT runner will print excessive outputs.
```
INFO | Creating test configuration file '/root/avocado/data/avocado-vt/backends/qemu/cfg/tests-example.cfg'
INFO | Assigning vcpu num '4'
INFO | Assigning memory size '7168'
INFO | Start to run test
INFO:root:Config provided, ignoring option --vt-guest-os
INFO:root:Config provided, ignoring option --vt-arch or config vt.common.arch
INFO:root:Config provided, ignoring option --vt-machine-type or config vt.common.machine_type
INFO:root:Config provided and no option --vt-qemu-img or config vt.qemu.qemu_img set. Not trying to automatically set qemu bin
INFO:root:Config provided, ignoring config vt.qemu.nettype
INFO:root:Config provided, ignoring monitor setting
INFO:root:Config provided, ignoring config vt.qemu.smp
INFO:root:Config provided, ignoring config vt.qemu.image_type
INFO:root:Config provided, ignoring config vt.qemu.nic_model
INFO:root:Config provided, ignoring config vt.qemu.disk_bus
INFO:root:Config provided, ignoring config vt.qemu.vhost
INFO:root:Config provided, ignoring config vt.qemu.sandbox
JOB ID     : 41847374093671a4b3b5e6c4bde8b520ef09d3d5
JOB LOG    : /root/avocado/job-results/job-2021-08-06T08.55-4184737/job.log
 (1/1) Host_RHEL.m8.u5.product_av.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.5.0.s390x.io-github-autotest-qemu.boot.s390-virtio:  INFO:root:Avocado-VT 'unknown'
DEBUG:root:Searching for test modules that match 'type = boot' and 'provider = io-github-autotest-qemu' on this cartesian dict
DEBUG:root:Found subtest module /root/avocado/data/avocado-vt/virttest/test-providers.d/downloads/io-github-autotest-qemu/generic/tests/boot.py
WARNING:root:Creating new, empty env file
```

I tried to fix it so I add `logging.root = logging.getLogger('avocado')` in loader.py.
```diff
--- a/avocado_vt/loader.py
+++ b/avocado_vt/loader.py
[...]

PR #3179, opened 2021-08-16
Redefine the logger to avoid excessive output at the LOG_UI

PaulYuuu:
As we discussed at avocado-framework/avocado#4797 and #3169, the
current logger will display a lot of excessive output on the LOG_UI. These
outputs are the same as debug.log, they should only be in logs if we
don't use verbose mode, so this is unexpected behavior. Beraldo Leal
provides a good solution in #3169, the code works for me, so I
borrowed his solution to solve this problem accordingly.
    
Some VT modules will use "logging.warning" to print messages outside
the class/function, so when import these modules, the logging.RootLogger
will be initialized before the test execution, which will make the
RootLogger is out of control. Use the LOG_UI is a better choice.

### DIFF
--- a/avocado_vt/options.py
+++ b/avocado_vt/options.py
@@ -34,6 +34,9 @@ from virttest.standalone_test import SUPPORTED_NIC_MODELS
 from virttest.standalone_test import SUPPORTED_TEST_TYPES
 
 
+LOG = logging.getLogger('avocado.vt.options')
+
+
 class VirtTestOptionsProcess(object):
 
     """
@@ -124,8 +127,8 @@ class VirtTestOptionsProcess(object):
                             'config vt.qemu.qemu_bin')
         if (get_opt(self.config, 'vt.config') and
                 get_opt(self.config, 'vt.qemu.qemu_bin') is None):
-            logging.info("Config provided and no %s set. Not trying "
-                         "to automatically set qemu bin.", qemu_bin_setting)
+            LOG.info("Config provided and no %s set. Not trying "
+                     "to automatically set qemu bin.", qemu_bin_setting)
         else:
             (qemu_bin_path, qemu_img_path, qemu_io_path,
              qemu_dst_bin_path) = standalone_test.find_default_qemu_paths(
@@ -146,8 +149,8 @@ class VirtTestOptionsProcess(object):
                             'config vt.qemu.qemu_img')
         if (get_opt(self.config, 'vt.config') and
                 get_opt(self.config, 'vt.qemu.bin') is None):
-            logging.info("Config provided and no %s set. Not trying "
-                         "to automatically set qemu bin", qemu_img_setting)
+            LOG.info("Config provided and no %s set. Not trying "
+                     "to automatically set qemu bin", qemu_img_setting)
         else:
             (_, qemu_img_path,
              _, _) = standalone_test.find_default_qemu_paths(
@@ -191,7 +194,7 @@ class VirtTestOptionsProcess(object):
             elif get_opt(self.config, 'vt.common.nettype') == 'user':
                 self.cartesian_parser.assign("nettype", "user")
         else:
-            logging.info("Config provided, ignoring %s", nettype_setting)
+            LOG.info("Config provided, ignoring %s", nettype_setting)
 
     def _process_monitor(self):
         if not get_opt(self.config, 'vt.config'):
@@ -202,7 +205,7 @@ class VirtTestOptionsProcess(object):
             elif get_opt(self.config, 'vt.qemu.monitor') == 'human':
                 self.cartesian_parser.assign("monitor_type", "human")
         else:
-            logging.info("Config provided, ignoring monitor setting")
+            LOG.info("Config provided, ignoring monitor setting")
 
     def _process_smp(self):
         smp_setting = 'config vt.qemu.smp'
@@ -220,12 +223,12 @@ class VirtTestOptionsProcess(object):
                     raise ValueError("Invalid %s '%s'. Valid value: (1, 2, "
                                      "or integer)" % get_opt(self.config, 'vt.qemu.smp'))
         else:
-            logging.info("Config provided, ignoring %s", smp_setting)
+            LOG.info("Config provided, ignoring %s", smp_setting)
 
     def _process_arch(self):
         if get_opt(self.config, 'vt.config'):
             arch_setting = "option --vt-arch or config vt.common.arch"
-            logging.info("Config provided, ignoring %s", arch_setting)
+            LOG.info("Config provided, ignoring %s", arch_setting)
             return
 
         arch = get_opt(self.config, 'vt.common.arch')
@@ -247,7 +250,7 @@ class VirtTestOptionsProcess(object):
                 self.cartesian_parser.only_filter(get_opt(self.config,
                                                           'vt.common.machine_type'))
         else:
-            logging.info("Config provided, ignoring %s", machine_type_setting)
+            LOG.info("Config provided, ignoring %s", machine_type_setting)
 
     def _process_image_type(self):
         image_type_setting = 'config vt.qemu.image_type'
@@ -260,7 +263,7 @@ class VirtTestOptionsProcess(object):
                 self.cartesian_parser.assign("image_format",
                                              get_opt(self.config, 'vt.qemu.image_type'))
         else:
-            logging.info("Config provided, ignoring %s", image_type_setting)
+            LOG.info("Config provided, ignoring %s", image_type_setting)
 
     def _process_nic_model(self):
         nic_model_setting = 'config vt.qemu.nic_model'
@@ -272,7 +275,7 @@ class VirtTestOptionsProcess(object):
                 self.cartesian_parser.assign(
                     "nic_model", get_opt(self.config, 'vt.qemu.nic_model'))
         else:
-            logging.info("Config provided, ignoring %s", nic_model_setting)
+            LOG.info("Config provided, ignoring %s", nic_model_setting)
 
     def _process_disk_buses(self):
         disk_bus_setting = 'config vt.qemu.disk_bus'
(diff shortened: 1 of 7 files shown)


avocado_vt/options.py
```diff
@@ -34,6 +34,9 @@
 from virttest.standalone_test import SUPPORTED_TEST_TYPES
 
 
+LOG = logging.getLogger('avocado.vt.options')
```
PaulYuuu, line 37: I have not idea how to name this logger is best, so I chose the file name first. Please let me know if have any concerns.

PaulYuuu:
Ah, seems that this solution does not work on python3.9, but works on python3.6.

I am not sure if this is the bug of python3.9 itself, so set the status to "Draft" temporarily.

PaulYuuu:
> Ah, seems that this solution does not work on python3.9, but works on python3.6.
> 
> I am not sure if this is the bug of python3.9 itself, so set the status to "Draft" temporarily.

It's my mistake. Test results are the same in python3.9 and python3.6, I missed installing a required package so the RootLogger has been initialized on the module import phage. For instances:
https://github.com/avocado-framework/avocado-vt/blob/d4c323aac396667cde94a3d22ba419647c8e6751/virttest/virsh.py#L70-L75
https://github.com/avocado-framework/avocado-vt/blob/d4c323aac396667cde94a3d22ba419647c8e6751/virttest/env_process.py#L58-L64

To avoid the RootLogger being initialized before test execution, Use the `avocado.output.LOG_UI`/`logging.getLogger('avocado.app')` is a good choice.

PaulYuuu:
Hello @balamuruhans @pevogam @luckyh, as I mentioned above, this patch fixes two problems, redefine the logger in the first entry(options.py) and avoid the RootLogger being initialized before test execution.
Can you help to review it? Thanks in advance.

avocado_vt/options.py
pevogam, line 37: It would be better if we use small caps here - it is both more standard and we might confuse LOG with one of the avocado outputs like LOG_UI and LOG_JOB. Unless you maybe mean "LOG_VT_OPTS" or something but a simple `log = logging.getLogger(...)` is better IMO.

virttest/env_process.py
```diff
@@ -58,10 +58,11 @@
 try:
     import PIL.Image
 except ImportError:
-    logging.warning('No python imaging library installed. PPM image '
-                    'conversion to JPEG disabled. In order to enable it, '
-                    'please install python-imaging or the equivalent for your '
-                    'distro.')
+    LOG_UI = logging.getLogger('avocado.app')
+    LOG_UI.warning('No python imaging library installed. PPM image '
```
pevogam, line 62: Can you explain what is the reason for choosing the UI output for this? I think it was only previously printed in the log files.

virttest/guest_agent.py
```diff
@@ -12,8 +12,9 @@
 try:
     import json
 except ImportError:
-    logging.warning("Could not import json module. "
-                    "virt agent functionality disabled.")
+    LOG_UI = logging.getLogger('avocado.app')
+    LOG_UI.warning("Could not import json module. "
+                   "virt agent functionality disabled.")
```
pevogam, line 17: Same here and below

pevogam (commented):
I left a quick review mostly with questions and small suggestions, let me know what you think.

avocado_vt/options.py
PaulYuuu: Okay, that makes sense for me, will update later.

virttest/env_process.py
PaulYuuu: Well, I just want to find a suitable logger to handle these messages, so I think highlighting those warnings in the UI is a notification for users. but it's okay for both LOG_UI and LOG_JOB.
Actually, as I discussed with @luckyh offline, a logger provided by VT is a good enhancement to handle some logs, avocado should improve the behavior of handling the RootLogger as well.
Anyway, they are another topic, but it is an optimization indeed that we can consider.

pevogam: I guess it could be useful on the UI side too as long as we make sure that these imports are always needed. There are some libvirt or qemu specific dependencies for instance that some of us don't want to install to begin with and it could be too noisy to look at a warning message like this on each run if you don't have to install the given dependency for your own runs.

PaulYuuu: Thank you @pevogam for your clarification. And after my investigation, these messages will not be saved to the log files if using other loggers, because avocado will only save logging messages after `run_test`, so the messages generated during the import phase are lost. I will fall back to use LOG_UI.

virttest/virt_admin.py
```diff
@@ -62,8 +62,9 @@
 try:
     VIRTADMIN_EXEC = path.find_command("virt-admin")
 except path.CmdNotFoundError:
-    logging.warning("virt-admin executable not set or found on path, "
-                    "virtadmin-admin module will not function normally")
+    logging.getLogger('avocado.app').warning(
```
pevogam, line 65: @PaulYuuu So I guess these warnings are not included in the job log? Or where does the `avocado.app` logger's handlers point to?

PaulYuuu: `avocado.app` is same as `LOG_UI`
https://github.com/avocado-framework/avocado/blob/41b5626cf2d409110338486db2489d8d224f3bab/avocado/core/output.py#L34

You are right, they are not included in the job.log. Actually, before avocado is initialized, messages called by logging will not be saved to any log files if I understand correctly, even use `LOG_JOB`
https://github.com/avocado-framework/avocado/blob/41b5626cf2d409110338486db2489d8d224f3bab/avocado/core/output.py#L36

pevogam: > `avocado.app` is same as `LOG_UI`
> https://github.com/avocado-framework/avocado/blob/41b5626cf2d409110338486db2489d8d224f3bab/avocado/core/output.py#L34

Alright, then this is indeed better approach.

> You are right, they are not included in the job.log. Actually, before avocado is initialized, messages called by logging will not be saved to any log files if I understand correctly, even use `LOG_JOB`
> https://github.com/avocado-framework/avocado/blob/41b5626cf2d409110338486db2489d8d224f3bab/avocado/core/output.py#L36

My only remaining worry is if some of these modules affect people that don't use them. I can do some testing once we have more reviews, feel free to ping me then.

avocado_vt/options.py
PaulYuuu: After considering it again, since it's a global variable, I will fall back to use big caps, but I will use another name.

avocado_vt/options.py
```diff
@@ -34,6 +34,9 @@
 from virttest.standalone_test import SUPPORTED_TEST_TYPES
 
 
+LOG_OPTS = logging.getLogger('avocado.vt.options')
```
luckyh, line 37: I think we can use the name `LOG` instead because it is the default logger for this module.

luckyh (commented):
LGTM, otherwise

luckyh (approved):
Thanks.

pevogam:
> Thanks.

Thanks @luckyh, I will do some testing today (as I was waiting for more reviews first) and also approve if the messages are rightly placed.

luckyh:
> Thanks @luckyh, I will do some testing today (as I was waiting for more reviews first) and also approve if the messages are rightly placed.

@pevogam ok, thank you as well :)

pevogam:
So I did some testing and my fears were correct. We disabled some messages on our side which were previously in the logs for Qemu users and were not supposed to be as we don't use libvirt type of vms:
```diff
diff --git a/virttest/utils_libvirtd.py b/virttest/utils_libvirtd.py
index df31d8a0..df6df664 100644
--- a/virttest/utils_libvirtd.py
+++ b/virttest/utils_libvirtd.py
@@ -51,7 +51,7 @@ class Libvirtd(object):
         self.service_list = []
 
         if LIBVIRTD is None:
-            logging.warning("Libvirtd service is not available in host, "
+            logging.debug("Libvirtd service is not available in host, "
                             "utils_libvirtd module will not function normally")
 
         self.service_name = "libvirtd" if not service_name else service_name
diff --git a/virttest/virsh.py b/virttest/virsh.py
index dcc15225..6b2d1056 100644
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -66,8 +66,8 @@ VIRSH_COMMAND_GROUP_CACHE_NO_DETAIL = False
 try:
     VIRSH_EXEC = path.find_command("virsh")
 except path.CmdNotFoundError:
-    logging.warning("Virsh executable not set or found on path, "
-                    "virsh module will not function normally")
+    logging.debug("Virsh executable not set or found on path, "
+                  "virsh module will not function normally")
     VIRSH_EXEC = '/bin/true'
```
and the second one overlaps with the changes here. What this means is that these warnings will be made even worse now as they will flood the UI itself. @PaulYuuu What would be optimal is to not show libvirt related warnings if the users are only using qemu vms but if not then could these libvirt-related warnings at least be redirected towards the logs and not the UI?

PaulYuuu:
Hello @pevogam, unfortunately, I can't find a solution to redirect those warnings to the log file with the current avocado version. They are out of any class/function, so when avocado loads these modules, these related codes will be executed before avocado-vt is initialized. avocado should handle the RootLogger as well and I think the root cause you are discussing in https://github.com/avocado-framework/avocado/issues/4758.

A possible way is to move/wrap them so that they do not become global calls, to avoid they will be executed when loading the module. Users should call them when they really need them, this requires other libvirt stakeholders to improve the logic.

Therefore, before solving the root cause, we should avoid `RootLogger` being initialized during avocado initializing, otherwise a lot of excessive output will be printed in LOG_UI.

pevogam:
> Hello @pevogam, unfortunately, I can't find a solution to redirect those warnings to the log file with the current avocado version. They are out of any class/function, so when avocado loads these modules, these related codes will be executed before avocado-vt is initialized. avocado should handle the RootLogger as well and I think the root cause you are discussing in [avocado-framework/avocado#4758](https://github.com/avocado-framework/avocado/issues/4758).
> 
> A possible way is to move/wrap them so that they do not become global calls, to avoid they will be executed when loading the module. Users should call them when they really need them, this requires other libvirt stakeholders to improve the logic.
> 
> Therefore, before solving the root cause, we should avoid `RootLogger` being initialized during avocado initializing, otherwise a lot of excessive output will be printed in LOG_UI.

@beraldoleal If you agree with @PaulYuuu's conclusions here then I am fine with this and we can transfer this issue to avocado-framework/avocado#4758 as well.

beraldoleal (approved):
LGTM, thank you.

pevogam (approved):
Ok, LGTM too. Thanks for your patience @PaulYuuu.

luckyh:
Thanks to all.

PR #3180, opened 2021-08-16
Add the function to install stressapptest cmd

iccaszhulili:

### DIFF
--- a/virttest/utils_stress.py
+++ b/virttest/utils_stress.py
@@ -5,10 +5,12 @@ import threading
 import logging
 
 from avocado.utils import cpu
+from avocado.core import exceptions
 
 from virttest import virsh
 from virttest import cpu as cpuutil
 from virttest import utils_net
+from virttest import utils_package
 from virttest.utils_test import libvirt
 from virttest.libvirt_xml.devices.disk import Disk
 
@@ -197,3 +199,30 @@ class VMStressEvents():
             else:
                 raise NotImplementedError
             time.sleep(self.itr_sleep_time)
+
+
+def install_stressapptest(vm):
+    """
+    Install stressapptest cmd
+
+    :param vm: the vm to be installed with stressapptest
+    """
+    session = vm.wait_for_login(timeout=360)
+    name = ["git", "gcc", "gcc-c++", "make"]
+    if not utils_package.package_install(name, session, timeout=300):
+        raise exceptions.TestError("Installation of packages %s in guest "
+                                   "failed" % name)
+
+    app_repo = "git clone https://github.com/stressapptest/" \
+               "stressapptest.git"
+    stressapptest_install_cmd = "rm -rf stressapptest " \
+                                "&& %s" \
+                                " && cd stressapptest " \
+                                "&& ./configure " \
+                                "&& make " \
+                                "&& make install" % app_repo
+    s, o = session.cmd_status_output(stressapptest_install_cmd)
+    if s:
+        raise exceptions.TestError("Failed to install stressapptest "
+                                   "in guest: '%s'" % o)
+    session.close()


iccaszhulili:
JOB ID : 6c876a6c20265f631aafc9b77c23bb11d2fd3972
JOB LOG : /root/avocado/job-results/job-2021-08-20T04.40-6c876a6/job.log
(1/1) type_specific.io-github-autotest-libvirt.virsh.domdirtyrate_calc.positive: PASS (69.05 s)
RESULTS : PASS 1 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME : 69.92 s

iccaszhulili:
@dzhengfy please help to review. Thx

virttest/utils_stress.py
```diff
@@ -197,3 +199,33 @@ def host_stress_event(self, event):
...
+    :param vm: the vm to be installed with stressapptest
+    """
+    session = vm.wait_for_login()
+    name = "git gcc gcc-c++ make"
+    try:
+        if not utils_package.package_install(name, session, timeout=300):
+            raise virt_vm.VMError("Installation of packages %s failed" %
+                                  name)
+    except Exception as detail:
+        raise detail
```
kylazhang, line 217: I think no need to use try..except here?

virttest/utils_stress.py
```diff
@@ -197,3 +199,33 @@ def host_stress_event(self, event):
...
+
+
+def install_stressapptest(vm):
+    """
+    Install stressapptest cmd
+
+    :param vm: the vm to be installed with stressapptest
+    """
+    session = vm.wait_for_login()
+    name = "git gcc gcc-c++ make"
```
kylazhang, line 211: I think you should use list here

kylazhang:
Also please resolve check failure issue.

virttest/utils_stress.py
```diff
@@ -197,3 +199,33 @@ def host_stress_event(self, event):
...
+               "stressapptest.git"
+    stressapptest_install_cmd = "rm -rf stressapptest " \
+                                "&& %s" \
+                                " && cd stressapptest " \
+                                "&& ./configure " \
+                                "&& make " \
+                                "&& make install" % app_repo
+    s, o = session.cmd_status_output(stressapptest_install_cmd)
+    if s:
+        logging.error("Failed to install stressapptest "
```
chunfuwen, line 229: perhaps we can not allow silent failure,but throw exception here

dzhengfy:
@iccaszhulili Could you help to reply to the last comment?

virttest/utils_stress.py
iccaszhulili: updated

iccaszhulili:
@kylazhang @chunfuwen updated, please help to review. Thanks very much

virttest/utils_stress.py
```diff
@@ -197,3 +200,29 @@ def host_stress_event(self, event):
...
+
+def install_stressapptest(vm):
+    """
+    Install stressapptest cmd
+
+    :param vm: the vm to be installed with stressapptest
+    """
+    session = vm.wait_for_login(timeout=360)
+    name = ["git", "gcc", "gcc-c++", "make"]
+    if not utils_package.package_install(name, session, timeout=300):
```
kamilvarga, line 213: Not sure if we should install packages in avocado-vt. It is not recommended in the tp-libvirt as you can see for example in this PR: https://github.com/autotest/tp-libvirt/pull/3741.

virttest/utils_stress.py
```diff
@@ -197,3 +200,29 @@ def host_stress_event(self, event):
...
+        raise virt_vm.VMError("Installation of packages %s failed" % name)
```
dzhengfy, line 214: I don't think it is necessary for us to import virt_vm module to make more dependencies only because we need to raise an VMError. Here it is ok to simply raise exceptions.TestError.

iccaszhulili: @dzhengfy updated, thx

PR #3187, opened 2021-08-19
Enable ssh connection for migration

Yingshun:
SSH pwdless connection needs for some storage migration cases.

Test result:
https://github.com/autotest/tp-libvirt/pull/3716#issue-715788540

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -52,6 +52,7 @@ from virttest import utils_qemu
 from virttest import migration
 from virttest import utils_kernel_module
 from virttest import arch
+from virttest.utils_conn import SSHConnection
 from virttest.utils_version import VersionInterval
 from virttest.staging import service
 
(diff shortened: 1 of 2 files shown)


Yingshun:
@cliping @dzhengfy would you please help review it? Thanks!

virttest/utils_conn.py
```diff
@@ -576,6 +576,9 @@ def __init__(self, *args, **dargs):
                                                             port='22',
                                                             remote_path='/root/.ssh/authorized_keys')
 
+    def __getstate__(self):
```
Yingshun, line 579: It's for making the object pickable. If not, it'll throw a warning - 'Unable to save environment: a class that defines __slots__ without defining __getstate__ cannot be pickled'.

cliping (approved):
LGTM

virttest/env_process.py
```diff
@@ -1147,6 +1148,11 @@ def preprocess(test, params, env):
             session.close()
             test.cancel("Failed to map hostname and ipaddress of target host")
         session.close()
+        if params.get("setup_ssh") == "yes":
```
chunfuwen, line 1151: perhaps move those logic change to provider under tp-libvirt considering it is specific libvirt migration feature

Yingshun: No, ssh passwordless connection is required by preprocess and postprocess. Moving to tp-libvirt does not work in this case.

chunfuwen (approved):
LGTM

kylazhang (approved):

PR #3195, opened 2021-08-26
Replace with cmd_output to workaround cmd_output_safe issue

leidwang:

ID:1997415

### DIFF
--- a/virttest/utils_netperf.py
+++ b/virttest/utils_netperf.py
@@ -217,7 +217,7 @@ class Netperf(object):
         if self.client == "nc":
             list_cmd = "wmic process where name='%s' list" % target
         try:
-            output = self.session.cmd_output_safe(list_cmd, timeout=120)
+            output = self.session.cmd_output(list_cmd, timeout=120)
             check_reg = re.compile(r"%s" % target, re.I | re.M)
             return bool(check_reg.findall(output))
         except Exception as err:


yanglei-rh:
> > > > > > Looks like a different issue to me. The aexpect patch `5221d6c00ff91d982453eb5a481188b9587ea298` handle the issue that some output is not captured. However, this case failed due to `is_target_running` returns true, means the output contains something that's not supposed to have or the process is actually still running.
> > > > > 
> > > > > 
> > > > > In fact, this error occurred in the wait_for function. After unexpectedly exiting wait_for, the status was queried again, and the result was True, so the logic here is also a little problematic.
> > > > 
> > > > 
> > > > Since I don't have a machine that can be used for testing, I can't know the real reason. Inferring only from the log information, the reason for jumping out of ”utils_misc.wait_for()“ is that when checking the netperf process, there is not correct return value and the function was jumped out. In fact, the netperf process still exists at this time.
> > > > @leidwang Please correct me if I am was wrong.
> > > 
> > > 
> > > Yes, as you said, the `wait_for` function unexpectedly jumped out. I mean, the prompt error in this place is not "netperf still running, netperf hangs". My understanding should be that `wait_for` ends normally and `n_client.is_netperf_running( )` return value is True, it should be this error prompt.If jump out of `wait_for` ahead of time, it is another kind of error.
> > 
> > 
> > @leidwang Ok, I get it you. It's just that I think the current error is an abnormal situation. It is not a problem with netperf itself. So I think there is no need to specifically capture the wrong information. Capturing it may cause unnecessary misunderstandings on the final execution result of netperf. And it will make the entire code seem messy. On the contrary, based on the current code. Through log and some debugging methods, we can find other issue. For example, the current problem is a good example.
> 
> @yanglei-rh Of course, this patch is not intended to correct the inaccuracy of the error message. I think it might be more appropriate to change the code like this.
> 
> ```
>             if test_duration - 5 <= run_time:
>                 if n_client.is_netperf_running():
>                     test.fail("netperf still running,netperf hangs")
>                 else:
>                     logging.info("netperf runs successfully")
>             else:
>                 test.fail("netperf terminated unexpectedly,executed %ss" % run_time)
> ```

@leidwang If the timeout period is very short, changing to this logic may miss product bugs.

huangyum:
> > Looks like a different issue to me. The aexpect patch `5221d6c00ff91d982453eb5a481188b9587ea298` handle the issue that some output is not captured. However, this case failed due to `is_target_running` returns true, means the output contains something that's not supposed to have or the process is actually still running.
> 
> In fact, this error occurred in the wait_for function. After unexpectedly exiting wait_for, the status was queried again, and the result was True, so the logic here is also a little problematic.

Yeah, you are right. I had a try, cmd_output_safe() sometimes return empty causing wait_for return, then it returns non-empty output causing `is_target_running` return true. I think we should get the real output inside guest to check what's going wrong. BTW, why does the session log contain so less information now? cc @YongxueHong since he fixed the session log issue before.

PR #3197, opened 2021-08-31
VTDiscoverer fix

richtja:
The c0430a30d0050eed1205a03161b581f38fd4a61c broke the initialization of
VTDiscoverer. Because of this change, the `avocado list --resolver` and
`avocado run --vt-config` commands not working with nrunner. This is the
fix of that problem.


willianrampazzo (approved):

richtja:
This is no more needed, because  #3211 fixed this problem.

PR #3203, opened 2021-09-02
utils_disk: new added parts

dzhengfy:
This is to provide a new utitily function which is very basic and
can be invoked by many other utitily functions.

### DIFF
--- a/virttest/utils_disk.py
+++ b/virttest/utils_disk.py
@@ -1102,6 +1102,20 @@ def get_parts_list(session=None):
     return parts
 
 
+def get_added_parts(session, old_parts):
+    """
+    Get newly added partition list comparing to old parts
+
+    :param session: the vm session
+    :param old_parts: list, the old partition list
+    :return: list, the newly added partition list
+    """
+    new_parts = get_parts_list(session)
+    added_parts = list(set(new_parts).difference(set(old_parts)))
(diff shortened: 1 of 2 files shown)


chunfuwen (approved):
lgtm

dzhengfy:
@Yingshun could you help review? Thanks.

virttest/utils_disk.py
```diff
@@ -1102,6 +1102,20 @@ def get_parts_list(session=None):
     return parts
 
 
+def get_added_parts(session, old_parts):
```
Yingshun, line 1105: How about setting a default value for session? I think this function can also be used in localhost, just like get_parts_list(). What do you think?

dzhengfy: My original thought is that we will not test scenario to add a new disk to a host. So I only apply this function to operate in vm.

Yingshun: Ok, I see. But I still think it's better to use the same way as get_parts_list(). Anyway, it's up to you!

dzhengfy: I would not like to write codes which we do not expect to use. However, I will be happy to enhance it once we need it in future. Thanks.

Yingshun: Ok, got it.

Yingshun (approved):

PR #3205, opened 2021-09-03
[patch set #1] Remove the use of autotest 

luckyh:
This is a part of #2963


pevogam (commented):
Hi @luckyh, Autotest does support Python 3 but on a particular branch that we could merge to preserve this compatibility. It could be too much trouble though so we can also go on with this PR. What is your opinion?

PR #3212, opened 2021-09-08
Add function to fetch attrs of an xml object

chloerh:
Along with unittest of controller and interface xml objects

### DIFF
--- a/virttest/libvirt_xml/base.py
+++ b/virttest/libvirt_xml/base.py
@@ -320,6 +320,70 @@ class LibvirtXMLBase(propcan.PropCanBase):
                 target_obj.setup_attrs(**value)
                 setattr(self, key, target_obj)
 
+    def fetch_attrs(self):
+        """
+        Fetch attributes of xml object
+
+        :return: dict-type attributes, keys were pre-defined in xml classes
+
+        Example:
+        A dimm device 'dimm_xml' has xml like this:
+            <memory model="dimm">
+                <target>
+                    <size unit="KiB">524288</size><node>0</node>
+                </target>
+                <alias name="dimm2" />
+                <address base="0x120000000" slot="2" type="dimm" />
+            </memory>
+
+        dimm_xml.fetch_attrs()
+
+        Output should be:
+            {
+                'target': {
+                    'node': 0, 'size': 524288, 'size_unit': 'KiB'
+                    },
+                'alias': {'name': 'dimm2'},
+                'address': {
+                    'attrs': {
+                        'base': '0x120000000', 'slot': '2', 'type': 'dimm'
+                        },
+                    'type_name': 'dimm'
+                    },
+                'mem_model': 'dimm'
+            }
+
+        """
+        attrs = {}
+        slots = set(self.__all_slots__) - set(self.__uncompareable__) - {'device_tag'}
+        for key in slots:
+            try:
+                # Try to get values of each attribute, slot by slot
+                value = self[key]
+            except (xcepts.LibvirtXMLAccessorError,
+                    xcepts.LibvirtXMLNotFoundError,
+                    AttributeError,
+                    KeyError):
+                continue
+            else:
+                # Got value and check the type of value. If it's an nested xml
+                # instance, keep fetching attributes of it.
+                if isinstance(value, LibvirtXMLBase):
+                    value = value.fetch_attrs()
+                # If it's a list, return the list if the elements of the list
+                # are not xml instances
+                elif isinstance(value, list):
+                    if not value:
+                        continue
+                    # If the elements are xml instances, keep fetching
+                    # attributes of each xml instance and assign a list
+                    # of dict-type attrs to target slot
+                    elif isinstance(value[0], LibvirtXMLBase):
+                        value = [v.fetch_attrs() for v in value]
+                attrs[key] = value
+
+        return attrs
+
 
 def load_xml_module(path, name, type_list):
     """
--- a/virttest/libvirt_xml/devices/controller.py
+++ b/virttest/libvirt_xml/devices/controller.py
@@ -14,10 +14,11 @@ class Controller(base.TypedDeviceBase):
                  'address', 'pcihole64', 'target', 'alias', 'model_name',
                  'node')
 
-    def __init__(self, type_name, virsh_instance=base.base.virsh):
+    def __init__(self, type_name='pci', virsh_instance=base.base.virsh):
         super(Controller, self).__init__(device_tag='controller',
                                          type_name=type_name,
                                          virsh_instance=virsh_instance)
+        # TODO: Remove 'type' since it's duplicated with 'type_name'
         accessors.XMLAttribute('type', self, parent_xpath='/',
                                tag_name='controller', attribute='type')
         accessors.XMLAttribute('index', self, parent_xpath='/',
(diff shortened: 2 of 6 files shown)


chloerh:
(1/6) test_controller_xml.TestcontrollerXML.test_setup_controller_pci: PASS (0.28 s)
 (2/6) test_controller_xml.TestcontrollerXML.test_fetch_attrs_controller_pci: PASS (0.27 s)
 (3/6) test_interface_xml.TestcontrollerXML.test_setup_iface_default: PASS (0.30 s)
 (4/6) test_interface_xml.TestcontrollerXML.test_fetch_attrs_iface_default: PASS (0.28 s)
 (5/6) test_memory_xml.TestMemoryXML.test_setup_memory_default: PASS (0.30 s)
 (6/6) test_memory_xml.TestMemoryXML.test_fetch_attrs_memory_default: PASS (0.28 s)
RESULTS    : PASS 6 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 1.99 s

chloerh:
@dzhengfy @kylazhang @Yingshun Could you help review this? Thanks.

virttest/libvirt_xml/base.py
```diff
@@ -320,6 +320,70 @@ def setup_attrs(self, **attrs):
                 target_obj.setup_attrs(**value)
                 setattr(self, key, target_obj)
 
+    def fetch_attrs(self):
```
chunfuwen, line 323: Could you elaborate a little bit more about what issue this method want to address?
I assume we previously have some common sense that regarding xml management, we firstly get mapped object regarding relevant xml, then we can manipulate the object to get attrs or others

virttest/libvirt_xml/base.py
```diff
@@ -320,6 +320,70 @@ def setup_attrs(self, **attrs):
...
+                'mem_model': 'dimm'
+            }
+
+        """
+        attrs = {}
+        slots = set(self.__all_slots__) - set(self.__uncompareable__) - {'device_tag'}
+        for key in slots:
+            try:
+                # Try to get values of each attribute, slot by slot
+                value = self[key]
```
chunfuwen, line 362: align with line 321 using setattr, it sounds like more proper to use get_attr() here

virttest/libvirt_xml/base.py
```diff
@@ -320,6 +320,70 @@ def setup_attrs(self, **attrs):
...
+            except (xcepts.LibvirtXMLAccessorError,
+                    xcepts.LibvirtXMLNotFoundError,
+                    AttributeError,
+                    KeyError):
+                continue
```
chunfuwen, line 367: it is helpful for debugging if exception details can be print out

virttest/libvirt_xml/base.py
chloerh: @chunfuwen  Actually, no, get_attr() will easily conflict with pre-defined getter function.

virttest/libvirt_xml/base.py
chloerh: @chunfuwen Actually, no. We do not want that many logging here. It's not useful, only distracting.

virttest/libvirt_xml/base.py
chunfuwen: > @chunfuwen Actually, no, get_attr() will easily conflict with pre-defined getter function.

@chloerh ,here I mean use getattr(self, key).
Quto'get_attr() will easily conflict with pre-defined getter function', could you share what spcific pre-defined getter function it easily conflict with  ? since getattr is build-in function

virttest/libvirt_xml/base.py
chunfuwen: OK

virttest/libvirt_xml/base.py
```diff
@@ -320,6 +320,70 @@ def setup_attrs(self, **attrs):
...
+        attrs = {}
+        slots = set(self.__all_slots__) - set(self.__uncompareable__) - {'device_tag'}
+        for key in slots:
+            try:
+                # Try to get values of each attribute, slot by slot
+                value = self[key]
+            except (xcepts.LibvirtXMLAccessorError,
+                    xcepts.LibvirtXMLNotFoundError,
+                    AttributeError,
+                    KeyError):
```
chunfuwen, line 366: For exception handling ,after python3 afterwards, it suggest throw except as object
```
  try:
      xx()
    except Exception as e:
         xx()
```

virttest/libvirt_xml/base.py
chloerh: @chunfuwen Any xml class that has attribute named 'attr' will have a pre-defined function named 'get_attr'. 
If you still don't understand, please contact me, I'll explain more clearly face to face.

virttest/libvirt_xml/base.py
chloerh: @chunfuwen But we're not going to throw the exception... We just want to except them

virttest/libvirt_xml/base.py
chloerh: To be more clear, 

Any xml class that has attribute named 'attr' will have **3** pre-defined functions, they are:
get_attr()
set_attr()
del_attr()

I believe it's more clear now. If not, please contact me.

chunfuwen: @chloerh 
I assume you means 3-pre-defined are :__getattr__(), __setattr__(), __del_attr__() rather than get_attr(), set_attr(), del_attr().
Moreover, If we use 3-pref-defined metholds, generally we call : object._getattr__('key'), but if use build-in method,it is :getattr(object, 'key'). There are clear difference.

However, It is not blocker, it is in your hand to decide whether adopt it or not

virttest/libvirt_xml/base.py
chunfuwen: @kylazhang 
I remember in the past  we ever spent some of efforts in tp-libvirt to change from simple throw exception to throw exception as object

chunfuwen: it is long term issue we need address slowly, but not a blocker

virttest/libvirt_xml/base.py
chunfuwen, line 323: @chloerh 

I notice currently this method is mainly used by  unittests/libvirt_xml/device_xml/test_xx.py. It sounds like it aims to meet test validation requirement. Do you intend to recommend its usage on tp-libvirt? since previously we have some common sense that we get embedded objects first and further get attribute from object as default option when dealing with xml.

chunfuwen (approved):
LGTM

kylazhang (approved):

Yingshun (approved):
LGTM

virttest/libvirt_xml/base.py
dzhengfy, line 367: Is 'pass' better than 'continue' here because there is no other codes to be executed after this line in the loop and the loop will continue by default?

dzhengfy (commented):
Others LGTM

virttest/libvirt_xml/base.py
dzhengfy: After discussing offline, I understand the origin purpose, so I accept we use 'continue' instead of 'pass' to forcefully repeat the loop without caring about its following logic.

dzhengfy (approved):

dzhengfy:
@chloerh Thanks for providing the unit tests for so many devices.

Issue #3213, opened 2021-09-08
Missing screendumps and other important files for VT tests with the new VT runner


pevogam:
Currently only a simple debug.log could be found in the results folder for each test with lots of previously available files that are useful for debugging now missing:
```
# ll /mnt/local/results/latest/test-results/0q-mw101-original.unattended_install.cdrom.in_cdrom_ks.default_install.aio_threads.mw101.smp2.Win10.x86_64/
total 200
-rw-r--r-- 1 root root    22 Sep  8 11:06 background-error.log
-rw-r--r-- 1 root root 16561 Sep  8 11:06 catch_monitor-mis1-pid-unknown.log
-rw-r--r-- 1 root root 16561 Sep  8 11:06 catch_monitor-mis2-pid-unknown.log
-rw-r--r-- 1 root root 16280 Sep  8 11:06 catch_monitor-mw101-pid-84225.log
drwxr-xr-x 2 root root  4096 Sep  8 11:05 data
-rw-r--r-- 1 root root 56151 Sep  8 11:06 debug.log
-rw-r--r-- 1 root root 14788 Sep  8 11:05 hmp1-mis1-pid-unknown.log
-rw-r--r-- 1 root root 14788 Sep  8 11:05 hmp1-mis2-pid-unknown.log
-rw-r--r-- 1 root root 19449 Sep  8 11:06 hmp1-mw101-pid-84225.log
-rw-r--r-- 1 root root  1336 Sep  8 11:06 ip-sniffer.log
-rw-r--r-- 1 root root     0 Sep  8 11:05 output
drwxr-xr-x 2 root root  4096 Sep  8 11:06 screendumps_mis1_31349_iter0
drwxr-xr-x 2 root root  4096 Sep  8 11:06 screendumps_mis2_31480_iter0
drwxr-xr-x 2 root root  4096 Sep  8 11:06 screendumps_mw101_84225_iter0
-rw-r--r-- 1 root root   726 Sep  8 11:06 stderr
-rw-r--r-- 1 root root  5232 Sep  8 11:06 stdout
```
The currently available files are:
```
[...]

PR #3214, opened 2021-09-09
 Additional log files for vt-runner

richtja:
This commit copies content from the temporary vt logdir to the avocado
test-results. This will bring more useful debug files about test
runtime, which are available on legacy runner.

Reference: #3213

### DIFF
--- a/avocado_vt/plugins/vt_runner.py
+++ b/avocado_vt/plugins/vt_runner.py
@@ -37,6 +37,7 @@ class VirtTest(utils.TestUtils):
         self.outputdir = path.init_dir(self.logdir, 'data')
         self.bindir = data_dir.get_root_dir()
         self.virtdir = os.path.join(self.bindir, 'shared')
+        utils_misc.set_log_file_dir(self.logdir)
 
     @property
     def params(self):
@@ -58,7 +59,27 @@ class VirtTest(utils.TestUtils):
             status = "ERROR"
             self.queue.put(messages.StderrMessage.get(traceback.format_exc()))
         finally:
+            self.save_log_dir()
             self.queue.put(messages.FinishedMessage.get(status))
+
+    def save_log_dir(self):
+        """
+        Sends the content of vt logdir to avocado logdir
+        """
+        for root, _, files in os.walk(self.logdir, topdown=False):
+            basedir = os.path.relpath(root, start=self.logdir)
+            for file in files:
+                file_path = os.path.join(root, file)
+                with open(file_path, 'rb') as f:
+                    base_path = os.path.join(basedir, file)
+                    while True:
+                        # Read data in manageable chunks rather than
+                        # all at once.
+                        in_data = f.read(200000)
+                        if not in_data:
+                            break
+                        self.queue.put(messages.FileMessage.get(in_data,
+                                                                base_path))
 
     def _run_test(self):
         params = self.params
--- a/avocado_vt/plugins/vt_runner.py
+++ b/avocado_vt/plugins/vt_runner.py
@@ -34,6 +34,7 @@ class VirtTest(utils.TestUtils):
         # clear existing error events
         self.background_errors.clear()
         self.debugdir = self.logdir
+        self.outputdir = path.init_dir(self.logdir, 'data')
         self.bindir = data_dir.get_root_dir()
         self.virtdir = os.path.join(self.bindir, 'shared')
 


willianrampazzo (approved):

richtja:
Hi @pevogam this is a fix for your issue #3213

pevogam:
Interesting that I didn't spot this before pushing my PR for the unattended install. Will look into this right away.

pevogam:
Is there a reason for the files to be copied right in the end? This breaks some behavior where tests dump additional files and simply use the test folder for the purpose and prevents early inspection while the test is running.

Also while some folders are indeed there at the end of the test. You can see I describe many more missing files in #3213.

richtja:
> Is there a reason for the files to be copied right in the end? This breaks some behavior where tests dump additional files and simply use the test folder for the purpose and prevents early inspection while the test is running.
  
Hi @pevogam unfortunately I don't understand how this breaks  prevention of early inspection while the test is running. The files are copied after the testing is finished.

> 
> Also while some folders are indeed there at the end of the test. You can see I describe many more missing files in #3213.

Yes you are right, I found more problems with this solution. So for now I converted to the draft.

pevogam:
> > Is there a reason for the files to be copied right in the end? This breaks some behavior where tests dump additional files and simply use the test folder for the purpose and prevents early inspection while the test is running.
> 
> Hi @pevogam unfortunately I don't understand how this breaks prevention of early inspection while the test is running. The files are copied after the testing is finished.

Yes, maybe not completely preventing because technically one should still be able to check a custom log-searched tmp path where the results are stored before they are copied. My problem with this approach is that one has to dig into a particular temporary path and thus deal with temporary and less accessible storage internals instead of the already well established test result folders.

richtja:
> > > Is there a reason for the files to be copied right in the end? This breaks some behavior where tests dump additional files and simply use the test folder for the purpose and prevents early inspection while the test is running.
> > 
> > 
> > Hi @pevogam unfortunately I don't understand how this breaks prevention of early inspection while the test is running. The files are copied after the testing is finished.
> 
> Yes, maybe not completely preventing because technically one should still be able to check a custom log-searched tmp path where the results are stored before they are copied. My problem with this approach is that one has to dig into a particular temporary path and thus deal with temporary and less accessible storage internals instead of the already well established test result folders.

I am not sure if I understand, you need the easy access to those files during the test runtime? Such solution mid be quite tricky, there is so many possible channels for saving to the output directory and vt-runner can't have any control of them. So it will be difficult to track the changes and send them to avocado results' directory.

pevogam:
> > > > Is there a reason for the files to be copied right in the end? This breaks some behavior where tests dump additional files and simply use the test folder for the purpose and prevents early inspection while the test is running.
> > > 
> > > 
> > > Hi @pevogam unfortunately I don't understand how this breaks prevention of early inspection while the test is running. The files are copied after the testing is finished.
> > 
> > 
> > Yes, maybe not completely preventing because technically one should still be able to check a custom log-searched tmp path where the results are stored before they are copied. My problem with this approach is that one has to dig into a particular temporary path and thus deal with temporary and less accessible storage internals instead of the already well established test result folders.
> 
> I am not sure if I understand, you need the easy access to those files during the test runtime? Such solution mid be quite tricky, there is so many possible channels for saving to the output directory and vt-runner can't have any control of them. So it will be difficult to track the changes and send them to avocado results' directory.

Hmmm, maybe I am missing something but with the serial runner before this was fairly straightforward and all files were not copied as diffs but directly written to the test results directory. What are the limitations of the VT runner that make it impossible for it to just directly create the files in the test logdir? Could this be a general Avocado limitation instead but one that relates in some way to the next runner?

richtja:
> > > > > Is there a reason for the files to be copied right in the end? This breaks some behavior where tests dump additional files and simply use the test folder for the purpose and prevents early inspection while the test is running.
> > > > 
> > > > 
> > > > Hi @pevogam unfortunately I don't understand how this breaks prevention of early inspection while the test is running. The files are copied after the testing is finished.
> > > 
> > > 
> > > Yes, maybe not completely preventing because technically one should still be able to check a custom log-searched tmp path where the results are stored before they are copied. My problem with this approach is that one has to dig into a particular temporary path and thus deal with temporary and less accessible storage internals instead of the already well established test result folders.
> > 
> > 
> > I am not sure if I understand, you need the easy access to those files during the test runtime? Such solution mid be quite tricky, there is so many possible channels for saving to the output directory and vt-runner can't have any control of them. So it will be difficult to track the changes and send them to avocado results' directory.
> 
> Hmmm, maybe I am missing something but with the serial runner before this was fairly straightforward and all files were not copied as diffs but directly written to the test results directory. What are the limitations of the VT runner that make it impossible for it to just directly create the files in the test logdir? Could this be a general Avocado limitation instead but one that relates in some way to the next runner?

Unfortunately this can't be done same as in legacy runner, because the new runner architecture is modular, which means that every nrunner like vr-runner is standalone task and might be run even on different machine then avocado itself. So the vr-runner outputdir and avocado test resultdir don't have to be in the same file system.

pevogam:
> > > > > > Is there a reason for the files to be copied right in the end? This breaks some behavior where tests dump additional files and simply use the test folder for the purpose and prevents early inspection while the test is running.
> > > > > 
> > > > > 
> > > > > Hi @pevogam unfortunately I don't understand how this breaks prevention of early inspection while the test is running. The files are copied after the testing is finished.
> > > > 
> > > > 
> > > > Yes, maybe not completely preventing because technically one should still be able to check a custom log-searched tmp path where the results are stored before they are copied. My problem with this approach is that one has to dig into a particular temporary path and thus deal with temporary and less accessible storage internals instead of the already well established test result folders.
> > > 
> > > 
> > > I am not sure if I understand, you need the easy access to those files during the test runtime? Such solution mid be quite tricky, there is so many possible channels for saving to the output directory and vt-runner can't have any control of them. So it will be difficult to track the changes and send them to avocado results' directory.
> > 
> > 
> > Hmmm, maybe I am missing something but with the serial runner before this was fairly straightforward and all files were not copied as diffs but directly written to the test results directory. What are the limitations of the VT runner that make it impossible for it to just directly create the files in the test logdir? Could this be a general Avocado limitation instead but one that relates in some way to the next runner?
> 
> Unfortunately this can't be done same as in legacy runner, because the new runner architecture is modular, which means that every nrunner like vr-runner is standalone task and might be run even on different machine then avocado itself. So the vr-runner outputdir and avocado test resultdir don't have to be in the same file system.

I see. While this might work well for unit tests or small tests I am not that confident it won't create lots of headaches with VT tests where some of them might take tens of minutes to half an hour. Take unattended install tests for instance - the new runner practically makes it if not impossible then very difficult to know how the installation process goes until the very end. In case where some VT tests could hang on boot or get stuck at other large timeout checks, it could take an hour before seeing any progress. But I guess this is beyond the scope of this issue which is just about the additional log files. We could perhaps discuss this problem elsewhere later on.

pevogam:
@richtja Just to clarify something on my side - I don't consider pushes a ping (not sure if you do) so if you need me to test and check this again, just let me know.

richtja:
> > > > > > > Is there a reason for the files to be copied right in the end? This breaks some behavior where tests dump additional files and simply use the test folder for the purpose and prevents early inspection while the test is running.
> > > > > > 
> > > > > > 
> > > > > > Hi @pevogam unfortunately I don't understand how this breaks prevention of early inspection while the test is running. The files are copied after the testing is finished.
> > > > > 
> > > > > 
> > > > > Yes, maybe not completely preventing because technically one should still be able to check a custom log-searched tmp path where the results are stored before they are copied. My problem with this approach is that one has to dig into a particular temporary path and thus deal with temporary and less accessible storage internals instead of the already well established test result folders.
> > > > 
> > > > 
> > > > I am not sure if I understand, you need the easy access to those files during the test runtime? Such solution mid be quite tricky, there is so many possible channels for saving to the output directory and vt-runner can't have any control of them. So it will be difficult to track the changes and send them to avocado results' directory.
> > > 
> > > 
> > > Hmmm, maybe I am missing something but with the serial runner before this was fairly straightforward and all files were not copied as diffs but directly written to the test results directory. What are the limitations of the VT runner that make it impossible for it to just directly create the files in the test logdir? Could this be a general Avocado limitation instead but one that relates in some way to the next runner?
> > 
> > 
> > Unfortunately this can't be done same as in legacy runner, because the new runner architecture is modular, which means that every nrunner like vr-runner is standalone task and might be run even on different machine then avocado itself. So the vr-runner outputdir and avocado test resultdir don't have to be in the same file system.
> 
> I see. While this might work well for unit tests or small tests I am not that confident it won't create lots of headaches with VT tests where some of them might take tens of minutes to half an hour. Take unattended install tests for instance - the new runner practically makes it if not impossible then very difficult to know how the installation process goes until the very end. In case where some VT tests could hang on boot or get stuck at other large timeout checks, it could take an hour before seeing any progress. But I guess this is beyond the scope of this issue which is just about the additional log files. We could perhaps discuss this problem elsewhere later on.

Now I see, and I understand how this can be useful, but first of all I would like to make those files visible for users at least after the test, and then we can discuss how to make it better.

richtja:
Closing this because, because the issue #3213 is better solved in #3226.

PR #3218, opened 2021-09-10
Update the unattended install test in accordance with the new VT runner

pevogam:
These are adaptations for the unattended install tests to run again with the new VT runner.

### DIFF
--- a/avocado_vt/utils.py
+++ b/avocado_vt/utils.py
@@ -207,4 +207,4 @@ class TestUtils:
             msg.append("s are" if len(bg_errors) > 1 else " is")
             msg.append((" detected, please refer to file: "
                         "'%s' for more details.") % BG_ERR_FILE)
-            self.error(''.join(msg))
+            self.log.error(''.join(msg))


virttest/shared/deps/run_autotest/kernel_install/kernelinstall.py
```diff
@@ -83,7 +83,7 @@ def _kernel_install_src(self, base_tree, config=None, config_list=None,
             base_tree = os.path.join(self.bindir, base_tree)
         if not utils.is_url(config):
             config = os.path.join(self.bindir, config)
-        kernel = self.job.kernel(base_tree, self.outputdir)
+        kernel = self.job.kernel(base_tree, self.logdir)
```
richtja, line 86: Hi @pevogam, the problem with `outputdir` is fixed by #3214. So I think this change is not necessary.

richtja (changes requested):

virttest/shared/deps/run_autotest/kernel_install/kernelinstall.py
pevogam: Thanks for talking a look, this is just some fixes I collected that were lacking on master. Alright, I will take a look into your PR first and then see what is left here.

pevogam:
@richtja Do you think the currently remaining diff is still a valid fix? If not I will close this and if yes I will rebase it.

richtja:
@pevogam after the #3226 this is no more needed. Actually I think we can move the `TestUtils` functionality back to the `test.VirtTest` class, because now there isn't any other usage of it elsewhere.

pevogam:
Alright, in this case I am closing this.

Regarding later refactoring of the VT runner, it is probably better to wait after the LTS release and perform regression fixes only.

PR #3224, opened 2021-09-16
qdevices.py: add chardev parameters

nanliu-r:
ID: 1986691

abstract=on|off specifies the use of the abstract
socket namespace, rather than the filesystem. Optional,
defaults to false. tight=on|off sets the socket length
of abstract sockets to their minimum, rather than the
full sun_path length. Optional, defaults to true.

related tp-qemu patch: [tp-qemu/3008](https://github.com/autotest/tp-qemu/pull/3008)

Signed-off-by: Nana Liu <nanliu@redhat.com>

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -2663,6 +2663,10 @@ class DevContainer(object):
             chardev_param.update({'path': file_name})
             if (backend == 'pipe' and params.get('auto_create_pipe', 'yes') == "yes"):
                 process.system("mkfifo %s" % file_name)
+            if backend == 'unix_socket':
+                chardev_param.update(
+                    {'abstract': params.get('chardev_abstract'),
+                     'tight': params.get('chardev_tight')})
         elif backend in ['udp', 'tcp_socket']:
             chardev_param.update(
                 {'host': params['chardev_host'],
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -1586,7 +1586,8 @@ class CharDevice(QCustomDevice):
         elif backend in ["socket"]:
             common_opts += ["server", "wait", "reconnect"]
             special_opts = ["host", "port", "to", "ipv4",
-                            "ipv6", "nodelay", "path"]
+                            "ipv6", "nodelay", "path",
+                            "abstract", "tight"]
 
         elif backend == 'udp':
             special_opts = ["host", "port", "localaddr",
--- a/virttest/shared/deps/serial/serial_host_send_receive.py
+++ b/virttest/shared/deps/serial/serial_host_send_receive.py
@@ -222,6 +222,9 @@ def main():
     chardev_type = options.type
     if options.socket:
         device = options.socket
+        # support abstract unix socket address
+        if '@' in device:
+            device = device.replace('@', '\0')
     else:
         parser.error("Please set -s parameter.")
 


virttest/qemu_devices/qcontainer.py
```diff
@@ -2674,7 +2674,9 @@ def chardev_define_by_params(self, chardev_id, params, file_name=None):
         if 'socket' in backend:  # tcp_socket & unix_socket
             chardev_param.update(
                 {'server': params.get('chardev_server', 'on'),
-                 'wait': params.get('chardev_wait', 'off')})
+                 'wait': params.get('chardev_wait', 'off'),
+                 'abstract': params.get('chardev_abstract', ''),
+                 'tight': params.get('chardev_tight', '')})
```
yanan-fu, line 2679: tcp soscket does not support these two options:
```
-chardev socket,id=id[,host=host],port=port[,to=to][,ipv4=on|off][,ipv6=on|off][,nodelay=on|off][,reconnect=seconds]
         [,server=on|off][,wait=on|off][,telnet=on|off][,websocket=on|off][,reconnect=seconds][,mux=on|off]
         [,logfile=PATH][,logappend=on|off][,tls-creds=ID][,tls-authz=ID] (tcp)
```

nanliu-r: > tcp socket does not support these two options:
> 
> ```
> -chardev socket,id=id[,host=host],port=port[,to=to][,ipv4=on|off][,ipv6=on|off][,nodelay=on|off][,reconnect=seconds]
>          [,server=on|off][,wait=on|off][,telnet=on|off][,websocket=on|off][,reconnect=seconds][,mux=on|off]
>          [,logfile=PATH][,logappend=on|off][,tls-creds=ID][,tls-authz=ID] (tcp)
> ```

Thanks and update.

virttest/qemu_devices/qdevices.py
```diff
@@ -1586,7 +1586,8 @@ def get_supported_options(self, backend):
         elif backend in ["socket"]:
             common_opts += ["server", "wait", "reconnect"]
             special_opts = ["host", "port", "to", "ipv4",
-                            "ipv6", "nodelay", "path"]
+                            "ipv6", "nodelay", "path",
+                            "abstract", "tight"]
```
yanan-fu, line 1590: Need be updated too,  for unix_socket only

virttest/qemu_devices/qcontainer.py
```diff
@@ -2671,6 +2671,10 @@ def chardev_define_by_params(self, chardev_id, params, file_name=None):
                  'ipv6': params.get('chardev_ipv6')})
         if backend == 'tcp_socket':
             chardev_param.update({'to': params.get('chardev_to')})
+        if backend == 'unix_socket':
```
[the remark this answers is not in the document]
nanliu-r: > Recommend to move after L2665.

Seems here is more reasonable, it can keep consistence with 'tcp_socket'.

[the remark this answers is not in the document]
yanan-fu: I do not think so,  L2672~L2673 is better to move under the `elif` blocker in L2666, considering it is involved by your patch so i did not raise it.

virttest/qemu_devices/qdevices.py
nanliu-r: > Need be updated too, for unix_socket only

update and thanks.

virttest/qemu_devices/qcontainer.py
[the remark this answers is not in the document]
nanliu-r: > I do not think so, L2672~L2673 is better to move under the `elif` blocker in L2666, considering it is involved by your patch so i did not raise it.

update and thanks.

virttest/shared/deps/serial/serial_host_send_receive.py
```diff
@@ -222,6 +222,8 @@ def main():
     chardev_type = options.type
     if options.socket:
         device = options.socket
+        if '@' in device:
+            device = device.replace('@', '\0')
```
yanan-fu, line 226: It is better to use a separate commit and explain the reason for this change.

yanan-fu:
> > @nanliu-r Use `socket_type` to handle the exception for the options you added in this patch is not suitable. To support this fully, i am afraid `CharDevice` need be reorganized.
> > It is my fault i did not go through the whole CharDevice before. I think it is not necessary to reorganize CharDevice for these two new option, keep qdevice and update qcontainer is enough I think.
> 
> @yanan-fu We also need to get the supported options for different backend from qdevice, How about putting the two options into the 'special_opts'? 

yes, that is what i mean. Distinguish tcp_socket and unix_socket in qdevices need reorganize the CharDevice, so just keep the current design and add the two new options is ok.

> Seems we store the options of both unix_socket and tcp_socket into this list.

nanliu-r:
> > > @nanliu-r Use `socket_type` to handle the exception for the options you added in this patch is not suitable. To support this fully, i am afraid `CharDevice` need be reorganized.
> > > It is my fault i did not go through the whole CharDevice before. I think it is not necessary to reorganize CharDevice for these two new option, keep qdevice and update qcontainer is enough I think.
> > 
> > 
> > @yanan-fu We also need to get the supported options for different backend from qdevice, How about putting the two options into the 'special_opts'?
> 
> yes, that is what i mean. Distinguish tcp_socket and unix_socket in qdevices need reorganize the CharDevice, so just keep the current design and add the two new options is ok.
> 
> > Seems we store the options of both unix_socket and tcp_socket into this list.

update, please help to review, thanks.

yanan-fu:
Ack with the change of add two new options for chardev.
For the change of `serial_host_send_receive.py`, under debugging, thanks~

virttest/shared/deps/serial/serial_host_send_receive.py
nanliu-r: > @nanliu-r We need further discussion on this, seems you didn't find the root cause of your issue.

@vivianQizhu update and please help to review again. Thanks.

vivianQizhu:
Please fix CI failure.

nanliu-r:
> @nanliu-r For the CI error, the root cause is the Author name is not same with the signed-off-by: Author: nanliu [nanliu@redhat.com](mailto:nanliu@redhat.com) Signed-off-by: Nana Liu [nanliu@redhat.com](mailto:nanliu@redhat.com)
> 
> Please update it, thanks!

@yanan-fu  update, thanks a lot!

PR #3229, opened 2021-09-17
qemu_vm: Add designation of nic driver model for different platforms

yanglei-rh:
ID:2005273

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -3877,6 +3877,15 @@ class VM(virt_vm.BaseVM):
         :param mac: Optional MAC address, None to randomly generate.
         """
         # returns existing or new nic object
+        if params['nic_model'] == 'virtio':
+            machine_type = self.params.get("machine_type")
+            if "s390" in machine_type:
+                model = "virtio-net-ccw"
+            elif '-mmio:' in machine_type:
+                model = "virtio-net-device"
+            else:
+                model = "virtio-net-pci"
+            params['nic_model'] = model
         nic = super(VM, self).add_nic(**params)
         nic_index = self.virtnet.nic_name_index(nic.nic_name)
         nic.set_if_none('vlan', str(nic_index))


yanglei-rh:
Hello @PaulYuuu Could you please help me review it? Thanks in advance.

virttest/shared/cfg/guest-hw.cfg
```diff
@@ -24,7 +24,9 @@ variants:
         required_qemu = [2.6.0, )
         nic_model = e1000e
     - virtio_net:
-        nic_model = virtio
+        nic_model = virtio-net-pci
```
PaulYuuu, line 27: You should consider different naming in different platforms.
https://github.com/avocado-framework/avocado-vt/blob/11a2c5a3eded430db30740b642d02f3d8a1279eb/virttest/qemu_vm.py#L606-L613

yanglei-rh: Thanks for your reminder. After you reminded me yesterday, I forgot it when I updated it. Thank you so much.

yanglei-rh:
Hi @vivianQizhu Could you please help me review it? Thanks in advance.

vivianQizhu:
@yanglei-rh Does this change for libvirt? Because qemu_kvm.py change it to `virtio-net-xxx`. And if only for `Host_RHEL.m5` it is `virtio`, I think it can be removed, together with the those qemu_kvm.py translations. And, would you like to update subtest.cfg together?

yanglei-rh:
> @yanglei-rh Does this change for libvirt? Because qemu_kvm.py change it to `virtio-net-xxx`. And if only for `Host_RHEL.m5` it is `virtio`, I think it can be removed, together with the those qemu_kvm.py translations. And, would you like to update subtest.cfg together?

Hi @vivianQizhu This change does not affect libvirt. Only `Host_RHEL.m5 `uses `virtio`, I will delete it in the next update. Does subtest.cfg also need to be changed? At present, when I run the case, this file will automatically get the configuration.

yanglei-rh:
@vivianQizhu Could you please help me review it again, thanks a lot.

vivianQizhu:
> > @yanglei-rh Does this change for libvirt? Because qemu_kvm.py change it to `virtio-net-xxx`. And if only for `Host_RHEL.m5` it is `virtio`, I think it can be removed, together with the those qemu_kvm.py translations. And, would you like to update subtest.cfg together?
> 
> Hi @vivianQizhu This change does not affect libvirt. Only `Host_RHEL.m5 `uses `virtio`, I will delete it in the next update. Does subtest.cfg also need to be changed? At present, when I run the case, this file will automatically get the configuration.

libvirt_vm.py also consumes `nic_model`, it should be affected and I am not sure if it is correct. If the modification is not for libvirt_vm, would you help explain why we need this change? From my understanding qemu_vm already change it to virtio-net-xxx when it is defined to `virtio`, why this change is now necessary?

yanglei-rh:
> > > @yanglei-rh Does this change for libvirt? Because qemu_kvm.py change it to `virtio-net-xxx`. And if only for `Host_RHEL.m5` it is `virtio`, I think it can be removed, together with the those qemu_kvm.py translations. And, would you like to update subtest.cfg together?
> > 
> > 
> > Hi @vivianQizhu This change does not affect libvirt. Only `Host_RHEL.m5 `uses `virtio`, I will delete it in the next update. Does subtest.cfg also need to be changed? At present, when I run the case, this file will automatically get the configuration.
> 
> libvirt_vm.py also consumes `nic_model`, it should be affected and I am not sure if it is correct. If the modification is not for libvirt_vm, would you help explain why we need this change? From my understanding qemu_vm already change it to virtio-net-xxx when it is defined to `virtio`, why this change is now necessary?

Hi @vivianQizhu  Since when I automating a new case, I wrote `nic_model=virtio-net-pci `in the cfg file. But when python is executed, the driver of the hot plug to the guest is always virtio. So I changed this file.

PaulYuuu:
Hello @vivianQizhu, the root cause is hotplug_nic cannot parse the nic_nodel automatically, so @yanglei-rh submit this one.
https://github.com/avocado-framework/avocado-vt/blob/503d28b6123c39127c1652c641cba44a8cc575a4/virttest/qemu_vm.py#L4005-L4010

vivianQizhu:
@yanglei-rh Please check here: https://github.com/avocado-framework/avocado-vt/blob/master/virttest/libvirt_vm.py#L884 I am not sure if it would be correct for libvirt if you modify it this way. Only if you can confirm that it is also okay for libvirt_vm, otherwise I suppose you can only modify it inside hotplug process.

yanglei-rh:
> @yanglei-rh Please check here: https://github.com/avocado-framework/avocado-vt/blob/master/virttest/libvirt_vm.py#L884 I am not sure if it would be correct for libvirt if you modify it this way. Only if you can confirm that it is also okay for libvirt_vm, otherwise I suppose you can only modify it inside hotplug process.

Hi @vivianQizhu  I tried to test this patch with tp-libvirt, and the test passed. But I only tested the case related to hotplug. I am not sure if the current changes will affect other cases. If this is a concern, I think we can close this patch, and then I change the files involved in hotplug. Is it ok?

PaulYuuu:
After double-checked, the network model of `virt-install` is `virtio`, so to avoid affecting libvirt, we should parse virtio-net device in hotplug_nic part.
```
# virt-install -h | grep network
  --pxe                 Boot from the network using the PXE protocol
  -w NETWORK, --network NETWORK
                        Configure a guest network interface. Ex:
                        --network bridge=mybr0
                        --network network=my_libvirt_virtual_net
                        --network network=mynet,model=virtio,mac=00:11...
```

yanglei-rh:
> hotplug_nic

Hi @PaulYuuu Thanks for your help, I discussed the above problem with @vivianQizhu  on IRC. For function qemu_vm.hotplug_nic(self, **params), I don't think I have the ability to rewrite it. Could you provide some help here? Or I still use `pci_model = vrirtio-net-pci `in the `nic_acpi_index` case.

PaulYuuu:
As I mentioned above, I think you can try to fix this issue in:
https://github.com/avocado-framework/avocado-vt/blob/503d28b6123c39127c1652c641cba44a8cc575a4/virttest/qemu_vm.py#L3989-L4006
or:
https://github.com/avocado-framework/avocado-vt/blob/503d28b6123c39127c1652c641cba44a8cc575a4/virttest/qemu_vm.py#L3869-L3888

The second one is a better choice.

yanglei-rh:
```
def add_nic(self, **params):
        """
        Add new or setup existing NIC, optionally creating netdev if None
        """
        # returns existing or new nic object
        nic = super(VM, self).add_nic(**params)
        nic_index = self.virtnet.nic_name_index(nic.nic_name)
        nic.set_if_none('vlan', str(nic_index))
        nic.set_if_none('device_id', utils_misc.generate_random_id())
        nic.set_if_none('queues', '1')
        if 'netdev_id' not in nic:
            # virtnet items are lists that act like dicts
            nic.netdev_id = self.add_netdev(**dict(nic))
        def get_nic_model():
            machine_type = self.params.get("machine_type")
            if "s390" in machine_type:
                nic_model = "virtio-net-ccw"
            elif '-mmio:' in machine_type:
                nic_model = "virtio-net-device"
            else:
                nic_model = "virtio-net-pci"
            return nic_model
        nic.set_if_none('nic_model', get_nic_model())
        nic.set_if_none('queues', params.get('queues', '1'))
        if params.get("enable_msix_vectors") == "yes" and int(nic.queues) > 1:
            nic.set_if_none('vectors', 2 * int(nic.queues) + 2)
        return nic
```
@PaulYuuu I tried to make the above changes. But in the cases, the driver of hotplug is still virtio. The reason for this problem is that my starting point is wrong, or there are still other restrictions in the framework code？ Could you provide further help? Thanks in advance.

yanglei-rh:
> @yanglei-rh, `nic = super(VM, self).add_nic(**params)` will set nic.nic_model to `virtio`, so `nic.set_if_none('nic_model', get_nic_model())` not works. You should use `nic.nic_model = get_nic_model()` or handle it at the beginning.

@PaulYuuu Thanks for your help.

yanglei-rh:
Hello @vivianQizhu @PaulYuuu Could you please help me review it again? Thanks in advance.

PaulYuuu (approved):

vivianQizhu:
@yanglei-rh Please update your commit title and commit message. Thanks.

yanglei-rh:
> @yanglei-rh Please update your commit title and commit message. Thanks.

Hi @vivianQizhu Updated, please help me review it again, thanks a lot.

vivianQizhu:
@yanglei-rh We might have the modified module name as the prefix of the commit title: `qemu_vm: Add designation of nic driver model for different platforms`

vivianQizhu:
@yanglei-rh @PaulYuuu Thanks.

PR #3232, opened 2021-09-22
Updated connect_libvirtd to support su

kamilvarga:
Since the su did not work for some special cases, the new terminal
has to be opened and therefore the code was updated to support this.

Signed-off-by: kvarga <kvarga@redhat.com>

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -3087,7 +3087,7 @@ def connect_libvirtd(uri, read_only="", virsh_cmd="list", auth_user=None,
                                               uri, virsh_cmd, vm_name)
     # allow specific user to run virsh command
     if su_user != "":
-        command = "su %s -c '%s'" % (su_user, command)
+        command = "su %s -P -c '%s'" % (su_user, command)
 
     logging.info("Execute %s", command)
     # setup shell session


virttest/utils_test/libvirt.py
```diff
@@ -3087,7 +3087,7 @@ def connect_libvirtd(uri, read_only="", virsh_cmd="list", auth_user=None,
                                               uri, virsh_cmd, vm_name)
     # allow specific user to run virsh command
     if su_user != "":
-        command = "su %s -c '%s'" % (su_user, command)
+        command = "su %s -P -c '%s'" % (su_user, command)
```
chunfuwen, line 3090: is capital letter 'P' or  lower case 'p' ? I assume here it should be lower case 'p'

kamilvarga: It is capital P to create a new pseudoterminal as per: https://man7.org/linux/man-pages/man1/su.1.html

chunfuwen: On my local environment, I met some issue when use su -P:
```
For more details see su(1).
[root@chunfulap2 pytestwcf]# su -P
su: invalid option -- 'P'


```

Yingshun: It works for me. you should add 'su_user', right?

kamilvarga: > On my local environment, I met some issue when use su -P:
> 
> ```
> For more details see su(1).
> [root@chunfulap2 pytestwcf]# su -P
> su: invalid option -- 'P'
> ```

As mentioned by @Yingshun you probably missed new 'su_user' in your command. It works on my local Fedora as well as RHEL 8.4 and beaker RHEL9.0 machines.

Issue #3239, opened 2021-09-29
VT runner customization hook


pevogam:
This is to specify all the remaining unsolved VT runner problems I encountered migrating from the serial to the parallel runner. They all boil down to the isolated memory space for the VT runner and previous assumptions of shared memory addresses.

1) Using the `env_process` hooks is no longer possible. In the past we could customize pre- and post-processing behaviors by assigning certain hook functions to
```python
preprocess_vm_off_hook = None
preprocess_vm_on_hook = None
postprocess_vm_on_hook = None
postprocess_vm_off_hook = None
```
from Avocado plugins on the job level. This is now no longer possible because the VT runner using the env process module is its own process.

2) Importing or changing system paths to provide the VT runner with more utilities cannot be done. An avocado plugin could so far do `sys.path.insert(1, utility_path)` and rely on the fact that all extra test utilities will be provided within the test code. This is no longer the case since changing the system paths on the job level does not change anything on the test runner level which has its own process.

3) Setting or changing logging behavior from external libraries can only be done within tests and not at a single location for all of them. If a test uses an external library that logs messages and wants to change the logging level or some other tweak, doing so on the job level will not affect anything on the test level as the VT runner performs its own imports and initial logging configuration. We can't for instance do this:
```python
import logging
log = logging.getLogger("Pyro4.core")
log.setLevel(99)
```
and have to do the same within the VT runner in order to affect the tests.

Considering all these problems have the same root cause, I am wondering what do you think about adding ability to customize the VT runner like run some initial script/code that configures its independent memory space. Something like this could support all limitations described above.

pevogam:
@richtja @clebergnu Tagging you here since this might be a useful summary to you with all remaining problems related to the new test isolation I have collected.

PR #3242, opened 2021-10-01
Add a possibility to customize VT tests via pre-run hooks

pevogam:
This solves remaining process-isolation issues and background described in 

https://github.com/avocado-framework/avocado-vt/issues/3239

### DIFF
--- a/avocado_vt/test.py
+++ b/avocado_vt/test.py
@@ -202,6 +202,10 @@ class VirtTest(test.Test, utils.TestUtils):
 
     def _runTest(self):
         params = self.params
+        if params.get("test_pre_hook"):
+            self.log.info(f"Executing VT test pre-hook from {params['test_pre_hook']}")
+            with open(params["test_pre_hook"]) as f:
+                exec(f.read())
 
         # Report virt test version
         logging.info(version.get_pretty_version_info())


pevogam:
@richtja, @clebergnu, it would be great if we could review and eventually merge a fix like this before the LTS since it will remove all the remaining process isolation issues we have like env hooks, custom logging, etc.

avocado_vt/test.py
```diff
@@ -202,6 +202,9 @@ def runTest(self):
 
     def _runTest(self):
         params = self.params
+        if params.get("test_pre_hook"):
+            self.log.info(f"Executing VT test pre-hook from {params['test_pre_hook']}")
+            exec(open(params["test_pre_hook"]).read())
```
richtja, line 207: Hi @pevogam I am not quite understand how this should work. Can you give me an example how the `params["test_pre_hook"]` value should look like? Also I have a feeling that this code might be possible to move to the `env_process.preprocess` method and copy the hooks from `params` to the preprocess_vm_off_hook and other variables. What do you think?

pevogam: What I used for testing here was a path to a small customization script as you could probably guess from `exec(open(params["test_pre_hook"]).read())`. This is where I could do all of the steps like:

1) Changing logging levels for some external libraries like logging.getLogger("extlib").setLevel(99)
2) Inserting some special utility paths like `sys.path.insert(1, utility_path)`.
3) Setting multiple env processing hooks like
```python
env_process.preprocess_vm_off_hook = off_state(ss.get_states)
env_process.preprocess_vm_on_hook = on_state(ss.get_states)
env_process.postprocess_vm_on_hook = on_state(ss.set_states)
env_process.postprocess_vm_off_hook = off_state(ss.set_states)
```
As you can notice, we might use some `params` keys for the env process hooks but this is just a minor part of what we actually need to do in order to customize the runner behavior.

samiraguiar: My only remark is regarding not explicitly closing the file, which puts the burden on the GC and in some rare cases can lead to issues (such as running ou of file descriptors or preventing unmount operations).

pevogam: I would have preferred the one liner knowing this since there isn't much danger of unmounting in this stage of the code and GC works just fine (and it was the main recommended way on some source forums) but we cannot sacrifice some better patterns for the sake of prettier syntax and API limitations so I went ahead with a context manager.

PR #3247, opened 2021-10-08
virttest/qemu_vm.py: Change the command that add cpu device.

6-dehan:
virttest/qemu_vm.py: change the outdate command "cpu-add" to
 "device_add".

ID: 1926584


vivianQizhu:
@nanliu-r Request for your review, thanks.

PR #3250, opened 2021-10-09
Add a decorator to shield the daemon mode for signal_process 

jinqi7:

### DIFF
--- a/virttest/utils_split_daemons.py
+++ b/virttest/utils_split_daemons.py
@@ -419,3 +419,25 @@ def is_modular_daemon(session=None):
         else:
             IS_MODULAR_DAEMON[host_key] = False
     return IS_MODULAR_DAEMON[host_key]
+
+
+def adopt_func_on_libvirtd(func):
+    """
+        Decorator method to handle process in different daemon modes
+
+        :params func: function to be wrapped
+    """
+
+    def signal_libvirtd(*args, **kwargs):
+        """
+        Wrap method of the decorator
+
+        :params *args: arguments of the function
+        :params **kwargs: dict args of the function
+        """
+        list_args = list(args)
+        if is_modular_daemon() and args[0] == 'libvirtd':
+            list_args[0] = 'virtqemud'
+
+        return func(*list_args, **kwargs)
+    return signal_libvirtd


jinqi7:
Related to tp-libvirt PR https://github.com/autotest/tp-libvirt/pull/3813.
avocado run --vt-type libvirt libvirt_hooks.vm.daemon_t.exit_0
No python imaging library installed. Screendump and Windows guest BSOD detection are disabled. In order to enable it, please install python-imaging or the equivalent for your distro.
No python imaging library installed. PPM image conversion to JPEG disabled. In order to enable it, please install python-imaging or the equivalent for your distro.
JOB ID : cb5f5a23a4ffa4773ff04bf0799b5d8bc687d970
JOB LOG : /root/avocado/job-results/job-2021-10-09T02.12-cb5f5a2/job.log
(1/1) type_specific.io-github-autotest-libvirt.libvirt_hooks.vm.daemon_t.exit_0: PASS (31.97 s)
RESULTS : PASS 1 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME : 33.37 s

virttest/utils_misc.py
```diff
@@ -2230,6 +2231,7 @@ def program_is_alive(program_name, pid_files_dir=None):
     return pid_is_alive(pid)
 
 
+@utils_split_daemons.adopt_func_on_libvirtd
```
chunfuwen, line 2234: it sounds like this method is also used by tp-qemu, so it is not proper to apply libvirtd related logic decorator on this method. One way to move forward is to copy similar function to utils lib  exclusively reserved for libvirt

chunfuwen:
@kylazhang 
@Yingshun 
please help take a look

virttest/utils_split_daemons.py
```diff
@@ -419,3 +419,25 @@ def is_modular_daemon(session=None):
...
+        Wrap method of the decorator
+
+        :params *args: arguments of the function
+        :params **kwargs: dict args of the function
+        """
+        if is_modular_daemon() and args[0] == 'libvirtd':
+            list_args = list(args)
+            list_args[0] = 'virtqemud'
+
+        return func(*list_args, **kwargs)
```
chunfuwen, line 442: it looks like there is potential risk here, if is_modular_daemon() return false,  call func(*list_args, **kwargs) may throw list_args is unknown variable error. Therefore, pulling list_args = list(args) out of scope of if statemnet could be one way to fix this risk

jinqi7: Updated.

virttest/utils_misc.py
jinqi7: If so, I'm not quite understand what's the meaning for the comment in tp-libvirt PR 3813.

virttest/utils_split_daemons.py
```diff
@@ -419,3 +419,25 @@ def is_modular_daemon(session=None):
         else:
             IS_MODULAR_DAEMON[host_key] = False
     return IS_MODULAR_DAEMON[host_key]
+
+
+def adopt_func_on_libvirtd(func):
```
chunfuwen, line 424: adopt_func_on_libvirtd --> adapt_func_on_libvirtd ?

virttest/utils_misc.py
chunfuwen: @jinqi7 
Since signal_program is shared by both qemu and libvirt, so it is not proper to apply libvirt specific decorator here.
But I believe decorator adopt_func_on_libvirtd is still useful, so you can still use this PR to add adopt_func_on_libvirtd only.
Regarding  tp-libvirt PR 3813, there are two ways moving forward:
1)One is to port signal_program into your code in your tp-libvirt, then apply decorator on that
2)Two is to still use current signal_program in avocado-vt, but go back to use old way in previous tp-libvirt  implementation.

jinqi7: @chunfuwen, For the first way,  do you mean keep duplicate functions in the script? Maybe it's not good.
For the second way, do you mean restore it back to the script below?  And also keep "adopt_func_on_libvirtd"  function in avocado-vt?
            if utils_split_daemons.is_modular_daemon():
                utils_misc.signal_program('libvirtd', 1,
                                      '/var/run')
           else:
               utils_misc.signal_program('virtqemud', 1,
                                      '/var/run')

chunfuwen: Yes, For second way, you can restore it back to the script as mentioned previously in tp-libvirt.

jinqi7: Updated.  Shall we still keep the "adopt_func_on_libvirtd" without being used?@chunfuwen

jinqi7:
Since the patch is not used by tp-libvirt, I'll close it for now. If it's needed, we can open it again.

PR #3253, opened 2021-10-14
qemu_vm: change the smp pattern to match maxcpus correctly

yanan-fu:
Upstream qemu change with commit 553dc36b38, it tweak
[, maxcpus=cpus] to [, maxcpus=maxcpus] in qemu-options

id: 2013885

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -586,7 +586,7 @@ class VM(virt_vm.BaseVM):
 
         def add_smp(devices):
             smp_str = " -smp %d" % self.cpuinfo.smp
-            smp_pattern = "smp .*\[,maxcpus=cpus\].*"
+            smp_pattern = "smp .*\[,maxcpus=.*\].*"
             if devices.has_option(smp_pattern):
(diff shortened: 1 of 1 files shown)


YongxueHong (approved):
LGTM.

virttest/qemu_vm.py
```diff
@@ -586,7 +586,7 @@ def add_log_anaconda(devices, pci_bus='pci.0'):
 
         def add_smp(devices):
             smp_str = " -smp %d" % self.cpuinfo.smp
-            smp_pattern = "smp .*\[,maxcpus=cpus\].*"
+            smp_pattern = "smp .*\[,maxcpus=.*\].*"
```
PaulYuuu, line 589: How about `"smp .*\[,maxcpus=\w*cpus\].*"`

PaulYuuu: Yes, both are OK. But `.*\]` may leak to other properties' namespace.
BTW, for the easiest way to find maxcpus support, it can be `smp .*maxcpus=.*`? If we ignore all risks.

PR #3263, opened 2021-11-02
libvirt_nested: not control daemon

dzhengfy:
Previous restarting libvirtd daemon method requires to recognize if the
environment is a monolithic vs modular daemons as default. But after
the libvirt packages are installed in L1 guest, those daemons are not
expected to be started automatically. So preovious restarting method
can not work correctly on right daemons. To make the right daemons
start correctly, it is easy to reboot the vm directly or explicitly
restart it by name in certain case which will be more flexible. So
this fix is to remove that part and leave the control to the invoker.


Yingshun (approved):

cliping (approved):
LGTM

PR #3270, opened 2021-11-10
backends: add new test type virttools

smitterl:
Adds a new test type 'virttools' intended to be used for tools offered by the 'virt-manager' repository:

- virt-manager
- virt-clone
- virt-install
- virt-xml

A passing test case for this new type is available in https://github.com/autotest/tp-libvirt/pull/3923

### DIFF
--- a/virttest/backends/virttools/cfg/tests.cfg
+++ b/virttest/backends/virttools/cfg/tests.cfg
@@ -1,0 +1,26 @@
+# File reserved for avocado-vt use, don't modify.
+include tests-shared.cfg
+
+variants:
+    - @virttools_quick:
+        start_vm = no
+        virt_install_binary = /usr/bin/virt-install
+        qemu_img_binary = /usr/bin/qemu-img
+        hvm_or_pv = hvm
+        use_os_variant = yes
+        use_os_type = yes
+        only qcow2
+        only virtio_net
+        only virtio_blk
+        only smp2
+        only no_9p_export
+        only no_virtio_rng
+        only no_pci_assignable
+        only (image_backend=filesystem)
+        only smallpages
+        only default_bios
+        only bridge
+
+
+
+only virttools_quick
--- a/virttest/backends/virttools/cfg/drivers.cfg
+++ b/virttest/backends/virttools/cfg/drivers.cfg
@@ -1,0 +1,3 @@
+# Configuration files for some special variables in specific uri.
+variants:
+    - qemu:
--- a/virttest/backends/virttools/cfg/tests-shared.cfg
+++ b/virttest/backends/virttools/cfg/tests-shared.cfg
@@ -1,0 +1,34 @@
+# Copy this file to tests-shared.cfg and edit it.
+#
+# This file contains the base test set definitions, shared among single host
+# and multi host jobs.
+
+# Virtualization type (qemu or libvirt)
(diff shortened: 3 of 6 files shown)


smitterl:
@chunfuwen @Yingshun @dzhengfy Please, can you help review?

virttest/backends/virttools/cfg/tests.cfg
```diff
@@ -0,0 +1,24 @@
+# File reserved for avocado-vt use, don't modify.
+include tests-shared.cfg
+
+variants:
+    - @libvirt_quick:
```
chunfuwen, line 5: could be virttools_quick to syn up with your intention?

virttest/backends/virttools/cfg/tests.cfg
```diff
@@ -0,0 +1,24 @@
...
+        start_vm = no
+        virt_install_binary = /usr/bin/virt-install
+        qemu_img_binary = /usr/bin/qemu-img
+        hvm_or_pv = hvm
+        use_os_variant = yes
+        use_os_type = yes
+        only qcow2
```
chunfuwen, line 12: Do we really need below filters when you enable virttools type?

dzhengfy (commented):
Others LGTM

virttest/backends/virttools/cfg/default_tests
```diff
@@ -0,0 +1 @@
+
```
Yingshun, line 1: This file only has a blank line, is it necessary? (Can we remove this line? or this file?

virttest/backends/virttools/cfg/tests.cfg
smitterl: @chunfuwen In short, yes, I believe they are necessary, at least I prefer to keep them. I tried to run tests without them and I conclude it's better to keep them.

Details:
-----------
I believe they are not necessary _functionally_. However, I've found they have a big impact regarding performance. When I tried test case without these filters, they never started in a reasonable time while I found the cartesian config was still creating its test dictionaries in the background (can be seen in the crash log after ^C).

I assume the reason is that the number of variants explodes without these standard filters.

Please notice that many of these filters usually don't have an impact on the tp-libvirt test functions but are rather tp-qemu specific. I took these filters to be the same as the ones the tp-libvirt/libvirt test type uses, where they possibly aren't functionally necessary either. I'm not sure.

In any case, as mentioned before, without these filters I never got the test to start in a reasonable time, so I propose let's keep them. :)

virttest/backends/virttools/cfg/default_tests
smitterl: I successfully ran the test case without the file so I removed it :)

dzhengfy (approved):

Yingshun (approved):
LGTM

chunfuwen (approved):
LGTM

chunfuwen:
@smitterl , some checks were not successful, perhaps you may take a look before it merged

chunfuwen:
The CI failed on fedora 34 AVOCADO_SRC with error message :ERROR: run() got an unexpected keyword argument 'allow_output_check' (1.89 s), which had been similarly fixed by https://github.com/avocado-framework/avocado-vt/pull/3300. 
But the fix may not be proprogated to container image format

PR #3277, opened 2021-11-18
is_qemu_feature_supported: Support to check if a test is supported

Yingshun:
Add a function to check if a test is supported in the installed qemu
version.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -2056,6 +2056,34 @@ def compare_qemu_version(major, minor, update, is_rhev=True, params={}):
     return True
 
 
+def is_qemu_function_supported(params, ignore_error=False):
+    """
+    Check whether the function is supported in this qemu-kvm version by comparing
+    the installed qemu-kvm version
+
+    :param params: Dictionary with the test parameters
+    :param ignore_error: Whether to raise an exception
+    :raise: When ignore_error is set False, raise TestCancel if the feature is
+        not supported
+    :return: True if the function is supported;
+        False if the function is not supported when ignore_error is set to True
+    """
+    supported_qemu_ver = eval(
+        params.get("func_supported_since_qemu_kvm_ver", '()'))
+    unsupported_err_msg = params.get("qemu_unsupported_err_msg",
+                                     "This qemu version doesn't support "
+                                     "this function.")
+
+    if supported_qemu_ver:
+        if not compare_qemu_version(*supported_qemu_ver, False):
+            if ignore_error:
+                LOG.error(unsupported_err_msg)
+                return False
+            else:
+                raise exceptions.TestCancel(unsupported_err_msg)
+    return True
+
+
 class ForAll(list):
 
     def __getattr__(self, name):


Yingshun:
**Test result:**
```
>>> d={'func_supported_since_qemu_kvm_ver': '(1,3,4)'}
>>> utils_misc.is_qemu_feature_supported(d)
True
>>> d1={'func_supported_since_qemu_kvm_ver': '(10,3,4)'}
>>> utils_misc.is_qemu_feature_supported(d1)
  File "/var/tmp/libvirt-ci/runtest/avocado-vt/avocado-vt/virttest/utils_misc.py", line 2083, in is_qemu_feature_supported
    raise exceptions.TestCancel(unsupported_err_msg)
avocado.core.exceptions.TestCancel: This qemu version doesn't support this function.
>>> utils_misc.is_qemu_feature_supported(d1, True)
This qemu version doesn't support this function.
False
>>> 
```

dzhengfy (approved):
LGTM

virttest/utils_misc.py
```diff
@@ -2056,6 +2056,34 @@ def compare_qemu_version(major, minor, update, is_rhev=True, params={}):
...
+
+    :param params: Dictionary with the test parameters
+    :param ignore_error: Whether to raise an exception
+    :raise: When ignore_error is set False, raise TestCancel if the feature is
+        not supported
+    :return: True if the feature is supported;
+        False if the feature is not supported when ignore_error is set to True
+    """
+    supported_qemu_ver = eval(
+        params.get("func_supported_since_qemu_kvm_ver", '()'))
```
chunfuwen, line 2072: func_supported_since_qemu_kvm_ver -->feature_supported_since_qemu_kvm_ver?

virttest/utils_misc.py
```diff
@@ -2056,6 +2056,34 @@ def compare_qemu_version(major, minor, update, is_rhev=True, params={}):
     return True
 
 
+def is_qemu_feature_supported(params, ignore_error=False):
+    """
+    Check whether the function is supported in this qemu-kvm version by comparing
```
chunfuwen, line 2061: function-->feature?

virttest/utils_misc.py
```diff
@@ -2056,6 +2056,34 @@ def compare_qemu_version(major, minor, update, is_rhev=True, params={}):
...
+    """
+    supported_qemu_ver = eval(
+        params.get("func_supported_since_qemu_kvm_ver", '()'))
+    unsupported_err_msg = params.get("qemu_unsupported_err_msg",
+                                     "This qemu version doesn't support "
+                                     "this function.")
+
+    if supported_qemu_ver:
+        if not compare_qemu_version(*supported_qemu_ver, False):
+            if ignore_error:
```
chunfuwen, line 2079: From function name implication meaning, not suggest add additional parameter : ignore_error to enable exceptions.TestCancel. Better just return pure bool value:True or False. You may use another high level function to enable exceptions.TestCancel

virttest/utils_misc.py
Yingshun: May I ask what the difference between 'function' and 'feature' is here?

Yingshun: I've checked with chunfu, updated function name to 'is_qemu_function_supported' as discussed.

virttest/utils_misc.py
Yingshun: As above.

virttest/utils_misc.py
Yingshun: I think this function can be a high level qemu version comparing function. We don't need to add another to raise cancel exception I think. And it also can return True/False, so we can continue to check other things if needed.  What do you think?

chunfuwen (approved):
LGTM

PR #3297, opened 2021-12-10
Fix the get_host_ip_address()

yalzhang:
Sometimes there are 2 default gateway on the host, which cause the
get_ip_address_by_interface() to fail with value error. Update it
to pass a single interface name.

Signed-off-by: Yalan Zhang <yalzhang@redhat.com>

PR #3305, opened 2021-12-15
Fix get iface error

kylazhang:
resolve the get method as ip route may display like this

default via 10.73.11.254 dev eno1
default via 10.73.11.254 dev eno1 proto dhcp metric 100

which cause error from "addr = netifaces.ifaddresses(ifname).get(ver)" like
ValueError: You must specify a valid interface name.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3885,7 +3885,7 @@ def get_default_gateway(iface_name=False, session=None):
     :rtype: string
     """
     if iface_name:
-        cmd = "ip route | awk '/default/ { print $5 }'"
+        cmd = "ip route | awk '/default.*proto/ { print $5 }'"
     else:
         cmd = "ip route | awk '/default/ { print $3 }'"
     try:


kylazhang:
```
(1/1) type_specific.io-github-autotest-libvirt.virtual_network.multivms.port_isolated.iface.update_iface.none_to_yes: PASS (66.62 s)
```

kylazhang:
@yalzhang please help to make sure, thanks

yalzhang:
> @yalzhang please help to make sure, thanks

Hi Kyla, It works.

yalzhang (approved):

yalzhang:
Hi kyla, actually, I have a pr to fix this issue as well https://github.com/avocado-framework/avocado-vt/pull/3297/files. I think this one is more robust after I got below outputs:
# ip route | grep default 
default via 192.168.122.1 dev br0 

with current patch, it can not get  a default route.

I have checked the meaning of 'proto' by "man ip-route", it is the routing protocol identifier of this route. I'm not sure if it is mandatory. But the above example shows it can be missing sometimes. And I have asked network QE and they said there can be 2 default route, and usually the first one will take effect.

kylazhang:
similar with https://github.com/avocado-framework/avocado-vt/pull/3297, but it can not cover the issue like
```
default via 192.168.122.1 dev br0
default via 192.168.0.1 dev enp4s0 proto dhcp metric 101
```
which want to get the first one. So close it. Thanks

PR #3306, opened 2021-12-16
test_setup: Introduce new variable 'extra_size' for HugePageConfig

huangyum:
The function get_target_hugepages() only calculates the vm memory,
if hotplug memory with hugepages after guest boot, it can cause
error due to lack of hugepages. Introduce 'extra_size' to allocate
hugepages beforehand to avoid such error.

ID: 2012092

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -406,6 +406,8 @@ class HugePageConfig(object):
         self.lowest_mem_per_vm = float(normalize_data_size(vm_mem_minimum))
         utils_memory.drop_caches()
         self.ext_hugepages_surp = utils_memory.get_num_huge_pages_surp()
+        self.extra_size = float(normalize_data_size(params.get("extra_hp_size",
+                                                               "0")))
 
         overcommit_hugepages = int(params.get("overcommit_hugepages", 0))
         if overcommit_hugepages != self.over_commit.proc_fs_value:
@@ -486,7 +488,7 @@ class HugePageConfig(object):
             self.vms = self.max_vms
         # memory of all VMs plus qemu overhead of 128MB per guest
         # (this value can be overridden in your cartesian config)
-        vmsm = self.vms * (self.mem + self.qemu_overhead)
+        vmsm = self.vms * (self.mem + self.qemu_overhead) + self.extra_size
         target_hugepages = int(vmsm * 1024 // self.hugepage_size)
 
         # FIXME Now the buddyinfo can not get chunk info which is bigger


huangyum:
(1/2) Host_RHEL.m9.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.x86_64.io-github-autotest-qemu.hotplug_mem_reserve.reserve_on: PASS (66.37 s)
 (2/2) Host_RHEL.m9.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.x86_64.io-github-autotest-qemu.hotplug_mem_reserve.reserve_off: PASS (62.74 s)
@MiriamDeng @yanan-fu Request review, thanks.

MiriamDeng:
LGTM
(1/2) Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.6.0.ppc64le.io-github-autotest-qemu.hotplug_mem_reserve.reserve_on: PASS (113.87 s)
(2/2) Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.6.0.ppc64le.io-github-autotest-qemu.hotplug_mem_reserve.reserve_off: PASS (140.09 s)

huangyum:
@yanan-fu @YongxueHong Request for review, thanks.

yanan-fu:
IMO, The existing logic to get the target hugepage is not accurate, but yes, no the problem of this patch.
This is patch is only a workaround for the existing logic error to make it work for your situation.

I think a real solution is better, thanks!

huangyum:
> IMO, The existing logic to get the target hugepage is not accurate, but yes, no the problem of this patch. This is patch is only a workaround for the existing logic error to make it work for your situation.
> 
> I think a real solution is better, thanks!
Yes, you are right. To get the accurate target hugepages is very complicated, as we may boot guest with a few numa nodes, and each numa node may have different memory backend. This patch is only setting a new variable for hotplugged memory. Can we only consider the hotplugged memory for now and handle the vm memory later? As far as I'm concerned, there is no a good way to fix all the things together now. And I don't want it to block the new case. What do you think?

yanan-fu:
Those mems be defined for hotplug is only test case level and be used when test running -- after test_setup. 
Without getting the accurate target hugepage size, but only use `mem` plus hotplug mem, i am ok with this change. Map hotplug mem to `extra_hp_size` in test case.
@YongxueHong  What's your opinion ? Thanks!

huangyum:
After discussed with @YongxueHong , will update on tp-qemu side.  As for the logic issue in vt, will not fix for now.  Thanks all.

PR #3308, opened 2021-12-17
ceph: add func to create keyring

dzhengfy:
When client to access the ceph server, the keyring file should be
exist in /etc/ceph/

### DIFF
--- a/virttest/ceph.py
+++ b/virttest/ceph.py
@@ -241,6 +241,27 @@ def create_config_file(ceph_monitor):
 
 
 @error_context.context_aware
+def create_keyring_file(client_name, client_key):
+    """
+    Create a ceph keyring file
+
+    :params client_name: the client name to access the ceph server
+    :params client_key: the client key to access the ceph server
+    :return: the keyring file path
+    """
+    ceph_dir = "/etc/ceph"
+    keyring_file = os.path.join(ceph_dir, "ceph.client.admin.keyring")
+    if not os.path.exists(ceph_dir):
+        os.makedirs(ceph_dir)
+    if os.path.exists(keyring_file):
+        os.remove(keyring_file)
+    with open(keyring_file, 'w') as f:
+        f.write('[%s]\n' % client_name)
+        f.write('        key = %s\n' % client_key)
+    return keyring_file
+
+
+@error_context.context_aware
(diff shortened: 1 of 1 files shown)


virttest/ceph.py
```diff
@@ -240,6 +240,27 @@ def create_config_file(ceph_monitor):
     return ceph_cfg
 
 
+@error_context.context_aware
+def create_keyring_file(client_name, client_key):
```
chunfuwen, line 244: In current ceph related cases, we create keyring_file in the tmp folder of tp-libvirt code, and specify with --keyring when using it. Do we really need override ceph keyring in /etc/ceph folder?

dzhengfy: @chunfuwen ,
Yes, we did add --keyring for rbd command, for example below:
`rbd -m libvirtceph.usersys.redhat.com --keyring /tmp/avocado_iknmvl8y/ceph.key info ovmf-pool/bios.qcow2 && rbd -m libvirtceph.usersys.redhat.com --keyring /tmp/avocado_iknmvl8y/ceph.key rm ovmf-pool/bios.qcow2`
But after that, we also has qemu-img command, for example below, if no keyring file, it will return 'Operation not permitted' error when I use another ceph server.
`qemu-img convert -O raw /tmp/avocado_iknmvl8y/released_os.qcow2 rbd:ovmf-pool/bios.qcow2:mon_host=<mon_ip>`

chunfuwen: Command 'rbd -m libvirtauthceph.usersys.redhat.com --keyring /var/lib/avocado/data/avocado-vt/ceph.key info virtual-disk-ceph-pool/attach.img 2> /dev/null|| qemu-img convert -O raw /var/lib/avocado/data/avocado-vt/avocado-vt-vm1_test.img rbd:virtual-disk-ceph-pool/attach.img:mon_host=libvirtauthceph.usersys.redhat.com:id=admin:key=AQA8Ij5f97SyNxAAePrE+BysFp4M3KCD1JZB4Q==' finished with 0 after 0.4634408950805664s

it is one of output from running ceph case, does it satisfy your request?

kylazhang (approved):

chunfuwen (approved):
LGTM

PR #3302, opened 2021-12-14
virsh: Fix can't get event info in some env

liuyd96:
Signed-off-by: Liu Yiding <liuyd.fnst@fujitsu.com>

PR #3310, opened 2021-12-20
virsh: Remove unnecessary change

liuyd96:
We did some changes in 4873a25 to fix error "Not found event tray-change after
40 seconds".
Since the error is hard to be reproduced locally and the error breaks a lot
cases in CI. So we merged that draft PR.

After further debug, we noticed that some changes are unnecessary or
inappropriate. Replacing time.sleep with utils_misc.wait_for is much better.

Fix: 4873a25f Debug draft
Signed-off-by: Liu Yiding <liuyd.fnst@fujitsu.com>

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -38,6 +38,7 @@ from functools import wraps
 import aexpect
 from aexpect import remote
 
+from avocado.core import exceptions
 from avocado.utils import path
 from avocado.utils import process
 
@@ -701,9 +702,7 @@ class EventTracker(object):
                     arg in inspect.signature(func).parameters else None
 
             def _get_event_output(session):
-                # Use get_output instead of get_stripped_output to avoid
-                # missing output.
-                output = session.get_output()
+                output = session.get_stripped_output()
                 LOG.debug(output)
                 return output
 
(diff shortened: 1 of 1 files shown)


virttest/virsh.py
```diff
@@ -713,9 +712,16 @@ def _get_event_output(session):
...
-                # Sometimes the output of virsh_session is missing, add a sleep
-                # for debug. Will remove it if it is unnecessary.
-                time.sleep(3)
+                # Sometimes the output of session can't be gotten immediately.
+                # Wait for a while to avoid this situation.
+                if not utils_misc.wait_for(
+                        lambda: re.search(
+                            "Welcome to virsh",
+                            _get_event_output(virsh_session)
+                        ), 3):
```
kylazhang, line 721: Maybe you can extend the timeout to a little longer, like 10s

kylazhang (commented):
Others are ok for me

kylazhang:
@chunfuwen please help to check it as it is the https://github.com/avocado-framework/avocado-vt/pull/3302 function optimization, thanks

virttest/virsh.py
liuyd96: > Maybe you can extend the timeout to a little longer, like 10s

I think it should not take 10s to get output from virsh session. Otherwise there maybe a regression I think.

chunfuwen: @liuyd96 , I suggest timeout can be 10 seconds considering some extreme cases on variant platform arch

chunfuwen (approved):
LGTM

kylazhang (approved):

PR #3335, opened 2022-01-25
test_setup: restore correct original_config in TransparentHugePageConfig

yanan-fu:
1. The 'original_config' be used to restore in cleanup() should be the config
   before setup, it is incorrect to get a new 'original_config' with a new
   TransparentHugePageConfig instance when doing cleanup.
2. Logically, khugepaged_test doesn't belong to setup(), so move it
   out. It can be called on demand.

id: 1654144

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -212,13 +212,14 @@ class SetupManager(object):
 
 class TransparentHugePageConfig(object):
 
-    def __init__(self, test, params, session=None):
+    def __init__(self, test, params, env, session=None):
         """
         Find paths for transparent hugepages and kugepaged configuration. Also,
         back up original host configuration so it can be restored during
         cleanup.
         """
         self.params = params
+        self.env = env
         self.session = session
 
         RH_THP_PATH = "/sys/kernel/mm/redhat_transparent_hugepage"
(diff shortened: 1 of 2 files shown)


yanan-fu:
@luckyh @zhenyzha  Could you please help review ? Thanks!

virttest/test_setup.py
```diff
@@ -239,14 +240,14 @@ def __init__(self, test, params, session=None):
         while len(tmp_list) > 0:
             tmp_cfg = tmp_list.pop()
             test_cfg[re.split(":", tmp_cfg)[0]] = re.split(":", tmp_cfg)[1]
-        # Save host current config, so we can restore it during cleanup
+        # Save host current config
         # We will only save the writeable part of the config files
-        original_config = {}
+        current_config = {}
```
luckyh, line 245: Would you like to elaborate on why this local variable name had to be modified?

yanan-fu: I think 'original_config' is the original configuration of the test env, can be used to restore env,  but not only one instance can be created during the test, so i think it is more suitable to use 'current' when create new instance.

zhenyzha:
(1/1) Host_RHEL.m9.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.aarch64.io-github-autotest-qemu.trans_hugepage.swapping.arm64-pci: PASS (241.77 s)

LGTM.Acked-by: Zhenyu Zhang [zhenyzha@redhat.com](mailto:zhenyzha@redhat.com)

luckyh (approved):

luckyh:
Thanks to all.

PR #3339, opened 2022-02-02
libvirt_vm: allow for skipping pm-utils installation

smitterl:
Though highly recommended to be used with the qemu guest agent,
the pm-utils package might not be available or the test scenario
might not want to install it.

Add a parameter to the `prepare_guest_agent` function to allow for
tests to skip the pm-utils installation. This speeds up testing
where pm-utils is not available.

Make installation of pm-utils the default option in order to
keep default behavior as-is and not require any changes from
consuming code.


chunfuwen (approved):
it makes sense

kylazhang (approved):

Issue #3323, opened 2022-01-11
DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12


luckyh:
Got the following warning with Python 3.10, which led me to open this issue for having the update to the corresponding parts in VT.
```
$ python
Python 3.10.1 (main, Dec  9 2021, 00:00:00) [GCC 11.2.1 20211203 (Red Hat 11.2.1-7)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import imp
<stdin>:1: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
```

Reference:
https://github.com/avocado-framework/avocado-vt/blob/2a9241a92ae88ddbbd0c306879be0ae94b87aed5/avocado_vt/utils.py#L128-L129
https://docs.python.org/3.10/library/importlib.html?highlight=importlib#module-importlib

ana:
@luckyh You might want to do something like this:
https://github.com/avocado-framework/avocado/commit/dfc340f1a0b0aeddbf6699e49e94b75ed080137c

luckyh:
@ana thank you for the information!

bgartzi:
I can't assign it to myself, but I'll take it.

bgartzi:
Hi all,

I had a quick check of the scope of the issue. It seems that there are a total of 10 (6+4) imp+importlib  uses. Generally speaking, they're used for 2 objectives:

1. Importing test cases that are going to be executed in the test loop.
2. Importing some autotest functionalities, which shouldn't be supported in the future (#2963, #3205).

The way of importing modules is quite heterogeneous so it seems quite reasonable to implement an _standarized_ way of doing this inside avocado-vt from now on.  One way could be using a similar approach to the one from avocado. Brief pseudocode:

```python
def import_module(module_name, module_dir=None):
    """
        desc
    """
    if module_dir:
        sys.path.insert(0, module_dir)
    try:
        module = importlib.import_module(module_name)
    except:
        raise whatever_ModuleNotFoundException
    finally:
        if module_dir:
            sys.path.pop(0) 
```
This could work, as it could mitigate the deprecation issue and would also fit with the autotest/py2 cases while they are completely removed from the code. 

The real problem comes with the _placement_ of this piece of code. In my opinion there are three ways of approaching this:

[...]

PR #3350, opened 2022-02-11
DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12

bgartzi:
Code changes in this PR work on the imp deprecation issue (#3323). It's divided in two commits:
1.  The `virttest._wrappers` module is created. The functionality needed for carrying on dynamic imports is added, as well as some related unittests.
2. The code from `avocado_vt` and `virttests` is changed from using imp/importlib to the new method.

Note:
The modules from `virttest.remote_commander` are not using `_wrappers.import_module`, but `importlib`. This has been left in this way since it can't be assumed that virttest is installed in the remote system in which the import is going to take place.

ID: 2042377

### DIFF
--- a/virttest/_wrappers.py
+++ b/virttest/_wrappers.py
@@ -1,0 +1,72 @@
+""" The file contains some wrappers of common python base-package workflow
+    sets that can be repeated throughout the avocado-vt code.
+    Please, remember that it is a private module.
+    Please, use this module ONLY from other virttest/avocado-vt modules.
+"""
+
+import sys
+import importlib
+from importlib.machinery import (SourceFileLoader, SOURCE_SUFFIXES,
+                                 SourcelessFileLoader, BYTECODE_SUFFIXES,
+                                 ExtensionFileLoader, EXTENSION_SUFFIXES)
+
+
+_LOADERS = ((SourceFileLoader, SOURCE_SUFFIXES),
+            (SourcelessFileLoader, BYTECODE_SUFFIXES),
+            (ExtensionFileLoader, EXTENSION_SUFFIXES))
+
+
+def _load_from_spec(spec):
+    """ Just for code refactoring, it takes a module spec and does the needed
+        things so as to import it into the execution environment.
+        :param spec: module spec.
+        :type spec: ModSpec
+        :returns : Imported module instance
+    """
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[spec.name] = module
+    spec.loader.exec_module(module)
+    return module
+
+
+def import_module(name, path=None):
+    """ Imports a module named <name> to the execution environment.
+        If the module is not in the PYTHONPATH or the local path,
+        its path can be determined by the <path> argument.
+        :param name: Name of the module that is going to be imported
+        :type name: String
+        :param path: Path in which the module is located.
+                          If None, it will find the module in the current dir
+                          and the PYTHONPATH.
+        :type path: String, list of strings or  None
+        :returns: The imported module
+    """
+    if path is None:
+        if name in sys.builtin_module_names:
+            spec = importlib.machinery.BuiltinImporter.find_spec(name, path)
+            return _load_from_spec(spec)
+        path = sys.path
+    elif isinstance(path, str):
+        path = [path]
+
+    for entry in path:
+        finder = importlib.machinery.FileFinder(entry, *_LOADERS)
+        spec = finder.find_spec(name)
+        if spec is not None:
+            break
+    else:
+        raise ImportError(f"Couldn't find any module named {name}")
+    return _load_from_spec(spec)
+
+
+def load_source(name, path):
+    """ Imports the contents of a source file from <path> to the
+        execution environment inside a module named <name>.
+        :param name: Name of the module that is going to be imported
+        :type name: String
+        :param path: Path of the source file being imported.
+        :type path: String
+        :returns: The imported module
+    """
+    spec = importlib.util.spec_from_file_location(name, path)
+    return _load_from_spec(spec)
--- a/virttest/remote_commander/remote_runner.py
+++ b/virttest/remote_commander/remote_runner.py
@@ -9,7 +9,6 @@ Created on Dec 6, 2013
 
 import os
 import sys
-import imp
 import importlib
 import select
 import time
@@ -746,7 +745,10 @@ class CommanderSlaveCmds(CommanderSlave):
         :return: module.run().return
         """
         assert os.path.isfile(test_path)
-        module = imp.load_source('ext_test', test_path)
+        spec = importlib.util.spec_from_file_location('ext_test', test_path)
+        module = importlib.util.module_from_spec(spec)
+        sys.modules['ext_test'] = module
+        spec.loader.exec_module(module)
         assert hasattr(module, "run")
         run = getattr(module, "run")
         helper = Helper(self)
--- a/avocado_vt/plugins/vt_init.py
+++ b/avocado_vt/plugins/vt_init.py
@@ -1,5 +1,3 @@
-import importlib
-
 from avocado.core import plugin_interfaces
 from avocado.core.loader import loader
 from avocado.core.settings import settings
@@ -14,6 +12,7 @@ from virttest.standalone_test import (SUPPORTED_DISK_BUSES,
                                       SUPPORTED_NIC_MODELS,
                                       SUPPORTED_TEST_TYPES,
                                       find_default_qemu_paths)
+from virttest._wrappers import import_module
 
 if hasattr(plugin_interfaces, 'Init'):
     class VtInit(plugin_interfaces.Init):
@@ -273,6 +272,6 @@ if hasattr(plugin_interfaces, 'Init'):
 
             settings.merge_with_configs()
 
-            virt_loader = getattr(importlib.import_module('avocado_vt.loader'),
+            virt_loader = getattr(import_module('avocado_vt.loader'),
                                   'VirtTestLoader')
             loader.register_plugin(virt_loader)
--- a/avocado_vt/plugins/vt_list.py
+++ b/avocado_vt/plugins/vt_list.py
@@ -25,6 +25,7 @@ from avocado.core.plugin_interfaces import CLI
 from virttest.compat import get_settings_value, add_option
 from .vt import add_basic_vt_options, add_qemu_bin_vt_option
 from ..loader import VirtTestLoader
+from virttest._wrappers import load_source
 
 
 # The original virt-test runner supports using autotest from a git checkout,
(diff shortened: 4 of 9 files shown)


bgartzi:
@luckyh Could you please help me reviewing this PR and selecting another reviewer?
Thank you!

PD: sorry for the title error and several check errors :(

virttest/remote_commander/remote_runner.py
```diff
@@ -746,7 +745,9 @@ def python_file_run_with_helper(self, test_path):
         :return: module.run().return
         """
         assert os.path.isfile(test_path)
-        module = imp.load_source('ext_test', test_path)
+        sys.path.insert(0, test_path)
+        module = importlib.import_module('ext_test')
+        sys.path.pop(0)
```
luckyh, line 750: For the replacement of `imp.load_source`, looks at least we have to take care about these points:
```python
$ cat /tmp/t.py
def hi():
    print('hi')

# a normal importing case
$ python
Python 3.10.4 (main, Mar 25 2022, 00:00:00) [GCC 11.2.1 20220127 (Red Hat 11.2.1-9)] on linux
>>> import sys
>>> import imp
<stdin>:1: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
>>> t = imp.load_source('t', '/tmp/t.py')
>>> sys.modules['t']
<module 't' from '/tmp/t.py'>
>>> t.hi()
hi
```

1. `imp.load_source` requires the 2nd argument to be the file path of the given python file, which means for our new approach there is some processing work had to be done before passing it to `sys.path.insert`.
```python
>>> imp.load_source('t', '/tmp')
  File "/usr/lib64/python3.10/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 719, in _load
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 879, in exec_module
  File "<frozen importlib._bootstrap_external>", line 1016, in get_code
  File "/usr/lib64/python3.10/imp.py", line 158, in get_data
    return super().get_data(path)
  File "<frozen importlib._bootstrap_external>", line 1073, in get_data
IsADirectoryError: [Errno 21] Is a directory: '/tmp'
```
```python
>>> import sys
>>> import importlib
>>> sys.path.insert(0, '/tmp/t.py')
>>> t = importlib.import_module('t')
  File "/usr/lib64/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 't'
>>> sys.path.pop(0)
'/tmp/t.py'
>>> sys.path.insert(0, '/tmp')
>>> t = importlib.import_module('t')
>>> t.hi()
hi
```

2. the 1st argument of `imp.load_source` which is the name of the module doesn't have to be the same name to the python file, which is kind of tricky to the new approach.
```python
>>> foo = imp.load_source('foo', '/tmp/t.py')
>>> sys.modules['foo']
<module 'foo' from '/tmp/t.py'>
>>> foo.hi()
hi
```
```python
>>> sys.path.insert(0, '/tmp')
>>> importlib.import_module('foo')
  File "/usr/lib64/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'foo'
```

3. when the python file is not existed for whatever reason, the new approach could not give the expected result in the case that there was other module having the same module name stored in `sys.path`.
```python
>>> imp.load_source('os', '/tmp/nonexistfile.py')
  File "/usr/lib64/python3.10/imp.py", line 170, in load_source
    module = _exec(spec, sys.modules[name])
  File "<frozen importlib._bootstrap>", line 619, in _exec
  File "<frozen importlib._bootstrap_external>", line 879, in exec_module
  File "<frozen importlib._bootstrap_external>", line 1016, in get_code
  File "/usr/lib64/python3.10/imp.py", line 158, in get_data
    return super().get_data(path)
  File "<frozen importlib._bootstrap_external>", line 1073, in get_data
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/nonexistfile.py'
```
```python
>>> sys.path.insert(0, '/tmp')
>>> m = importlib.import_module('os')
>>> m.__file__
'/usr/lib64/python3.10/os.py'
>>> m.hi()
AttributeError: module 'os' has no attribute 'hi'
```

luckyh: Maybe we can could some inspiration from the two places below?

[1] https://github.com/python/cpython/blob/5397b5afc1f594dc0ba3b7743351d595e637bf24/Lib/imp.py#L166-L177
[2] https://docs.python.org/3.10/library/importlib.html?highlight=importlib#importing-a-source-file-directly

bgartzi:
I introduced a new method inside the `_wrappers` module called `load_source`, which imitates the workflow of `imp.load_source`. The implementation is based on the recipe showed in [0]. I separated the test cases to test both methods accordingly, and made sure that the `_wrappers.import_module` approach gives the same results as the `imp.find_module` + `imp.load_module` workflow gave.

Then, I updated the `virttest`/`avocado_vt` code accordingly so as to make the old `imp`/`importlib` based workflows follow the just introduced ones. 

[0] https://docs.python.org/3.10/library/importlib.html?highlight=importlib#importing-a-source-file-directly

virttest/_wrappers.py
```diff
@@ -0,0 +1,47 @@
...
+        :returns: The imported module
+    """
+    if directory:
+        sys.path.insert(0, directory)
+    # Just in case the module source has been dropped after
+    # the interpreter execution had begun
+    importlib.invalidate_caches()
+    module = importlib.import_module(name)
+    if directory:
+        sys.path.pop(0)
```
luckyh, line 30: To ensure the given directory must be removed from `sys.path` after importing a module it seems the above two lines have to be put into a `finally` clause, for example:
```python
try:
    importlib.invalidate_caches()
    module = importlib.import_module(name)
finally:
    if directory:
        sys.path.pop(0)
```

virttest/_wrappers.py
```diff
@@ -0,0 +1,47 @@
...
+    return module
```
luckyh, line 31: Looks we have to make the entire routine thread-safe as well, otherwise it might result in getting wrong priority of paths while importing modules within more than one thread context.

virttest/_wrappers.py
bgartzi: Adding this now, thank you for yet another catch! :)

virttest/_wrappers.py
bgartzi: I've been thinking about this one more deeply. It's in fact needed to ensure that the path priority is preserved, so I'll be adding a lock in the file to be acquired/released when modifying the `sys.path` during the process of importing a module.

However, after thinking about this issue for a bit, I met another concern: the fact that other modules running in other threads could be also modifying and inserting elements in `sys.path[0]` as well. Those threads wouldn't be aware of our (yet hypothetical) `_wrappers.local_lock` used to modify `sys.path` safely inside the `virttest._wrappers` module. In that case, we could be removing an arbitrary value when calling `sys.path.pop(0)` instead of the one that we actually want at the end of the import_module routine.

I've been trying to find if there is any lock that can help us protect that variable, but haven't found nothing that is really accessible and used by other modules that could be accessing `sys.path`. Instead, it seems that the most reasonable way of avoiding that case, is by calling `sys.path.remove(directory)` although its time complexity won't be O(1).

What do you think?

bgartzi:
Hi @luckyh! I'm adding the changes discussed above. However, I'm still concerned about one exact case related to `sys.path`. Although the `pop` case has been covered, it seems that we could still have some issues coming from other threads inserting values into `sys.path[0]` right after the `import_module` method does it. As other modules are not aware of the lock our import method is using, it could happen that other threads (doing unrelated things) could insert a value before ours, and mix up the path search priority during the module import.

luckyh (commented):
Hi @bgartzi, thank you for the update! I merely have the two comments below to the latest version, thanks!

luckyh (approved):
LGTM

luckyh:
@YongxueHong @dzhengfy @chunfuwen @kylazhang could you please help review this PR? thanks!

YongxueHong (approved):

virttest/_wrappers.py
```diff
@@ -0,0 +1,73 @@
...
+                          and the PYTHONPATH.
+        :type path: String, list of strings or  None
+        :returns: The imported module
+    """
+    if path is None:
+        if name in sys.builtin_module_names:
+            spec = importlib.machinery.BuiltinImporter.find_spec(name, path)
+            return _load_from_spec(spec)
+        else:
+            path = sys.path
```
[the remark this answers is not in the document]
bgartzi: It makes it look way cleaner, thank you!

bgartzi:
After a long time without having put attention on the topic, I'm adding two commits to this pull request to fix the error raised by the pylint check.

pylint was raising an E1102 (not-callable) error on `virttest/utils_test/__init__.py:L1820`. In that scope, the `run_func` variable was  causing the error. That variable comes form the `utils_misc.get_test_entrypoint_func` method, which takes the module that has been imported by the import_module wrapper introduced in the previous commit and returns its run attribute.

The first commit adds another check to `utils_misc.get_test_entrypoint_func` to ensure the returned object is in fact callable and that is not pointing to any other kind of variable.

The second commit adds an inline comment to disable the E1102 error that pylint was still raising on that line.

luckyh:
> With another comment to disable the fake positive raised by pylint now it seems the checks have passed. However, I'm starting to feel like [utils_misc: assert run is callable](https://github.com/avocado-framework/avocado-vt/pull/3350/commits/a02cbd029a8661bbc0dc1caa83339439df929923) would better fit a separated pull request since it doesn't really match the scope of this one.
> 
> What do you think?

Make sense, let's create a separated PR, thanks!

bgartzi:
Got the checks to assert that the object returned by `utils_misc.get_test_entrypoint_func` was actually callable split into PR #3539.

The comments to ignore the not-callable errors raised by pylint introduced in e20e6611609271b8f439defde511ff41a023a6b0 are not safe until the previous #3539 is merged.

PR #3358, opened 2022-02-21
utils_net: additional checks for arp table parsing

bgartzi:
Adding additional check when parsing the arp table (reading /proc/net/arp).
Previously the method was not checking whether the read line was empty/parseable or not.

ID: 2053188

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3315,9 +3315,15 @@ def parse_arp(session=None, timeout=60.0, **dargs):
             arp_cache = arp_file.read()
 
     for line in arp_cache.splitlines():
-        mac = line.split()[3]
-        ip = line.split()[0]
-        flag = line.split()[2]
+        parsed_elements = line.split()
+        try:
+            mac = parsed_elements[3]
+            ip = parsed_elements[0]
+            flag = parsed_elements[2]
+        except IndexError:
+            # Check if data is missing
+            raise OSError("Read invalid data from %s host arp file" %
+                          "remote" if session else "local")
 
         # Skip the header
         if mac.count(":") != 5:
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3306,13 +3306,15 @@ def parse_arp(session=None, timeout=60.0, **dargs):
     :return: dict mapping MAC to IP
     """
     ret = {}
-    func = process.getoutput
+    arp_file_path = "/proc/net/arp"
     if session:
-        func = session.cmd_output
-    arp_cache = func('cat /proc/net/arp', timeout=timeout,
-                     **dargs).strip().split('\n')
+        arp_cache = session.cmd_output("cat %s" % arp_file_path,
+                                       timeout=timeout, **dargs)
+    else:
+        with open(arp_file_path, 'r') as arp_file:
+            arp_cache = arp_file.read()
 
-    for line in arp_cache:
+    for line in arp_cache.splitlines():
         mac = line.split()[3]
         ip = line.split()[0]
         flag = line.split()[2]


bgartzi:
Hi @luckyh could you please help me review this PR? Thank you!

virttest/utils_net.py
```diff
@@ -3313,9 +3313,14 @@ def parse_arp(session=None, timeout=60.0, **dargs):
...
-        ip = line.split()[0]
-        flag = line.split()[2]
+        parsed_elements = line.split()
+        try:
+            mac = parsed_elements[3]
+            ip = parsed_elements[0]
+            flag = parsed_elements[2]
+        except IndexError:
+            # Skip possible empty lines
+            continue
```
luckyh, line 3323: Normally, it looks not possible to get empty lines here unless some bad thing is preventing us for reading the entry file properly, therefore I'd suggest raising an exact error (`RuntimeError` or `OSError`?) instead, what do you think?

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv4/arp.c?id=038101e6b2cd5c55f888f85db42ea2ad3aecb4b6#n1385

bgartzi: Yes, you're right. This way we'd end up shadowing the root cause, and raising another exception which wouldn't be directly related to the real issue. I think that a `VMAddressVerificationError` would be the one being raised at the end, and effectively, it doesn't give much information about the reason behind the failure.


I was having a look on different `OSError` and `RuntimeErrors` and I'm not sure if I could use them to give the needed proper information about the issue. Maybe we should be implementing another exception on `virttest.utils_net`? Something that would go like:
``` python
ArpParseError(NetError):

    def __init__(self, remote=False):
        """ :param remote: Indicates which was the arp table
                           that was being parsed
                           (the guest's or the host's)
            :type remote: bool
        """
        self.remote = remote

    def __str__(self):
        return ("Error when parsing %s arp table. It seemed to be empty. "
                "An OS issue should be the reason behind this." % \
                ("remote" if self.remote else "local host"))
``` 
However, I don't think it would give us too much further information about the root cause of the failure, but that it would stop the execution and explain why it couldn't keep running. What do you think?

luckyh: > However, I don't think it would give us too much further information about the root cause of the failure, but that it would stop the execution and explain why it couldn't keep running. What do you think?

I just checked the commit history related to this function and found that at the first time the function was introduced without using `process.getoutput` as its command runner (https://github.com/avocado-framework/avocado-vt/commit/eb004798d6850137d4ebbd9e9cf4f43f177158d9), it just used Python's builtin file IO APIs to access the proc entry directly. At that stage, it would indicate to some critical OS failure which could not be recovered if we could get an `IndexError` from the above lines, and thus I think an `OSError` would be something we want to tell to the end users.

Afterward, https://github.com/avocado-framework/avocado-vt/commit/e1d50763511aa9848c6714adf66546102b6349bb changed the manner for the extension of some remote execution proposal. However, it brought the overhead and vulnerabilities from involving an extra child process (by invoking `process.getoutput`), which also makes us feel hard to give the end users further information about the root cause.

So, first of all I think maybe we can try to replace `process.getoutput` back to Python's builtin file IO APIs, to reduce the overhead. And then we can catch the `IndexError`s here and throw out `OSError`s for pointing that some really bad thing happened and nothing was received from the output but garbage.

Please let me know your opinion, thanks!

bgartzi: > At that stage, it would indicate to some critical OS failure which could not be recovered (...) I think an OSError would be something we want to tell to the end users.

I really see your point now. I think that what you said on
> ...involving an extra child process, which also **makes us feel hard to give the end users further information about the root cause**.

is key.

I'm using the python builtin IO API for reading the files in a local host, and I'll use the session for the case in which remote execution is needed.

 Thank you for the review!

virttest/utils_net.py
```diff
@@ -3315,9 +3315,15 @@ def parse_arp(session=None, timeout=60.0, **dargs):
...
-        ip = line.split()[0]
-        flag = line.split()[2]
+        parsed_elements = line.split()
+        try:
+            mac = parsed_elements[3]
+            ip = parsed_elements[0]
+            flag = parsed_elements[2]
+        except IndexError:
+            # Skip possible empty lines
+            raise OSError("arp file in %s host is empty" %
```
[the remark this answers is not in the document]
bgartzi: Yes, you're right. Give me a minute so I tweak it.
Also, I'm deleting the comment since it doesn't make sense.

luckyh (approved):
LGTM, thanks!

luckyh:
@PaulYuuu @yanglei-rh would you mind to help review this PR as well? thanks!

virttest/utils_net.py
```diff
@@ -3306,16 +3306,24 @@ def parse_arp(session=None, timeout=60.0, **dargs):
...
+                                       timeout=timeout, **dargs)
+    else:
+        with open(arp_file_path, 'r') as arp_file:
+            arp_cache = arp_file.read()
 
-    for line in arp_cache:
-        mac = line.split()[3]
-        ip = line.split()[0]
-        flag = line.split()[2]
+    for line in arp_cache.strip().split('\n'):
```
PaulYuuu, line 3317: `splitlines()` would be better?

bgartzi: Definitely!

yanglei-rh:
LGTM

PaulYuuu (approved):

luckyh:
Thanks to all.

bgartzi:
Thank you all for the review!

PR #3360, opened 2022-02-28
compressed qemu-nbd: dependencies development for testcase

bgartzi:
This pull request is trying to merge 3 commits that implement the dependencies that a test case had on virttest (see bz-2047527).
The first two commits add the code needed to support image-opt generation for storage filters. More specifically, the first commit adds the needed backbone for image-opt generation for filters, and the second commit adds the needed support for the compress filter. Support for more filters can be easily added by adding a new private function named `_<filter_type>` into `_FilterConfigGatherer` class which reads params and returns an `OrderedDict` containing the key/value pairs that are meaningful to that filter.
The last commit adds the needed changes so the `nbd` module uses the image opts when a filter is used when exporting an image.

I'm not adding the bz ID info since I'm adding it in the provider pull request, since the bz is related to the test case development and not the development of its dependencies.

ID: 2047527

### DIFF
--- a/virttest/nbd.py
+++ b/virttest/nbd.py
@@ -51,6 +51,8 @@ def export_image(qemu_nbd, filename, local_image, params):
         nbd_port: TCP port to listen on as a server (-p)
         nbd_server_tls_creds: path to TLS credentials
         nbd_export_bitmaps: bitmap names separated by space, e.g. 'bm1 bm2'
+        nbd_export_image_opts: image-opts string that will be used to describe
+                               the exported image instead of using filename
     :return: pid of qemu-nbd server or None
     """
     cmd_dict = {
@@ -67,12 +69,14 @@ def export_image(qemu_nbd, filename, local_image, params):
         "pid_file": "",
         "bitmap": "",
         "allocation_depth": "",
-        "export_readonly": ""
+        "export_readonly": "",
+        "image_opts": ""
     }
     export_cmd = (
         '{secret_object} {tls_creds} {allocation_depth} {export_readonly} '
         '{export_format} {persistent} {desc} {port} {bitmap} '
-        '{export_name} {fork} {pid_file} {unix_socket} {filename}')
+        '{export_name} {fork} {pid_file} {unix_socket} {filename} '
+        '{image_opts}')
 
     pid_file = utils_misc.generate_tmp_file_name('%s_nbd_server' % local_image,
                                                  'pid')
@@ -130,6 +134,14 @@ def export_image(qemu_nbd, filename, local_image, params):
         cmd_dict['bitmap'] = "".join(
             [" -B %s" % _ for _ in params['nbd_export_bitmaps'].split()]
         )
+
+    # Export image with a filter
+    image_opts = params.get("nbd_export_image_opts")
+    if image_opts:
+        # Call the cmd with image-opts
+        cmd_dict["export_format"] = ""
+        cmd_dict["filename"] = ""
+        cmd_dict["image_opts"] = "--image-opts %s" % image_opts
 
     qemu_nbd_pid = None
     cmdline = qemu_nbd + ' ' + string.Formatter().format(export_cmd,
--- a/virttest/qemu_storage.py
+++ b/virttest/qemu_storage.py
@@ -382,7 +382,7 @@ class _FilterConfigGatherer(object):
             :type params: Ordered Dict
             :return: Ordered Dict
         """
-        _config_gatherers = {}
+        _config_gatherers = {"compress": cls._compress}
         filter_params = params.object_params(filter_name)
         filter_type = filter_params.get("image_filter_driver_type")
         # Calls the adequate function for us
@@ -394,6 +394,13 @@ class _FilterConfigGatherer(object):
                                       "support hasn't been implemented for "
                                       f"{filter_type} filter yet.")
         return gather_func(filter_params)
+
+    @staticmethod
+    def _compress(params):
+        """ Gives an empty OrderedDict (as compress filter doesn't have
+            specific attributes related to the filter)
+        """
+        return collections.OrderedDict({"driver": "compress"})
 
 
 class _ParameterAssembler(string.Formatter):
--- a/virttest/qemu_storage.py
+++ b/virttest/qemu_storage.py
@@ -269,6 +269,15 @@ def _get_image_meta(image, params, root_dir):
             meta['file'].update({
                 k: v[1] for k, v in six.iteritems(mapping) if v[0]
             })
+    filters = params.get("image_filter_drivers", "").split()
+    # Stack them reversed, so the first filter on the list
+    # will be on top of the rest of filters
+    for img_filter in filters[::-1]:
+        tmp_meta = collections.OrderedDict()
+        tmp_meta["file"] = meta
+        meta = tmp_meta
+        filter_cfg = _FilterConfigGatherer.gather(img_filter, params)
+        meta.update(filter_cfg)
 
     return meta
 
@@ -345,6 +354,46 @@ def get_image_repr(image, params, root_dir, representation=None):
 
         func = mapping["json"] if image_secret or access_needed else mapping["filename"]
     return func(image, params, root_dir)
+
+
+class _FilterConfigGatherer(object):
+    """ This class can be used to gather filter-specific configuration
+        from test params to generate fully-valid image_meta dicts.
+        The filter-specific image_meta should be obtained by calling
+        the gather class method.
+        This class should contain as much private methods as filters that
+        should be supported for image_meta generation. In other words,
+        each supported filter should contain a _<filter_type> implemented
+        in this class as well as a "<filter_type>": <cfg_gather_func> key/value
+        entry in the _config_gatherers dict from the gather method.
+    """
+
+    @classmethod
+    def gather(cls, filter_name, params):
+        """ Gathers filter-specific config from params and collects
+            them properly on an OrderedDict.
+            This method is in charge of calling the appropriate private
+            class method
+            :param filter_name: Name/id of the filter for which data is being
+                                gathered
+            :type filter_name: str
+            :param params: Collection of parameters from which data will be
+                           gathered
+            :type params: Ordered Dict
+            :return: Ordered Dict
+        """
+        _config_gatherers = {}
+        filter_params = params.object_params(filter_name)
+        filter_type = filter_params.get("image_filter_driver_type")
+        # Calls the adequate function for us
+        try:
+            gather_func = _config_gatherers[filter_type]
+        except KeyError:
+            # Filter is not supported
+            raise NotImplementedError("Image configuration meta generation "
+                                      "support hasn't been implemented for "
+                                      f"{filter_type} filter yet.")
+        return gather_func(filter_params)
 
 
 class _ParameterAssembler(string.Formatter):
--- a/virttest/shared/cfg/base.cfg
+++ b/virttest/shared/cfg/base.cfg
@@ -968,3 +968,11 @@ nvdimm_uuid = ""
 
 # Set granularity for a bitmap(Only works when adding a new bitmap to an image)
 # bitmap_granularity = "131072"
+
+# Set below params to use the compress filter by imageopts on a image.
+# image_filter_drivers: to set a list of image_filter_driver ids/names
+#                       that will be stacked in left-to-right order.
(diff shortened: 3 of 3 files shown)


bgartzi:
@luckyh I'd appreaciate if you could help me reviewing this. Maybe your point of view, @zhencliu, could also be helpful on this. Thank you!

virttest/nbd.py
```diff
@@ -131,6 +135,14 @@ def export_image(qemu_nbd, filename, local_image, params):
             [" -B %s" % _ for _ in params['nbd_export_bitmaps'].split()]
         )
 
+    # Export image with a filter
+    image_opts = params.get("nbd_export_image_opts")
+    if image_opts:
+        # Call the cmd with image-opts
+        cmd_dict["image_format"] = ""
```
zhencliu, line 142: image_format -> export_format

virttest/qemu_storage.py
```diff
@@ -269,6 +269,16 @@ def _get_image_meta(image, params, root_dir):
             meta['file'].update({
                 k: v[1] for k, v in six.iteritems(mapping) if v[0]
             })
+    filters = params.get("image_filters", "").split()
```
zhencliu, line 272: Confirmed with feature owners, currently, we don't support multiple filters for a single node, but I think it's OK to use a 'stack' here in case we support it in future, @YongxueHong , would you pls make some comments?

virttest/qemu_storage.py
```diff
@@ -269,6 +269,16 @@ def _get_image_meta(image, params, root_dir):
...
+    # Stack them reversed, so the first filter on the list
+    # will be on top of the rest of filters
+    for img_filter in filters[::-1]:
+        tmp_meta = collections.OrderedDict()
+        tmp_meta["file"] = meta
+        meta = tmp_meta
+        meta["driver"] = img_filter
```
zhencliu, line 279: remove this line, let filter function itself set its filter name. e.g. copy-on-read filter, we cannot define a _copy-on-read() function, we can assign image filter name as 'cor', define the function as _cor, and let _cor() set {'driver': 'copy-on-read'}

virttest/qemu_storage.py
```diff
@@ -347,6 +357,48 @@ def get_image_repr(image, params, root_dir, representation=None):
     return func(image, params, root_dir)
 
 
+class _FilterConfigGatherer(object):
+    """ This class can be used to gather filter-specific configuration
+        from test params to generate fully-valid image_meta dicts.
+        The filter-specific image_meta should be obtained by calling
+        the gather_filter_config method.
```
zhencliu, line 364: no gather_filter_config method defined :)

virttest/qemu_storage.py
```diff
@@ -347,6 +357,48 @@ def get_image_repr(image, params, root_dir, representation=None):
...
+        """
+        # Calls the adequate function for us
+        try:
+            gather_func = getattr(cls, f"_{filter_type}")
+        except AttributeError:
+            # Filter is not supported
+            raise NotImplementedError("Image configuration meta generation "
+                                      "support hasn't been implemented for "
+                                      f"{filter_type} filter yet.")
+        return gather_func(filter_type, params)
```
zhencliu, line 392: filter_type is not required here, each private function knows very well what filter_type it is

virttest/qemu_storage.py
```diff
@@ -347,6 +357,48 @@ def get_image_repr(image, params, root_dir, representation=None):
...
+
+    @staticmethod
+    def _compress(filter_type, params):
+        """ Gives an empty OrderedDict (as compress filter doesn't have
+            specific attributes related to the filter)
+        """
+        return collections.OrderedDict()
```
zhencliu, line 399: return collections.OrderedDict({'driver': 'compress'})

virttest/qemu_storage.py
zhencliu: Also, please introduce the new params in virttest/shared/cfg/base.cfg

virttest/qemu_storage.py
bgartzi: True!

virttest/qemu_storage.py
bgartzi: This makes it much more clean and flexible, thanks!

bgartzi:
Thanks @zhencliu for the review. I already applied the suggestions, now the PR looks better.

hellohellenmao:
@luckyh @YongxueHong Could anyone of you please help to take a review here? Thanks.

virttest/qemu_storage.py
```diff
@@ -375,11 +375,12 @@ def gather(cls, filter_type, params):
...
-            :param params: collections.OrderedDict
-            :type params: collections.OrderedDict
-            :return: collections.OrderedDict
+            :param params: Collection of parameters from which data will be
+                           gathered
+            :type params: Ordered Dict
+            :return: Ordered Dict
         """
-        # Calls the adecuate function for us
+        # Calls the adequate function for us
```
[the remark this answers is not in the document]
bgartzi: Thank you for this catch! I'm fixing it now.

virttest/qemu_storage.py
bgartzi: AFAIK, filter nodes are the first layer on the block device backend node chain. They can be used on top of the format nodes (raw, qcow2...) which are used on top of the protocol nodes (file, iscsi...). The filter nodes allow to apply different data manipulation algorithms to the data that is being read/written by the guest from/to the block device that such I/O operation is targeting.

For example, using a throttling filter over a given block device allows limiting the I/O operation that such device will be exposed to. Other example is the compress filter, which allows compressing the data that is being written onto the block device while the I/O operation is being carried out.

@zhencliu, I'd appreciate it if you could correct me if I'm wrong and add any other thing I haven't mentioned.

YongxueHong: @bgartzi Thanks for your detailed explanations.

hellohellenmao:
@luckyh Could you please help to take a review here? Thanks.

YongxueHong (approved):
LGTM.

virttest/qemu_storage.py
```diff
@@ -269,6 +269,15 @@ def _get_image_meta(image, params, root_dir):
             meta['file'].update({
                 k: v[1] for k, v in six.iteritems(mapping) if v[0]
             })
+    filters = params.get("image_filters", "").split()
```
luckyh, line 272: What about using the name `image_filter_drivers`?

And also, I'm not sure if we can see this block driver [1] as a *filter*, so, to me an alternative name could be `image_block_drivers`.

@bgartzi @zhencliu @YongxueHong what do you think about it?

[1] https://github.com/qemu/qemu/commit/1c14eaabdb7

bgartzi: Thank you for pointing this out. `image_filter_drivers` seem more appropriate to me.

It seems that the block driver you mentioned would also fit in the same level of the block driver stack as filter drivers would. However, calling the config parameter `image_block_drivers` could feel a bit confusing to me, as it could be also referring to drivers from other layers of the block device backend, as protocols or formats. As we want to describe the stack of drivers located in the _filter level_ of the block device backend nodes with this parameter, I think that calling the parameter `image_filter_drivers` would be more appropriate than `image_block_drivers`.

What do you think? @luckyh @zhencliu @YongxueHong

YongxueHong: Hi @bgartzi 
Yes, your explanations make sense. The `image_filter_drivers` is okay for me.
Thanks.

zhencliu: It's OK from my side.

The only concern is:
take copy-on-read flter as an example, do we want to use
image_filter_drivers = copy-on-read
if so, we may need a mapping in gather function to map copy-on-read to the right handler(function)

luckyh: > It seems that the block driver you mentioned would also fit in the same level of the block driver stack as filter drivers would. However, calling the config parameter `image_block_drivers` could feel a bit confusing to me, as it could be also referring to drivers from other layers of the block device backend, as protocols or formats. As we want to describe the stack of drivers located in the _filter level_ of the block device backend nodes with this parameter, I think that calling the parameter `image_filter_drivers` would be more appropriate than `image_block_drivers`.

Yeah, that makes sense to me. Actually, I was wondering if there would the case of having two filter instances with the same filter type to a image, e.g.
```
driver=foo,arg1=abc,file.driver=foo,file.arg1=xyz,...
```
and if that would be true, then we have to consider to use object expression for filters, and it could be:
```
image_filters = foo bar
image_filter_driver_foo = foo
image_filter_foo_arg1_foo = abc
image_filter_driver_bar = foo
image_filter_foo_arg1_bar = xyz
```

bgartzi: Yes, you are all right, thank you for pointing these issues out. I'm applying them into the PR.
To sum up, we are
1. supporting filters named as `copy-on-read` or similar (names that can't be represented by python method names) by switching from the `getattr` method to a map/dictionary in  https://github.com/avocado-framework/avocado-vt/blob/4f7d6974a9ef586d6ae24b44eccdb5e13793d361/virttest/qemu_storage.py#L384-L390
2. considering object expressions for image filters. We can't stack a `compress` filter over other `compress`filter for the same image. However, as a proof of concept, `qemu-img` lets us stack a `copy-on-read` filter over other one, which means we might support the scenario you mentioned @luckyh for other filter types. 
``` {bash}
$ qemu-nbd --image-opts driver=compress,file.driver=compress,file.file.driver=qcow2,file.file.file.driver=file,file.file.file.filename=foo.qcow2 -t
qemu-nbd: Failed to blk_new_open 'driver=compress,file.driver=compress,file.file.driver=qcow2,file.file.file.driver=file,file.file.file.filename=foo.qcow2': Compression is not supported for underlying format: compress
```
fails, but
``` bash
$ qemu-nbd --image-opts driver=copy-on-read,file.driver=copy-on-read,file.file.driver=qcow2,file.file.file.driver=file,file.file.file.filename=foo.qcow2 -t
```
Seems to work fine. So I'm adding these changes too.

3. I'm also changing `image_filters` for `image_filter_drivers`, so the scenario you mentioned would now look like
```
image_filter_drivers = foo bar
image_filter_driver_type_foo = foo
image_filter_driver_foo_arg1_foo = abc
image_filter_driver_type_bar = foo
image_filter_driver_foo_arg1_bar = xyz
```
What do you think?

luckyh: >     3. I'm also changing `image_filters` for `image_filter_drivers`, so the scenario you mentioned would now look like
> 
> 
> ```
> image_filter_drivers = foo bar
> image_filter_driver_type_foo = foo
> image_filter_driver_foo_arg1_foo = abc
> image_filter_driver_type_bar = foo
> image_filter_driver_foo_arg1_bar = xyz
> ```
> 
> What do you think?

@bgartzi looks good to me, thanks!

bgartzi:
I added the changes that we talked about. I also added some comments in the `virttest/shared/cfg/base.cfg` file, which seems I forgot to add in previous changes.

The main changes from the previous PR version to this one is that the `_FilterConfigGatherer` is using a dictionary as a map between keywords and actual methods. I put that dictionary into the `gather` method. Do you think that this placement looks good? Maybe creating the dictionary instance inside an `__init__` method would be more appropriated? Of course, in that case the gather method couldn't longer be a `classmethod`.

I'm updating https://github.com/autotest/tp-qemu/pull/3234 accordingly, and trying to have a look on the CI failed checks.

luckyh (approved):
LGTM

zhencliu:
LGTM

virttest/qemu_storage.py
```diff
@@ -347,6 +356,53 @@ def get_image_repr(image, params, root_dir, representation=None):
...
+
+    @classmethod
+    def gather(cls, filter_name, params):
+        """ Gathers filter-specific config from params and collects
+            them properly on an OrderedDict.
+            This method is in charge of calling the appropriate private
+            class method
+            :param filter_name: Name/id of the filter for which data is being
+                                gathered
+            :type filter_type: str
```
YongxueHong, line 379:
```suggestion
            :type filter_name: str
```

YongxueHong:
Suggest adding the Bug ID(`ID: 2047527`) at the end of the comment 0 since our automation tool will scan the related patches of the bug then comment the commit message in the bugupdate the status of bug to `VERIFIED`.
Thanks.

YongxueHong:
LGTM.

YongxueHong:
Thanks all, let's merge it.

PR #3363, opened 2022-03-02
Add copy_on_read attribute support into disk creation

chunfuwen:
when call libvirt.create_disk_xml(params)


chunfuwen:
@dzhengfy  please take a look

dzhengfy (approved):
The codes LGTM.
But I am thinking the new attributes might be added in the future and we have to modify this function time by time. We may use setup_attrs() in new cases which should be easier.

chunfuwen:
So then we need think more about  why setup_attrs() is better, in which case this is better, and what is specific benefits that setup_attrs() can bring

PR #3367, opened 2022-03-09
Remove md5sum value for windows iso

peixiu:
Remove the default md5sum value as it will change frequently.

id: 2056338


peixiu:
@leidwang @vivianQizhu Could you help to review this patch?

Thanks~
Peixiu

vivianQizhu:
@peixiu Maybe we need to update the commit log, to explain why we don't want to keep md5sum check here. That's not really for  test mode, as we can still keep it but we will need to change it back and forth in future. So we would like to remove the checking as we don't see the necessary for it now.

leidwang:
> @peixiu Maybe we need to update the commit log, to explain why we don't want to keep md5sum check here. That's not really for test mode, as we can still keep it but we will need to change it back and forth in future. So we would like to remove the checking as we don't see the necessary for it now.

Yes,I agree with Qianqian.
Btw,is it possible to keep the md5sum check here? Because we can easily get the md5 value.

leidwang:
LGTM.

Acked-by: leidwang@redhat.com

vivianQizhu (approved):

vivianQizhu:
Thanks @peixiu @leidwang

PR #3370, opened 2022-03-15
implement new function create_secret in libvirt_secret.py

nanli1:
write new function create_secret in libvirt_secret.py
     this func is implemented by translating keys in params to new ones,we'd better directly pass a dict with secret settingsSo we use setup_attr to rewrite function

### DIFF
--- a/virttest/utils_libvirt/libvirt_secret.py
+++ b/virttest/utils_libvirt/libvirt_secret.py
@@ -3,10 +3,64 @@ Secret related utility functions
 """
 
 import logging
+import re
+
+from aexpect import remote
+
+from avocado.core import exceptions
 
 from virttest import virsh
+from virttest.utils_test import libvirt
+from virttest.libvirt_xml import secret_xml
 
 LOG = logging.getLogger('avocado.' + __name__)
+
+
+def create_secret(sec_dict, remote_args=None):
+    """
+    Create a secret with 'virsh secret-define'
+
+    :param sec_dict: dict to create secret
+    :param remote_args: Parameters for remote host
+    :return: UUID of the secret
+
+    Example:
+
+        secret_dict = {'secret_ephemeral': 'no',
+                    'secret_private': 'yes',
+                    'description': 'secret desc',
+                    'usage': 'volume',
+                    'volume': '/path/to/volume'}
+        sec_uuid = libvirt_secret.create_secret(sec_dict=secret_dict)
+    """
+
+    sec_xml = secret_xml.SecretXML()
+    sec_xml.setup_attrs(**sec_dict)
+    sec_xml.xmltreefile.write()
+
+    # define the secret and get its uuid
+    if remote_args:
+        server_ip = remote_args.get("remote_ip", "")
+        server_user = remote_args.get("remote_user", "")
+        server_pwd = remote_args.get("remote_pwd", "")
+        if not all([server_ip, server_user, server_pwd]):
+            raise exceptions.TestError("remote_[ip|user|pwd] are necessary!")
+        remote_virsh_session = virsh.VirshPersistent(**remote_args)
+        remote.scp_to_remote(server_ip, '22', server_user, server_pwd,
+                             sec_xml.xml, sec_xml.xml, limit="",
+                             log_filename=None, timeout=600, interface=None)
+        ret = remote_virsh_session.secret_define(sec_xml.xml)
+        remote_virsh_session.close_session()
+    else:
+        ret = virsh.secret_define(sec_xml.xml)
+        libvirt.check_exit_status(ret)
+    try:
+        sec_uuid = re.findall(r".+\S+(\ +\S+)\ +.+\S+",
+                              ret.stdout_text)[0].lstrip()
+    except IndexError:
+        raise exceptions.TestError("Fail to get newly created secret uuid")
+
+    return sec_uuid
 
 
 def get_secret_list(remote_virsh=None):


chunfuwen (approved):
LGTM

virttest/utils_libvirt/libvirt_secret.py
```diff
@@ -3,12 +3,86 @@
...
 
+def create_secret(params, remote_args=None):
+    """
+    Create a secret with 'virsh secret-define'
+
+    :param params: Test run params
+    :param remote_args: Parameters for remote host
+    :return: UUID of the secret
+    """
+    sec_usage_type = params.get("sec_usage", "volume")
```
Yingshun, line 27: Rather than moving the place, I would suggest using setup_attrs() to update xml parts. setup_attrs() is our recommend way to setup xml setting. Please check "Libvirt automation tips in upstream" -> "Recommendation for library usage or specific checkpoints in avocado-vt" -> item 15.

chunfuwen: That is what is doing in the line 42 
```
 sec_xml.setup_attrs(**sec_dict)
```

Yingshun: Thanks for your reminder ! But I think we'd better directly pass a dict with secret settings, instead of translating keys in params to new ones.

virttest/utils_test/libvirt.py
```diff
@@ -3461,75 +3460,6 @@ def virsh_cmd_has_option(cmd, option, raise_skip=True):
         return found
 
 
-def create_secret(params, remote_args=None):
```
Yingshun, line 3464: @nanli1 If this change will cost too much effort, I'd like to suggest taking  @dzhengfy 's proposal - Add deprecated log in libvirt.create_secret(). For details, please check https://github.com/autotest/tp-libvirt/pull/4081#discussion_r823636255

virttest/utils_libvirt/libvirt_secret.py
chunfuwen: I suggest we can move this PR forward in this way.
1)For current PR, we just focus on move this method into different file
2)Create separate one PR to create new method create_secret_xx() using so called pass a dict with secret settings, let user choose which one they want to use.

virttest/utils_test/libvirt.py
nanli1: Add deprecated log to create_secret func and not move function

virttest/utils_libvirt/libvirt_secret.py
nanli1: Add deprecated log to create_secret func and not move function

virttest/utils_libvirtd.py
```diff
@@ -441,6 +443,17 @@ def deprecation_warning():
                 "libvirtd service.")
 
 
+def deprecation_function_waring(func):
```
Yingshun, line 446: utils_libvirtd is a module to control libvirtd service. I don't think this decorator should be here, what do you think?

nanli1: OK, where should we put it?

Yingshun: Hmm, it must be a lower-level module if you want to make it to be generic. Otherwise, we can leave it to libvirt.py I think. 
@dzhengfy do you have any suggestions?

BTW, @nanli1 I think we'd better provide a replacement of the deprecated function. 
Please have a look at 
https://github.com/avocado-framework/avocado-vt/blob/6e3d2eb2d8390d3d0cd8d20c94f7e28940fb6e46/virttest/utils_hotplug.py#L59
https://github.com/avocado-framework/avocado-vt/blob/6e3d2eb2d8390d3d0cd8d20c94f7e28940fb6e46/virttest/utils_test/libvirt.py#L2405-L2406

virttest/utils_test/libvirt.py
chunfuwen: @dzhengfy 
@Yingshun 
Could you elaborate more details why deprecate this method?

virttest/utils_libvirtd.py
nanli1: Added commend
```

 This function is deprecated now.
    The recommended usage is below:
    Define a parameter in cfg like
        sec_dict = {"secret_ephemeral": "yes", "secret_private": "yes",
        "description": "desc", "usage": "volume"}
    Set up the device in python like
        sec_dict = eval(params.get('sec_dict', '{}'))
        sec_xml = secret_xml.SecretXML("no", "yes")
        sec_xml.setup_attrs(**sec_dict)

```

virttest/utils_test/libvirt.py
```diff
@@ -3465,10 +3465,23 @@ def create_secret(params, remote_args=None):
...
+    The recommended usage is below:
+    Define a parameter in cfg like
+        sec_dict = {"secret_ephemeral": "yes", "secret_private": "yes",
+        "description": "desc", "usage": "volume"}
+    Set up the device in python like
+        sec_dict = eval(params.get('sec_dict', '{}'))
+        sec_xml = secret_xml.SecretXML()
+        sec_xml.setup_attrs(**sec_dict)
+        sec_xml.xmltreefile.write()
+
```
dzhengfy, line 3478: There is 'remote_args' parameter in this function. How do you handle this?

chunfuwen (changes requested):
Please elaborated more details why deprecated this method?

nanli1:
@dzhengfy @Yingshun Please help re-review ,I write a new function

@chunfuwen 
I add deprecated log for def create_secret and rewrite new function in libvirt_secret.py
because this func is implemented by translating keys in params to new ones, we'd better directly pass a dict with secret settings, So we use setup_attr to rewrite function

virttest/utils_libvirt/libvirt_secret.py
```diff
@@ -3,12 +3,63 @@
...
+from avocado.core import exceptions
 
 from virttest import virsh
+from virttest.utils_test import libvirt
+from virttest.libvirt_xml import secret_xml
 
 LOG = logging.getLogger('avocado.' + __name__)
 
 
+def create_secret(params, remote_args=None):
```
Yingshun, line 19: Can we directly pass sec_dict?  I think it'll be simpler. What do you think?

dzhengfy: She has already used sec_dict.

Yingshun: params is something like {'sec_dict': {'sec_key1':'value1', 'sec_key2':'value2'}, 'other_key': 'other_value',...} now.
What I mean is that we may pass {'sec_key1':'value1', 'sec_key2':'value2'} directly.

virttest/utils_libvirt/libvirt_secret.py
```diff
@@ -3,12 +3,63 @@
...
+
+    # define the secret and get its uuid
+    if remote_args:
+        server_ip = remote_args.get("remote_ip", "")
+        server_user = remote_args.get("remote_user", "")
+        server_pwd = remote_args.get("remote_pwd", "")
+        if not all([server_ip, server_user, server_pwd]):
+            raise exceptions.TestError("remote_[ip|user|pwd] are necessary!")
+        remote_virsh_session = virsh.VirshPersistent(**remote_args)
+        remote.scp_to_remote(server_ip, '22', server_user, server_pwd,
```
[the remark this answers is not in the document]
Yingshun: XML file will be created in /tmp folder, so I think it'll be ok.
```
>>> from virttest.libvirt_xml import secret_xml
>>> sec_xml = secret_xml.SecretXML()
>>> sec_xml.xml
'/tmp/xml_utils_temp_vetclby1.xml'
>>> 
```

virttest/utils_libvirt/libvirt_secret.py
Yingshun: Oops, just found that it has been fixed in the latest version.

nanli1:
@Yingshun @dzhengfy Please help re-review

virttest/utils_libvirt/libvirt_secret.py
nanli1: Added sec_dict in  create_secret function as params

virttest/utils_libvirt/libvirt_secret.py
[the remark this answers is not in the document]
nanli1: Not update here

virttest/utils_libvirt/libvirt_secret.py
```diff
@@ -3,12 +3,64 @@
...
+def create_secret(sec_dict=None, remote_args=None):
+    """
+    Create a secret with 'virsh secret-define'
+
+    :param sec_dict: dict to create secret
+    :param remote_args: Parameters for remote host
+    :return: UUID of the secret
+    """
+
+    # Get dic from cfg file and give default dict value
```
Yingshun, line 28: Sorry, I just found it. Is this comment necessary? If no, please delete it.  The sec_dict is a parameter of this function, people should not care about where it comes from, right?

Yingshun (commented):
Others LGTM

virttest/utils_libvirt/libvirt_secret.py
nanli1: Yes ,it should be more simple ,I removed it

Yingshun (approved):
LGTM

virttest/utils_libvirt/libvirt_secret.py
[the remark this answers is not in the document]
dzhengfy: OK， that is fine

virttest/utils_libvirt/libvirt_secret.py
```diff
@@ -3,12 +3,62 @@
...
+from avocado.core import exceptions
 
 from virttest import virsh
+from virttest.utils_test import libvirt
+from virttest.libvirt_xml import secret_xml
 
 LOG = logging.getLogger('avocado.' + __name__)
 
 
+def create_secret(sec_dict=None, remote_args=None):
```
chunfuwen, line 19: Could someone elaborate more details why we pass monolithic sec_dict here? What benefits it can bring really?

nanli1: as the deafult value ,
```
    if not sec_dict:
        sec_dict = {'secret_ephemeral': 'no',
                    'secret_private': 'yes',
                    'description': 'secret desc',
                    'usage': 'volume',
                    'volume': '/path/to/volume'}
```

chunfuwen, line 19: To keep compatibility and agility, maybe below changes can be referred:
```
sec_default_dict = {'secret_ephemeral': 'no',
                    'secret_private': 'yes',
                    'description': 'secret desc',
                    'usage': 'volume',
                    'volume': '/path/to/volume'}

sec_default_dict.update(sec_dict)
secret_xml.set_attrs(sec_default_dict)
```

This can bring at least 3 positive effects:
1)sec_dict can be passed partially, not monolithic one
2)Even sec_dict is {}, it can use default value to create secret
3)I assume this help keep compatibility with old libvirt.create_secret() method

nanli1: @chunfuwen Thanks for your comment . As we discussed with dzheng ,yicui. I will remove default value and give example in doc string

virttest/utils_libvirt/libvirt_secret.py
```diff
@@ -3,12 +3,69 @@
...
```
Yingshun, line 19: sec_dict should not be a default value parameter, please delete '=None'.

virttest/utils_libvirt/libvirt_secret.py
```diff
@@ -3,12 +3,69 @@
...
+    Example:
+
+        set dict in cfg file:
+            secret_dict = {'secret_ephemeral': 'no',
+                            'secret_private': 'yes',
+                            'description': 'secret desc',
+                            'usage': 'volume',
+                            'volume': '/path/to/volume'}
+
+        secret_dict = eval(self.params.get("secret_dict", '{}'))
```
Yingshun, line 36: Are L29 and L36 necessary? can we delete them?

virttest/utils_libvirt/libvirt_secret.py
nanli1: done

virttest/utils_libvirt/libvirt_secret.py
nanli1: done

chunfuwen (approved):
LGTM

dzhengfy (approved):
LGTM

Yingshun (approved):

PR #3378, opened 2022-03-21
libvirt_numa: handle numa nodeset utils

dzhengfy:
Some formats of numa nodeset are provided in provider cases, but
those formats require the host numa nodes continouse. The truth is
host numa nodes have big differences among available numa hosts.
The host numa nodes might be not continuous, so certain format is
not support, like 'x-y'. In this case, we have to use 'x,y' instead.

### DIFF
--- a/virttest/utils_libvirt/libvirt_numa.py
+++ b/virttest/utils_libvirt/libvirt_numa.py
@@ -7,6 +7,8 @@ http://libvirt.org/formatdomain.html
 import logging
 import ast
 
+from avocado.core import exceptions
+
 from virttest.libvirt_xml import vm_xml
 
 LOG = logging.getLogger('avocado.' + __name__)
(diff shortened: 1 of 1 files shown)


kylazhang:
@dzhengfy do you plan to also support the situation like '1,2,3' & '1-3' & '1-3,^2'  ?

virttest/utils_libvirt/libvirt_numa.py
```diff
@@ -79,3 +81,57 @@ def create_hmat_xml(vmxml, params):
     vmxml.sync()
 
     return vmxml
+
+
+def get_continuous_numa_node_index(node_list):
```
chunfuwen, line 86: I assume you want to get the index of first adjacent numa node.
it sounds like it is a little bit clearer if method name changed to get_first_continuous_numa_node_index or get_first_adjacent_numa_node_index

dzhengfy:
> @dzhengfy do you plan to also support the situation like '1,2,3' & '1-3' & '1-3,^2' ?

no plan yet as the cases' requirement is not to test those complex format so far.

virttest/utils_libvirt/libvirt_numa.py
dzhengfy: Updated with rename and move the function to internal

chunfuwen (approved):
LGTM

Yingshun (approved):

PR #3379, opened 2022-03-22
libvirt_embedded_qemu: use cancel instead error

dzhengfy:


chunfuwen (approved):
it make sense that throw test cancel instead of test.fail  if not utils_split_daemons.is_modular_daemon()
LGTM

chunfuwen:
it is minor_fix, let's merge it

PR #3382, opened 2022-03-24
Remove root logger from avocado_vt package

richtja:
In `avocado-vg` package were still some logs using root logger. But
after https://github.com/avocado-framework/avocado/issues/5049 the
avocado logs only logs with namespace `avocado*` so the root logs are
not saved in avocado log files. This commit will fix the problem.


beraldoleal (approved):
LGTM, thanks.

chunfuwen (approved):
LGTM

PR #3387, opened 2022-03-30
Get the first interface name

yalzhang:
As kernel updates, the case will get error like
"Get nic enp94s0f0np0\neth0\neth1 operstate error". In fact, only
first one "enp94s0f0np0" is what we want, update the script to
support it.


yalzhang:
@Yingshun  Please help to review, Thank you!

Yingshun (approved):

PR #3389, opened 2022-03-31
libvirt_cpu: get cpu unavailable to pin

dzhengfy:
There are some host cpus which can not be pined by guest by default.
We do not know the reason, and we have to avoid those cpus pined by guest.
This function is to get the cpu list which can not be pined on the host.


harvey0100:

Issue #3391, opened 2022-04-06
Can't load avocado-vt modules into python console because vt_init fails


smitterl:
With the latest master today (avocado 94.0) I can't load avocado-vt modules into python3 session.
I tried using avocado 92lts and 82lts but this issue reproduces.

```
$ python3 --version
Python 3.6.8
$ cd avocado
$ pip install -r requirements-dev.txt
$ pip install -e .
$ cd ../avocado-vt
$ pip install -r 
$ pip install -r requirements.txt
$ pip install -e .
$ python3
>> from virttest import utils_misc
No python imaging library installed. Screendump and Windows guest BSOD detection are disabled. In order to enable it, please install python-imaging or the equivalent for your distro.
No python imaging library installed. PPM image conversion to JPEG disabled. In order to enable it, please install python-imaging or the equivalent for your distro.

Reproduced traceback from: /home/smitterl/NotBackedUp/avocado-tmp/avocado/avocado/core/extension_manager.py:216
Traceback (most recent call last):
  File "/home/smitterl/NotBackedUp/avocado-tmp/avocado-vt/avocado_vt/plugins/vt_init.py", line 276, in initialize
    virt_loader = getattr(importlib.import_module('avocado_vt.loader'),
  File "/usr/lib64/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/smitterl/NotBackedUp/avocado-tmp/avocado-vt/avocado_vt/loader.py", line 33, in <module>
    from .test import VirtTest
  File "/home/smitterl/NotBackedUp/avocado-tmp/avocado-vt/avocado_vt/test.py", line 33, in <module>
    from virttest import utils_env
  File "/home/smitterl/NotBackedUp/avocado-tmp/avocado-vt/virttest/utils_env.py", line 19, in <module>
    from virttest import ip_sniffing
[...]

PR #3392, opened 2022-04-06
Fix avocado-vt issue:https://github.com/avocado-framework/avocado-vt/…

chunfuwen:
…issues/3391

Error running method "initialize" of plugin "vt-init": cannot import name 'log_line' when
import from virttest import utils_misc in python terminal

virttest/utils_logfile.py and virttest/utils_misc.py duplicate the same code :
def log_line(filename, line) implementation
need change virttest/ip_sniffing.py to import log_line from virttest/utils_logfile.py

### DIFF
--- a/virttest/ip_sniffing.py
+++ b/virttest/ip_sniffing.py
@@ -18,7 +18,7 @@ from avocado.utils import process
 
 import six
 
-from virttest.utils_misc import log_line
+from virttest.utils_logfile import log_line
 from virttest.utils_version import VersionInterval
 
 LOG = logging.getLogger('avocado.' + __name__)


chunfuwen:
Fix the issue:https://github.com/avocado-framework/avocado-vt/issues/3391#

smitterl (changes requested):
Unfortunately, the manual test fails still for me:

```
Reproduced traceback from: /home/smitterl/NotBackedUp/avocado-tmp/avocado/avocado/core/extension_manager.py:216
  File "/home/smitterl/NotBackedUp/avocado-tmp/avocado-vt/avocado_vt/plugins/vt_init.py", line 276, in initialize
    virt_loader = getattr(importlib.import_module('avocado_vt.loader'),
  File "/usr/lib64/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/smitterl/NotBackedUp/avocado-tmp/avocado-vt/avocado_vt/loader.py", line 33, in <module>
    from .test import VirtTest
  File "/home/smitterl/NotBackedUp/avocado-tmp/avocado-vt/avocado_vt/test.py", line 33, in <module>
    from virttest import utils_env
  File "/home/smitterl/NotBackedUp/avocado-tmp/avocado-vt/virttest/utils_env.py", line 19, in <module>
    from virttest import ip_sniffing
  File "/home/smitterl/NotBackedUp/avocado-tmp/avocado-vt/virttest/ip_sniffing.py", line 21, in <module>
    from virttest.utils_logfile import log_line
ImportError: cannot import name 'log_line'

Error running method "initialize" of plugin "vt-init": cannot import name 'log_line'
```

smitterl:
I compared both functions `utils_log.log_line` and `utils_misc.log_line` they have the same definition - I think we can just remove `utils_misc.log_line` and make sure it's not used anymore

https://github.com/avocado-framework/avocado-vt/blob/master/virttest/utils_logfile.py#L45
https://github.com/avocado-framework/avocado-vt/blob/master/virttest/utils_misc.py#L417


https://github.com/avocado-framework/avocado-vt/search?q=log_line

smitterl:
Looks like what happened was that we tried to remove dependency on `utils_misc` but instead of moving the code we duplicated it - https://github.com/avocado-framework/avocado-vt/commit/3dffc0b93e125d26dacd0e0e2afd5221babb1f3e

smitterl:
> Looks like what happened was that we tried to remove dependency on `utils_misc` but instead of moving the code we duplicated it - [3dffc0b](https://github.com/avocado-framework/avocado-vt/commit/3dffc0b93e125d26dacd0e0e2afd5221babb1f3e)

@luckyh - Some code consumed by tp-qemu seems to be consuming the `log_line` function from the `utils_misc` module. It was moved to the module `utils_log` in referenced commit above. Do you have any concerns updating all code to use `utils_log` for this and remove it from `utils_misc`?

luckyh:
> Unfortunately, the manual test fails still for me, s. run below. I do not know which magic importlib does. Would it be safer to rename one of both `log_line` instances or if they are the same remove one of them?

I just tried to modify the code as the following and seems it could work properly (no ideas about how `importlib` dealing with it.)
```patch
diff --git a/virttest/ip_sniffing.py b/virttest/ip_sniffing.py
index 5c711f21..acddf291 100644
--- a/virttest/ip_sniffing.py
+++ b/virttest/ip_sniffing.py
@@ -18,7 +18,7 @@ from avocado.utils import process
 
 import six
 
-from virttest.utils_misc import log_line
+from virttest import utils_misc
 from virttest.utils_version import VersionInterval
 
@@ -174,7 +174,7 @@ class Sniffer(object):
 
     def _output_logger_handler(self, line):
         try:
-            log_line(self._logfile, line)
+            utils_misc.log_line(self._logfile, line)
         except Exception as e:
             LOG.warn("Can't log ip sniffer output: '%s'", e)
         if self._output_handler(line):
```
and the interesting thing is that I supposed it should work as well with using the one provided by `utils_logfile` but it's not. After changing all the `utils_misc` to `utils_logfile` from the above patch, the importing stuff can work, tests can finished properly, however, no corresponding log file (``) was generated 

> Do you have any concerns updating all code to use utils_log for this and remove it from utils_misc?

@smitterl so my concern is

luckyh:
Oops, just close this by mistaken, sorry for the inconvenience.

chunfuwen:
Close it firstly, since there is no further feedbacks

PR #3396, opened 2022-04-19
Log improvement for qemu testing

luckyh:
* qcontainer: remove the warning of q35
* remove the import check of json module
* utils_net: init ovs instance only if needed
* ~~qemu_vm: remove the log of QDevice/QBus specification~~
* don't create `background-error.log` until there was any bg_err
* qemu_vm: only log thread-id of vcpus for debugging

### DIFF
--- a/avocado_vt/utils.py
+++ b/avocado_vt/utils.py
@@ -195,13 +195,15 @@ class TestUtils:
         """
         err_file_path = os.path.join(self.logdir, BG_ERR_FILE)
         bg_errors = self.background_errors.get_all()
-        error_messages = ["BACKGROUND ERROR LIST:"]
+        error_messages = []
         for index, error in enumerate(bg_errors):
             error_messages.append(
                 "- ERROR #%d -\n%s" % (index, "".join(
                     traceback.format_exception(*error)
                     )))
-        genio.write_file(err_file_path, '\n'.join(error_messages))
+        if error_messages:
+            error_messages.insert(0, "BACKGROUND ERROR LIST:")
+            genio.write_file(err_file_path, '\n'.join(error_messages))
         if bg_errors:
             msg = ["Background error"]
             msg.append("s are" if len(bg_errors) > 1 else " is")
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -3219,10 +3219,11 @@ class VM(virt_vm.BaseVM):
                 raise e
 
             LOG.debug("VM appears to be alive with PID %s", self.get_pid())
-            # Record vcpu infos in debug log
+
             is_preconfig = params.get_boolean("qemu_preconfig")
             if not is_preconfig:
-                self.get_vcpu_pids(debug=True)
+                LOG.debug("vCPUs appear to be alive with Thread-IDs %s",
+                          ", ".join([str(tid) for tid in self.vcpu_threads]))
             vhost_thread_pattern = params.get("vhost_thread_pattern",
                                               r"\w+\s+(\d+)\s.*\[vhost-%s\]")
             self.vhost_threads = self.get_vhost_threads(vhost_thread_pattern)
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -1185,23 +1185,21 @@ __ovs = None
 __bridge = Bridge()
 
 
-def __init_openvswitch(func):
+def __init_openvswitch():
     """
-    Decorator used for late init of __ovs variable.
+    Initialize the __ovs variable.
     """
-    def wrap_init(*args, **kargs):
-        global __ovs
-        if __ovs is None:
-            try:
-                __ovs = factory(openvswitch.OpenVSwitchSystem)()
-                __ovs.init_system()
-                if (not __ovs.check()):
-                    raise Exception("Check of OpenVSwitch failed.")
-            except Exception as e:
-                LOG.debug("Host does not support OpenVSwitch: %s", e)
-
-        return func(*args, **kargs)
-    return wrap_init
+    global __ovs
+    if __ovs is None:
+        try:
+            __ovs = factory(openvswitch.OpenVSwitchSystem)()
+            __ovs.init_system()
+            if (not __ovs.check()):
+                raise Exception("Check of OpenVSwitch failed.")
+        except Exception as e:
+            LOG.error("Host does not support OpenVSwitch: %s", e)
+            __ovs = None
+    return __ovs
 
 
 def setup_ovs_vhostuser(hp_num, tmpdir, br_name, port_names,
@@ -1974,7 +1972,6 @@ def check_add_dnsmasq_to_br(br_name, tmpdir):
     return None
 
 
-@__init_openvswitch
 def find_bridge_manager(br_name, ovs=None):
     """
     Finds bridge which contain interface iface_name.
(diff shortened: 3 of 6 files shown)


luckyh:
Fixed the issue `sequence item 0: expected str instance, int found`.

luckyh:
@YongxueHong @yanan-fu @PaulYuuu would you like to help review this patch? thanks!

avocado_vt/utils.py
```diff
@@ -195,12 +195,14 @@ def verify_background_errors(self):
...
-        error_messages = ["BACKGROUND ERROR LIST:"]
+        error_messages = []
         for index, error in enumerate(bg_errors):
             error_messages.append(
                 "- ERROR #%d -\n%s" % (index, "".join(
                     traceback.format_exception(*error)
                     )))
+        if error_messages:
+            error_messages.insert(0, "BACKGROUND ERROR LIST:")
         genio.write_file(err_file_path, '\n'.join(error_messages))
```
yanan-fu, line 206: I think, write under `if error_messages`  is enough to avoid idle file.

virttest/utils_net.py
```diff
@@ -1185,23 +1185,21 @@ def get_stp_status(self, brname):
...
+    if __ovs is None:
+        try:
+            __ovs = factory(openvswitch.OpenVSwitchSystem)()
+            __ovs.init_system()
+            if (not __ovs.check()):
+                raise Exception("Check of OpenVSwitch failed.")
+        except Exception as e:
+            LOG.error("Host does not support OpenVSwitch: %s", e)
+            __ovs = None
+    return __ovs
```
yanan-fu, line 1202: As `__ovs` be defined as a global parameter,   how about delete `return __ovs`, and call `__init_openvswitch` directly without return value ?

avocado_vt/utils.py
luckyh: Good catch, will update it, thanks!

virttest/utils_net.py
luckyh: Considering the situation that there would likely be further improvements to the network related modules, it is better to not use the global var directly.

yanan-fu: Don't use global var is make sense to me.   So with the new change, remove decorator, the definition can be change too i think.

YongxueHong (approved):
LGTM

yanan-fu (approved):

avocado_vt/utils.py
yanan-fu: @luckyh  Seems this part have not be updated yet ?

virttest/utils_net.py
yanan-fu: Ack with this, thanks!

PaulYuuu:
@yanglei-rh, aarch64 does not support OVS, can you help run a few tests with OVS bridge?

yanglei-rh:
> @yanglei-rh, aarch64 does not support OVS, can you help run a few tests with OVS bridge?

@PaulYuuu Ok, I will test it on x86.

virttest/qemu_vm.py
```diff
@@ -3039,8 +3039,6 @@ def create(self, name=None, params=None, root_dir=None,
             try:
                 self.devices, self.spice_options = self.make_create_command()
                 self.update_vga_global_default(params, migration_mode)
-                LOG.debug(self.devices.str_short())
-                LOG.debug(self.devices.str_bus_short())
```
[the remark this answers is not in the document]
luckyh: @PaulYuuu ok, I see, but I'd like to know the use case a bit further. So, do you mean `DevContainer.str_bus_short` is quite useful when debugging or doing some test/framework development? and is it possible to replace this by directly reading the qemu command line?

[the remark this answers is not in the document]
PaulYuuu: Yes, I suppose only developing test cases need it, but actually, I only care about the PCI(e) bridge, other buses look uncomfortable.

yanglei-rh:
> > @yanglei-rh, aarch64 does not support OVS, can you help run a few tests with OVS bridge?
> 
> @PaulYuuu Ok, I will test it on x86.

Tried to apply this patch on x86 env and test with ovs bidge,there is no errors were encountered.

virttest/qemu_vm.py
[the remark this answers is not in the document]
luckyh: I see, let's consider the use case you mentioned a bit more, so this time I'd like to discard the relevant commit from the patch set, thanks.

luckyh:
Hi @YongxueHong @yanan-fu @PaulYuuu, I've updated this PR with removing one of the commits (`qemu_vm: remove the log of QDevice/QBus specification`) because per [@PaulYuuu ' comment](https://github.com/avocado-framework/avocado-vt/pull/3396#discussion_r855739525) I think we need to consider more about the related use cases. Would you please help check this again? thanks!

YongxueHong:
> Hi @YongxueHong @yanan-fu @PaulYuuu, I've updated this PR with removing one of the commits (`qemu_vm: remove the log of QDevice/QBus specification`) because per [@PaulYuuu ' comment](https://github.com/avocado-framework/avocado-vt/pull/3396#discussion_r855739525) I think we need to consider more about the related use cases. Would you please help check this again? thanks!

It looks good to me. Thanks.

PaulYuuu (approved):
LGTM

yanan-fu:
LGTM

luckyh:
Thanks to all.

pevogam:
Ah, kudos for this PR, I had this annoying OpenVSwitch error message for years now!

pevogam:
Just to add something to this, the following two log messages are also always showing for Qemu VM users that don't use libvirtd in any way and would probably prefer not to see them all the time:
```diff
diff --git a/virttest/utils_libvirtd.py b/virttest/utils_libvirtd.py
index 0642338b..b039320c 100644
--- a/virttest/utils_libvirtd.py
+++ b/virttest/utils_libvirtd.py
@@ -57,8 +57,8 @@ class Libvirtd(object):
         self.service_list = []
 
         if LIBVIRTD is None:
-            LOG.warning("Libvirtd service is not available in host, "
-                        "utils_libvirtd module will not function normally")
+            LOG.debug("Libvirtd service is not available in host, "
+                      "utils_libvirtd module will not function normally")
 
         self.service_name = "libvirtd" if not service_name else service_name
 
* Unmerged path virttest/utils_net.py
diff --git a/virttest/virsh.py b/virttest/virsh.py
index c73c5df2..000cacad 100644
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -71,7 +71,7 @@ VIRSH_COMMAND_GROUP_CACHE_NO_DETAIL = False
 try:
     VIRSH_EXEC = path.find_command("virsh")
 except path.CmdNotFoundError:
-    logging.getLogger('avocado.app').warning(
+    logging.getLogger('avocado.app').debug(
         "Virsh executable not set or found on path, virsh module will not "
         "function normally")
     VIRSH_EXEC = '/bin/true'
     ```

luckyh:
> Just to add something to this, the following two log messages are also always showing for Qemu VM users that don't use libvirtd in any way and would probably prefer not to see them all the time

@pevogam good catch! however my thought is slightly different and that is, instead of simply demoting the log level (which could be seen as a workaround to me) I'd suggest resolving the issue by either

* dynamically importing these modules depending on needs

or

* completely refactoring the reporting of these command/service missing issues (for example, replacing them by exceptions instead)

and that's because I'd consider it as a design defect, thanks.

pevogam:
> however my thought is slightly different and that is, instead of simply demoting the log level (which could be seen as a workaround to me) I'd suggest resolving the issue by either
> 
>     * dynamically importing these modules depending on needs
> 
> 
> or
> 
>     * completely refactoring the reporting of these command/service missing issues (for example, replacing them by exceptions instead)
> 
> 
> and that's because I'd consider it as a design defect, thanks.

Hmmm, dynamically importing them would probably require moving the imports to multiple more restricted scopes on various locations and replacing the above by exceptions will break the complete run for many people unless we throw an exception if a dynamic import fails. In short I don't see how (1) or (2) are separate options and not the same option more or less. But it could be that I misunderstand your exact intent.

Indeed, demoting the logging level is a mere workaround and a quick patch on my side, nothing really diving deeper into this design flaw in logging verbosity.

PR #3398, opened 2022-04-21
Add option to force setting socket value

leidwang:

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1638,6 +1638,7 @@ class VM(virt_vm.BaseVM):
         smp = params.get_numeric('smp')
         vcpu_maxcpus = params.get_numeric("vcpu_maxcpus")
         vcpu_sockets = params.get_numeric("vcpu_sockets")
+        win_max_vcpu_sockets = params.get_numeric("win_max_vcpu_sockets", 2)
         vcpu_cores = params.get_numeric("vcpu_cores")
         vcpu_threads = params.get_numeric("vcpu_threads")
         vcpu_dies = params.get("vcpu_dies", 0)
@@ -1650,9 +1651,9 @@ class VM(virt_vm.BaseVM):
 
         # Some versions of windows don't support more than 2 sockets of cpu,
         # here is a workaround to make all windows use only 2 sockets.
-        if (vcpu_sockets and vcpu_sockets > 2 and
-                params.get("os_type") == 'windows'):
-            vcpu_sockets = 2
+        if (vcpu_sockets and params.get("os_type") == 'windows' and
+                vcpu_sockets > win_max_vcpu_sockets):
+            vcpu_sockets = win_max_vcpu_sockets
 
         amd_vendor_string = params.get("amd_vendor_string")
         if not amd_vendor_string:


leidwang:
Hi @PaulYuuu,Could you help to review this patch?Thanks a lot.

virttest/qemu_vm.py
```diff
@@ -1638,6 +1638,7 @@ def sort_key(dev):
         smp = params.get_numeric('smp')
         vcpu_maxcpus = params.get_numeric("vcpu_maxcpus")
         vcpu_sockets = params.get_numeric("vcpu_sockets")
+        vcpu_sockets_force = params.get("vcpu_sockets_force", False)
```
PaulYuuu, line 1641:
```suggestion
        vcpu_sockets_force = params.get_boolean("vcpu_sockets_force", False)
```

And the name is not suitable, I mean: force what?

virttest/qemu_vm.py
```diff
@@ -1651,7 +1652,7 @@ def sort_key(dev):
         # Some versions of windows don't support more than 2 sockets of cpu,
         # here is a workaround to make all windows use only 2 sockets.
         if (vcpu_sockets and vcpu_sockets > 2 and
-                params.get("os_type") == 'windows'):
+                params.get("os_type") == 'windows' and vcpu_sockets_force is False):
```
PaulYuuu, line 1655: If we can collect the list of unsupported Windows versions, maybe we can remove this workaround?

@leidwang @vivianQizhu, what's your opinion?

virttest/qemu_vm.py
leidwang: Okay,I will change the name.Thanks!

leidwang: What do you think of `force_set_vcpu_sockets`?

virttest/qemu_vm.py
vivianQizhu: @PaulYuuu @leidwang There are too many different Windows versions and it will change yearly, I don't think it is necessary to maintain a list for it, but instead, I would suggest to have a configurable `max_vcpu_sockets`, like: `params.get('max_vcpu_sockets', 2)`, and provide users the opportunity to update it when needed, and we might update some of the Windows guest cfgs accordingly for those ones we believe necessary.

PaulYuuu: @vivianQizhu, this limitation is only for Windows guests, so I am not sure the name is suitable. BTW, for future windows releases, maybe the limitation is gone? For current releases, only `WinXP ~ Win11` and `Win2003 ~ Win2022`, So a list also doesn't change frequently?

vivianQizhu: @PaulYuuu
It might varies from different editions, not only versions, so it is not that easy to main such a list, we don't really track all edition changes frequently. Besides WinXP - Win8 have already EOL thus we will not have them, and Win8.1/2012/R2 are all very closed to EOL, we will need to add them and then remove them after they are EOL. Anyway, we will not be able to maintain such a list, if it does need, we might need a owner for it who need to check them on time, both @leidwang and me will not do that, that's out of our test requirement.
You are right it is only for Windows and the default should not be `2`, how about this:
I will remove `params.get("os_type") == 'windows'` and `vcpu_sockets > 2`
```
max_vcpu_sockets = params.get('max_vcpu_sockets')
if vcpu_sockets and max_vcpu_sockets and vcpu_sockets > max_vcpu_sockets:
    vcpu_sockets =  max_vcpu_sockets
```
And we set `max_vcpu_sockets = 2` in Windows.cfg, for any other specific requirement we can set them separately. This way it seems more reasonable and also stick to guest OS specification, and does not require extra effort to maintain(which I don't see the necessity).

leidwang: Hi @vivianQizhu I think we can not make the code like this:
```
max_vcpu_sockets = params.get('max_vcpu_sockets')
if vcpu_sockets and max_vcpu_sockets and vcpu_sockets > max_vcpu_sockets:
    vcpu_sockets =  max_vcpu_sockets
```
Because previous cases not set `max_vcpu_sockets`, only set `vcpu_sockets`, and the value will be limited by this workaround. If we changed the code as you provided,the workaround will be removed,so I think we'd better keep the workaround and add a configurable parameter.

vivianQizhu: @leidwang Not sure if I miss something, by setting `max_vcpu_sockets = 2` in Windows.cfg you would get exactly same thing with previous code.

leidwang: Okay,if so,we need to add the seting to windows.cfg too.I agree with you.

virttest/qemu_vm.py
```diff
@@ -1652,7 +1652,7 @@ def sort_key(dev):
         # here is a workaround to make all windows use only 2 sockets.
         if (vcpu_sockets and vcpu_sockets > 2 and
                 params.get("os_type") == 'windows'):
-            vcpu_sockets = 2
+            vcpu_sockets = params.get_numeric("max_vcpu_sockets", 2)
```
vivianQizhu, line 1655: I agree with @PaulYuuu that the name might not be that proper if it is Windows only, so either we can remove `params.get('os_type') == 'windows'` and specify it in Windows.cfg, or we might name it `win_max_vcpu_sockets`.

vivianQizhu: And the logic is not correct here, if `vcpu_sockets` is set to `4` and `max_vcpu_sockets` is `6`, the `vcpu_socket` will be set to `6`, which is not expected.
It should be `vcpu_sockets > max_vcpu_sockets` instead of a hard code `vcpu_sockets > 2`.

vivianQizhu:
@leidwang Please update Windows related cfg together: https://github.com/avocado-framework/avocado-vt/tree/master/virttest/shared/cfg/guest-os/Windows/, I suppose at least you will need to update those ones that needed by https://github.com/autotest/tp-qemu/pull/3290.

leidwang:
> @leidwang Please update Windows related cfg together: https://github.com/avocado-framework/avocado-vt/tree/master/virttest/shared/cfg/guest-os/Windows/, I suppose at least you will need to update those ones that needed by [autotest/tp-qemu#3290](https://github.com/autotest/tp-qemu/pull/3290).

Hi @vivianQizhu Thanks for you explanation.Currently, it's hard for us to gather all the windows sockets limits. I think we'd better provide a parameter(win_max_vcpu_sockets) here and keep the current workaround. This parameter can be defined by the user in the future.

vivianQizhu:
@leidwang Okay, let's focus on one issue fix for each single PR, LGTM now.
@PaulYuuu Would you like to have a re-review? Thanks a lot.

virttest/qemu_vm.py
```diff
@@ -1650,9 +1651,9 @@ def sort_key(dev):
 
         # Some versions of windows don't support more than 2 sockets of cpu,
         # here is a workaround to make all windows use only 2 sockets.
-        if (vcpu_sockets and vcpu_sockets > 2 and
-                params.get("os_type") == 'windows'):
-            vcpu_sockets = 2
+        if (vcpu_sockets and win_max_vcpu_sockets and
```
PaulYuuu, line 1654: Here you removed the `os_type` check, but also you set `win_max_vcpu_sockets ` and its default is 2, so for Linux guests, they will also go to this step.

leidwang: Okay,Thanks for your review.So we still need to check the os type now.

PaulYuuu: Restore the os_type check is okay for me, or leave `win_max_vcpu_sockets` empty for other OS, this should be the correct logic of the code. Anyway, this is up to you.

leidwang: Ok thanks @PaulYuuu.
I would like to keep the current code changes, your suggestion is reasonable, but I think it can be done in the future. Thanks!

vivianQizhu:
Thanks @leidwang @PaulYuuu

PR #3405, opened 2022-05-12
Update undefine() function to cover ovmf guest

meinaLi:
For compatibility with undefining ovmf guest and seabios guest, and no need to modify each python file one by one. I update the "def undefine()" in virsh.py and libvirt_vm.py. The libvirt_xml/vm_xml.py has already been implemented before.

Signed-off-by: Meina Li <meili@redhat.com>


virttest/virsh.py
```diff
@@ -1577,8 +1577,13 @@ def undefine(name, options=None, **dargs):
     :return: CmdResult object
     """
     cmd = "undefine %s" % name
-    if options is not None:
-        cmd += " %s" % options
+
+    vmxml = dumpxml(name).stdout_text
```
chunfuwen, line 1581: I assume we don't need this change due to --nvram, there is some example how to handle --nvram under current implemenation in file libvirt/tests/src/virsh_cmd/domain/virsh_undefine.py,

virttest/virsh.py
```diff
@@ -1577,8 +1577,16 @@ def undefine(name, options=None, **dargs):
...
+    if "nvram" in vmxml:
+        if options is None or "--nvram" not in options:
+            cmd += " --nvram"
+        else:
+            cmd += " %s" % options
+    else:
+        if options is not None:
+            cmd += " %s" % options
 
     if platform.machine() == "aarch64":
```
Yingshun, line 1591: I wonder if we should remove platform checking here to make it work for all archs?

Yingshun: Or add x86_64? I'm not quite sure about other archs, @dzhengfy  what's your opinion?

virttest/virsh.py
```diff
@@ -1577,8 +1577,16 @@ def undefine(name, options=None, **dargs):
...
+
+    vmxml = dumpxml(name).stdout_text
+    if "nvram" in vmxml:
+        if options is None or "--nvram" not in options:
+            cmd += " --nvram"
+        else:
+            cmd += " %s" % options
+    else:
+        if options is not None:
+            cmd += " %s" % options
```
dzhengfy, line 1589: As --nvram does not have side effect even on seabios guest, we can safely use it on both seabios and uefi guests. So let's remove above special checking. Thanks.

virttest/virsh.py
meinaLi: The implementation in virsh_undefine.py is so complicated and it will affect all other scripts which calls this function. So I directly update to x86_64 to cover seabios which will not have side effect.

virttest/virsh.py
```diff
@@ -1580,7 +1580,8 @@ def undefine(name, options=None, **dargs):
     if options is not None:
         cmd += " %s" % options
 
-    if platform.machine() == "aarch64":
+
+    if platform.machine() == "aarch64" or "x86_64":
```
dzhengfy, line 1584: Please help some descriptive message for --nvram purpose and make a statement , similar like below:  it will be applied on both seabios guest (a risk, suppose no side effect currently) and uefi guest...

Yingshun, line 1584:
```suggestion
    if platform.machine() in ["aarch64", "x86_64"]:
```

virttest/virsh.py
meinaLi: Updated

virttest/virsh.py
meinaLi: updated

meinaLi: updated

Yingshun (approved):

dzhengfy (approved):
LGTM

chloerh (approved):

chunfuwen (approved):
LGTM

qiankehan:
@meinaLi @chunfuwen Any reason to use `--nvram` by default?
The `virsh undefine` could **remove**(`--nvram`) or **keep**(`--keep-nvram`) the NVRAM. I cannot understand why removing `--nvram` is the default.
BTW, if I'd like to test `--keep-nvram`, the avocado-vt will add `--nvram`. That will cause this error:
```
error: Options --nvram and --keep-nvram are mutually exclusive
```

meinaLi:
> @meinaLi @chunfuwen Any reason to use `--nvram` by default? The `virsh undefine` could **remove**(`--nvram`) or **keep**(`--keep-nvram`) the NVRAM. I cannot understand why removing `--nvram` is the default. BTW, if I'd like to test `--keep-nvram`, the avocado-vt will add `--nvram`. That will cause this error:
> 
> ```
> error: Options --nvram and --keep-nvram are mutually exclusive
> ```

1. This is because we need to use ovmf guest in default to test. But most of the auto cases are using undefine command without --nvram. So all of them are undefined failed.
2. Actually I talked with you before about this question, for seabios and ovmf to ues --nvram. It won't affect seabios guest. But didn't consider the --keep-nvram. Now this PR had already been removed and directly updated in all related scrippts because of some other reasons.

PR #3406, opened 2022-05-13
libvirt: add new param

dzhengfy:
This is to give an oppotunity for the user to decide if an exception
or a boolean value is needed when checking fails.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -3815,7 +3815,8 @@ def customize_libvirt_config(params,
 
 
 def check_logfile(search_str, log_file, str_in_log=True,
-                  cmd_parms=None, runner_on_target=None):
+                  cmd_parms=None, runner_on_target=None,
+                  ignore_status=False):
     """
     Check if the given string exists in the log file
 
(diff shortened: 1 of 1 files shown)


chunfuwen (approved):
LGTM

virttest/utils_test/libvirt.py
```diff
@@ -3825,20 +3826,26 @@ def check_logfile(search_str, log_file, str_in_log=True,
...
     if str_in_log == bool(int(cmdRes.exit_status)):
-        raise exceptions.TestFail("The string '{}' {} included in {}"
-                                  .format(search_str, "is not" if str_in_log else "is",
-                                          log_file))
+        if not ignore_status:
+            raise exceptions.TestFail("The string '{}' {} included in {}"
+                                      .format(search_str, "is not" if str_in_log else "is",
+                                              log_file))
+        else:
+            return False
```
Yingshun, line 3845: How about logging an error message before returning false? So others can easily understand what happened.

dzhengfy: updated

Yingshun (approved):

PR #3408, opened 2022-05-17
qdevices: Make "pcie_direct_plug" work properly

PaulYuuu:
"pcie_direct_plug" cannot be kept in params all the time, otherwise it 
will be appended to the qemu command line. Setting this parameter as a
property when initializing a device is an easy fix but not the best 
solution.

ID: 2087062

### DIFF
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -89,7 +89,10 @@ class QBaseDevice(object):
         self.params = OrderedDict()    # various device params (id, name, ...)
         if params:
             for key, value in six.iteritems(params):
-                self.set_param(key, value)
+                if key == "pcie_direct_plug":
+                    self.pcie_direct_plug = value == "yes"
+                else:
+                    self.set_param(key, value)
 
     def add_child_bus(self, bus):
         """
@@ -2677,7 +2680,8 @@ class QPCIEBus(QPCIBus):
         self.__root_port_params = root_port_params
         self.__last_port_index = [0]
 
-    def is_direct_plug(self, device):
+    @staticmethod
+    def is_direct_plug(device):
         """
         Justify if the device should be plug to pcie bus directly,
         the device is not directly plugged only if it is a virtio
@@ -2686,8 +2690,9 @@ class QPCIEBus(QPCIBus):
         :param device: the QBaseDevice object
         :return: bool value for directly plug or not.
         """
-        not_direct_plug = device.get_param("pcie_direct_plug", "no") == "no"
-        if not_direct_plug and device.is_pcie_device():
+        direct_plug = (device.pcie_direct_plug
+                       if hasattr(device, "pcie_direct_plug") else False)
+        if not direct_plug and device.is_pcie_device():
             return False
         return True
 


PaulYuuu:
Hello @luckyh, can you help review? I saw some kata tests also want to use this parameter.

virttest/qemu_devices/qdevices.py
```diff
@@ -89,7 +89,10 @@ def __init__(self, dev_type="QBaseDevice", params=None, aobject=None,
         self.params = OrderedDict()    # various device params (id, name, ...)
         if params:
             for key, value in six.iteritems(params):
-                self.set_param(key, value)
+                if key == "pcie_direct_plug":
+                    self.pcie_direct_plug = value == "yes"
```
vivianQizhu, line 93: I understand how it comes, just not prefer to add such exception on fundamental definition.
How about `device.set_param("pcie_direct_plug", None)` so that it should be removed as I see in`def set_param()`:
```
elif value is None and option in self.params:
            del(self.params[option])
```

virttest/qemu_devices/qdevices.py
```diff
@@ -2686,8 +2690,9 @@ def is_direct_plug(self, device):
         :param device: the QBaseDevice object
         :return: bool value for directly plug or not.
         """
-        not_direct_plug = device.get_param("pcie_direct_plug", "no") == "no"
```
vivianQizhu, line 2689: I mean here, after get the value, remove it: `device.set_param("pcie_direct_plug", None)`. I don't really test it, just imagine it should work.

virttest/qemu_devices/qdevices.py
PaulYuuu: @vivianQizhu, if we use `set_param` to save `pcie_direct_plug` property, when the value is not `None`, it will be passed into the qemu command line, because this property is not a qemu device option.

virttest/qemu_devices/qdevices.py
PaulYuuu: Ah okay, but `is_direct_plug` will be called twice, so if set it to `None` at the first time, then the second time cannot get the correct value. And I can't guarantee if it will be called more times later.

virttest/qemu_devices/qdevices.py
vivianQizhu: @PaulYuuu Yes you are right, that might kill migration or repeatedly hotplug alike functions. But still I hope we can fix it but not introduce such weird exceptions. To make `pcie_direct_plug` a property is a good idea, I was thinking that to directly assign `device.pcie_direct_plug = False` when you get the param from cfg in qemu_vm.py, avoid calling `set_param`. But it seems that will require too much outside the pcie bus, which is also unexpected to me...Anyway I don't figure out a better solution so far, and I don't see any risk on your current solution. So I will approve it. Let's have @luckyh 's review, see if we can have a better idea.

vivianQizhu (approved):

PaulYuuu:
Hello @luckyh , any comments here?

luckyh (approved):
ACK

virttest/qemu_devices/qdevices.py
luckyh: My understanding is a refactoring of bus topology management is something must to have, so this time let use the current solution as a workaround to fix the issue, thanks.

luckyh:
Thanks to all.

PR #3411, opened 2022-05-20
Added the Chinese README file.

cheng191:


chunfuwen:
@clebergnu 

Please help confirm whether it  is common practice to add specific language support on readme.txt file only considering that all artifacts  in this repository are English only

PR #3420, opened 2022-06-06
libvirt_vmxml.py:Add function to create deivice if not get expected device

nanli1:
Need to add function to create device when not get device

From https://github.com/autotest/tp-libvirt/pull/4219

### DIFF
--- a/virttest/utils_libvirt/libvirt_vmxml.py
+++ b/virttest/utils_libvirt/libvirt_vmxml.py
@@ -11,6 +11,15 @@ from avocado.core import exceptions
 
 from virttest import virsh
 from virttest.libvirt_xml import vm_xml
+from virttest.libvirt_xml.devices import controller
+from virttest.libvirt_xml.devices import channel
+from virttest.libvirt_xml.devices import disk
+from virttest.libvirt_xml.devices import hostdev
+from virttest.libvirt_xml.devices import interface
+from virttest.libvirt_xml.devices import rng
+from virttest.libvirt_xml.devices import vsock
+from virttest.libvirt_xml.devices import watchdog
+from virttest.utils_test import libvirt
 
 LOG = logging.getLogger('avocado.' + __name__)
 
@@ -65,3 +74,60 @@ def remove_vm_devices_by_type(vm, device_type):
 
     if vm_was_running:
         vm.start()
+
+
+def create_vm_device_by_type(dev_type, dev_dict):
+    """
+    Create device by device type
+
+    :param dev_type: device type
+    :param dev_dict: dict for device
+    :return: device object
+    """
+    dev_obj = None
+    if dev_type == 'disk':
+        dev_obj = disk.Disk()
+    elif dev_type == 'controller':
+        dev_obj = controller.Controller()
+    elif dev_type == 'interface':
+        dev_obj = interface.Interface()
+    elif dev_type == 'channel':
+        dev_obj = channel.Channel()
+    elif dev_type == 'vsock':
+        dev_obj = vsock.Vsock()
+    elif dev_type == 'watchdog':
+        dev_obj = watchdog.Watchdog()
+    elif dev_type == 'rng':
+        dev_obj = rng.Rng()
+    elif dev_type == 'hostdev':
+        dev_obj = hostdev.Hostdev()
+
+    dev_obj.setup_attrs(**dev_dict)
+
+    return dev_obj
+
+
+def modify_vm_device(vmxml, dev_type, dev_dict=None, index=0):
+    """
+     Get specified device , update it with given dev_dict if the device exists,
+     Create the device if it does not exist
+
+    :param vmxml: domain VMXML instance
+    :param dev_type: device type
+    :param dev_dict: dict to create device
+    :param index: device index
+    :return: device object
+    """
+    dev_obj = None
+    try:
+        dev_obj, xml_devices = libvirt.get_vm_device(vmxml, dev_type, index=index)
+        dev_obj.setup_attrs(**dev_dict)
+
+        vmxml.devices = xml_devices
+        vmxml.xmltreefile.write()
+        vmxml.sync()
+    except IndexError:
+        dev_obj = create_vm_device_by_type(dev_type, dev_dict)
+        libvirt.add_vm_device(vmxml, dev_obj)
+
+    return dev_obj


nanli1:
@dzhengfy Hi dzheng ，Here is the pr  to update `def get_vm_device` Please help check
 Could you please help confirm which device needs to be added in` if else` branch

virttest/utils_test/libvirt.py
```diff
@@ -2371,19 +2372,46 @@ def create_tpm_dev(params):
...
+        elif dev_tag == 'watchdog':
+            new_device = watchdog.Watchdog()
+        elif dev_tag == 'rng':
+            new_device = rng.Rng()
+        elif dev_tag == 'hostdev':
+            new_device = hostdev.Hostdev()
+
+        new_device.setup_attrs(**dev_dict)
+        add_vm_device(vmxml, new_device)
+        return (new_device, xml_devices)
```
dzhengfy, line 2414: In the original codes, I do not know why we return all devices by `xml_devices`. However, I do not find any reference for this function in other codes. So should we only return one device object?

dzhengfy (commented):
Others LGTM. Thanks a lot

virttest/utils_test/libvirt.py
nanli1: yes , there is not any reference for this function in other codes.
I removed it `xml_devices`
BTW: Could you please help confirm which device needs to be added in` if else branch` or these is enough , because  there are many devices under the folder of  ` ./virttest/libvirt_xml/devices/*` @dzhengfy

virttest/utils_test/libvirt.py
```diff
@@ -2371,19 +2372,46 @@ def create_tpm_dev(params):
     return tpm_dev
 
 
-def get_vm_device(vmxml, dev_tag, index=0):
+def get_vm_device(vmxml, dev_tag, dev_dict={}, index=0):
     """
-    Get current vm device according to device tag
+    Get current vm device according to device tag,
+    if no this device , add one device
```
dzhengfy, line 2378: Could you help add a description about the fact that this function only supports to add a new device whose class __init__() has default parameter values.

virttest/utils_test/libvirt.py
dzhengfy: So far we do not need to add all devices. In future if anyone want to use, it is so easy to extend this function as well as test result .

virttest/utils_test/libvirt.py
nanli1: @dzhengfy Thank you dzheng,   I have updated it ,Is this OK

 ```
 if no this device, add a new device and only support the device class
     default parameter values.
```

**Review by dzhengfy — DISMISSED** [review]
LGTM

File: virttest/utils_test/libvirt.py [hunk 3]
```diff
@@ -2371,19 +2372,47 @@ def create_tpm_dev(params):
     return tpm_dev
 
 
-def get_vm_device(vmxml, dev_tag, index=0):
+def get_vm_device(vmxml, dev_tag, dev_dict={}, index=0):
```
Yingshun, line 2375: Could you please update the default value of dev_dict as None? It may lead to some unexpected behavior.
For details, please check https://github.com/avocado-framework/avocado-vt/pull/3420.

nanli1: @Yingshun Thank you :smile:  ,Yingshun ,  but  Did you give the correct pr link

nanli1: @Yingshun Great .Thanks yingshun,I have updated it ,And will learn from this link.

Yingshun (approved):

virttest/utils_test/libvirt.py
```diff
@@ -2371,19 +2372,47 @@ def create_tpm_dev(params):
     return tpm_dev
 
 
-def get_vm_device(vmxml, dev_tag, index=0):
+def get_vm_device(vmxml, dev_tag, dev_dict=None, index=0):
```
chunfuwen, line 2375: Command-query separation (CQS)
A principle of imperative computer programming. It was devised by Bertrand Meyer as part of his pioneering work on the Effective programming language. It states that every method should either be a command that performs an action, or a query that returns data to the caller, but not both

As per Command-query separation (CQS) principal, don't try to mix get and add operations together in one method.

nanli1: Yes ,Thank you chunfu ,That's a great principle :smile: , 
But I think this function only implement one operation even after my updated, which is to get one device you want. it follows the principle, right ? And this updated propose is gotten from dzheng @dzhengfy Please help check

chloerh, line 2375: How about 
```suggestion
def get_vm_device(vmxml, dev_tag, index=0, dev_dict=None):
```
Though I didn't find reference of this function in tp-libvirt, just to be safe, adding the new arg to the end of the list could reduce the risk. What do you say?

nanli1: Great ,Updated like this can avoid some potential risk ,Thanks your comment.
Updated.

chloerh (approved):

virttest/utils_test/libvirt.py
chunfuwen: If you want to create one method,which achieve the purpose of creating one if not present when call get_xx, I suggest you create one separate method. Don't just modify them on primitive get_vm_device.

nanli1:
@dzhengfy  @Yingshun @chloerh @chunfuwen Please help re-review,Thank you all

virttest/utils_libvirt/libvirt_vmxml.py
```diff
@@ -9,8 +9,18 @@
 
 from avocado.core import exceptions
 
+from virttest.libvirt_xml.devices import controller
```
dzhengfy, line 12: Please reorder these lines and move them after line 22?

virttest/utils_libvirt/libvirt_vmxml.py
```diff
@@ -65,3 +75,55 @@ def remove_vm_devices_by_type(vm, device_type):
...
+
+def create_vm_device_by_type(dev_type, dev_dict):
+    """
+    Create device by device type
+
+    :param dev_type: device type
+    :param dev_dict: dict for device
+    :return: device object
+    """
+    dev_obj = ''
```
dzhengfy, line 88: `dev_obj = None ` is it better?

virttest/utils_libvirt/libvirt_vmxml.py
```diff
@@ -65,3 +75,55 @@ def remove_vm_devices_by_type(vm, device_type):
...
+        dev_obj = hostdev.Hostdev()
+
+    dev_obj.setup_attrs(**dev_dict)
+
+    return dev_obj
+
+
+def get_or_create_vm_device_if_absent(vmxml, dev_type, dev_dict, index=0):
+    """
+     Get specified device ,create the device if it not exist
```
dzhengfy, line 113: if it does not exist

nanli1: done

virttest/utils_libvirt/libvirt_vmxml.py
nanli1: done

virttest/utils_libvirt/libvirt_vmxml.py
nanli1: done

virttest/utils_libvirt/libvirt_vmxml.py
```diff
@@ -65,3 +74,55 @@ def remove_vm_devices_by_type(vm, device_type):
...
+        dev_obj = rng.Rng()
+    elif dev_type == 'hostdev':
+        dev_obj = hostdev.Hostdev()
+
+    dev_obj.setup_attrs(**dev_dict)
+
+    return dev_obj
+
+
+def get_or_create_vm_device_if_absent(vmxml, dev_type, dev_dict, index=0):
```
dzhengfy, line 110: After further thinking, this function's purpose should be updating a vm device using an unified way instead of creating multiple updating functions for different devices. If the device exists, let's update it with given dev_dict, otherwise, let's create it with given dev_dict. 
So shall we rename it as 'modify_vm_device'? @chunfuwen @Yingshun @chloerh

virttest/utils_libvirt/libvirt_vmxml.py
```diff
@@ -65,3 +74,55 @@ def remove_vm_devices_by_type(vm, device_type):
...
+
+    :param vmxml: domain VMXML instance
+    :param dev_type: device type
+    :param dev_dict: dict to create device
+    :param index: device index
+    :return: device object
+    """
+
+    try:
+        return libvirt.get_vm_device(vmxml, dev_type, index=index)
```
dzhengfy, line 122: If rename, please use setup_attrs() to update this returned device.

dzhengfy (commented):
'modify_vm_device' might be more accurate name

chunfuwen (approved):
LGTM

virttest/utils_libvirt/libvirt_vmxml.py
Yingshun: I agree with you. It's really weird returning the existing device without any updates or comparisons in this function. Then, I think it'll be much more clear if the device will be updated with the given dev_dict.

virttest/utils_libvirt/libvirt_vmxml.py
nanli1: done

chunfuwen (approved):
LGTM

PR #3422, opened 2022-06-07
migration: Add a new param

cliping:
Add a param to check guest state on source host.

Signed-off-by: cliping <lcheng@redhat.com>

### DIFF
--- a/virttest/migration.py
+++ b/virttest/migration.py
@@ -46,7 +46,7 @@ class MigrationTest(object):
         # The format is: <func_object, func_return>
         self.func_ret = {}
 
-    def post_migration_check(self, vms, params, uptime=None, uri=None):
+    def post_migration_check(self, vms, params, uptime=None, dest_uri=None, src_uri=None):
         """
         Validating migration by performing checks in this method
         * check vm state after migration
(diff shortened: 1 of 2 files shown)


cliping:
Test result: https://github.com/autotest/tp-libvirt/pull/4252

cliping:
@dzhengfy  @Yingshun , please take a look, thanks.

virttest/migration.py
```diff
@@ -59,40 +59,47 @@ def post_migration_check(self, vms, params, uptime=None, uri=None):
...
         :param vms: VM objects of migrating vms
         :param uptime: uptime dict of vms before migration
-        :param uri: target virsh uri
+        :param dest_uri: target virsh uri
+        :param src_uri: source virsh uri
         :return: updated dict of uptime
         """
-        vm_state = params.get("virsh_migrated_state", "running")
+        vm_state_dest = params.get("virsh_migrate_dest_state", "running")
+        vm_state_src = params.get("virsh_migrate_src_state")
```
dzhengfy, line 67: Could we have 'Shut off' as default value for `vm_state_src`?

dzhengfy (commented):
Other LGTM

virttest/migration.py
cliping: Updated, thanks.

Yingshun (approved):

virttest/migration.py
```diff
@@ -59,40 +59,47 @@ def post_migration_check(self, vms, params, uptime=None, uri=None):
...
         :param vms: VM objects of migrating vms
         :param uptime: uptime dict of vms before migration
-        :param uri: target virsh uri
+        :param dest_uri: target virsh uri
+        :param src_uri: source virsh uri
         :return: updated dict of uptime
         """
-        vm_state = params.get("virsh_migrated_state", "running")
+        vm_state_dest = params.get("virsh_migrate_dest_state", "running")
+        vm_state_src = params.get("virsh_migrate_src_state", "shut off")
```
dzhengfy, line 67: Just a reminder. 
In existing tp-libvirt cases, some python files already use 'virsh_migrate_src_state' for another purpose like the prepared vm state before migration. So we need to be careful if those files use this function in future and we need rename the variable if any.

dzhengfy, line 67: Just a reminder.
In existing tp-libvirt cases, some python files already use 'virsh_migrate_src_state' for another purpose like the prepared vm state before migration. So we need to be careful if those files use this function in future and we need rename the variable if any. But so far so good.

dzhengfy (approved):

PR #3423, opened 2022-06-08
virsh: remove improper nvram added

dzhengfy:
--nvram option should not be added in this function because it is
very low level implementation and should be tranparent and clean.

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -31,7 +31,6 @@ import select
 import locale
 import base64
 import inspect
-import platform
 
 from functools import wraps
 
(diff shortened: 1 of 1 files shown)


virttest/virsh.py
```diff
@@ -1588,12 +1587,6 @@ def undefine(name, options=None, **dargs):
     if options is not None:
         cmd += " %s" % options
 
-    # When the platform is aarch64 or x86_64, both seabios and ovmf
-    # guests will use --nvram option
-    if platform.machine() in ["aarch64", "x86_64"]:
-        if options is None or "--nvram" not in options:
-            cmd += " --nvram"
-
     LOG.debug("Undefine VM %s", name)
```
chunfuwen, line 1590: I think it can be also removed as well

chloerh (approved):

virttest/virsh.py
dzhengfy: removed

chunfuwen (approved):
LGTM

virttest/virsh.py
dzhengfy: After discussion in the team, we decide to keep the log.

PR #3424, opened 2022-06-08
guest-hw: remove nfs settings

aliang123:
comment out remote_nfs expose settings, define them
    in user cases.

id:2094709

### DIFF
--- a/virttest/shared/cfg/guest-hw.cfg
+++ b/virttest/shared/cfg/guest-hw.cfg
@@ -402,8 +402,8 @@ variants image_backend:
         image_raw_device = yes
     - remote_nfs:
         storage_type = nfs
-        nfs_mount_dir = /mnt/nfs_images
-        nfs_mount_options = rw
+        #nfs_mount_dir = /mnt/nfs_images
+        #nfs_mount_options = rw
         # Please update nfs_mount_src base your nfs server
         #nfs_mount_src = 192.168.12.10:/vol/s2imagetest
     - local_nfs:


aliang123:
@zhencliu please help to review, thanks.

virttest/env_process.py
```diff
@@ -1089,8 +1089,8 @@ def preprocess(test, params, env):
             name_tag = "image_name_%s" % image_name
             if params.get(name_tag):
                 image_name_only = os.path.basename(params[name_tag])
-                params[name_tag] = os.path.join(image_nfs.mount_dir,
-                                                image_name_only)
+            params[name_tag] = os.path.join(image_nfs.mount_dir,
+                                            image_name_only)
```
zhencliu, line 1093: what about doing this in the following way:
1. remove line 1087
2. in the for loop
img_params = params.object_params(image_name)
if img_params['storage_type'] == 'nfs':
    name_tag = "image_name_%s" % image_name
    image_basename = os.path.basename(img_params['image_name'])
    params[name_tag] = os.path.join(image_nfs.mount_dir, image_basename)

What's more, we have to update virttest/shared/cfg/guest-hw.cfg, comment out the following params:
nfs_mount_dir
nfs_mount_options
export_dir
export_options

The reason for the change is:
1. these params are very specific, it's not a good idea to define them in the guest-hw.cfg file.
2. what if we define them in guest-hw.cfg? Discussed with yongxue and xu today, it's complicated, when we create the tests-example.cfg file, the order of the cfg files included(i.e. include xxx) is not correct, it's a heavy task to fix that, so currently, we made a decision to use this workaround.

Last, please also enhance the kar code to apply the nfs storage backends.

Thanks.

aliang123: Have updated, and img_params['storage_type'] == 'nfs': has been defined in line 1052.

zhencliu: hi aihua, I am afraid they are not the same...consider the following configuration:
only remote_nfs
storage_type = nfs
images += data
storage_type_data = nvme

i.e. we used a system image from nfs, and a local nvme device, which could be a valid configuration supported by VT, although currently VT only supports one storage backend.

zhencliu:
hi @YongxueHong and @luckyh , would you take a look at this PR, I checked the tp-libvirt, base.cfg will be updated with the related nfs settings before running test cases, e.g. nfs_mount_dir, so IMO, it's safe to comment out them in guest-hw.cfg

zhencliu:
LGTM

PR #3434, opened 2022-06-24
utils_env:register iscsidev to env

aliang123:
register iscsidev to env
    
    id:2100690

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1044,6 +1044,7 @@ def preprocess(test, params, env):
         iscsidev = qemu_storage.Iscsidev(params, base_dir, "iscsi")
         params["image_name"] = iscsidev.setup()
         params["image_raw_device"] = "yes"
+        env.register_iscsidev("iscsi_%s" % params["main_vm"], iscsidev)
 
     if params.get("storage_type") == "lvm":
         lvmdev = qemu_storage.LVMdev(params, base_dir, "lvm")
@@ -1817,11 +1818,12 @@ def postprocess(test, params, env):
 
     if params.get("storage_type") == "iscsi":
         try:
-            iscsidev = qemu_storage.Iscsidev(params, base_dir, "iscsi")
+            iscsidev = env.get_iscsidev("iscsi_%s" % params["main_vm"])
             iscsidev.cleanup()
         except Exception as details:
             err += "\niscsi cleanup: %s" % str(details).replace('\\n', '\n  ')
             LOG.error(details)
+        env.unregister_iscsidev("iscsi_%s" % params["main_vm"])
 
     if params.get("storage_type") == "lvm":
         try:
--- a/virttest/utils_env.py
+++ b/virttest/utils_env.py
@@ -200,6 +200,34 @@ class Env(IterableUserDict):
         return self.data.get("sync__%s" % port)
 
     @lock_safe
+    def register_iscsidev(self, name, iscsi_dev):
+        """
+        Register iscsi device object into env;
+
+        :param name: name of register iscsi device object
+        :param iscsi_dev: iscsi device object;
+        """
+        self.data["iscsidev__%s" % name] = iscsi_dev
+
+    def get_iscsidev(self, name):
+        """
+        Get iscsi device object by name from env;
+
+        :param name: iscsi device object name;
+        :return: iscsi device object
+        """
+        return self.data.get("iscsidev__%s" % name)
+
+    @lock_safe
+    def unregister_iscsidev(self, name):
+        """
+        Remove iscsi device object from env;
+
+        :param name: name of iscsi device object;
+        """
+        del self.data["iscsidev__%s" % name]
+
+    @lock_safe
     def register_lvmdev(self, name, lvmdev):
         """
         Register lvm device object into env;


aliang123:

YongxueHong:
Hi @aliang123 
Would you like to explain the reason why register the iscsidev to env?
Thanks.

aliang123:
> Hi @aliang123 Would you like to explain the reason why register the iscsidev to env? Thanks.
In original code,
  In pre stage, we define an iscsidev instance to do some setup steps, then in post stage, we re-define an iscsidev instance to do some cleanup steps, that's not reasonable, and we should operate the same iscsidev instance in pre and post stage.
 
So after discussion with zhenchao, I update code as it is now:
   In pre stage, we define an iscsidev instance to do some setup steps, then we store it in the env, and get it in the post stage to do some cleanup steps.

zhencliu:
> > Hi @aliang123 Would you like to explain the reason why register the iscsidev to env? Thanks.
> > In original code,
> > In pre stage, we define an iscsidev instance to do some setup steps, then in post stage, we re-define an iscsidev instance to do some cleanup steps, that's not reasonable, and we should operate the same iscsidev instance in pre and post stage.
> 
> So after discussion with zhenchao, I update code as it is now: In pre stage, we define an iscsidev instance to do some setup steps, then we store it in the env, and get it in the post stage to do some cleanup steps.

yes, just handle iscsidev in the same way to handle lvmdev in VT. 
But I just found that env can be 'pickled', do you have concern about that? @YongxueHong

YongxueHong:
> > > Hi @aliang123 Would you like to explain the reason why register the iscsidev to env? Thanks.
> > > In original code,
> > > In pre stage, we define an iscsidev instance to do some setup steps, then in post stage, we re-define an iscsidev instance to do some cleanup steps, that's not reasonable, and we should operate the same iscsidev instance in pre and post stage.
> > 
> > 
> > So after discussion with zhenchao, I update code as it is now: In pre stage, we define an iscsidev instance to do some setup steps, then we store it in the env, and get it in the post stage to do some cleanup steps.
> 
> yes, just handle iscsidev in the same way to handle lvmdev in VT. But I just found that env can be 'pickled', do you have concern about that? @YongxueHong

Hi @aliang123 CC @zhencliu 
You mean the pre stage and post stage is the job stage or the test stage?
From the your patch, I think your requirements focus on the test stage which is to set up the environment for the test, so I suggest that we could use the `_setup_manager = test_setup.SetupManager()` then register your setup.
You can refer to:
```
    _setup_manager.initialize(test, params, env)
    _setup_manager.register(test_setup.UlimitConfig)
    _setup_manager.do_setup()
```
The registering objects in the `env` is aim to keep the objects to be calling among the multiple jobs, like vm object.
So I think it is not appropriate for registering in the `env`.

YongxueHong:
> > > Hi @aliang123 Would you like to explain the reason why register the iscsidev to env? Thanks.
> > > In original code,
> > > In pre stage, we define an iscsidev instance to do some setup steps, then in post stage, we re-define an iscsidev instance to do some cleanup steps, that's not reasonable, and we should operate the same iscsidev instance in pre and post stage.
> > 
> > 
> > So after discussion with zhenchao, I update code as it is now: In pre stage, we define an iscsidev instance to do some setup steps, then we store it in the env, and get it in the post stage to do some cleanup steps.
> 
> yes, just handle iscsidev in the same way to handle lvmdev in VT. But I just found that env can be 'pickled', do you have concern about that? @YongxueHong

Hi @aliang123 CC @zhencliu 
You mean the pre stage and post stage is the job stage or the test stage?
From the your patch, I think your requirements focus on the test stage which is to set up the environment for the test, so I suggest that we could use the `_setup_manager = test_setup.SetupManager()` then register your setup.
You can refer to:
```
    _setup_manager.initialize(test, params, env)
    _setup_manager.register(test_setup.UlimitConfig)
    _setup_manager.do_setup()
```
The registering objects in the `env` is aim to keep the objects to be calling among the multiple jobs, like vm object.
So I think it is not appropriate for registering in the `env`.

zhencliu:
> > > > Hi @aliang123 Would you like to explain the reason why register the iscsidev to env? Thanks.
> > > > In original code,
> > > > In pre stage, we define an iscsidev instance to do some setup steps, then in post stage, we re-define an iscsidev instance to do some cleanup steps, that's not reasonable, and we should operate the same iscsidev instance in pre and post stage.
> > > 
> > > 
> > > So after discussion with zhenchao, I update code as it is now: In pre stage, we define an iscsidev instance to do some setup steps, then we store it in the env, and get it in the post stage to do some cleanup steps.
> > 
> > 
> > yes, just handle iscsidev in the same way to handle lvmdev in VT. But I just found that env can be 'pickled', do you have concern about that? @YongxueHong
> 
> Hi @aliang123 CC @zhencliu You mean the pre stage and post stage is the job stage or the test stage? From the your patch, I think your requirements focus on the test stage which is to set up the environment for the test, so I suggest that we could use the `_setup_manager = test_setup.SetupManager()` then register your setup. You can refer to:
> 
> ```
>     _setup_manager.initialize(test, params, env)
>     _setup_manager.register(test_setup.UlimitConfig)
>     _setup_manager.do_setup()
> ```
> 
> The registering objects in the `env` is aim to keep the objects to be calling among the multiple jobs, like vm object. So I think it is not appropriate for registering in the `env`.
hi @YongxueHong , consider this situation:
job1: run the job with iscsi backend, then save env(iscsidev) to a file
job2: run the job with nfs backend, and restore env from job1, then the only way we can get iscsi backend info is to restore it from env. But I don't think we need to cover that test scenario, @aliang123 what do you think?
And a 3rd way is to keep the code unchanged, just do a small code change for aihua's tests.

aliang123:
> _setup_manager

Hi, @YongxueHong 
   From the test aspect, iscsi setup and clean should be in job related stage, not in test case stage. So I think, add it to env is reasonable.
  cc @zhencliu

YongxueHong:
> > _setup_manager
> 
> Hi, @YongxueHong From the test aspect, iscsi setup and clean should be in job related stage, not in test case stage. So I think, add it to env is reasonable. cc @zhencliu


Hi @aliang123 
From your code, I see that you unregister the iscsi during the postprocess stage.
```
    if params.get("storage_type") == "iscsi":
        ...
        env.unregister_iscsidev("iscsi_%s" % params["main_vm"])
```
I mean however the `env` will always unregister the iscsi dev after running the test case, maybe for the next job, it can not get the previous iscsi dev, seems like it can not reach the goal of the job-related stage.

aliang123:

PR #3435, opened 2022-06-24
iscsi:new function iscsi_get_device_by_lun

aliang123:
function:iscsi_get_device_by_lun() to get
logged-in iscsi device by lun.

id:2100698

### DIFF
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -40,6 +40,24 @@ def get_image_filename(portal, target, lun=0, user=None, password=None):
     return uri.format(auth=auth, portal=portal, target=target, lun=lun)
 
 
+def iscsi_get_device_by_lun(target, lun):
+    """
+    Get image filename of iscsi device.
+    """
+    cmd = "iscsiadm -m session -P 3"
+    pattern = r"Lun:\s(\d+).*?scsi\sdisk\s(\w+)\s+State:\srunning"
+    devices = None
+    output = process.run(cmd).stdout_text
+    targets = output.split('Target: ')[1:]
+    for item in targets:
+        if target in item:
+            devices = dict(re.findall(pattern, target, re.S))
+            for key, value in devices.items():
+                dname = "/dev/%s" % value
+                devices[key] = dname
+    return devices[lun]
+
+
 def restart_iscsid(reset_failed=True):
     """
     Restart iscsid service.


aliang123:

virttest/iscsi.py
```diff
@@ -330,14 +330,18 @@ def get_device_name(self):
...
             output = process.run(cmd).stdout_text
             targets = output.split('Target: ')[1:]
             for target in targets:
                 if self.target in target:
-                    device_name = re.findall(pattern, target, re.S)
+                    self.devices = dict(re.findall(pattern, target, re.S))
+                    device_name.append(self.devices["0"])
+                    for lun, device in self.devices.items():
+                        dname = "/dev/%s" % device
+                        self.devices[lun] = dname
```
zhencliu, line 344: hi Aihua, I just noticed that we need to call get_device_name first, then we can get self.devices, i.e. self.devices depends on get_device_name function, that might not be a good idea. So what about introducing a new interface, like get_devices, which returns the {lun: dev}. In this way, we don't need self.devices, every time we use it, we call the function to get it. For the devices could change(think about that we add a new lun in server, we have to refresh devices)

aliang123: updated, @zhencliu, please help to review again, thanks.

virttest/iscsi.py
```diff
@@ -347,6 +347,26 @@ def get_device_name(self):
             LOG.error("Session is not logged in yet.")
         return device_name
 
+    def get_devices(self):
+        """
+        Get devices from the target name.
+        """
+        cmd = "iscsiadm -m session -P 3"
+        pattern = r"Lun:\s(\d+).*?scsi\sdisk\s(\w+)\s+State:\srunning"
```
zhencliu, line 355: the pattern doesn't work well for the following:
Attached SCSI devices:
************************
Host Number: 17 State: running
scsi17 Channel 00 Id 0 Lun: 0
Attached scsi disk sdf          State: running
scsi17 Channel 00 Id 0 Lun: 1
Attached scsi disk sdi          State: ing         >>>>>>not running here
scsi17 Channel 00 Id 0 Lun: 2
Attached scsi disk sdh          State: running
scsi17 Channel 00 Id 0 Lun: 3
Attached scsi disk sdg          State: running

I cannot find an easy way to handle this, just fyi.
pattern = r"Lun:\s(\d+).*?scsi\sdisk\s(\w+)\s+State:\s(\w+)"
devices = {t[0]:t[1] for t in re.findall(pattern, c, re.S) if t and t[2] == 'running'}

virttest/iscsi.py
```diff
@@ -347,6 +347,26 @@ def get_device_name(self):
...
+        devices = None
+        if self.logged_in():
+            output = process.run(cmd).stdout_text
+            targets = output.split('Target: ')[1:]
+            for target in targets:
+                if self.target in target:
+                    devices = dict(re.findall(pattern, target, re.S))
+                    for lun, device in devices.items():
+                        dname = "/dev/%s" % device
+                        devices[lun] = dname
```
zhencliu, line 365: do we want to break out the for target in targets loop inside if?

aliang123:

aliang123:

PR #3436, opened 2022-06-24
env_process:extend iscsi support on image_name_*

aliang123:
Extend backend iscsi support on image_name_*

id:2100705

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1044,6 +1044,10 @@ def preprocess(test, params, env):
         iscsidev = qemu_storage.Iscsidev(params, base_dir, "iscsi")
         params["image_name"] = iscsidev.setup()
         params["image_raw_device"] = "yes"
+        for image_name in params.objects("images"):
+            name_tag = "image_name_%s" % image_name
+            if params.get(name_tag):
+                params["image_raw_device_%s" % image_name] = "yes"
 
     if params.get("storage_type") == "lvm":
         lvmdev = qemu_storage.LVMdev(params, base_dir, "lvm")


aliang123:
depends on https://github.com/avocado-framework/avocado-vt/pull/3435

aliang123:
depends on https://github.com/avocado-framework/avocado-vt/pull/3435

aliang123:

virttest/env_process.py
```diff
@@ -1044,6 +1044,13 @@ def preprocess(test, params, env):
         iscsidev = qemu_storage.Iscsidev(params, base_dir, "iscsi")
         params["image_name"] = iscsidev.setup()
         params["image_raw_device"] = "yes"
+        for image_name in params.objects("images"):
+            name_tag = "image_name_%s" % image_name
+            if params.get(name_tag):
```
zhencliu, line 1049: if not params.get(name_tag)  here?

aliang123:
@zhencliu , updated, please help to review again.

virttest/env_process.py
```diff
@@ -1044,6 +1044,14 @@ def preprocess(test, params, env):
         iscsidev = qemu_storage.Iscsidev(params, base_dir, "iscsi")
         params["image_name"] = iscsidev.setup()
         params["image_raw_device"] = "yes"
+        for image_name in params.objects("images"):
+            lun_tag = "lun_%s" % image_name
+            if params.get(lun_tag):
+                image_params = params.object_params(image_name)
+                lun = image_params.get("lun", "0")
```
zhencliu, line 1051: image_params['lun'], we should always get the value of 'lun' here

virttest/env_process.py
```diff
@@ -1044,6 +1044,14 @@ def preprocess(test, params, env):
...
+                name_tag = "image_name_%s" % image_name
+                params[name_tag] = iscsidev.iscsidevice.get_devices[lun]
```
zhencliu, line 1053: get_devices->get_devices()

virttest/env_process.py
```diff
@@ -1044,6 +1044,14 @@ def preprocess(test, params, env):
...
         params["image_name"] = iscsidev.setup()
         params["image_raw_device"] = "yes"
+        for image_name in params.objects("images"):
+            lun_tag = "lun_%s" % image_name
+            if params.get(lun_tag):
+                image_params = params.object_params(image_name)
+                lun = image_params.get("lun", "0")
+                name_tag = "image_name_%s" % image_name
+                params[name_tag] = iscsidev.iscsidevice.get_devices()[lun]
+                params["image_raw_device_%s" % image_name] = "yes"
```
zhencliu, line 1054: I just noticed that image_raw_device is set to raw in guest-hw.cfg, so here we may remove this line

virttest/env_process.py
aliang123: We use lun 0 for default, so we don't define lun_image1 = 0 in our cfgs, and for some needs, we set image_name_image1 = ${image_ name} in https://gitlab.cee.redhat.com/kvm-qe/kar/-/merge_requests/924, so lun = image_params.get("lun", "0") is better here.

virttest/env_process.py
aliang123: have updated, thanks.

virttest/env_process.py
aliang123: Although we define image_raw_device = yes in guest-hw.cfg, it doesn't take effect when rebase. Dut to the code limitation,  we define image_raw_device_$snapshot = no, in qemu_storage.rebase, we get image_raw_device = no for base image, that's not reasonable.

If we define params["image_raw_device_%s" % image_name] = "yes" in env_process, we can get image_raw_device = yes for base image when we do rebasing.

virttest/env_process.py
zhencliu: I mean the if statement(1049) has already confirmed 'lun' is defined, so we can always get image_params['lun'] here, the default value '0' will never be used.

but what you mentioned here is right, we need to consider lun_image1, way may also have to give it a default value, just like what you've done for remote_nfs

virttest/env_process.py
zhencliu: Sure, I got that pain-point, then it's OK for me for the update

virttest/env_process.py
aliang123: oh, I see. I don't define lun_image1=0 and in patch:924 I set image_name_image1. So I need to revert the code to the original, use "if params.get(name_tag):" instead of  "if params.get(lun_tag):" here, or else, the setting for image_name_image1 will be missed, thanks for reminding.

zhencliu: what about adding lun_image1 = 0 in 924, for it meaningless to assign an image name to a block device image

zhencliu: consider someone is now using hard conf:
image_name_data = /dev/sdc
Then the change now can break above.

If we use lun, which is not used for iscsi backend currently, then we don't break the settings above.

aliang123:

PR #3437, opened 2022-06-24
storage:get correct image name for block device

aliang123:
Get correct image name for block device

id:2100707

### DIFF
--- a/virttest/storage.py
+++ b/virttest/storage.py
@@ -180,6 +180,7 @@ def get_image_filename(params, root_dir, basename=False):
     enable_nvme = params.get("enable_nvme") == "yes"
     image_name = params.get("image_name")
     storage_type = params.get("storage_type")
+    image_raw_device = params.get("image_raw_device")
     if image_name:
         image_format = params.get("image_format", "qcow2")
         if enable_curl:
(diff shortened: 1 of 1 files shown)


virttest/storage.py
```diff
@@ -233,6 +234,8 @@ def get_image_filename(params, root_dir, basename=False):
 
             return storage_ssh.get_image_filename(server, ssh_image_path,
                                                   user, port, host_key_check)
+        if storage_type == 'iscsi' and image_raw_device == 'yes':
+            return image_name
```
zhencliu, line 238: this depends on the env_process, qemu_storage/storage can be used without starting vm, so we cannot assume that params has been updated in env_process here.

if storage_type == 'iscsi':
     return iscsi.get_image_filename_iscsi(params)

In iscsi.get_image_filename_iscsi, we log into iscsi server and scan the devices(get_devcies) then we get the image name (e.g. /dev/sdb) with lun id.

The same issue may happen on nfs, please make a note for future fix.

aliang123: Have updated, please help to check again.

aliang123:

PR #3441, opened 2022-06-28
fetch_attrs:Handle XMLElementBool type value

chloerh:
A value of XMLElementBool means the tag exists when the value is
True, doesn't exist when it's False. Therefore the False value
should not be returned when calling the function fetch_attrs()


chloerh:
@Yingshun @dzhengfy Could you help review this fix? Thanks.

Yingshun (approved):

dzhengfy (approved):
LGTM

PR #3444, opened 2022-06-30
Add debug print to check_cmd_output function

vasekhodina:
In file virttest/utils_test/libvirt.py in function check_cmd_output a
debug log print has been added, to help with debugging. It shows you
what is the command that has been run and what is the output of the
command.

I just thought this would be helpful when I was debugging an issue and
needed to know what exectly avocado is executing.

It could save time by not having to add debugging printsouts.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -3896,6 +3896,7 @@ def check_cmd_output(cmd, content, err_ignore=False, session=None):
     s, cmd_output = utils_misc.cmd_status_output(cmd, shell=True,
                                                  ignore_status=err_ignore, session=session)
     pattern_list = [content] if not isinstance(content, list) else content
+    LOG.debug('Ran command "%s" with output "%s"', cmd, cmd_output)
     for item in pattern_list:
         if not re.search(r'%s' % item, cmd_output):
             if err_ignore:


virttest/utils_test/libvirt.py
```diff
@@ -3896,6 +3896,7 @@ def check_cmd_output(cmd, content, err_ignore=False, session=None):
     s, cmd_output = utils_misc.cmd_status_output(cmd, shell=True,
                                                  ignore_status=err_ignore, session=session)
     pattern_list = [content] if not isinstance(content, list) else content
+    LOG.debug('Ran command "%s" with output "%s"' % cmd, cmd_output)
```
dzhengfy, line 3899: utils_misc.cmd_status_output(verbose=True) already supports to log the message your want when no session, so if you want to enhance this, it is better to add logging in utils_misc.cmd_status_output() when there is a  session passed in.

For the `LOG.debug(),` we need to pay attention to the usage of ',' instead of '%'. So it should be 
`LOG.debug('Ran command "%s" with output "%s"',  % (cmd, cmd_output))`

dzhengfy:
@vasekhodina please update your commit with your signature by using  'git commit -s '

virttest/utils_test/libvirt.py
dzhengfy: Sorry, I made a mistake in above comment for LOG.debug.
 It should be `LOG.debug('Ran command "%s" with output "%s"', cmd, cmd_output)`, right?

dzhengfy, line 3899: Sorry, I made a mistake in above comment for LOG.debug.
It should be LOG.debug('Ran command "%s" with output "%s"', cmd, cmd_output), right?

vasekhodina: Fixed.

vasekhodina: Fixed.

dzhengfy (approved):
LGTM

PR #3445, opened 2022-07-01
qemu_vm: modify/optimize the code to create qemu command in json style

nickzhq:
ID: 2048026

Signed-off-by: Houqi (Nick) Zuo <hzuo@redhat.com>

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -42,6 +42,7 @@ from virttest.qemu_devices import qdevices
 from virttest.qemu_devices.utils import (DeviceError, DeviceHotplugError,
                                          DeviceInsertError, DeviceRemoveError,
                                          DeviceUnplugError, none_or_int)
+from virttest.qemu_devices.utils import set_cmdline_format_by_cfg
 from virttest.utils_params import Params
 from virttest.qemu_capabilities import Flags, Capabilities, MigrationParams
 from virttest.utils_version import VersionInterval
@@ -188,10 +189,12 @@ class DevContainer(object):
         """:return: qemu version, e.g. 5.2.0"""
         return self.__qemu_ver
 
-    def initialize_iothread_manager(self, params, guestcpuinfo):
+    def initialize_iothread_manager(self, params, guestcpuinfo,
+                                    cmdline_format_cfg=None):
         """Initialize iothread manager.
         :param params: vt params
         :param guestcpuinfo: Cpuinfo object that stores guest cpu info
+        :param cmdline_format_cfg: command style in dict
         """
         iothreads_lst = params.objects("iothreads")
         iothread_scheme = params.get("iothread_scheme")
@@ -211,6 +214,7 @@ class DevContainer(object):
             }
             iothread = qdevices.QIOThread(iothread_id=iothread,
                                           params=iothread_dict)
+            set_cmdline_format_by_cfg(iothread, cmdline_format_cfg, "images")
             iothreads.append(iothread)
 
         scheme_to_manager = {
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -6,6 +6,7 @@ interact with qemu qdev structure.
 
 :copyright: 2012-2013 Red Hat Inc.
 """
+import json
 import logging
 import os
 import time
@@ -16,7 +17,7 @@ from pathlib import Path
 
 import aexpect
 
-from virttest import qemu_monitor
+from virttest import qemu_monitor, utils_numeric
 from virttest import utils_misc
 from virttest.qemu_devices.utils import DeviceError
 from virttest.qemu_devices.utils import none_or_int
@@ -88,6 +89,7 @@ class QBaseDevice(object):
                 self.add_child_bus(bus)
         self.dynamic_params = []
         self.params = OrderedDict()    # various device params (id, name, ...)
+        self.cmdline_format = "raw"
         if params:
             for key, value in six.iteritems(params):
                 if key == "pcie_direct_plug":
@@ -247,7 +249,20 @@ class QBaseDevice(object):
 
     def cmdline(self):
         """ :return: cmdline command to define this device """
-        raise NotImplementedError
+        _cmdline = {"json": self._cmdline_json,
+                    "raw": self._cmdline_raw}
+        if self.cmdline_format not in _cmdline:
+            raise ValueError("The input of qemu-kvm command format is NOT "
+                             "supported!")
+        return _cmdline.get(self.cmdline_format)()
+
+    def _cmdline_raw(self):
+        """ :return: cmdline command to define this device in raw format"""
+        return NotImplementedError
+
+    def _cmdline_json(self):
+        """ :return: cmdline command to define this device in json format"""
+        return NotImplementedError
 
     def cmdline_nd(self):
         """
@@ -417,8 +432,8 @@ class QCustomDevice(QBaseDevice):
         else:
             self.__backend = None
 
-    def cmdline(self):
-        """ :return: cmdline command to define this device """
+    def _cmdline_raw(self):
+        """ :return: cmdline command to define this device in raw format"""
         if self.__backend and self.params.get(self.__backend):
             out = "-%s %s," % (self.type, self.params.get(self.__backend))
             params = self.params.copy()
@@ -824,6 +839,12 @@ class QBlockdevNode(QCustomDevice):
         """ Get the node name instead of qemu id. """
         return self.params.get('node-name')
 
+    def _cmdline_json(self):
+        params = self.params.copy()
+        out = "-%s " % self.type
+        new_args = self._convert_blkdev_args(params)
+        return out + '\'' + json.dumps(new_args) + '\''
+
 
 class QBlockdevFormatNode(QBlockdevNode):
     """ New a format type blockdev node. """
@@ -1105,6 +1126,42 @@ class QDevice(QCustomDevice):
         else:
             return False
 
+    def _cmdline_json(self):
+        command_dict = {}
+        out = "-%s " % self.type
+
+        usb_storage_driver = self.params.get("driver") == "usb-storage"
+
+        expect_string_val = ["write-cache", "disable-legacy"]
+
+        for key, val in self.params.items():
+            # wwn needs to be presented as hexadecimal
+            if key in ["wwn"]:
+                command_dict[key] = int(val, 16)
+            # physical_block_size from device ("driver": "scsi-hd")
+            # logical_block_size from device ("driver": "scsi-hd")
+            # bootindex from device ("driver": "scsi-hd")
+            # max_sectors from device ("driver": "virtio-scsi-pci")
+            # num_queues from device ("driver": "virtio-scsi-pci")
+            # virtqueue_size from device ("driver": "virtio-scsi-pci")
+            elif key in ["physical_block_size", "logical_block_size",
+                         "bootindex", "max_sectors", "num_queues",
+                         "virtqueue_size"]:
+                command_dict[key] = int(val)
+            # port from device ("driver": "usb-storage")
+            elif key == "port" and usb_storage_driver:
+                command_dict[key] = str(val)
+            # disable-legacy from device ("driver": "virtio-scsi-pci")
+            # write-cache from device ("driver": "scsi-hd")
+            elif val in ('on', 'yes') and key not in expect_string_val:
+                command_dict[key] = True
+            elif val in ('off', 'no') and key not in expect_string_val:
+                command_dict[key] = False
+            else:
+                command_dict[key] = val
+
+        return out + '\'' + json.dumps(command_dict) + '\''
+
 
 class QGlobal(QBaseDevice):
 
@@ -1293,6 +1350,13 @@ class QObject(QCustomDevice):
     # pylint: disable=E0202
     def verify_hotplug(self, out, monitor):
         return len(out) == 0
+
+    def _cmdline_json(self):
+        command_dict = {}
+        out = "-%s " % self.type
+        params = self.params.copy()
+        command_dict["qom-type"] = params.pop("backend")
+        return out + '\'' + json.dumps(dict(command_dict, **params)) + '\''
 
 
 class QIOThread(QObject):
@@ -1503,6 +1567,29 @@ class Memory(QObject):
                             return False
                 return True
         return False
+
+    def _cmdline_json(self):
+        command_dict = {}
+        out = "-%s " % self.type
+        params = self.params.copy()
+        self._convert_memobj_args(params)
+        command_dict["qom-type"] = params.pop("backend")
+        return out + '\'' + json.dumps(dict(command_dict, **params)) + '\''
+
+    @staticmethod
+    def _convert_memobj_args(args):
+        """
+        Convert string to uint64,
+        and convert "size": "14336M"  to "size": 15032385536 (bytes)
+
+        :param args: Dictionary with the qmp parameters.
+        :type args: dict
+        :return: Converted args.
+        :rtype: dict
+        """
+        if "size" in args:
+            args["size"] = int(utils_numeric.normalize_data_size(args["size"],
+                                                                 "B"))
 
 
 class Dimm(QDevice):
--- a/virttest/qemu_devices/utils.py
+++ b/virttest/qemu_devices/utils.py
@@ -73,3 +73,17 @@ def none_or_int(value):
         return int(value)
     else:
         raise TypeError("This parameter has to be int or none")
+
+
+def set_cmdline_format_by_cfg(dev, config, key):
+    """
+    Mark the cmdline format based on the settings.
+
+    :param dev: The device needed to be marked
+    :param config: Dict with setting
+    :param key: The field of device
+    """
+    if key in config:
+        name = dev.type
+        if name in config[key]:
+            dev.cmdline_format = config[key][name]
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -49,7 +49,7 @@ from virttest import error_context
 from virttest import utils_vsock
 from virttest import error_event
 from virttest.qemu_devices import qdevices, qcontainer
-from virttest.qemu_devices.utils import DeviceError
+from virttest.qemu_devices.utils import DeviceError, set_cmdline_format_by_cfg
 from virttest.qemu_capabilities import Flags
 from virttest.utils_params import Params
 from virttest.utils_version import VersionInterval
@@ -224,6 +224,37 @@ class VM(virt_vm.BaseVM):
         self.start_monotonic_time = 0.0
         self.last_boot_index = 0
         self.last_driver_index = 0
+
+    def _get_cmdline_format_cfg(self):
+        """
+        get data from file or input from parameter and then convert data to dict
+        :return: style data in dict
+        """
+        def _file(filepath):
+            if not filepath:
+                raise ValueError("The filepath is empty!")
+            if not os.path.isabs(filepath):
+                vt_type = self.params.get("vt_type", "qemu")
+                filepath = data_dir.get_backend_cfg_path(vt_type, filepath)
+            with open(filepath, "r") as f:
+                content = f.read()
+            return content
+
+        def _string(content):
+            return content
+
+        def _default(dummy):
+            return "{}"
+
+        handler, value = self.params.get("qemu_cmdline_format_cfg", ":").split(
+            ":", 1)
+        get_func = {"file": _file,
+                    "string": _string,
+                    "": _default}
+        if handler not in get_func:
+            LOG.warning("Unknown qemu cmdline format config...ignoring!")
+            handler, value = "", ""
+        return json.loads(get_func.get(handler)(value))
 
     @property
     def vcpu_threads(self):
@@ -1027,6 +1058,8 @@ class VM(virt_vm.BaseVM):
                 backend_param = Params(backend_options)
                 dev = devices.memory_object_define_by_params(backend_param,
                                                              name)
+                set_cmdline_format_by_cfg(dev, self._get_cmdline_format_cfg(),
+                                          "mem_devs")
                 devs.append(dev)
             else:
                 if params.get("hugepage_path") \
@@ -2072,12 +2105,16 @@ class VM(virt_vm.BaseVM):
             devices.insert(devices.usb_by_params(usb_dev, usb_dev_params))
 
         # initialize iothread manager
-        devices.initialize_iothread_manager(params, self.cpuinfo)
+        devices.initialize_iothread_manager(params,
+                                            self.cpuinfo,
+                                            self._get_cmdline_format_cfg())
 
         # Add object throttle group
         for group in params.objects("throttle_groups"):
             group_params = params.object_params(group)
             dev = devices.throttle_group_define_by_params(group_params, group)
+            set_cmdline_format_by_cfg(dev, self._get_cmdline_format_cfg(),
+                                      "images")
             devices.insert(dev)
 
         # Add images (harddrives)
@@ -2121,6 +2158,8 @@ class VM(virt_vm.BaseVM):
                                                    image_bootindex,
                                                    pci_bus=parent_bus)
             for _ in devs:
+                set_cmdline_format_by_cfg(_, self._get_cmdline_format_cfg(),
+                                          "images")
                 devices.insert(_)
 
         # Add filesystems
@@ -2336,6 +2375,9 @@ class VM(virt_vm.BaseVM):
                                                        image_bootindex,
                                                        pci_bus=parent_bus)
                 for _ in devs:
+                    set_cmdline_format_by_cfg(_,
+                                              self._get_cmdline_format_cfg(),
+                                              "images")
                     devices.insert(_)
 
         add_floppy(devices, params)
--- a/virttest/shared/cfg/base.cfg
+++ b/virttest/shared/cfg/base.cfg
@@ -1001,3 +1001,13 @@ uuid_dimm = ""
 # image_filter_drivers = comp1
 # image_filter_driver_type_comp1 = compress
 # compress filter doesnt have params
+
+# The parameter qemu_cmdline_format_cfg is used to point expression
+# of special devices in qemu command based on libvirt version.
+# For now, there are 2 kinds of format supported: raw and json
+# This default setting is raw format( qemu_cmdline_format_cfg = string:{} )
+# It can be set by user in cfg file.
+# This parameter may support multi-type value, including path, json
+# Example:
+# qemu_cmdline_format_cfg = file:/home/user/libvirt-latest.json
+# qemu_cmdline_format_cfg = string:{"images":{"blockdev": "json"}}


nickzhq:
(1/1) Host_RHEL.m9.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.x86_64.io-github-autotest-qemu.blockdev_full_backup.during_reboot.with_data_plane.q35: STARTED
 (1/1) Host_RHEL.m9.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.x86_64.io-github-autotest-qemu.blockdev_full_backup.during_reboot.with_data_plane.q35: PASS (172.34 s)

virttest/qemu_devices/utils.py
```diff
@@ -73,3 +73,18 @@ def none_or_int(value):
...
+def mark_qemu_command_style(config, key, dev):
+    """
+    mark the qemu command style based on the settings
+    :param config - dict with setting
+    :param key - the field of device
+    :param dev - the device needed to be marked
+    :return dev - device with qemu command style
+    """
+    if key in config:
+        name = dev.cmdline().split(" ")[0].split("-")[1]
```
luckyh, line 87: Instead of getting the cmdline category by rendering the raw format cmdline every time, which I think is not that efficient, I'd suggest using other approaches, for example, via getting the class type.

virttest/qemu_devices/utils.py
```diff
@@ -73,3 +73,18 @@ def none_or_int(value):
         return int(value)
     else:
         raise TypeError("This parameter has to be int or none")
+
+
+def mark_qemu_command_style(config, key, dev):
```
luckyh, line 78: Then I'd understand this call as `set_cmdline_format_by_cfg`.

virttest/qemu_vm.py
```diff
@@ -223,6 +223,35 @@ def __init__(self, name, params, root_dir, address_cache, state=None):
         self.last_boot_index = 0
         self.last_driver_index = 0
 
+    def get_command_style(self):
```
luckyh, line 226: Suggest being a private method.

virttest/qemu_vm.py
```diff
@@ -223,6 +223,35 @@ def __init__(self, name, params, root_dir, address_cache, state=None):
...
+                res = json.loads(line)
+            return res
+
+        command_style_param = self.params.get("qemu_command_style")
+        flag, cont = command_style_param.split(":", 1)
+        sw = {"path": __path,
+              "json": __json}
+
+        if flag not in sw:
+            LOG.error("The format is NOT support! ")
```
luckyh, line 252: Should we also consider to either:
* fallback to use raw format?

or
* raise an `ValueError` accordingly?

virttest/qemu_devices/qcontainer.py
```diff
@@ -210,6 +212,9 @@ def initialize_iothread_manager(self, params, guestcpuinfo):
             }
             iothread = qdevices.QIOThread(iothread_id=iothread,
                                           params=iothread_dict)
+            iothread = mark_qemu_command_style(command_style,
+                                               "mem_devs",
```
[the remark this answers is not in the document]
nickzhq: > IMO, _iothread_ would more likely be a part of storage devices therefore using `"images"` seems to be more reasonable, what do you think?
Hello @luckyh , in fact, both object in mem_devs and object in images need to be marked. According to the context of this line, which is about the memory, I prefer to put "mem_devs" at here.  
Thanks!

virttest/qemu_devices/qdevices.py
```diff
@@ -87,6 +88,7 @@ def __init__(self, dev_type="QBaseDevice", params=None, aobject=None,
                 self.add_child_bus(bus)
         self.dynamic_params = []
         self.params = OrderedDict()    # various device params (id, name, ...)
+        self.qemu_command_style = "raw"  # the raw type is default value
```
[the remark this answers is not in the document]
nickzhq: > Suggest the attr name `cmdline_format`.

ok

virttest/qemu_devices/qdevices.py
```diff
@@ -418,21 +420,38 @@ def __init__(self, dev_type, params=None, aobject=None,
...
+                    out += "%s=%s," % (key, value)
+                else:
+                    out += "%s," % key
+            if out[-1] == ',':
+                out = out[:-1]
+            return out
+        elif self.qemu_command_style == "json":
+            return self.__json_format_line()
+
+    def __json_format_line(self):
```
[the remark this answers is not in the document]
nickzhq: > Suggest the method name `_cmdline_json`.

ok

nickzhq:
@luckyh Please check this patch again! But, I believe we should schedule a meeting for some comments if you are available tomorrow. Thanks!

nickzhq:
@luckyh @zhencliu  Please check this again, thanks!

nickzhq:
(1/1) Host_RHEL.m9.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.x86_64.io-github-autotest-qemu.blockdev_full_backup.during_reboot.with_data_plane.q35: STARTED
 (1/1) Host_RHEL.m9.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.x86_64.io-github-autotest-qemu.blockdev_full_backup.during_reboot.with_data_plane.q35: PASS (139.77 s)

nickzhq:
@luckyh @zhencliu Hello, guys! Please take a look, thanks!

virttest/qemu_devices/qdevices.py
```diff
@@ -1104,6 +1123,10 @@ def verify_hotplug(self, out, monitor):
         else:
             return False
 
+    def _cmdline_json(self):
```
zhencliu, line 1126: hi Nick, do you think this function can cover the following situation?
 -device {"driver":"scsi-hd","bus":"scsi0.0","channel":0,"scsi-id":0,"lun":0,"device_id":"drive-scsi0-0-0-0","drive":"libvirt-1-format","id":"scsi0-0-0-0","bootindex":1,"write-cache":"on"}
As you can see, there are also some 'digits' inline. If we set the param with a digit instead of a string, I should be OK here. Please make a double check, thanks

requirements.txt
```diff
@@ -5,4 +5,4 @@ simplejson>=3.5.3
 netaddr<=0.7.19; python_version < '3.0'
 netaddr>=0.7.19; python_version >= '3.0'
 zipp<=1.2.0; python_version < '3.0'
-netifaces>=0.10.5
+netifaces>=0.10.5
```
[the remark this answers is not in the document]
zhencliu: I guess you want to re-format the file by removing the last space? Then this maybe not related to the commit, if you want to change it, it's better to submit a new commit, let's keep it as it is in your commit

virttest/qemu_devices/qdevices.py
```diff
@@ -246,7 +248,18 @@ def get_children(self):
 
     def cmdline(self):
         """ :return: cmdline command to define this device """
-        raise NotImplementedError
+        if self.cmdline_format == "json":
+            return self._cmdline_json()
+        else:
+            return self._cmdline_raw()
```
zhencliu, line 254: do we want to check the exact string 'raw' here in else, and raise a ValueError exception in other cases?
then if there is a typo, e.g. 'jons', i.e. if we want json format, but give a typo, the cmd will be in raw format.
But, as the format is defined in templates, and it seems there is no test scenario to cover the format test, I think it's OK too.

virttest/qemu_devices/qdevices.py
```diff
@@ -459,6 +475,20 @@ def cmdline_nd(self):
...
+        Convert string type of 'on' and 'off' to boolean, and create new dict
+        (e.g: 'cache': {'direct': 'true'}) from key which include symbol '.'
+        (e.g: 'cache.direct': 'true') to adhere to the blockdev qmp syntax.
+
+        :param args: Dictionary with the qmp parameters.
+        :type args: dict
+        :return: Converted args.
+        :rtype: dict
+        """
+        return NotImplementedError
```
vivianQizhu, line 490: @nickzhq One quick comment, I think here should not be `NotImplementedError`, I suppose we should implement json format at the same level with raw format. We have to find out the regular pattern of json format among same type command line(so here in virttest they are in same QxxDevice class), and then implement them at a common level. If the current params structure can not fit the new format pattern, we need to twist params structure instead of implement  command line generator separately.
We could discuss it offline.

virttest/qemu_devices/qdevices.py
```diff
@@ -417,8 +433,8 @@ def __init__(self, dev_type, params=None, aobject=None,
         else:
             self.__backend = None
 
-    def cmdline(self):
-        """ :return: cmdline command to define this device """
+    def _cmdline_raw(self):
```
zhencliu, line 436: it seems we don't have a _cmdline_json implementation for class QObject nor its parent QCustomDevice.

jingzhao84:
Acked: jinzhao@redhat.com

All related test had been passed

Thanks
Jing

nickzhq:
@luckyh  please take a review, thanks!

virttest/qemu_vm.py
```diff
@@ -906,6 +936,8 @@ def add_pcidevice(devices, host, params, device_driver="pci-assign",
                        " It only support following parameter:\n %s" %
                        (", ".join(fail_param), pcidevice_help))
                 LOG.warn(msg)
+            set_cmdline_format_by_cfg(dev, self._get_cmdline_format_cfg(),
+                                      "images")
```
luckyh, line 940: The `pcidevice`/`vfio-pci` device is not a kind of images, could you please double confirm that?

virttest/qemu_vm.py
```diff
@@ -225,6 +225,36 @@ def __init__(self, name, params, root_dir, address_cache, state=None):
...
         self.last_driver_index = 0
 
+    def _get_cmdline_format_cfg(self):
+        """
+        get data from file or input from parameter and then convert data to dict
+        :return: style data in dict
+        """
+        def _file(filepath):
+            if not filepath:
+                raise ValueError(...)
```
luckyh, line 235: Wrong syntax here.

virttest/qemu_devices/qdevices.py
```diff
@@ -460,6 +476,20 @@ def cmdline_nd(self):
             out = out[:-1]
         return out
 
+    @staticmethod
+    def _convert_raw_style_to_json_style_args(args):
```
luckyh, line 480: Looks it is unnecessary to introduce this virtual prototype to the `QCustomDevice` so I'd suggest removing it.

virttest/qemu_devices/qdevices.py
```diff
@@ -729,17 +759,7 @@ def clear_child_nodes(self):
...
-        Convert string type of 'on' and 'off' to boolean, and create new dict
-        (e.g: 'cache': {'direct': 'true'}) from key which include symbol '.'
-        (e.g: 'cache.direct': 'true') to adhere to the blockdev qmp syntax.
-
-        :param args: Dictionary with the qmp parameters.
-        :type args: dict
-        :return: Converted args.
-        :rtype: dict
-        """
+    def _convert_raw_style_to_json_style_args(args):
```
luckyh, line 762: Suggest keeping the previous name until we have a better approach for that, because the actual procedure being called here is about parsing the given params to get desired arguments rather than doing the conversion from raw to json.

virttest/qemu_devices/qdevices.py
```diff
@@ -1105,6 +1131,45 @@ def verify_hotplug(self, out, monitor):
...
+            elif key == "port":
+                command_dict[key] = str(val)
+            else:
+                command_dict[key] = val
+
+        command_dict = self._convert_raw_style_to_json_style_args(command_dict)
+        return out + '\'' + json.dumps(command_dict) + '\''
+
+    @staticmethod
+    def _convert_raw_style_to_json_style_args(args):
```
luckyh, line 1161: Suggest squashing this method into `_cmdline_json`.

virttest/qemu_devices/utils.py
```diff
@@ -73,3 +73,16 @@ def none_or_int(value):
         return int(value)
     else:
         raise TypeError("This parameter has to be int or none")
+
+
+def set_cmdline_format_by_cfg(dev, config, key):
```
YongxueHong, line 78: Do we need to expose this function as a public API? I mean it makes sense for the level of the caller from the test case?
From my understanding, I think it just is useful for creating devices during calling `vm.create`, if I misunderstand, please point out it, thanks.
I suggest that we can assign it as a private API in the module `qcontainer`.

YongxueHong:
Please remove the period at the end of the commit header.
Thanks.

virttest/qemu_vm.py
nickzhq: > The `pcidevice`/`vfio-pci` device is not a kind of images, could you please double confirm that?

After I check the xml file, I may find "controller". I'm NOT sure which I use at here, "images" or "controller"?

virttest/qemu_devices/utils.py
nickzhq: > Do we need to expose this function as a public API? I mean it makes sense for the level of the caller from the test case? From my understanding, I think it just is useful for creating devices during calling `vm.create`, if I misunderstand, please point out it, thanks. I suggest that we can assign it as a private API in the module `qcontainer`.

From my opinion, the function may belong to the module `utils` ( This function has been used at `qemu_vm.py` and `qcontainer.py` ). Obviously, it's a shared function.  
In addition, once this function was declared as private, how other developers call it at other .py file?

YongxueHong: > > Do we need to expose this function as a public API? I mean it makes sense for the level of the caller from the test case? From my understanding, I think it just is useful for creating devices during calling `vm.create`, if I misunderstand, please point out it, thanks. I suggest that we can assign it as a private API in the module `qcontainer`.
> 
> From my opinion, the function may belong to the module `utils` ( This function has been used at `qemu_vm.py` and `qcontainer.py` ). Obviously, it's a shared function. In addition, once this function was declared as private, how other developers call it at other .py file?

Yes, I am afraid that other .py file call it in an invalid way since this function will change the `dev.cmdline_format`, I think it can not be exposed to the case level at least.

nickzhq: > > > Do we need to expose this function as a public API? I mean it makes sense for the level of the caller from the test case? From my understanding, I think it just is useful for creating devices during calling `vm.create`, if I misunderstand, please point out it, thanks. I suggest that we can assign it as a private API in the module `qcontainer`.
> > 
> > 
> > From my opinion, the function may belong to the module `utils` ( This function has been used at `qemu_vm.py` and `qcontainer.py` ). Obviously, it's a shared function. In addition, once this function was declared as private, how other developers call it at other .py file?
> 
> Yes, I am afraid that other .py file call it in an invalid way since this function will change the `dev.cmdline_format`, I think it can not be exposed to the case level at least.

Basically, the 'dev' is an argument passed obviously by user. And if user uses this function, the user has to get the 'dev' first. 
In case level, if user really use this function, I believe that's a necessary test step.

YongxueHong: > > > > Do we need to expose this function as a public API? I mean it makes sense for the level of the caller from the test case? From my understanding, I think it just is useful for creating devices during calling `vm.create`, if I misunderstand, please point out it, thanks. I suggest that we can assign it as a private API in the module `qcontainer`.
> > > 
> > > 
> > > From my opinion, the function may belong to the module `utils` ( This function has been used at `qemu_vm.py` and `qcontainer.py` ). Obviously, it's a shared function. In addition, once this function was declared as private, how other developers call it at other .py file?
> > 
> > 
> > Yes, I am afraid that other .py file call it in an invalid way since this function will change the `dev.cmdline_format`, I think it can not be exposed to the case level at least.
> 
> Basically, the 'dev' is an argument passed obviously by user. And if user uses this function, the user has to get the 'dev' first. In case level, if user really use this function, I believe that's a necessary test step.

Yes, I agree with you but what kind of scenario needs the user call this function?

nickzhq: > > > > > Do we need to expose this function as a public API? I mean it makes sense for the level of the caller from the test case? From my understanding, I think it just is useful for creating devices during calling `vm.create`, if I misunderstand, please point out it, thanks. I suggest that we can assign it as a private API in the module `qcontainer`.
> > > > 
> > > > 
> > > > From my opinion, the function may belong to the module `utils` ( This function has been used at `qemu_vm.py` and `qcontainer.py` ). Obviously, it's a shared function. In addition, once this function was declared as private, how other developers call it at other .py file?
> > > 
> > > 
> > > Yes, I am afraid that other .py file call it in an invalid way since this function will change the `dev.cmdline_format`, I think it can not be exposed to the case level at least.
> > 
> > 
> > Basically, the 'dev' is an argument passed obviously by user. And if user uses this function, the user has to get the 'dev' first. In case level, if user really use this function, I believe that's a necessary test step.
> 
> Yes, I agree with you but what kind of scenario needs the user call this function?

It depends on owner, I'm NOT sure which scenario calls this function.
On top of that, this function can't be declared to private in the 'qcontainer.py' since both the 'qemu_vm.py' and the 'qcontainer.py' call this function.

virttest/qemu_devices/qdevices.py
```diff
@@ -460,6 +475,10 @@ def cmdline_nd(self):
             out = out[:-1]
         return out
 
+    @staticmethod
+    def _convert_blkdev_args(args):
+        return NotImplementedError
+
```
luckyh, line 481: Suggest removing this declaration as blkdev is irrelevant to `QCustomDevice`.

virttest/qemu_devices/qdevices.py
```diff
@@ -247,7 +249,21 @@ def get_children(self):
...
     def cmdline(self):
         """ :return: cmdline command to define this device """
-        raise NotImplementedError
+        if self.cmdline_format == "json":
+            return self._cmdline_json()
+        elif self.cmdline_format == "raw":
+            return self._cmdline_raw()
+        else:
+            raise ValueError("The input of qemu-kvm command format is NOT "
+                             "supported!")
```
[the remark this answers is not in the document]
luckyh: For several reasons, we should avoid to store the static outcome of the command line into the local variable. Thus I'd suggest as the following.
```python
_cmdline = {"json": self._cmdline_json, "raw": self._cmdline_raw}
if self.cmdline_format not in _cmdline:
    ...
return _cmdline.get(self.cmdline_format)()
```

nickzhq:
@luckyh Please check again, thanks!

virttest/qemu_devices/qdevices.py
```diff
@@ -1504,6 +1568,36 @@ def verify_hotplug(self, out, monitor):
...
+        (e.g: "size": "14336M" ) from old dict(e.g: "size": 15032385536)
+        to qmp syntax.
+
+        :param args: Dictionary with the qmp parameters.
+        :type args: dict
+        :return: Converted args.
+        :rtype: dict
+        """
+        new_args = dict()
+        for key, value in six.iteritems(args):
```
zhencliu, line 1593: hi Nick, as now python2 is not considered, here we can use args.items instead.

virttest/qemu_devices/qdevices.py
```diff
@@ -1105,6 +1116,35 @@ def verify_hotplug(self, out, monitor):
         else:
             return False
 
+    def _cmdline_json(self):
+        command_dict = {}
+        out = "-%s " % self.type
+        for key, val in six.iteritems(self.params):
```
zhencliu, line 1122: no need to consider python2 here

virttest/qemu_devices/qdevices.py
```diff
@@ -1105,6 +1116,35 @@ def verify_hotplug(self, out, monitor):
...
+            # physical_block_size from device ("driver": "scsi-hd")
+            # logical_block_size from device ("driver": "scsi-hd")
+            # bootindex from device ("driver": "scsi-hd")
+            # max_sectors from device ("driver": "virtio-scsi-pci")
+            # num_queues from device ("driver": "virtio-scsi-pci")
+            # virtqueue_size from device ("driver": "virtio-scsi-pci")
+            elif key in ["physical_block_size", "logical_block_size",
+                         "bootindex", "max_sectors", "num_queues",
+                         "virtqueue_size"]:
+                command_dict[key] = int(val.split(" ")[0])
```
zhencliu, line 1135: why we need a split(" ") here? the space chars should be removed and then assigned to cartesian param

virttest/qemu_devices/qdevices.py
```diff
@@ -1105,6 +1116,35 @@ def verify_hotplug(self, out, monitor):
...
+            # port from device ("driver": "usb-storage")
+            elif key == "port":
+                command_dict[key] = str(val)
```
zhencliu, line 1138: Do we want to add a check of usb-storage here? in case there is some other 'port' params defined, and a int number is required.

nickzhq: > Do we want to add a check of usb-storage here? in case there is some other 'port' params defined, and a int number is required.

So far, it's NOT necessary.
Based on the devices supported, those codes work well, which are confirmed by owner.

virttest/qemu_devices/qdevices.py
nickzhq: > why we need a split(" ") here? the space chars should be removed and then assigned to cartesian param

While I debugging the code, I had found some space chars left at param. That's why those code is here.
However, I've tried the code `command_dict[key] = int(val)` works fine as well.

nickzhq:
@YongxueHong @luckyh @zhencliu Please check again, thanks!

virttest/qemu_devices/qdevices.py
```diff
@@ -1504,6 +1573,36 @@ def verify_hotplug(self, out, monitor):
...
+        params = self._convert_memobj_args(params)
+        command_dict["qom-type"] = params["backend"]
+        del params["backend"]
+        return out + '\'' + json.dumps(dict(command_dict, **params)) + '\''
+
+    @staticmethod
+    def _convert_memobj_args(args):
+        """
+        Convert string type of string to uint64, and create new dict
+        (e.g: "size": "14336M" ) from old dict(e.g: "size": 15032385536)
```
zhencliu, line 1589: Refer to your code below, I guessed you'd like to convert 14366M to 15032385536 (Bytes)

virttest/qemu_devices/qdevices.py
```diff
@@ -1504,6 +1573,36 @@ def verify_hotplug(self, out, monitor):
...
+        :rtype: dict
+        """
+        new_args = dict()
+        for key, value in args.items():
+            if key == "size":
+                new_args[key] = int(utils_numeric.normalize_data_size(value,
+                                                                      "B"))
+            else:
+                new_args[key] = value
+        return new_args
```
zhencliu, line 1604: if 'size' in args:
    args['size'] = int(utils_numeric.normalize_data_size(args['size'], B)) ?

So we don't need a loop copy of other items?

zhencliu: Sorry that I lost this part in my previous review, anyway, yours is still OK for me.

nickzhq:
@zhencliu @YongxueHong @luckyh  Please check again, thanks!

nickzhq:
@YongxueHong @zhencliu @luckyh  Please check again, thanks!

virttest/qemu_vm.py
```diff
@@ -225,6 +225,36 @@ def __init__(self, name, params, root_dir, address_cache, state=None):
...
+    def _get_cmdline_format_cfg(self):
+        """
+        get data from file or input from parameter and then convert data to dict
+        :return: style data in dict
+        """
+        def _file(filepath):
+            if not filepath:
+                raise ValueError("The filepath is empty!")
+            if not os.path.isabs(filepath):
+                filepath = os.path.join(data_dir.get_data_dir(), filepath)
```
yanan-fu, line 237: I guess the conf file are saved and backend not data dir:  `data_dir.get_backend_cfg_path('qemu', filepath)`.

nickzhq:
Hello @yanan-fu , please check again, thanks!

yanan-fu:
> Hello @yanan-fu , please check again, thanks!

The update with get the expected filepath LGTM, thanks!

zhencliu (approved):

virttest/qemu_vm.py
```diff
@@ -225,6 +225,37 @@ def __init__(self, name, params, root_dir, address_cache, state=None):
...
+                content = f.read()
+            return content
+
+        def _string(content):
+            return content
+
+        def _default(dummy):
+            return "{}"
+
+        handler, value = self.params.get("qemu_cmdline_format_cfg", ":").split(
```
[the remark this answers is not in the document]
nickzhq: > Please add the docstring for this parameter `qemu_cmdline_format_cfg` in the base.cfg https://github.com/avocado-framework/avocado-vt/blob/master/virttest/shared/cfg/base.cfg

Hello @YongxueHong , the `qemu_cmdline_format_cfg` was added at Kar project instead of avocado-vt project.  
Thanks!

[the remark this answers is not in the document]
YongxueHong: Hi @nickzhq 
yeah, I saw that, but we are adding this function in the avocado-vt project which is also an upstream project, so as to should give the docstring about it for the upstream.

virttest/shared/cfg/base.cfg
```diff
@@ -1001,3 +1001,14 @@ uuid_dimm = ""
...
+# qemu_cmdline_format_cfg
+# The parameter is used to point expression of special devices in
+# qemu command based on libvirt version.
+# For now, there are 2 kinds of format supported: raw and json
+# This default setting is raw format( qemu_cmdline_format_cfg = string:{} )
+# It can be set by user in cfg file or in kar command line.
+# This parameter may support multi-type value, including path, json
+# Example:
+#   qemu_cmdline_format_cfg = file:/home/user/libvirt-latest.json
+#   qemu_cmdline_format_cfg = string:{"images":{"blockdev": "json"}}
```
YongxueHong, line 1014:
```suggestion
# qemu_cmdline_format_cfg = file:/home/user/libvirt-latest.json
# qemu_cmdline_format_cfg = string:{"images":{"blockdev": "json"}}
```

virttest/shared/cfg/base.cfg
```diff
@@ -1001,3 +1001,14 @@ uuid_dimm = ""
 # image_filter_drivers = comp1
 # image_filter_driver_type_comp1 = compress
 # compress filter doesnt have params
+
+# qemu_cmdline_format_cfg
+# The parameter is used to point expression of special devices in
+# qemu command based on libvirt version.
+# For now, there are 2 kinds of format supported: raw and json
+# This default setting is raw format( qemu_cmdline_format_cfg = string:{} )
+# It can be set by user in cfg file or in kar command line.
```
YongxueHong, line 1010:
```suggestion
# It can be set by user in cfg file.
```
The keyword `kar` does not make sense for upstream.

virttest/shared/cfg/base.cfg
```diff
@@ -1001,3 +1001,14 @@ uuid_dimm = ""
 # image_filter_drivers = comp1
 # image_filter_driver_type_comp1 = compress
 # compress filter doesnt have params
+
+# qemu_cmdline_format_cfg
```
YongxueHong, line 1005:
```suggestion
```

virttest/shared/cfg/base.cfg
```diff
@@ -1001,3 +1001,14 @@ uuid_dimm = ""
...
+# The parameter is used to point expression of special devices in
+# qemu command based on libvirt version.
```
YongxueHong, line 1007:
```suggestion
# The parameter qemu_cmdline_format_cfg is used to point expression 
# of special devices in qemu command based on libvirt version.
```

nickzhq:
@YongxueHong Please check again, thanks!

YongxueHong (approved):
LGTM.

zhencliu (approved):

luckyh (approved):

luckyh:
Thanks to all.

PR #3450, opened 2022-07-07
Remove the check for config object

cliping:
If config type is 'qemu', the config object should be
'LibvirtQemuConfig'. So need to remove this check.

Signed-off-by: cliping <lcheng@redhat.com>

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -3783,8 +3783,6 @@ def customize_libvirt_config(params,
             libvirtd.restart()
         obj_conf = target_conf
     else:
-        if not isinstance(config_object, utils_config.LibvirtConfigCommon):
-            return None
         # Handle local libvirtd
         config_object.restore()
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -3783,8 +3783,6 @@ def customize_libvirt_config(params,
             libvirtd.restart()
         obj_conf = target_conf
     else:
-        if not isinstance(config_object, utils_config.LibvirtConfigCommon):
```
dzhengfy, line 3786: class LibvirtQemuConfig(LibvirtConfigCommon) 
Above definition should support LibvirtQemuConfig object. Could you give more reasons for the removing?

cliping: Sorry, I rerun https://github.com/autotest/tp-libvirt/pull/4274 in new test env and found we don't need to remove this check. So close it.

PR #3463, opened 2022-07-15
add source file for create_reuse_snap function

nanli1:
Need source file before create snap with --reuse-external  option

Note:
`fedora_35 AVOCADO_SRC: SETUP:setup.py develop --user VT_TYPE:qemu `error not related my code ,Please help review

### DIFF
--- a/virttest/utils_libvirt/libvirt_disk.py
+++ b/virttest/utils_libvirt/libvirt_disk.py
@@ -227,7 +227,10 @@ def make_relative_path_backing_files(vm, pre_set_root_dir=None, origin_image=Non
 
 
 def create_reuse_external_snapshots(vm, pre_set_root_dir=None, skip_first_one=False,
-                                    disk_target="vda", disk_type="file", snapshot_chain_lenghth=4):
+                                    disk_target="vda", disk_type="file",
+                                    snapshot_chain_lenghth=4,
+                                    create_source_file=False, snap_extra='',
+                                    disk_format='qcow2'):
     """
     Create reuse external snapshots
 
@@ -236,6 +239,9 @@ def create_reuse_external_snapshots(vm, pre_set_root_dir=None, skip_first_one=Fa
     :param skip_first_one: whether skip first image file
     :param disk_target: disk target
     :param snapshot_chain_lenghth : snapshot length
+    :param create_source_file: the flag to prepare reuse source file or not
+    :param snap_extra: create snapshot extra option
+    :param disk_format: disk format
     :return: absolute root path of backing files and snapshot list
     """
     if pre_set_root_dir is None:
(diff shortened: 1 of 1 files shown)


nanli1:
@chunfuwen hi chunfu ,Please help review

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -258,11 +262,13 @@ def create_reuse_external_snapshots(vm, pre_set_root_dir=None, skip_first_one=Fa
         backing_file_path = os.path.join(root_dir, key)
         external_snap_shot = "%s/%s" % (backing_file_path, value)
         snapshot_external_disks.append(external_snap_shot)
+        if create_source_file:
+            libvirt.create_local_disk(disk_type, disk_format='qcow2', path=external_snap_shot)
```
chunfuwen, line 266: Do we really need add creating disk logic into this method? as the method name imply, it just aims to create snapshots. And here it also hardcode disk_format='qcow2'

nanli1: hi chunfu ,Thanks for your comment  ：D

1. I think we could add this scenario that create reuse snap with source file , because the function name is just create_reuse_external_snapshots, which means create reuse snapshots. @dzhengfy hi dzheng, could you please give some comments? :D
2. disk_format='qcow2' related , updated

dzhengfy (approved):
LGTM

virttest/utils_libvirt/libvirt_disk.py
dzhengfy: I think creating external snapshot with source file is a concrete and special case and could belong to the create_reuse_external_snapshots() function. Thanks.

chunfuwen: @dzhengfy 
@nanli1 
As per clean code practices suggestion, one function is better do one thing. Creating local file is separate operation to create external snapshots. If you want to achieve both, suggest create one separate function to wrap them, rather than pull lots of unrelated stuff as one function here.

chunfuwen (commented):
see inline comments

virttest/utils_libvirt/libvirt_disk.py
nanli1: @chunfuwen Thanks your comment chunfu, Here are some concerns.

1.  I don't think create reuse snapshots with local file is one separate operation.  The function's name `create_reuse_external_snapshots` , which means only aim to create reuse snapshots, If you insist on separating the create reuse snap with source file and no source file ,you may need to update the function name.
2. If I create a new function , I think these two function are almost same codes, Is that also kind of redundant?
3.

chunfuwen: @nanli1 we need have one concept how one big function is built up from small functions, rather than wrap all thins in one function as we move project forward in the long run.
In terms of this PR specifically, I am tolerate to approve it since your other PR need it urgently.

chunfuwen (approved):
LGTM

PR #3471, opened 2022-07-28
qemu_vm: Avoid overwrite existing virtio properties

PaulYuuu:
1. "virtio_dev_xxx" are global settings for virtio devices, but they will
    always be set if we define them, even if the device has set these
    properties. The property setting by the device should be higher than
the global one.
2. `virtio_dev_xxx` will set properties to whole virtio devices, but some
    devices cannot work well that property such as "vhost-user-fs-pci" +
    "iommu_platform", so the new parameter `virtio_dev_filter` can filter
    out some devices which we want to set those properties.

ID: 2111781

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -2760,8 +2760,10 @@ class VM(virt_vm.BaseVM):
             attr_info = [None, params["keyboard_layout"], None]
             add_qemu_option(devices, "k", [attr_info])
 
-        # Add options for all virtio devices
-        virtio_devices = filter(lambda x: re.search(r"(?:^virtio-)|(?:^vhost-)",
+        # Add options for virtio devices which match the pattern
+        virtio_dev_filter = params.get("virtio_dev_filter",
+                                       "(?:^virtio-)|(?:^vhost-)")
+        virtio_devices = filter(lambda x: re.search(virtio_dev_filter,
                                                     x.get_param('driver', '')),
                                 devices)
         for device in virtio_devices:
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -2778,7 +2778,7 @@ class VM(virt_vm.BaseVM):
                 "ats": params.get("virtio_dev_ats"),
                 "aer": params.get("virtio_dev_aer")}
             for key, value in properties_to_be_set.items():
-                if value and key in dev_properties:
+                if key not in device.params and value and key in dev_properties:
                     device.set_param(key, value)
 
         # set tag for pcic
--- a/virttest/shared/cfg/base.cfg
+++ b/virttest/shared/cfg/base.cfg
@@ -122,6 +122,10 @@ flexible_nic_index = no
 # Set guest kernel iommu option (iommu=off/force/pt e.g.)
 # guest_iommu_option = pt
 
+# Specify a filter to make virtio_dev_xxx params only handle those devices
+# The following filter will capture all devices except balloon and iommu devices
+# virtio_dev_filter = '^(?:(?:virtio-)|(?:vhost-))(?!(?:balloon)|(?:user))'
+
 # List of block device object names (whitespace separated)
 images = image1
 # List of block device object names with order (whitespace separated)


PaulYuuu:
Hello @vivianQizhu @yanan-fu @yiqianwei, would you help review? Thanks in advance.

PaulYuuu:
> Hello @vivianQizhu @yanan-fu @yiqianwei, would you help review? Thanks in advance.

Hey, any comments?

yanan-fu:
With this implementation, there are two different sets of logic that can change the same device configuration.  I think it's going to be a little confused.  But yes, the existing logic is for all virtio device, does not allow device specific setting.  
Could you please share more about the requirements ?  Why does it needs different setting for the device you want to change with device.params. Thanks!

PaulYuuu:
> With this implementation, there are two different sets of logic that can change the same device configuration. I think it's going to be a little confused. But yes, the existing logic is for all virtio device, does not allow device specific setting. Could you please share more about the requirements ? Why does it needs different setting for the device you want to change with device.params. Thanks!

OK, in one case, most virtio devices want to enable iommu_platform, and only one or a few devices cannot enable it because of the limitation, such as virtio-iommu, virtio-balloon. Actually the `iommu_platform` should not display in the device help output, but developers think this usage can avoid in the libvirt layer and it's a wrong usage, so not plan to correct it.
If we use `xxx_extra_params`, we need to set each device, but `virtio_dev_xxx` can set all virtio devices, as long as this parameter appears in the help output. So with this change, the virtio parameters can be controlled relatively flexibly, even if it's a workaround.

yanan-fu:
> OK, in one case, most virtio devices want to enable iommu_platform, and only one or a few devices cannot enable it because of the limitation, such as virtio-iommu, virtio-balloon. Actually the `iommu_platform` should not display in the device help output, but developers think this usage can avoid in the libvirt layer and it's a wrong usage, so not plan to correct it. 

Okay,  If those virtio devices which show iommu_platform in help info but does not support it actually, how do you think about defining blacklist for them ?

PaulYuuu:
> Okay, If those virtio devices which show iommu_platform in help info but does not support it actually, how do you think about defining blacklist for them ?

I thought about it before, but it's hard to maintain, I mean we don't know how many virtio devices will be introduced in the future, and what the support status of these properties of them.

yanan-fu:
> > Okay, If those virtio devices which show iommu_platform in help info but does not support it actually, how do you think about defining blacklist for them ?
> 
> I thought about it before, but it's hard to maintain, I mean we don't know how many virtio devices will be introduced in the future, and what the support status of these properties of them.

I think it is same for test case level to handle `xxx_extra_params`, when new virtio device be introduced and have the same problem, case need be updated accordingly IMO, and may be not only one cases.

PaulYuuu:
> I think it is same for test case level to handle `xxx_extra_params`, when new virtio device be introduced and have the same problem, case need be updated accordingly IMO, and may be not only one cases.

In case users use compiled qemu themselves, there has a lot of virtio devices, we hard to maintain them in a white/black list. This just provides a flexible method to control virtio_dev_xxx parameters, this change will keep the previous usage and support _extra_params as well. 

However, not many people/feature owners will set iommu-related parameters as a global setting.

PaulYuuu:
@yanan-fu updated according to our offline discussion, I think these are the minimal changes based on the conversation.

virttest/qemu_vm.py
```diff
@@ -2651,7 +2651,9 @@ def add_secure_guest_object(params):
             add_qemu_option(devices, "k", [attr_info])
 
         # Add options for all virtio devices
-        virtio_devices = filter(lambda x: re.search(r"(?:^virtio-)|(?:^vhost-)",
+        virtio_dev_filter = params.get("virtio_dev_filter",
+                                       "(?:^virtio-)|(?:^vhost-)")
+        virtio_devices = filter(lambda x: re.search(virtio_dev_filter,
```
yanan-fu, line 2656: In this way, user define a whitelist. However, i think it is more convenient to define a blacklist.  As I think user does not know all the devices (such as virtio-net that are defined in framework), but they know the devices that be used in the case but do not need the default virtio option.  What do you think about this ?

By the way,  it is better to change the the comment with L2653, not `all virtio devices` now :)

PaulYuuu:
Test:
```python
>>> import re
>>> devices = [{'dev': 'virtio-net-pci'}, {'dev': 'vhost-vsock'}, {'dev': 'pxb-pcie'}, {'dev': 'virtio-balloon-pci'}, {'dev': 'virtio-iommu-pci'}, {'dev': 'vhost-user-fs-pci'}]
>>> virtio_dev_filter = '^(?:(?:virtio-)|(?:vhost-))(?!(?:balloon)|(?:iommu))'
>>> list(filter(lambda x: re.search(virtio_dev_filter, x.get('dev')), devices))
[{'dev': 'virtio-net-pci'}, {'dev': 'vhost-vsock'}, {'dev': 'vhost-user-fs-pci'}]
>>> virtio_dev_filter = '^(?:(?:virtio-)|(?:vhost-))(?!(?:user)|(?:net))'
>>> list(filter(lambda x: re.search(virtio_dev_filter, x.get('dev')), devices))
[{'dev': 'vhost-vsock'}, {'dev': 'virtio-balloon-pci'}, {'dev': 'virtio-iommu-pci'}]
>>> virtio_dev_filter = '^(?:(?:virtio-)|(?:vhost-))(?:net)'
>>> list(filter(lambda x: re.search(virtio_dev_filter, x.get('dev')), devices))
[{'dev': 'virtio-net-pci'}]
```

virttest/qemu_vm.py
PaulYuuu: Sure, I already updated L2653, and after I rethought it, the filter can also define a blacklist, you can check the following test results.

virttest/qemu_vm.py
```diff
@@ -2693,7 +2695,7 @@ def add_secure_guest_object(params):
                 "ats": params.get("virtio_dev_ats"),
                 "aer": params.get("virtio_dev_aer")}
             for key, value in properties_to_be_set.items():
-                if value and key in dev_properties:
+                if key not in device.params and value and key in dev_properties:
```
yanan-fu, line 2698: @PaulYuuu  Do we need keep this one after introduce the customer `virtio_dev_filter`  ?

PaulYuuu: It's not a must but better to keep it to avoid wrong operations.

yanan-fu: I prefer to use one way to define the device options, same with what we discussed before introduce the filter :) 
Request for a correct `virtio_dev_filter` as the solution to define device options as user need.

PaulYuuu: Yes, this would be safer, but we cannot guarantee how users will use, because there 2 parameters that can be used together, but they should have primary and secondary order.

yanan-fu: okay, you convinced me. thanks!

PaulYuuu:
Re-push to fix the CI failure.

PaulYuuu:
Hello @vivianQizhu, could you also help to review?

vivianQizhu (approved):

PR #3478, opened 2022-08-01
Update "--nvram" to "--keep-nvram" during undefine

yalzhang:
For UEFI guest, we can not undefine it successfully without any nvram
options. With "--nvram", it will delete the nvram file when undefine the
guest. When we define the guest again, it will reset before normal
boot. As serail login will input something during the reset process, it
will trigger the guest wait for some manual intervention which cause auto
case to fail. Update "--nvram" to "--keep-nvram" to workaround it.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -296,13 +296,15 @@ class VM(virt_vm.BaseVM):
         """
         Undefine the VM.
         """
-        # If the current machine contains nvram, we have to set --nvram
+        # If the current machine contains nvram, we have to set --keep-nvram
+        # Adding "--keep-nvram" can undefine a vm with nvram successfully and
+        # also skip the reset process during next boot
         if self.params.get("vir_domain_undefine_nvram") == "yes":
             if options is None:
-                options = "--nvram"
+                options = "--keep-nvram"
             else:
-                if "--nvram" not in options:
-                    options += " --nvram"
+                if "nvram" not in options:
+                    options += " --keep-nvram"
         try:
             virsh.undefine(self.name, options=options, uri=self.connect_uri,
                            ignore_status=False)
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -687,9 +687,9 @@ class VMXML(VMXMLBase):
         nvram = any([os_attrs.get('os_firmware') == "efi", os_attrs.get('nvram')])
         if nvram:
             if options is None:
-                options = "--nvram"
-            if "--nvram" not in options:
-                options += " --nvram"
+                options = "--keep-nvram"
+            if "nvram" not in options:
+                options += "--keep-nvram"
 
         return virsh_instance.remove_domain(self.vm_name, options)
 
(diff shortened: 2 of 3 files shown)


chunfuwen:
Maybe we can try another alternative attempt to fix it from login part.
If I understand this issue correctly, it touch vm.wait_for_serial_login(timeout=timeout), which is defined in line 1372 of virttest/virt_vm.py
1) vm.wait_for_serail_login will  call serial_login in the same file virttest/virt_vm.py
```
    @error_context.context_aware
    def serial_login(self, timeout=LOGIN_TIMEOUT,
                     username=None, password=None, virtio=False):
        """
        Log into the guest via the serial console.
        If timeout expires while waiting for output from the guest (e.g. a
        password prompt or a shell prompt) -- fail.

        :param timeout: Time (seconds) before giving up logging into the guest.
        :param virtio: is a console virtio console (deprecated).
        :return: ConsoleSession instance.
        """
        error_context.context("Logging into '%s' via serial console." %
                              self.name)
        if not username:
            username = self.params.get("username", "")
        if not password:
            password = self.params.get("password", "")

        prompt = self.params.get("shell_prompt", r"[\#\$]\s*$")
        linesep = eval("'%s'" % self.params.get("shell_linesep", r"\n"))
        status_test_command = self.params.get("status_test_command", "")

        return self.console_manager.create_session(linesep,
                                                   status_test_command,
                                                   prompt,
                                                   username,
                                                   password,
                                                   timeout)

```
2)Actually self.console_manager.create_session will call create_session in vt_console.py
 ```
    def create_session(self, linesep, status_test_command,
                       prompt, username, password, timeout):
        """
        Return a console session with itself as the manager.

        :param linesep: line separator for console
        :param status_test_command: status test command for console
        :param prompt: console prompt pattern
        :param username: login username
        :param password: login password
        :param timeout: time (seconds) before giving up logging into the guest
        """
        if self._console is None:
            raise NoConsoleError("No console available.")
        self.__login(linesep, status_test_command, prompt,
                     username, password, timeout)
        return ConsoleSession(self)
```

3) so what we can do here is to add additional logic to handle system reset screen in self.__login
```
    @lock
    def __login(self, linesep, status_test_command,
                prompt, username, password, timeout):
        self._console.set_linesep(linesep)
        self._console.set_status_test_command(status_test_command)

```
        while True:
                 try:
                     match, text = self._console.read_until_last_line_matches(
                       r"\(or (press|type) Control-D to continue\):\s*$",
                      r"[Cc]reset screen xx",
                      timeout=timeout, internal_timeout=0.5)
                                 output += text
                      if match == 0:  # "Are you sure you want to press any key"
                          if debug:
                              LOG.debug("Got 'Press any key...', sending 'enter key'")
                          session.sendline("\n")
              except ExpectTimeoutError as error:
           
        except ExpectProcessTerminatedError as error:
            raise LoginProcessTerminatedError(error.status, error.output) from error
        remote.handle_prompts(self._console, username, password,
                              prompt, timeout)
```
        remote.handle_prompts(self._console, username, password,
                              prompt, timeout)
```

yalzhang:
https://github.com/avocado-framework/avocado-vt/pull/3482   fixed the issue in serial login, close this pr.

PR #3482, opened 2022-08-03
vt_console:Make sure console is responsive before trying to login

chloerh:
There could be other interactive interface shown on vm before
login prompt which we would like to skip. Making sure console is
responsive before sending any key to it could meet our needs.

### DIFF
--- a/virttest/vt_console.py
+++ b/virttest/vt_console.py
@@ -1,4 +1,5 @@
 import threading
+import time
 
 
 from aexpect import remote
@@ -18,6 +19,13 @@ class NoConsoleError(ConsoleError):
 class ConsoleBusyError(ConsoleError):
     """Raise when session tries to communicate with a console in use."""
 
+    pass
+
+
+class ConsoleNotResponsiveError(ConsoleError):
+    """
+    Raise when console is not responsive
+    """
     pass
 
 
@@ -60,6 +68,16 @@ class ConsoleManager(object):
                 prompt, username, password, timeout):
         self._console.set_linesep(linesep)
         self._console.set_status_test_command(status_test_command)
+
+        is_responsive = False
+        end_time = time.time() + timeout
+        while time.time() < end_time:
+            if self._console.is_responsive():
+                is_responsive = True
+                break
+        if not is_responsive:
+            raise ConsoleNotResponsiveError('Console is not responsive.')
+
         remote.handle_prompts(self._console, username, password,
                               prompt, timeout)
 
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -1401,6 +1401,9 @@ class BaseVM(object):
                     self.verify_alive()
                 time.sleep(0.5)
                 continue
+            except vt_console.ConsoleNotResponsiveError as why:
+                LOG.error(str(why))
+                time.sleep(1)
         else:
             raise remote.LoginTimeoutError('exceeded %s s timeout' %
                                            timeout)


yalzhang (approved):
Try several cases, the issue is fixed.

Yingshun (approved):

chloerh:
Test result:
Running

virttest/vt_console.py
```diff
@@ -44,6 +52,7 @@ class ConsoleManager(object):
     def __init__(self):
         self._console = None
         self.status_test_command = None
+        self.is_responsive = None
```
chunfuwen, line 55: I assume is_responsive is boolean variable, so self.is_responsive = False??

chunfuwen:
@luckyh 
It looks that tp-qemu also use vm.wait_for_serial_login(), so loop you for more visibility

virttest/vt_console.py
chloerh: I'd prefer None, which means it's unset. Thanks.

chunfuwen (approved):
LGTM

virttest/vt_console.py
```diff
@@ -60,6 +69,15 @@ def __login(self, linesep, status_test_command,
                 prompt, username, password, timeout):
         self._console.set_linesep(linesep)
         self._console.set_status_test_command(status_test_command)
+
+        end_time = time.time() + timeout
+        while time.time() < end_time:
+            if self._console.is_responsive():
+                self.is_responsive = True
```
luckyh, line 76: Hi @chloerh is there some reason to make this as an attribute of the `ConsoleManager` class?

chloerh: Hi @luckyh . My thought was there was going to be a flag to mark the responsive status of console, and it occurred to me that this flag might be used by other functions of console. Then I decided to put it to the attribute list.

Thanks for reviewing. Actually this was kind of a workaround. Have you guys met this 'Reseting System' issue(Press any key to stop resetting system)? We would very much appreciate your suggestions or any other solutions. It has been bothering us for at least 2 weeks. Thanks again.

luckyh: > Hi @luckyh . My thought was there was going to be a flag to mark the responsive status of console, and it occurred to me that this flag might be used by other functions of console. Then I decided to put it to the attribute list.

I see. However, I have the concerns below about that

* Because the `ConsoleManager` is the class that you're going to have the attribute associated with, I'm wondering if it would make sense about the meaning of "console-manager is responsive" (since IMO, console-managers should always be responsive once created). Therefore, I'd like to request a better definition to that attribute.
* Due to consoles may change the state of responsiveness from time to time, it's not suggested for setting a static value as the cache.
 
> Thanks for reviewing. Actually this was kind of a workaround. Have you guys met this 'Reseting System' issue(Press any key to stop resetting system)? We would very much appreciate your suggestions or any other solutions. It has been bothering us for at least 2 weeks. Thanks again.

Uh, I haven't heard about this issue from the internal qemu testing users so far. Btw, previously we have solved a very similar issue (needing a confirmation by user to continue to boot from cdrom) by adding a step to call `vm.send_key` at boot stage, but I'm not sure if that would help for the case you are handling.

chloerh: > > Hi @luckyh . My thought was there was going to be a flag to mark the responsive status of console, and it occurred to me that this flag might be used by other functions of console. Then I decided to put it to the attribute list.
> 
> I see. However, I have the concerns below about that
> 
> * Because the `ConsoleManager` is the class that you're going to have the attribute associated with, I'm wondering if it would make sense about the meaning of "console-manager is responsive" (since IMO, console-managers should always be responsive once created). Therefore, I'd like to request a better definition to that attribute.
> * Due to consoles may change the state of responsiveness from time to time, it's not suggested for setting a static value as the cache.
> 

You're right. Thanks for the detailed comment. I'll change it to a local variable.

> > Thanks for reviewing. Actually this was kind of a workaround. Have you guys met this 'Reseting System' issue(Press any key to stop resetting system)? We would very much appreciate your suggestions or any other solutions. It has been bothering us for at least 2 weeks. Thanks again.
> 
> Uh, I haven't heard about this issue from the internal qemu testing users so far. Btw, previously we have solved a very similar issue (needing a confirmation by user to continue to boot from cdrom) by adding a step to call `vm.send_key` at boot stage, but I'm not sure if that would help for the case you are handling.

It does sound similar. I'll look it up.

Thanks again. I'll make the update a little later.

luckyh (approved):
LGTM

chunfuwen:
Let's merge it and monitor whether it can help fix reset screen issue

PR #3486, opened 2022-08-05
libvirt_bios: remove for efi

dzhengfy:
Remove some items for efi to avoid conflicted configuration with test case

### DIFF
--- a/virttest/utils_libvirt/libvirt_bios.py
+++ b/virttest/utils_libvirt/libvirt_bios.py
@@ -1,0 +1,30 @@
+"""
+Libvirt BIOS related utilities.
+
+:copyright: 2022 Red Hat Inc.
+"""
+
+import logging
+
+
+LOG = logging.getLogger('avocado.' + __name__)
+
+
+def remove_bootconfig_items_from_vmos(osxml):
+    """
+    Remove efi firmware attribute and loader/nvram elements
+
+    :param osxml: VMOSXML object
+    :return: VMOSXML, the updated object
+    """
+    # Remove efi firmware attribute and loader/nvram elements
+    # if they exist which may affect newly added same elements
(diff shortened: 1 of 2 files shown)


dzhengfy:
See test result in https://github.com/autotest/tp-libvirt/pull/4392#issuecomment-1206179304

meinaLi (approved):

virttest/utils_libvirt/libvirt_bios.py
```diff
@@ -0,0 +1,31 @@
...
+
+LOG = logging.getLogger('avocado.' + __name__)
+
+
+def remove_items_for_efi(vmxml, params):
+    """
+    Remove efi firmware attribute and loader/nvram elements
+
+    :param vmxml: vm xml
+    :param params: dict for test
```
chunfuwen, line 18: params parameter is never used in the code implementation

dzhengfy: Updated, thanks.

virttest/utils_libvirt/libvirt_bios.py
```diff
@@ -0,0 +1,30 @@
...
+:copyright: 2022 Red Hat Inc.
+"""
+
+import logging
+
+
+LOG = logging.getLogger('avocado.' + __name__)
+
+
+def remove_items_for_efi(vmxml):
```
chunfuwen, line 13: it looks like all deletion operates on  os_xml, so maybe passing os_xml looks a little natural,rather than vmxml.
And the method name could be remove_bootconfig_items_from_vmos ??

dzhengfy: updated

chunfuwen (approved):
LGTM

PR #3489, opened 2022-08-09
Fix a bug about order of importing config files

yolkfull:
As we support override of parameters defined in different config fils,
the parameters in config files imported later should override those
imported previously. But the parameters in subtests.cfg should take
effect anyway finally since they are the definite ones that we want
to test.

This patch fixes the bug that config of smp and memory defined in
subtests.cfg are overridden by guest-os.cfg by mistake.

Signed-off-by: Yingfu Zhou <yolkfull@hotmail.com>


lubinszARM:
cc

PR #3495, opened 2022-08-17
Fix the migration downtime parameter type and default value

pevogam:
Providing a python float even when converted to string will result
in a float number (with a decimal zero) to be passed to the qemu
monitor which will complain with

Error: Parameter 'downtime-limit' expects int64

and break the migration thus leading to incorrect system states
that were paused but cannot be continued and other harder to debug
issues down the line. It is better to convert to integer explicitly
here even if we multiply by a 1000.

Also use the correct default float downtime value which should be
300 ms (downtime-limit: 300 ms) when multiplied by 1000 and not
30 ms which is what the float would imply.


pevogam:
@YongxueHong @luckyh I would be curious to hear your opinion here.

luckyh (approved):
LGTM

YongxueHong (approved):
LGTM.

PR #3496, opened 2022-08-19
Add read permission to a bit of log files

PaulYuuu:
Let me take libvirtd.log as the example. There is an issue reported to
libvirt project: https://gitlab.com/libvirt/libvirt/-/issues/71, the use
case is people will use non-root user to analyze the libvirt instance
log file in another system, but by default the permission of libvirtd
log files are 0600, which reject other users to open the log for
analyzing.

For logs collection in testing, they should not contain confidential
information or not exposed to the outside, so add read permission for
all users should not a dangerous operation.

ID: 2119630

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -14,6 +14,7 @@ import resource
 
 from abc import ABCMeta
 from abc import abstractmethod
+from pathlib import Path
 
 from aexpect import remote
 
@@ -2462,11 +2463,13 @@ class LibvirtdDebugLog(object):
     """
     Enable libvirtd log for testcase in case
     with the use of param "enable_libvirtd_debug_log",
-    with additional params log level("libvirtd_debug_level")
-    and log file path("libvirtd_debug_file") can be controlled.
+    with additional params log level("libvirtd_debug_level"), log
+    filter("libvirtd_debug_filters") and log file
+    path("libvirtd_debug_file") can be controlled.
     """
 
-    def __init__(self, test, log_level="1", log_file="", log_filters="1:*"):
+    def __init__(self, test, log_level="1", log_file="", log_filters="1:*",
+                 log_permission=None):
         """
         initialize variables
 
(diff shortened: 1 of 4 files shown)


PaulYuuu:
Until libvirt/swtpm introduces a new option/knob to control the log permission, I would turn this as a workaround for log files.

PaulYuuu:
Hello @luckyh @chloerh, do you have any concerns about the log permission issue?

virttest/test_setup.py
```diff
@@ -2523,6 +2524,7 @@ def enable(self):
                                                 "libvirtd.log")
         # param used during libvirtd cleanup
         self.test.params["libvirtd_debug_file"] = self.log_file
+        Path(self.log_file).touch(mode=0o644, exist_ok=True)
```
chunfuwen, line 2527: Could we reuse current module method os.chmod(), rather than Path to achieve since os module is already imported?

chloerh:
Thanks @PaulYuuu . I'll invite a few colleagues who's more familiar with logs to take a look and then get back to you.

virttest/test_setup.py
PaulYuuu: To use `os.chmod()`, we should ensure the file exists. But before the daemon is restarted and collects log contents, we cannot assume it exists. A possible location is when restoring the daemon, https://github.com/avocado-framework/avocado-vt/blob/9dfc42808ac5ed2deab20f2b91c3bad5e126b75e/virttest/test_setup.py#L2545-L2549

But it seems no reason to custom log permission when disabling daemon, and LibvirtdDebugLog.enable in `preprocess` and `LibvirtdDebugLog.disable` in `postprocess`.

PaulYuuu:
Guys, after the update, any comments?

chunfuwen (approved):
Check tp-libvirt related cases, there is no permission check on libvirtd log file.
So the change will not impact current tp-libvirt
LGTM

luckyh (approved):
LGTM

PR #3510, opened 2022-09-05
Includes virtio-mem device support

mcasquer:
Includes virtio-mem device support and also
for virtio-mem option/attributes

ID: 2121669

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -2916,7 +2916,6 @@ class DevContainer(object):
             if key.endswith(suffix):
                 new_key = key.rsplit(suffix)[0]
                 dimm_params[new_key] = params[key]
-
         dimm_type = "nvdimm" if params.get("nv_backend") else "pc-dimm"
         attrs = qdevices.Dimm.__attributes__[dimm_type][:]
         dimm_uuid = dimm_params.get("uuid")
@@ -2935,22 +2934,43 @@ class DevContainer(object):
 
     def memory_define_by_params(self, params, name):
         """
-        Create memory modules by params, include memory object and
-        pc-dimm devices.
+        Create memory modules by params, include memory object,
+        pc-dimm, nvdimm and virtio-mem devices.
         """
         params = params.object_params(name)
         devices = []
-        if not self.has_device("pc-dimm"):
-            LOG.warn("'PC-DIMM' does not support by your qemu")
-            return devices
+        mem_devtype = params.get("vm_memdev_model", "dimm")
         mem = self.memory_object_define_by_params(params, name)
         if mem:
             devices.append(mem)
             use_mem = params.object_params(name).get("use_mem", "yes")
             if use_mem == "yes":
-                dimm = self.dimm_device_define_by_params(params, name)
-                dimm.set_param("memdev", mem.get_qid())
-                devices.append(dimm)
+                if mem_devtype == "dimm":
+                    dimm = self.dimm_device_define_by_params(params, name)
+                    dimm.set_param("memdev", mem.get_qid())
+                    devices.append(dimm)
+                elif mem_devtype == "virtio-mem":
+                    virtio_mem_params = Params()
+                    virtio_mem_bus = {'aobject': params.get('pci_bus', 'pci.0')}
+                    vmem_model = "virtio-mem-pci"
+                    suffix = "_memory"
+                    for key in list(params.keys()):
+                        if key.endswith(suffix):
+                            new_key = key.rsplit(suffix)[0]
+                            virtio_mem_params[new_key] = params[key]
+                    supported = ['any_layout', 'block-size', 'event_idx', 'indirect_desc', 'iommu_platform', 'memaddr', 'memdev', 'node', 'notify_on_empty',
+                                 'packed', 'prealloc', 'requested-size', 'size', 'unplugged-inaccessible', 'use-disabled-flag', 'use-started', 'x-disable-legacy-check']
+                    virtio_mem_params = virtio_mem_params.copy_from_keys(supported)
+                    if '-mmio:' in params.get("machine_type"):
+                        vmem_model = "virtio-mem-device"
+                        virtio_mem_bus = None
+                    if not self.has_device(vmem_model):
+                        raise exceptions.TestSkipError("%s device is not available" % vmem_model)
+                    virtio_mem = qdevices.QDevice(driver=vmem_model, parent_bus=virtio_mem_bus, params=virtio_mem_params)
+                    virtio_mem.set_param("id", "%s-%s" % ("virtio_mem", name))
+                    devices.append(virtio_mem)
+                else:
+                    raise DeviceError("Unsupported memory device type")
         return devices
 
     def input_define_by_params(self, params, name, bus=None):
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1000,14 +1000,13 @@ class VM(virt_vm.BaseVM):
 
             params["mem"] = str(int(mem_size_m))
             options.append(params["mem"])
-            if devices.has_device("pc-dimm"):
-                if mem_params.get("maxmem"):
-                    options.append("maxmem=%s" % mem_params["maxmem"])
-                    if mem_params.get("slots"):
-                        options.append("slots=%s" % mem_params["slots"])
-                for name in params.objects("mem_devs"):
-                    dev = devices.memory_define_by_params(params, name)
-                    devs.extend(dev)
+            if mem_params.get("maxmem"):
+                options.append("maxmem=%s" % mem_params["maxmem"])
+                if mem_params.get("slots"):
+                    options.append("slots=%s" % mem_params["slots"])
+            for name in params.objects("mem_devs"):
+                dev = devices.memory_define_by_params(params, name)
+                devs.extend(dev)
             machine_dev = devices.get_by_properties({"type": "machine"})[0]
             machine_cmd = machine_dev.cmdline_nd()
             output = re.findall(r",memory-backend=mem-([\w|-]+)", machine_cmd)
--- a/spell.ignore
+++ b/spell.ignore
@@ -483,6 +483,7 @@ mtab
 nodeset
 nr
 Numa
+nvdimm
 nvram
 passphrase
 pdb
--- a/virttest/shared/cfg/base.cfg
+++ b/virttest/shared/cfg/base.cfg
@@ -865,6 +865,16 @@ Linux:
 # Add nvdimm device and specify the backend
 # nv_backend = /tmp/nvdimm0
 
+# Include virtio-mem device instead of a dimm device
+# vm_memdev_model = "virtio-mem"
+# To make memory measurements effective with a virtio-mem device following kernel extra param needs to be included
+# kernel_extra_params_add = "memhp_default_state=online_movable"
+# Configure requested-size for a virtio-mem device
+# requested-size_memory_vmem0 = 1G
+# Define a specific numa node for a virtio-mem device
+# node_memory_vmem0 = "0"
+# All other virtio-mem device configurations apply the same rules as for pc-dimm
+
 # List of input device object names (whitespace separated)
 inputs = ""
 # Type of input device (currently supports mouse, keyboard and tablet)


mcasquer:
@luckyh @PaulYuuu could you review this PR? Thanks!

luckyh:
+@yanan-fu

vivianQizhu:
@menli820

PaulYuuu:
@zhenyzha

virttest/qemu_vm.py
```diff
@@ -2560,6 +2577,20 @@ def sort_key(dev):
...
+            options = {}
+            options["id"] = virtio_mem_device
+            requested_size = virtio_mem_params.get("virtio_mem_requested_size")
+            options["requested-size"] = requested_size
+            memdev = virtio_mem_params.get("virtio_mem_memdev")
+            options["memdev"] = memdev
+            node = virtio_mem_params.get("virtio_mem_node")
+            options["node"] = node
+            bus = virtio_mem_params.get("virtio_mem_bus")
+            options["bus"] = bus
```
PaulYuuu, line 2591: An important thing is: `bus`  should be set with `parent_bus=bus` in `QDevice` and it should not be defined by users, otherwise `QPCI(E)Bus` could not generate a correct address.

PaulYuuu (commented):
@mcasquer, the first thing, I think is to check if `virtio-mem` is a DIMM device or a new type of memory device, and then we can review accordingly based on the conclusion, please refer to the PDF from KVM forum or confirm it with someone.

virttest/qemu_devices/qcontainer.py
```diff
@@ -2916,8 +2916,11 @@ def dimm_device_define_by_params(self, params, name):
             if key.endswith(suffix):
                 new_key = key.rsplit(suffix)[0]
                 dimm_params[new_key] = params[key]
-
-        dimm_type = "nvdimm" if params.get("nv_backend") else "pc-dimm"
+        dimm_type = "pc-dimm"
+        if params.get("nv_backend"):
+            dimm_type = "nvdimm"
+        elif params.get("virtio_mem"):
+            dimm_type = "virtio-mem-pci"
```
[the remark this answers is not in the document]
mcasquer: Do you think it makes sense here to separate virtio-mem device in a different function?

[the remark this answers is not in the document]
PaulYuuu: I cannot provide an answer right now, if it's a DIMM device, we can put it here, but if not, then it should be in another function.

virttest/qemu_vm.py
mcasquer: @PaulYuuu so how can I set to the device which bus will I use?

PaulYuuu: https://github.com/avocado-framework/avocado-vt/blob/6e92a55fd4ce018df49d5bbf77bf994377ada0fd/virttest/qemu_vm.py?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L1468

https://github.com/avocado-framework/avocado-vt/blob/6e92a55fd4ce018df49d5bbf77bf994377ada0fd/virttest/qemu_vm.py?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L3758-L3787

mcasquer:
> @mcasquer, the first thing, I think is to check if `virtio-mem` is a DIMM device or a new type of memory device, and then we can review accordingly based on the conclusion, please refer to the PDF from KVM forum or confirm it with someone.

@PaulYuuu after reading the documents you shared, I consider it is not a DIMM device so I've finally separated from them, anyways I will wait for the developer to be back from KVM Forum and confirm this :smiley:

virttest/qemu_devices/qcontainer.py
```diff
@@ -2935,22 +2934,41 @@ def dimm_device_define_by_params(self, params, name):
...
             devices.append(mem)
             use_mem = params.object_params(name).get("use_mem", "yes")
+            use_virtio_mem = params.object_params(name).get("use_virtio_mem", "no")
             if use_mem == "yes":
                 dimm = self.dimm_device_define_by_params(params, name)
                 dimm.set_param("memdev", mem.get_qid())
                 devices.append(dimm)
+            if use_virtio_mem == "yes":
+                virtio_mem_params = Params()
+                suffix = "_virtio_mem"
```
PaulYuuu, line 2959: Keep my opinion, @luckyh @yanan-fu, you can give your point, if we want to align to `_dimm`?

virttest/qemu_vm.py
```diff
@@ -1000,7 +1000,7 @@ def add_memorys(devices, params):
 
             params["mem"] = str(int(mem_size_m))
             options.append(params["mem"])
-            if devices.has_device("pc-dimm"):
```
PaulYuuu, line 1003: Maybe this is also outdated...

virttest/qemu_devices/qcontainer.py
luckyh: That's a good point. My thought is we can consider to give a common suffix for all types of memory devices (e.g. `_memory` or `_memdev`), so that the related params being parsed in an unified way. And this time we can only apply it for virtio-mem, and leave the modification for dimm devices to the next stage.

mcasquer: @luckyh that's a good point! I can apply whatever (`_memory` or `_memdev`) if you all agree @yanan-fu @PaulYuuu :blush:

virttest/qemu_devices/qcontainer.py
```diff
@@ -2935,22 +2934,41 @@ def dimm_device_define_by_params(self, params, name):
...
+        pc-dimm and virtio-mem devices.
         """
         params = params.object_params(name)
         devices = []
         if not self.has_device("pc-dimm"):
             LOG.warn("'PC-DIMM' does not support by your qemu")
             return devices
+        if not self.has_device("virtio-mem"):
+            LOG.warn("'virtio-mem' does not support by your qemu")
+            return devices
```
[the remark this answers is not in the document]
mcasquer: @PaulYuuu  So can I delete it then?

mcasquer:
```bash
virttest/qemu_devices/qcontainer.py:2938:0: C0402: Wrong spelling of a word 'nvdimm' in a docstring:
        pc-dimm, nvdimm and virtio-mem devices.
                 ^^^^^^
```
@luckyh @PaulYuuu this seems to be a problem with the CI checks, that does not contains that word :confused:

PaulYuuu:
> ```shell
> virttest/qemu_devices/qcontainer.py:2938:0: C0402: Wrong spelling of a word 'nvdimm' in a docstring:
>         pc-dimm, nvdimm and virtio-mem devices.
>                  ^^^^^^
> ```
> 
> @luckyh @PaulYuuu this seems to be a problem with the CI checks, that do not contain that word 😕

So you need to add `nvdimm` word into `avocado-vt/spell.ignore` to ignore the spelling check.

virttest/qemu_devices/qcontainer.py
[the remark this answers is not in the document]
PaulYuuu: I think you can change to check the backend, if `no self.has_device(backend)` then return it. The backend can be `pc-dimm`, `nvdimm`, and `virtio-mem` IIRC.

virttest/qemu_devices/qcontainer.py
PaulYuuu: Yes, I prefer to pick `_memdev` as the suffix. But you can check if the usage exists.

mcasquer: @luckyh @PaulYuuu 
Actually `memdev` is in use on tp-qemu to make reference to the property `memdev=` from the numa node object, so this could lead to a conflict...

virttest/qemu_devices/qcontainer.py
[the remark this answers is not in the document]
PaulYuuu: Okay, I missed something. first I want to use `dimm_type = "nvdimm" if params.get("nv_backend") else "pc-dimm"` to judge the backend(type), but this line is only for DIMM devices, and has some limitations. So maybe we can introduce a new one like `memdev_type = dimm/virtio`, and then call the related function.

virttest/qemu_devices/qcontainer.py
PaulYuuu: OK, thank you. So let's have a try of `_memory`, it makes sense in `memory_define_by_params`

mcasquer: Great!

virttest/qemu_devices/qcontainer.py
```diff
@@ -2935,22 +2934,42 @@ def dimm_device_define_by_params(self, params, name):
...
             use_mem = params.object_params(name).get("use_mem", "yes")
+            use_virtio_mem = params.object_params(name).get("use_virtio_mem", "no")
             if use_mem == "yes":
                 dimm = self.dimm_device_define_by_params(params, name)
                 dimm.set_param("memdev", mem.get_qid())
                 devices.append(dimm)
+            if use_virtio_mem == "yes":
+                virtio_mem_params = Params()
+                virtio_mem_bus = {'aobject': params.get('pci_bus', 'pci.0')}
+                suffix = "_virtio_mem"
```
PaulYuuu, line 2960: Here is my example:
```python
mem_devtype = params.get("mem_devtype", "dimm") # Just an example
mem = self.memory_object_define_by_params(params, name)
if mem:
    devices.append(mem)
    use_mem = params.object_params(name).get("use_mem", "yes")
    if use_mem == "yes":
        if mem_devtype == "dimm":
            dimm = self.dimm_device_define_by_params(params, name)
            ......
            ......
            devices.append(dimm)
        elif mem_devtype == "virtio":
            vmem_model = "virtio-mem-pci"
            if "-mmio" in params.get("machine_type"):
                vmem_model = "virtio-mem-device"
            ......
            ......
            if not self.has_device(vmem_model):
                raise xxx
            virtio_mem = qdevices.QDevice(xxx)
            devices.append(virtio_mem)
        else:
           raise DeviceError("Unsupported memory device type")
```

mcasquer:
Seems the `_memory` suffix makes reference to the following `qxl_dev_memory` option: 

- https://gitlab.cee.redhat.com/kvm-qe/kar/-/blob/master/internal_cfg/base-internal.cfg#L71 

Including it on the qemu-kvm cmd line, which is not correct because virtio-mem device hasn't got it. So as confirmed with @PaulYuuu the framework now doesn't use it so it could be safely removed from kar repository or propose a new suffix name instead... what's your idea? :blush: @luckyh @yanan-fu

virttest/qemu_devices/qcontainer.py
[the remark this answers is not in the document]
mcasquer: @PaulYuuu ok got it, what about the `if not self.has_device("pc-dimm"):` is it really outdated too? Or I can just leave it as it is now

virttest/qemu_devices/qcontainer.py
mcasquer: @PaulYuuu thanks a lot for the example!

luckyh:
> Including it on the qemu-kvm cmd line, which is not correct because virtio-mem device hasn't got it... what's your idea?

Well, actually using suffixes to name the params which represent the attributes of classes had been proved to be a bad idea, resulting in naming conflicts would be the primary issue, as you've already seen. In fact, we should avoid to doing like so for the development of virtio-mem, but long story short, as we can consider it to be a legacy problem of the memdev stuff we don't have to solve it this time, and that's why I didn't mention that in the previous comments.

So, there are at least two approaches I'd prefer:

1. to restrict on what set of params can be taken for getting a virtio-mem object instance, just like the way of dimm devices being handled (by introducing a set of the supported options.)
2. to use prefix instead of suffix (e.g. `vm_memdev_<foo>`), because there must be no params having the exact same prefix the problem can be fixed completely. But we have to keep in mind that the routine of handling dimm devices is required to be modified accordingly in the further development, to align to the approach we are choosing.

Let me know your opinion, thanks!

mcasquer:
> So, there are at least two approaches I'd prefer:
> 
> 1. to restrict on what set of params can be taken for getting a virtio-mem object instance, just like the way of dimm devices being handled (by introducing a set of the supported options.)

@luckyh 
The first option sounds good to me, but I don't get how do you exactly handle a param to be specific for virtio-mem. Also if choosing this option, dimm devices code can remain untouched?

luckyh:
> @luckyh The first option sounds good to me, but I don't get how do you exactly handle a param to be specific for virtio-mem.

@mcasquer for example, by hardcoding a static list to provide all of the supported options' name of the virtio-mem device, like this:

```patch
diff --git a/virttest/qemu_devices/qcontainer.py b/virttest/qemu_devices/qcontainer.py
index f05d8b86..245ee61b 100644
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -2961,6 +2961,8 @@ class DevContainer(object):
                         if key.endswith(suffix):
                             new_key = key.rsplit(suffix)[0]
                             virtio_mem_params[new_key] = params[key]
+                    supported = [...]
+                    virtio_mem_params = virtio_mem_params.copy_from_keys(supported)
                     if '-mmio:' in params.get("machine_type"):
                         vmem_model = "virtio-mem-device"
                     if not self.has_device(vmem_model):
```

> Also if choosing this option, dimm devices code can remain untouched?

I think so, as what has been already addressed in [this thread](https://github.com/avocado-framework/avocado-vt/pull/3510#discussion_r994159139), we can treat it as a separated issue.

virttest/qemu_devices/qcontainer.py
```diff
@@ -2935,22 +2934,45 @@ def dimm_device_define_by_params(self, params, name):
...
             use_mem = params.object_params(name).get("use_mem", "yes")
             if use_mem == "yes":
-                dimm = self.dimm_device_define_by_params(params, name)
-                dimm.set_param("memdev", mem.get_qid())
-                devices.append(dimm)
+                if mem_devtype == "dimm":
+                    dimm = self.dimm_device_define_by_params(params, name)
+                    dimm.set_param("memdev", mem.get_qid())
+                    devices.append(dimm)
+                elif mem_devtype == "virtio":
```
luckyh, line 2955:
```suggestion
                elif mem_devtype == "virtio-mem":
```
since qemu has another virtio kind memdev - [virtio-pmem](https://qemu.readthedocs.io/en/latest/system/devices/virtio-pmem.html).

virttest/qemu_devices/qcontainer.py
```diff
@@ -2935,22 +2934,45 @@ def dimm_device_define_by_params(self, params, name):
...
-        pc-dimm devices.
+        Create memory modules by params, include memory object,
+        pc-dimm, nvdimm and virtio-mem devices.
         """
         params = params.object_params(name)
         devices = []
         if not self.has_device("pc-dimm"):
             LOG.warn("'PC-DIMM' does not support by your qemu")
             return devices
+        mem_devtype = params.get("mem_devtype", "dimm")
```
luckyh, line 2945: Suggest `vm_memdev_model` as the name for this new param, because:

1. to avoid the potential naming collisions it'd better to have `vm` as the namespace
2. to align to the naming style of nic devices (`nic_model`)

virttest/qemu_devices/qcontainer.py
```diff
@@ -2935,22 +2934,45 @@ def dimm_device_define_by_params(self, params, name):
...
+                    suffix = "_memory"
+                    for key in list(params.keys()):
+                        if key.endswith(suffix):
+                            new_key = key.rsplit(suffix)[0]
+                            virtio_mem_params[new_key] = params[key]
+                    supported = ['any_layout', 'block-size', 'event_idx', 'indirect_desc', 'iommu_platform', 'memaddr', 'memdev', 'node', 'notify_on_empty',
+                                 'packed', 'prealloc', 'requested-size', 'size', 'unplugged-inaccessible', 'use-disabled-flag', 'use-started', 'x-disable-legacy-check']
+                    virtio_mem_params = virtio_mem_params.copy_from_keys(supported)
+                    if '-mmio:' in params.get("machine_type"):
+                        vmem_model = "virtio-mem-device"
```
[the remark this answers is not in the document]
mcasquer: Included now, thanks!

virttest/qemu_devices/qcontainer.py
mcasquer: Included now, thanks!

virttest/qemu_devices/qcontainer.py
mcasquer: Included now, thanks!

luckyh (approved):
LGTM

virttest/qemu_devices/qcontainer.py
```diff
@@ -2935,22 +2934,46 @@ def dimm_device_define_by_params(self, params, name):
...
-        Create memory modules by params, include memory object and
-        pc-dimm devices.
+        Create memory modules by params, include memory object,
+        pc-dimm, nvdimm and virtio-mem devices.
         """
         params = params.object_params(name)
         devices = []
         if not self.has_device("pc-dimm"):
             LOG.warn("'PC-DIMM' does not support by your qemu")
             return devices
```
PaulYuuu, line 2944: Do we also need this now? Because we use `vm_memdev_model` to decide which one(dimm/virtio-mem) we want to use, it's so strange to check `pc-dimm` first and then attach a virtio-mem device. Another place need to be considered is https://github.com/avocado-framework/avocado-vt/blob/c7d21357b3a53295c0a5b819a1d1b81733e1796b/virttest/qemu_vm.py?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L1006-L1013

AFAICT, qemu support `pc-dimm` very early.

PaulYuuu:
Besides this, I think is time that feature owners can help to test and review the patch, and providing generated qemu command would be better.

virttest/qemu_devices/qcontainer.py
mcasquer: Well, I agree with removing pc-dimm checks

mcasquer: Changed!

mcasquer:
@zhenyzha @menli820 could you help on testing this PR? Thanks!

zhenyzha:
> @zhenyzha @menli820 could you help on testing this PR? Thanks!

@mcasquer hello I see tp-qemu has one case for testing virtio-mem.
https://github.com/autotest/tp-qemu/pull/3479/files

Command failed: Invalid CPU topology: maxcpus must be equal to or greater than smp
need add  `vcpu_maxcpus = ${smp}`

And unfortunately, this test requires two host numa and I need to deploy a new environment. update later.
On the other hand, we need to do some regression testing, it will take some time, update later.

mcasquer:
> @mcasquer hello I see tp-qemu has one case for testing virtio-mem.

Yes, and it should work but regarding the merging point of view it's still not 100% ready I will have to do some modifications :blush: 

> Command failed: Invalid CPU topology: maxcpus must be equal to or greater than smp
need add vcpu_maxcpus = ${smp}

That happened running the auto case from the virtio-mem tp-qemu PR?

zhenyzha:
> That happened running the auto case from the virtio-mem tp-qemu PR?

yes, The default value of vcpu_maxcpus may be different for each platform, so if you set `smp = 8` you'd better-set vcpu_maxcpus as well :)

mcasquer:
> yes, The default value of vcpu_maxcpus may be different for each platform, so if you set `smp = 8` you'd better-set vcpu_maxcpus as well :)

@zhenyzha Included now in [tp-qemu PR](https://github.com/autotest/tp-qemu/pull/3479/files#diff-1ad348c4b73c93aebea577c05447c7c89514de6880906e90cd9db21407b1782fR11) as I mentioned, the case works fine but it still needs changes :blush:

zhenyzha:
> @zhenyzha Included now in [tp-qemu PR](https://github.com/autotest/tp-qemu/pull/3479/files#diff-1ad348c4b73c93aebea577c05447c7c89514de6880906e90cd9db21407b1782fR11) as I mentioned, the case works fine but it still needs changes 😊

@mcasquer  Hello 
The regression test pass did not introduce new problems,
RESULTS    : PASS 215 | ERROR 6 | FAIL 0 | SKIP 4 | WARN 0 | INTERRUPT 0 | CANCEL 5

But the virtio_mem_numa_basic cases still fail, I don't see the successful generation of "virtio-mem-pci" in the cli cmd I will post the log in the bug, could you help me?

mcasquer:
> But the virtio_mem_numa_basic cases still fail, I don't see the successful generation of "virtio-mem-pci" in the cli cmd I will post the log in the bug, could you help me?

Sure!

menli820:
@mcasquer  hit the same issue as @zhenyzha  mentioned above,  for case 'virtio_mem_numa_basic cases ', no  '-device virtio-mem-pci ' generate,   I checked the log @zhenyzha pasted, it is the same, so not paste it again.

menli820:
Test on x86-64 platform.
@mcasquer  hit the same issue as @zhenyzha  mentioned above,  for case 'virtio_mem_numa_basic cases ', no  '-device virtio-mem-pci ' generate,   I checked the log @zhenyzha pasted, and it is the same, so not paste it again.

mcasquer:
> Test on x86-64 platform.
> @mcasquer hit the same issue as @zhenyzha mentioned above, for case 'virtio_mem_numa_basic cases ', no '-device virtio-mem-pci ' generate, I checked the log @zhenyzha pasted, and the error is the same, so not paste it again.

@menli820 the case has been updated, could you give one more try? Thanks! :blush:

zhenyzha:
` (1/1) Host_RHEL.m9.u2.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.virtio_mem_numa_basic.arm64-pci: FAIL: Error, requested-size: 0 does not match with plugged memory: 128`

This method can already be called correctly, FAIL is because the product takes longer to release on arm, and the avocado-vt part does not need to be changed.
Since the avocado-vt code has not changed after the regression test, so ack for the arm platform.

zhenyzha (approved):

PaulYuuu (approved):
Besides virtio-mem, could you also run a regression test of pc-dimm/nvdimm? We remove 1 check, so would be better to test those DIMM devices to avoid introducing a regression.

menli820:
> Besides virtio-mem, could you also run a regression test of pc-dimm/nvdimm? We remove 1 check, so would be better to test those DIMM devices to avoid introducing a regression.

Thanks for the suggesion, I am working on it, it will need some time to get the result, will update it later.

vivianQizhu:
> > Besides virtio-mem, could you also run a regression test of pc-dimm/nvdimm? We remove 1 check, so would be better to test those DIMM devices to avoid introducing a regression.
> 
> Thanks for the suggesion, I am working on it, it will need some time to get the result, will update it later.

@menli820 I believe this request is for @mcasquer, let's wait for @mcasquer's response :)

mcasquer:
No new bugs have been found during nvdimm + pc-dimm (mem_hotplug) regressions:

http://fileshare.englab.nay.redhat.com/pub/logs/mcasquer/rhel92_nvdimm/
http://fileshare.englab.nay.redhat.com/pub/logs/mcasquer/rhel92_mem_hotplug/

FYI @PaulYuuu

PaulYuuu:
ACK.

PaulYuuu:
Oh, @mcasquer, I forgot one thing, could you modify/add parameters in the `base.cfg` to let users know how to insert a virtio-mem device? Like `# vm_memdev_model = "virtio-mem"` and other comment lines.

PaulYuuu (approved):

luckyh (approved):

luckyh:
Thanks to all.

PR #3513, opened 2022-09-08
vm_xml: support new iothreadids element

dzhengfy:
In the past, there is only one attribute (id) supported in iothread, like
<iothread id="6"/>. But now thread_pool_min and thread_pool_max attributes
are added. And maybe more come in future. So old implementation can not meet
this requirement. This fix is to make this class extensible in future.

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -3697,20 +3697,20 @@ class VMIothreadidsXML(base.LibvirtXMLBase):
         """
         Convert a string to iothread tag and attributes.
         """
-        del index
-        del libvirtxml
-        return ('iothread', {'id': item})
+        if not isinstance(item, dict):
+            raise xcepts.LibvirtXMLError("Expected a dictionary of iothread "
+                                         "attributes, not a %s"
+                                         % str(item))
+        return ('iothread', dict(item))
 
     @staticmethod
     def marshal_to_iothreads(tag, attr_dict, index, libvirtxml):
         """
         Convert a iothread tag and attributes to a string.
         """
-        del index
-        del libvirtxml
         if tag != 'iothread':
             return None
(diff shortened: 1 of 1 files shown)


dzhengfy:
```
>>> ids_xml.iothread
[]
>>> ids_xml.setup_attrs(** {'iothread': [{'id': '2', 'thread-pool-max': '100', 'thread-pool-min': '10'}, {'id': '1'}]})
>>> ids_xml.fetch_attrs()
{'iothread': [{'id': '2', 'thread-pool-max': '100', 'thread-pool-min': '10'}, {'id': '1'}]}
>>> ids_xml.iothread
[{'id': '2', 'thread-pool-max': '100', 'thread-pool-min': '10'}, {'id': '1'}]
>>> 

```

chloerh (approved):

virttest/libvirt_xml/vm_xml.py
```diff
@@ -3697,20 +3697,20 @@ def marshal_from_iothreads(item, index, libvirtxml):
...
     def marshal_to_iothreads(tag, attr_dict, index, libvirtxml):
         """
         Convert a iothread tag and attributes to a string.
         """
-        del index
-        del libvirtxml
         if tag != 'iothread':
             return None
-        return attr_dict['id']
+        return dict(attr_dict)
```
chunfuwen, line 3713: libvirt/tests/src/virsh_cmd/domain/virsh_iothreadadd.py
there are some python files to depend on VMIothreadidsXML in tp-libvirt,
Could you check whether the change make it compatible with current usage in tp-libvirt ? since in my local attempt, there may not

dzhengfy: @chunfuwen , see above history. https://github.com/autotest/tp-libvirt/pull/4502 is for other cases' updates to align with this enhancement.

PR #3365, opened 2022-03-04
[WIP] migration: Support multiple hosts migration

YongxueHong:
Signed-off-by: Yongxue Hong <yhong@redhat.com>

PR #3515, opened 2022-09-08
Provide image specific parameters for each QemuImg instance

pevogam:
These are necessary for vms with multiple images and other cases where we would like to manipulate specific image and/or its snapshot.

### DIFF
--- a/virttest/storage.py
+++ b/virttest/storage.py
@@ -764,6 +764,7 @@ class QemuImg(object):
         :param root_dir: Base directory for relative filenames.
         :param tag: Image tag defined in parameter images.
         """
+        params = params.object_params(tag)
         self.params = params
         self.image_filename = get_image_filename(params, root_dir)
         self.image_format = params.get("image_format", "qcow2")


pevogam:
@luckyh @YongxueHong: I have a few more incoming PR-s regarding the qemu-img functionality that I am wondering whether to push now all at once or to wait for this and PR https://github.com/avocado-framework/avocado-vt/pull/3495 to be merged first. What do you think?

pevogam:
An example on how the above change is useful:
```
>>> from virttest.utils_params import Params
>>> from virttest.qemu_storage import QemuImg
>>>
>>> image="image1"
>>> state = "on_customize"
>>>
>>> params = Params()
>>> params["image_chain"] = f"snapshot {image}"
>>> params["image_name_snapshot"] = state
>>> params["image_format_snapshot"] = "qcow2"
>>> params["images_base_dir"] = "loc"
>>>
>>> params["image_name_image1"] = "test"
>>> qemu_img = QemuImg(params, params["images_base_dir"], "image1")
>>> qemu_img.image_filename
'loc/test.qcow2'
>>>
>>> qemu_img = QemuImg(params, params["images_base_dir"], "snapshot")
>>> qemu_img.image_filename
'loc/on_customize.qcow2'
```

pevogam:
Alright, in interest of saving time I would go ahead with all remaining pull requests:
- https://github.com/avocado-framework/avocado-vt/pull/3518
- https://github.com/avocado-framework/avocado-vt/pull/3519
- https://github.com/avocado-framework/avocado-vt/pull/3521

And the somewhat less related:
- https://github.com/avocado-framework/avocado-vt/pull/3520

I would really appreciate if anyone of you could take a look or suggest a maintainer that is responsible and/or knowledgeable about these areas of the code base. Thanks!

luckyh:
> Alright, in interest of saving time I would go ahead with all remaining pull requests:
> 
>     * [Wait for proper monitor status before sending cont to monitor #3518](https://github.com/avocado-framework/avocado-vt/pull/3518)
> 
>     * [Do not resume but verify paused status #3519](https://github.com/avocado-framework/avocado-vt/pull/3519)
> 
>     * [Add timeout and allow space containing paths for "qemu-img convert" #3521](https://github.com/avocado-framework/avocado-vt/pull/3521)
> 
> 
> And the somewhat less related:
> 
>     * [Make background errors nonfatal as they often register unrelated events #3520](https://github.com/avocado-framework/avocado-vt/pull/3520)
> 
> 
> I would really appreciate if anyone of you could take a look or suggest a maintainer that is responsible and/or knowledgeable about these areas of the code base. Thanks!

OK, I'll take a look on the above ones in the next couple weeks, thanks.

virttest/storage.py
```diff
@@ -764,6 +764,7 @@ def __init__(self, params, root_dir, tag):
         :param root_dir: Base directory for relative filenames.
         :param tag: Image tag defined in parameter images.
         """
+        params = params.object_params(tag)
```
luckyh, line 767: Actually some of the callers had already passed "image_params" to get the instances of the `QemuImg` class.

https://github.com/avocado-framework/avocado-vt/blob/6e92a55fd4ce018df49d5bbf77bf994377ada0fd/virttest/env_process.py#L744-L745
https://github.com/avocado-framework/avocado-vt/blob/6e92a55fd4ce018df49d5bbf77bf994377ada0fd/virttest/qemu_devices/qcontainer.py#L1080-L1083

Even though it seems to be harmless to convert to a "image_params" from itself via the `.object_params` method call, I'd still suggest not doing as so. Based on that I'd like to know about in which context(s) you found this issue, would you please elaborate more? thanks!

pevogam: > Actually some of the callers had already passed "image_params" to get the instances of the `QemuImg` class.

Hmmm, if some other methods are already passing `image_params` I would advise to simply do this once in the actual method with the current change instead of relying on each method doing this by themselves (and some then naturally forgetting to do so). What I mean is that calling `QemuImg(..., some_image)` already indicates that `some_image` will be the one to use and its parameters will be needed so no calling method needs to manually filter by suffix as some would obviously forget and it is a manual step that should instead be taken care of by the `QemuImg` constructor.

> Based on that I'd like to know about in which context(s) you found this issue, would you please elaborate more? thanks!

You can see my comment in https://github.com/avocado-framework/avocado-vt/pull/3515#issuecomment-1240535895 which will not be possible without this change. In particular, we can pass the test parameters and work with all images stated in `image_chain` just by specifying the image tag.

pevogam:
> @luckyh @YongxueHong: I have a few more incoming PR-s regarding the qemu-img functionality that I am wondering whether to push now all at once or to wait for this and PR #3495 to be merged first. What do you think?

Just to clarify, this pull request is not blocked by any other above, they are simply independent pull requests that followed this one.

virttest/storage.py
luckyh: @pevogam sorry for the delayed response.

> Hmmm, if some other methods are already passing `image_params` I would advise to simply do this once in the actual method with the current change instead of relying on each method doing this by themselves (and some then naturally forgetting to do so). What I mean is that calling `QemuImg(..., some_image)` already indicates that `some_image` will be the one to use and its parameters will be needed so no calling method needs to manually filter by suffix as some would obviously forget and it is a manual step that should instead be taken care of by the `QemuImg` constructor.

I see, however, I would advise to introduce a new method (like the below one) for handling the use case you mentioned. IMO, this approach would be more welcome by:

* all of the current callers, because the "calling convention" has not been changes.
* some potential improvement/enhancement to the `QemuImg` class, for instance I was wondering about decoupling its implementation from cartesian params.

```python
@classmethod
def from_params(cls, params, root_dir, tag):
    image_params = params.object_params(tag)
    return cls(image_params, root_dir, tag)
```

pevogam: Hi @luckyh and thanks for replying. I am still not convinced about your proposal and I think the reason is that you are likely not seeing the full picture of the comment https://github.com/avocado-framework/avocado-vt/pull/3515#issuecomment-1240535895 the way I wanted to portray it. Let me elaborate:

The above comment https://github.com/avocado-framework/avocado-vt/pull/3515#issuecomment-124053589 is the behavior we get when introducing the change of this PR while on the current master branch the same snippet will result in:
```
>>> from virttest.utils_params import Params
>>> from virttest.qemu_storage import QemuImg
>>> image = "image1"
>>> state = "on_customize"
>>> params = Params()
>>> params["image_chain"] = f"snapshot {image}"
>>> params["image_name_snapshot"] = state
>>> params["image_format_snapshot"] = "qcow2"
>>> params["images_base_dir"] = "loc"
>>> params["image_name_image1"] = "test"
>>> 
>>> qemu_img = QemuImg(params, params["images_base_dir"], "image1")
image_name parameter not set.
>>> qemu_img.image_filename
>>> qemu_img2 = QemuImg(params, params["images_base_dir"], "snapshot")
image_name parameter not set.
>>> qemu_img2.image_filename
```
What this means is that the current code does not work as intended and this pull request is rather fixing a bug rather than trying to make the parameters usage comfortable for a few callers out there. The behavior described in https://github.com/avocado-framework/avocado-vt/pull/3515#issuecomment-124053589 is in fact the right way we would expect the `QemuImg` constructor to work on the first place. Thus the parameter restriction is in fact necessary to fix this. If one restricts the parameters in advance like
```
>>> params2 = params.object_params("snapshot")
>>> qemu_img3 = QemuImg(params2, params2["images_base_dir"], "snapshot")
>>> qemu_img3.image_filename
'/mnt/local/intra2net-qa/loc/on_customize.qcow2'
```
then IMO this defeats the idea of passing "snapshot" to the QemuImg constructor on the first place. So on my side I would rather suggest we adapt all callers in Avocado VT rather than the other way around. Or do you think that I misunderstand the way the constructor was meant to be used? I got this perception from inspecting the constructor implementation IIRC.

Refactoring QemuImg not to use parameters at all (not sure if possible though) is a larger change that goes beyond this problem, which could be solved with much less considering the rest of the current code base.

luckyh: > Let me first say here that tradition or not assuming non-suffixed parameters are the "default" or "master" image is a dangerous proposition. It is much more explicit to simply make sure that each parametric object (here image) has its own values always properly stored via its suffix. In this way we won't have to worry about storing `abc` and `abc_image1` overwriting it due to unforeseen circumstances if we wanted to store the default image1's parameters without suffixes.

I totally agree with you at this point, but unfortunately the manner had already be chosen as so even before I started working on this project. A large effort would be took if we want to improve that, I guess (it would be a task about refactoring.)

> So I agree with your historical arguments and let's close this pull request unless I should simply modify it to fix this docstring at least? In this way it will be clear to readers of the API doc that image parameters are intended to be passed to the constructor and they won't have to search for use cases and historical accounts in order to figure this out. Does that sounds good?

Either of them sounds good to me, I mean, to close this or to correct the docstring. Actually, my team had the plan to refactor the entire image/storage management stuff in the future (the decoupling between `params` and `QemuImg` I mentioned in the previous comment may be one issue of that), which is a part of an epic about supporting test execution onto distributed systems (i.e. a cluster), see [this PR](https://github.com/avocado-framework/avocado-vt/pull/3365) as a glance if you are interested in this topic. So let's hope we could get a better readability and user experience after that.

luckyh: > > Let me first say here that tradition or not assuming non-suffixed parameters are the "default" or "master" image is a dangerous proposition. It is much more explicit to simply make sure that each parametric object (here image) has its own values always properly stored via its suffix. In this way we won't have to worry about storing `abc` and `abc_image1` overwriting it due to unforeseen circumstances if we wanted to store the default image1's parameters without suffixes.
> 
> I totally agree with you at this point, but unfortunately the manner had already be chosen as so even before I started working on this project. A large effort would be took if we want to improve that, I guess (it would be a task about refactoring.)
> 
> > So I agree with your historical arguments and let's close this pull request unless I should simply modify it to fix this docstring at least? In this way it will be clear to readers of the API doc that image parameters are intended to be passed to the constructor and they won't have to search for use cases and historical accounts in order to figure this out. Does that sounds good?
> 
> Either of them sounds good to me, I mean, to close this or to correct the docstring. Actually, my team had the plan to refactor the entire image/storage management stuff in the future (the decoupling between `params` and `QemuImg` I mentioned in the previous comment may be one issue of that), which is a part of an epic about supporting test execution onto distributed systems (i.e. a cluster), see [this PR](https://github.com/avocado-framework/avocado-vt/pull/3365) as a glance if you are interested in this topic. So let's hope we could get a better readability and user experience afterwards.

pevogam: > I totally agree with you at this point, but unfortunately the manner had already be chosen as so even before I started working on this project. A large effort would be took if we want to improve that, I guess (it would be a task about refactoring.)

Alright, then let's just close this pull request since a proper effort to improve the class is already underway and that might or might not include ridding of the complete params there (i.e. the current pull request won't change much).



> Actually, my team had the plan to refactor the entire image/storage management stuff in the future (the decoupling between `params` and `QemuImg` I mentioned in the previous comment may be one issue of that), which is a part of an epic about supporting test execution onto distributed systems (i.e. a cluster), see [this PR](https://github.com/avocado-framework/avocado-vt/pull/3365) as a glance if you are interested in this topic. So let's hope we could get a better readability and user experience afterwards.

This sounds pretty exciting actually and our Avocado I2N plugin already moves in the same direction. Glad to hear we make the same feature wishes! If I can be of any help let me know.

PR #3517, opened 2022-09-09
virsh.py:set event cmd as variable in start_get_event func

nanli1:
def start_get_event needs to update event_cmd as param

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -642,15 +642,15 @@ class EventNotFoundError(Exception):
 class EventTracker(object):
 
     @staticmethod
-    def start_get_event(vm_name):
+    def start_get_event(vm_name, event_cmd='event {} --all --loop'):
         """
         Use a virsh session with subcommand 'event' to catch events
-
         :param vm_name: name of the vm to be catched
+        :param event_cmd: cmd to check event
         :return: the virsh session with 'event'
         """
         virsh_session = aexpect.ShellSession(VIRSH_EXEC)
-        event_cmd = 'event %s --all --loop' % vm_name
+        event_cmd = event_cmd.format(vm_name)
         LOG.info('Sending "%s" to virsh shell', event_cmd)
         virsh_session.sendline(event_cmd)
         # Sometimes the output of session can't be gotten immediately,


virttest/virsh.py
```diff
@@ -642,15 +642,15 @@ def __str__(self):
...
         """
         Use a virsh session with subcommand 'event' to catch events
-
         :param vm_name: name of the vm to be catched
+        :param event_cmd: cmd to check event
         :return: the virsh session with 'event'
         """
         virsh_session = aexpect.ShellSession(VIRSH_EXEC)
-        event_cmd = 'event %s --all --loop' % vm_name
+        event_cmd = event_cmd % vm_name
```
Yingshun, line 653: I wonder if event_cmd need vm_name every time or not... Should it be event_cmd = event_cmd.format(vm_name)? What do you think?

nanli1: Hi yingshun ,Thanks for your patient. Could u please help to tell what's the difference `between "format" and "%s"` here. I find that if we use empty vm_name for "%s" type , we can also get ` event  --all --loop` @Yingshun

Yingshun: start_get_event() will be used by wait_event() and some virsh utilities will use this decorator, so I don't think it's reasonable to pass an empty vm_name here. The only difference between 'format' and '%' here is that there will no exception if no need to pass vm_name to 'event_cmd'.
For eg:
```
>>> event_cmd = "event --all --loop"
>>> my_vm = 'test1'
>>> event_cmd % my_vm
TypeError: not all arguments converted during string formatting
>>> event_cmd.format(my_vm)
'event --all --loop'
>>> 
```

virttest/virsh.py
```diff
@@ -642,15 +642,15 @@ def __str__(self):
 class EventTracker(object):
 
     @staticmethod
-    def start_get_event(vm_name):
+    def start_get_event(vm_name, event_cmd='event %s --all --loop'):
```
chunfuwen, line 645: event_cmd --> event_options, it looks like 'virsh event' is mandatory for any event action, and event_options e.g --all-loop is flexible, and will affect virsh event behaviour.

nanli1: Hi chunfu, Thanks  for your review. But we have met this event cmd ` "qemu-monitor-event vm_name --loop" `in some cases, so  'virsh event' may not be mandatory.

virttest/virsh.py
nanli1: updated yingshun,Thanks for your review. Did I get your point?

nanli1:
@chloerh hi haijiao,Could you please help re-review this pr ,This is related start_get_event and wait event :D

chloerh (approved):

Yingshun (approved):

chunfuwen (approved):
LGTM

chunfuwen:
merged it since all reviewer approve this request

PR #3518, opened 2022-09-09
Wait for proper monitor status before sending cont to monitor

pevogam:
It seems that if the vm is still in "paused (inmigrate)" status and similar, its monitor will completely ignore the "cont" monitor command and thus the vm will remain paused indefinitely, leading to later monitor validation no longer passing and breaking tests as a result.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -4863,7 +4863,13 @@ class VM(virt_vm.BaseVM):
         self.create(name=self.name, params=self.params, root_dir=self.root_dir,
                     timeout=self.MIGRATE_TIMEOUT, migration_mode="exec",
                     migration_exec_cmd="cat " + path, mac_source=self)
-        self.verify_status('paused')  # Throws exception if not
+
+        if utils_misc.wait_for(lambda: not self.monitor.verify_status('inmigrate'),
+                               timeout=self.MIGRATE_TIMEOUT, step=1):
+            self.verify_status('paused')  # Throws exception if not
+        else:
+            LOG.error(f"Monitor status {self.monitor.get_status()} is still in migration")
+            raise ValueError("Still paused inmigrate monitor status")
 
     def savevm(self, tag_name):
         """
@@ -4895,6 +4901,7 @@ class VM(virt_vm.BaseVM):
         """
         Resume the VM operation in case it's stopped.
         """
+        # NOTE: inmigrate, prelaunch, etc. might not register the "cont" command
         self.monitor.cmd("cont")
         if timeout:
             if not self.wait_for_status('running', timeout, step=0.1):
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -4839,16 +4839,20 @@ class VM(virt_vm.BaseVM):
         LOG.debug("Saving VM %s to %s" % (self.name, path))
         # Can only check status if background migration
         self.monitor.migrate("exec:cat>%s" % path, wait=False)
-        utils_misc.wait_for(
-            # no monitor.migrate-status method
-            lambda:
-            re.search("(status.*completed)",
-                      str(self.monitor.info("migrate")), re.M),
-            self.MIGRATE_TIMEOUT, 2, 2,
-            "Waiting for save to %s to complete" % path)
-        # Restore the speed and downtime to default values
-        qemu_migration.set_speed(self, str(32 << 20))
-        qemu_migration.set_downtime(self, 0.3)
+        try:
+            if not utils_misc.wait_for(
+                    # no monitor.migrate-status method
+                    lambda:
+                    re.search("(status.*completed)",
+                              str(self.monitor.info("migrate")), re.M),
+                    self.MIGRATE_TIMEOUT, 2, 2,
+                    "Waiting for save to %s to complete" % path):
+                raise virt_vm.VMMigrateTimeoutError("Timeout expired while waiting"
+                                                    " for saving to %s to finish" % path)
+        finally:
+            # Restore the speed and downtime to default values
+            qemu_migration.set_speed(self, str(32 << 20))
+            qemu_migration.set_downtime(self, 0.3)
         # Base class defines VM must be off after a save
         self.monitor.cmd("system_reset")
         self.verify_status('paused')  # Throws exception if not
--- a/spell.ignore
+++ b/spell.ignore
@@ -1533,6 +1533,7 @@ prebuild
 preconfig
 precopy
 Predefine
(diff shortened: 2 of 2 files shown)


pevogam:
The dictionary checker did not find the word for the output I actually receive from the qemu monitor side and thus does not depend on me, any advise where to add a new word for it?

luckyh:
> The dictionary checker did not find the word for the output I actually receive from the qemu monitor side and thus does not depend on me, any advise where to add a new word for it?

Can be added into [this file](https://github.com/avocado-framework/avocado-vt/blob/master/spell.ignore).

luckyh (changes requested):
@pevogam from my perspective, the check you're going to add should be handled by the migration routines (both synchronous and asynchronous ones), as it is waiting for migration being accomplished.

pevogam:
> @pevogam from my perspective, the check you're going to add should be handled by the migration routines (both synchronous and asynchronous ones), as it is waiting for migration being accomplished.

@luckyh, can you be a bit more specific which functionality are you referring to or at least confirm if what I propose below is right?

If it is the `qemu_migration` module then this module simply sets three parameters and most of the migration is still handled on the `qemu_vm` module side with `qemu_vm.VM` methods like `save_to_file()`, `restore_from_file()`, etc. so I assume not. Looking around it seems that the class also has methods like `wait_for_migration` and `migrate()` which handle some of the same situation slightly differently. I have an idea to move this to `restore_from_file()` function but:

1) I need https://github.com/avocado-framework/avocado-vt/pull/3519 merged in order to not include its changes here.
2) I need https://github.com/avocado-framework/avocado-vt/pull/3495 merged in order to not include fixes on migration parameters.

So I guess I won't be able to update this pull request before the other two.

pevogam:
Pull request updated now that previous dependencies have been merged, however I am not sure about the situation with the static checks. They complain about one overindented and one underindented lines but these lines are not part of the diff. Am I missing something obvious?

luckyh:
> Pull request updated now that previous dependencies have been merged, however I am not sure about the situation with the static checks. They complain about one overindented and one underindented lines but these lines are not part of the diff. Am I missing something obvious?

Seems that it indicated to the two places below (while the line numbers were not given correctly), you could have a check.

https://github.com/avocado-framework/avocado-vt/pull/3518/files#diff-274c35be00ea6170de3ec316e8ed20eca958a5de150a49770287fd1852a4e8feR4847
https://github.com/avocado-framework/avocado-vt/pull/3518/files#diff-274c35be00ea6170de3ec316e8ed20eca958a5de150a49770287fd1852a4e8feR4872

luckyh:
> @luckyh, can you be a bit more specific which functionality are you referring to or at least confirm if what I propose below is right?

Since I was not sure in which case/scenario you encountered into this issue, I could just give a very general conclusion about the resolution. It'd be helpful if you could elaborate more details, thanks.

> If it is the `qemu_migration` module then this module simply sets three parameters and most of the migration is still handled on the `qemu_vm` module side with `qemu_vm.VM` methods like `save_to_file()`, `restore_from_file()`, etc. so I assume not. Looking around it seems that the class also has methods like `wait_for_migration` and `migrate()` which handle some of the same situation slightly differently. I have an idea to move this to `restore_from_file()` function.

Yeah, to my understanding, `VM.migrate` has already handled this thing well, therefore `restore_from_file` might be the (only?) one in VT that you're interested.

virttest/qemu_vm.py
```diff
@@ -4839,16 +4839,20 @@ def save_to_file(self, path):
...
+                    re.search("(status.*completed)",
+                            str(self.monitor.info("migrate")), re.M),
+                    self.MIGRATE_TIMEOUT, 2, 2,
+                    "Waiting for save to %s to complete" % path):
+                raise virt_vm.VMMigrateTimeoutError("Timeout expired while waiting"
+                                                    " for saving to %s to finish" % path)
+        finally:
+            # Restore the speed and downtime to default values
+            qemu_migration.set_speed(self, str(32 << 20))
+            qemu_migration.set_downtime(self, 0.3)
```
luckyh: Suggest splitting this into a separated commit as it is another issue.

pevogam:
> > Pull request updated now that previous dependencies have been merged, however I am not sure about the situation with the static checks. They complain about one overindented and one underindented lines but these lines are not part of the diff. Am I missing something obvious?
> 
> Seems that it indicated to the two places below (while the line numbers were not given correctly), you could have a check.
> 
> https://github.com/avocado-framework/avocado-vt/pull/3518/files#diff-274c35be00ea6170de3ec316e8ed20eca958a5de150a49770287fd1852a4e8feR4847 https://github.com/avocado-framework/avocado-vt/pull/3518/files#diff-274c35be00ea6170de3ec316e8ed20eca958a5de150a49770287fd1852a4e8feR4872

Thanks, not sure why the line numbers were incorrect though. I tried to find them myself but thought the line numbers being outside the diff range meant something is borked (and still seems to be) in the CI.

> > @luckyh, can you be a bit more specific which functionality are you referring to or at least confirm if what I propose below is right?
> 
> Since I was not sure in which case/scenario you encountered into this issue, I could just give a very general conclusion about the resolution. It'd be helpful if you could elaborate more details, thanks.

Running (qemu) `vm.resume()` would send a `cont` to the qemu monitor which will be ignored if the status is `paused (inmigrate)` or similar and the `is_paused()` check will also return true because it will only loosely match the "paused" keyword above. So if I run the sequence `restore_from_file()` followed by `vm.resume()` this will lead to problems down the line.

> > If it is the `qemu_migration` module then this module simply sets three parameters and most of the migration is still handled on the `qemu_vm` module side with `qemu_vm.VM` methods like `save_to_file()`, `restore_from_file()`, etc. so I assume not. Looking around it seems that the class also has methods like `wait_for_migration` and `migrate()` which handle some of the same situation slightly differently. I have an idea to move this to `restore_from_file()` function.
> 
> Yeah, to my understanding, `VM.migrate` has already handled this thing well, therefore `restore_from_file` might be the (only?) one in VT that you're interested.

Yes, I mostly mimicked the `VM.migrate` in the implementation above so that it also applies to these methods which didn't seem to be updated back in the time the `VM.migrate` was.

Alright, I also split the commits as per your latest review.

virttest/qemu_vm.py
pevogam: Reusing the `migrate` logic some more seems rather on the same direction of changes to me but I will split them.

luckyh (approved):
LGTM

pevogam:
@YongxueHong Do you mind helping with a review? ^^

PR #3519, opened 2022-09-09
Do not resume but verify paused status

pevogam:
This is more compatible behavior with the implementation of the inverse method save_to_file() and with it both methods will now only verify paused status before and after they execute.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -4857,7 +4857,7 @@ class VM(virt_vm.BaseVM):
         self.create(name=self.name, params=self.params, root_dir=self.root_dir,
                     timeout=self.MIGRATE_TIMEOUT, migration_mode="exec",
                     migration_exec_cmd="cat " + path, mac_source=self)
-        self.resume()
+        self.verify_status('paused')  # Throws exception if not
 
     def savevm(self, tag_name):
         """


pevogam:
> This pull request **introduces 1 alert** when merging [b2a267d](https://github.com/avocado-framework/avocado-vt/commit/b2a267d2412911688feca72da4af27699766e3b3) into [f34a8a1](https://github.com/avocado-framework/avocado-vt/commit/f34a8a1e485e3b5b93759d2e48e908452ff68e7a) - [view on LGTM.com](https://lgtm.com/projects/g/avocado-framework/avocado-vt/rev/pr-b889b09a62660c7f914dc004f2e4d1e765b8033e)
> 
> **new alerts:**
> 
>     * 1 for Signature mismatch in overriding method

Hmmm, is this issue caused by me? The overriding method assumes argument that I don't think I touched on my side.

virttest/qemu_vm.py
```diff
@@ -4857,7 +4857,7 @@ def restore_from_file(self, path):
         self.create(name=self.name, params=self.params, root_dir=self.root_dir,
                     timeout=self.MIGRATE_TIMEOUT, migration_mode="exec",
                     migration_exec_cmd="cat " + path, mac_source=self)
-        self.resume()
+        self.verify_status('paused')  # Throws exception if not
```
luckyh, line 4860: Please note that the following test has to be updated accordingly.

https://github.com/autotest/tp-qemu/blob/19113f9526b359c9d6d3c88d436dad6d2b3adde6/qemu/tests/save_restore_vm.py#L43-L44

luckyh (approved):
LGTM

virttest/qemu_vm.py
pevogam: Hmmm, I don't have a clone of the repo from what I recall and perhaps and since a potential update there must be synced with potential update here perhaps is better to ping a maintainer and thus bring to more awareness about the change?

As I have a few more pull requests I might push touching upon the same functionality after sufficient testing and analysis perhaps I can open an issue there to track required updates for the test that I can then propose all at the same time? In short, I am looking for ways to decouple this pull request from a pull request there as much as possible.

pevogam:
@YongxueHong Could you help reviewing this? Thanks!

YongxueHong (approved):
LGTM.

pevogam:
Is there any chance we can merge this to free https://github.com/avocado-framework/avocado-vt/pull/3518 for reviewing/refactoring since it has two approvals already?

virttest/qemu_vm.py
pevogam: Ok, opened a todo there about this update: https://github.com/autotest/tp-qemu/issues/3523#issue-1432944738

Issue #3048, opened 2021-05-11
Demote background errors to warnings


pevogam:
We have many cases where an error like
```
01:59:16 ERROR| Reproduced traceback from: /usr/lib/python3.9/site-packages/avocado_vt/test.py:380
01:59:17 ERROR| Traceback (most recent call last):
01:59:17 ERROR|   File "/var/lib/avocado/data/avocado-vt/virttest/test-providers.d/downloads/io-github-autotest-qemu/generic/tests/unattended_install.py", line 25, in run
01:59:17 ERROR|     unattended_install.run(test, params, env)
01:59:17 ERROR|   File "/usr/lib/python3.9/site-packages/virttest/error_context.py", line 135, in new_fn
01:59:17 ERROR|     return fn(*args, **kwargs)
01:59:17 ERROR|   File "/usr/lib/python3.9/site-packages/virttest/tests/unattended_install.py", line 1423, in run
01:59:17 ERROR|     raise e
01:59:17 ERROR|   File "/usr/lib/python3.9/site-packages/virttest/tests/unattended_install.py", line 1419, in run
01:59:17 ERROR|     test.verify_background_errors()
01:59:17 ERROR|   File "/usr/lib/python3.9/site-packages/avocado_vt/test.py", line 206, in verify_background_errors
01:59:17 ERROR|     self.error(''.join(msg))
01:59:17 ERROR|   File "/usr/lib/python3.9/site-packages/avocado/core/test.py", line 970, in error
01:59:17 ERROR|     raise exceptions.TestError(message)
01:59:17 ERROR| avocado.core.exceptions.TestError: Background error is detected, please refer to file: 'background-error.log' for more details.
```
either hides an actual error or might happen if some test takes too long (even if within the test timeout). I would suggest that we demote these to warnings or something else that doesn't harm the actual test evaluation.

PR #3520, opened 2022-09-09
Make background errors nonfatal as they often register unrelated events

pevogam:
Related to issue 3048:

https://github.com/avocado-framework/avocado-vt/issues/3048

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -3189,7 +3189,7 @@ class VM(virt_vm.BaseVM):
                          qemu_command.replace(" -", " \\\n    -"))
                 self.qemu_command = qemu_command
                 monitor_exit_status = \
-                    params.get("vm_monitor_exit_status", "yes") == "yes"
+                    params.get_boolean("vm_monitor_exit_status", False)
                 self.process = aexpect.run_tail(
                     qemu_command,
                     partial(qemu_proc_term_handler, self,
--- a/virttest/shared/cfg/base.cfg
+++ b/virttest/shared/cfg/base.cfg
@@ -903,7 +903,7 @@ gluster_managed_by_test = "yes"
 #
 # vcpu_cputune = "0-2 N 1,3"
 
-# temporarily disable monitor QEMU vm exit status
+# Default behavior for treating monitor QEMU vm exit status as an error
 vm_monitor_exit_status = no
 
 # Bug `reboot` param from the kickstart is not actually restarts


chunfuwen (approved):
LGTM

avocado_vt/utils.py
```diff
@@ -209,7 +209,7 @@ def verify_background_errors(self):
             msg.append("s are" if len(bg_errors) > 1 else " is")
             msg.append((" detected, please refer to file: "
                         "'%s' for more details.") % BG_ERR_FILE)
-            self.error(''.join(msg))
+            logging.error(''.join(msg))
```
luckyh, line 212: In case of some users do have the demand of failing their tests upon any background error been detected, I'd suggest doing the extension to this function instead. That is we could let users specify which action would be taken at here by introducing a new param (e.g. `test_background_errors_resolution`): background errors will be suppressed when setting the value as, for example, `ignore`; background errors will be treated as fatal when setting the value as `error`.

What do you think?

pevogam: I am not sure if some of these extensions don't add too much complexity because in the years Avocado VT has ended up with many switches and parameters that we are not aware of anyone using, all for the sake of keeping previous behavior and only adding layers of additional logic on top of it. Can you think of a common situation where failing on these background errors (which have lots of false positives) is useful to users? If yes how about a simpler switch like `fail_on_background_error=yes/no`?

luckyh: > Can you think of a common situation where failing on these background errors (which have lots of false positives) is useful to users?

My point is if we are giving the name as "errors" then we should literally treat them as errors, otherwise, completely removing this facility would sound more reasonable to me because usually no one would care about the warning/error messages of passed tests.

There are a few monitors/watchers not running in the main thread which are serving different purposes during test execution. Due to the previous technical limitation I guess, the issues captured by them were handled as "background errors" instead of `test.error` (one of the terminated status of avocado test model), otherwise it would be a completely different topic than what we are discussing about.

One thing I'd agree with you is that we need to avoid those false positives. Currently looks we have two use cases:

1. the qemu vm exit status monitor

https://github.com/avocado-framework/avocado-vt/blob/c7d0371c9aa504b8ac8ecdbf7b74bd15fadadba6/virttest/qemu_vm.py#L142-L146

https://github.com/avocado-framework/avocado-vt/blob/c7d0371c9aa504b8ac8ecdbf7b74bd15fadadba6/virttest/qemu_vm.py#L3196-L3197

it is already configurable, and we can change the setting to not enable it by default in order to being friendly to the scenario not cares about the exit status of qemu processes.

2. the screen inactivity watcher

https://github.com/avocado-framework/avocado-vt/blob/c7d0371c9aa504b8ac8ecdbf7b74bd15fadadba6/virttest/env_process.py#L1989-L2003

https://github.com/avocado-framework/avocado-vt/blob/c7d0371c9aa504b8ac8ecdbf7b74bd15fadadba6/virttest/env_process.py#L1944-L1945

https://github.com/avocado-framework/avocado-vt/blob/c7d0371c9aa504b8ac8ecdbf7b74bd15fadadba6/virttest/shared/cfg/base.cfg#L756-L757

it is configurable as well and the default value already looks good to us.

What do you think?

> If yes how about a simpler switch like `fail_on_background_error=yes/no`?

Sounds good to me, thanks.

pevogam: In that case let's opt-in for better defaults instead of extra parameters and variation of logical execution paths. I will do some large scale retesting and then I will be able to say more if this is a good enough solution, if so I would push this commit as a change to the default monitor exit status value. If not, I will update my comments to the pull request accordingly.

luckyh (approved):
LGTM

chunfuwen (approved):
LGTM

PR #3521, opened 2022-09-09
Add timeout for "qemu-img convert"

pevogam:

### DIFF
--- a/virttest/qemu_storage.py
+++ b/virttest/qemu_storage.py
@@ -986,9 +986,13 @@ class QemuImg(storage.QemuImg):
         convert_cmd = self.image_cmd + " " + \
             self._cmd_formatter.format(self.convert_cmd, **cmd_dict)
 
-        LOG.info("Convert image %s from %s to %s", self.image_filename,
-                 self.image_format, convert_image.image_format)
-        process.run(convert_cmd)
+        timeout = convert_params.get_numeric("image_conversion_timeout", -1)
+        timeout = None if timeout == -1 else timeout
+
+        LOG.info("Convert image %s from %s to %s with timeout %s",
+                 self.image_filename, self.image_format,
+                 convert_image.image_format, timeout)
+        process.run(convert_cmd, timeout=timeout)
         if convert_image.encryption_config.key_secret:
             convert_image.encryption_config.key_secret.save_to_file()
 


virttest/qemu_storage.py
```diff
@@ -986,9 +986,12 @@ def convert(self, params, root_dir, cache_mode=None,
...
 
-        LOG.info("Convert image %s from %s to %s", self.image_filename,
-                 self.image_format, convert_image.image_format)
-        process.run(convert_cmd)
+        timeout = params.get_numeric("conversion_timeout", 12000)
+
+        LOG.info("Convert image %s from %s to %s with timeout %s",
+                 self.image_filename, self.image_format,
+                 convert_image.image_format, timeout)
+        process.run(convert_cmd, timeout=timeout)
```
chunfuwen, line 994: Do we need do assert about process.run() exit status ?

pevogam: Thanks for taking a look @chunfuwen! In this case I would say "no" because the `process.run` would fail on nonzero status to begin with. This means that we will be able to detect the conversion did not go well the way we do with all other commands we run via the `process` module.

chunfuwen (approved):
LGTM

virttest/qemu_storage.py
```diff
@@ -986,9 +986,12 @@ def convert(self, params, root_dir, cache_mode=None,
         convert_cmd = self.image_cmd + " " + \
             self._cmd_formatter.format(self.convert_cmd, **cmd_dict)
 
-        LOG.info("Convert image %s from %s to %s", self.image_filename,
-                 self.image_format, convert_image.image_format)
-        process.run(convert_cmd)
+        timeout = params.get_numeric("conversion_timeout", 12000)
```
luckyh, line 989: Shall we use `convert_params` instead of `params` to get a fine-grained manipulation?

luckyh, line 989: Suggest the param name `image_conversion_timeout` to avoid potential naming collisions.

luckyh, line 989: Do we consider to support the case of not setting timeout? I'm afraid the no timeout case won't be parsed properly by calling `Param.get_numeric`. Otherwise, as a workaround, users have to set a very large value to simulate that.

luckyh, line 989: Also, I'd suggest setting *no timeout* as the default value if we plan to support it, which is the current behavior.

virttest/qemu_storage.py
```diff
@@ -492,7 +492,7 @@ class QemuImg(storage.QemuImg):
                    "{target_image_format} {options} {convert_sparse_size} "
                    "{rate_limit} {convert_target_is_zero!b} "
                    "{convert_backing_file} "
-                   "{image_filename} {target_image_filename} "
+                   "'{image_filename}' '{target_image_filename}' "
```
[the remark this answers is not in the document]
pevogam: Indeed, this is also a common problem to be improved in multiple commands where the file names might contain spaces and are not currently handled. So I might as well drop this step as it is not the main change of the pull request. Is it worth registering an issue about the space-including filenames here?

virttest/qemu_storage.py
pevogam: IIRC my probelm with `convert_params` was that it wasn't exactly filtered version of params but a smaller set of parameters introduced locally. But even then what you say would make sense so I will check in more detail.

pevogam: Why is that? By default no timeout will be represented as no `conversion_timeout` key here which would result in `get_numeric` returning 12000, am I missing something? Or by "not parsed properly" do you mean "not handled properly" as in "forced to 12000 seconds"?

pevogam: How about a very large number as a default value? Can you think of practical use cases where waiting infinitely long would make sense?

virttest/qemu_storage.py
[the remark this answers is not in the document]
luckyh: > Is it worth registering an issue about the space-including filenames here?

Yeah, make sense, thanks!

virttest/qemu_storage.py
luckyh: I mean to exactly support setting no timeout rather than a certain time limit. In that case, no timeout would be logically treated as NaN, which should not be parsed by `get_numeric`.

luckyh: As [*test timeout*](https://avocado-framework.readthedocs.io/en/latest/guides/writer/chapters/writing.html#setting-a-test-timeout) will take effect eventually I believe it won't be really waiting "infinitely" long.

From my POV, the critical issues were:

* how to set a rational default timeout for each single operation of a test?
* how to handle the relationship between that and test timeout?

since how long a image conversion would take depends on a couple of factors (image size, hardware storage performance, system load status, etc.), it's hard for me to say if 12000s as the default is better or not comparing with using the test timeout for the **most** cases. What if someone would request to enlarge/reduce the number in future, how to evaluate that? I can't answer that this moment, so I just suggested what I put in the previous comment. I think users know their case better than us so I'd like to let them handle the thing independently.

virttest/qemu_storage.py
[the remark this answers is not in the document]
pevogam: Alright, registered as https://github.com/avocado-framework/avocado-vt/issues/3546 and will drop the quotes in the next push.

virttest/qemu_storage.py
pevogam: Then I propose (and will push) an option to do
```python
        timeout = convert_params.get_numeric("image_conversion_timeout", -1)
        timeout = None if timeout == -1 else timeout
```
even though I can think of no practical situation where infinite duration is reasonable as a use case. Of course, it is just as hard to guess large enough values for all sorts of QCOW2 conversions and sizes so I will simply push that change, treating -1 as no timeout.

pevogam: Indeed, see above for the solution I would push.

virttest/qemu_storage.py
```diff
@@ -986,9 +986,13 @@ def convert(self, params, root_dir, cache_mode=None,
         convert_cmd = self.image_cmd + " " + \
             self._cmd_formatter.format(self.convert_cmd, **cmd_dict)
 
-        LOG.info("Convert image %s from %s to %s", self.image_filename,
-                 self.image_format, convert_image.image_format)
-        process.run(convert_cmd)
+        timeout = convert_params.get_numeric("image_conversion_timeout", -1)
+        timeout = None if timeout == -1 else timeout
```
luckyh, line 990: Then maybe it seems to be the time to introduce a dedicated method to `Param` for getting timeouts, how does that sound? And later we may consider to add more features to support:

* setting timeouts in different units of time (e.g. `5m`, `1h`)
* basic calculations (with constants, e.g. `TEST_TIMEOUT * 0.8`)

pevogam: Hmm, I am not sure how `Params` object enhancements relates to the current pull request which aims specifically at `qemu-img` but bringing myself to talk about this new topic I would still say I am not sure whether introducing such methods in `Params` is not too data specific. The current type-handling methods are general enough and a small easy to maintain selection. Returning -1 is a very easy way to handle not just timeouts but various numeric parameters that might end up including infinity and the above is a simple solution that only uses -1 default value, something I hardly see as needing an entire new method with lots of added assumptions.

luckyh: > Returning -1 is a very easy way to handle not just timeouts but various numeric parameters that might end up including infinity and the above is a simple solution that only uses -1 default value, something I hardly see as needing an entire new method with lots of added assumptions.

Then, almost all the cases of handling timeouts may just have like this two lines approach (let's recall the situation of handling numerics before having `get_numeric`). Anyway, I understand that it's not a must to this PR, so I'd leave the proposal for future discussion, thanks.

luckyh (approved):
LGTM

pevogam:
@chunfuwen Thanks for rechecking and reviewing again the previous pull request, this one also has two approvals. Thanks!

pevogam:
From what I can see this pull request now has enough many approvals. Thanks both!

pevogam:
> From what I can see this pull request now has enough many approvals. Thanks both!

ping @luckyh

luckyh:
Alright, let's merge it.

PR #3524, opened 2022-09-19
Add libvirt audit related utils

chunfuwen:
Add libvirt audit related utils

enable/clean audit log and ensure auditd, those operations are common to libvirt audit related feature

### DIFF
--- a/virttest/utils_libvirt/libvirt_audit.py
+++ b/virttest/utils_libvirt/libvirt_audit.py
@@ -1,0 +1,39 @@
+"""
+libvirt audit related utility functions
+"""
+
+from avocado.utils import process
+from avocado.utils import service
+
+from virttest import utils_libvirtd
+
+
+def enable_audit_log(libvirtd_config, audit_level=1):
+    """
+    Configure audit log level
+
+    :param libvirtd_config: libvirtd config object
+    :param audit_level: audit level
+    """
+    libvirtd_config.audit_level = audit_level
+    libvirtd_config.audit_logging = 1
+    utils_libvirtd.Libvirtd('virtqemud').restart()
+
+
+def clean_up_audit_log_file():
+    """
+    Clean up audit message in log file.
+    """
+    cmd = "truncate -s 0  /var/log/audit/audit.log*"
+    process.run(cmd, shell=True)
+
+
+def ensure_auditd_started():
+    """
+    Check audit service status and start it if it's not running
+    """
+    service_name = 'auditd'
+    service_mgr = service.ServiceManager()
+    status = service_mgr.status(service_name)
+    if not status:
+        service_mgr.start(service_name)


chunfuwen:
test result can refer to tp-libvirt PR: https://github.com/autotest/tp-libvirt/pull/4529

chunfuwen:
@dzhengfy 
@Yingshun 
please take a look

virttest/utils_libvirt/libvirt_audit.py
```diff
@@ -0,0 +1,38 @@
...
+
+from avocado.utils import process
+from avocado.utils import service
+
+from virttest import utils_libvirtd
+
+
+def enable_audit_log(libvirtd_config):
+    """
+    Configure audit log level"
```
Yingshun, line 13: Do you need to delete the '"' at the end of the line?

virttest/utils_libvirt/libvirt_audit.py
```diff
@@ -0,0 +1,38 @@
...
+
+    :param libvirtd_config: libvirtd config object
+    """
+    libvirtd_config.audit_level = 1
+    libvirtd_config.audit_logging = 1
```
Yingshun, line 18: Is it better to set these values as needed? I think we'd better avoid hard-codes here. What do you think?

virttest/utils_libvirt/libvirt_audit.py
```diff
@@ -0,0 +1,38 @@
...
+
+def clean_up_audit_log_file():
+    """
+    Clean up audit message in log file.
+    """
+    cmd = "truncate -s 0  /var/log/audit/audit.log*"
+    process.run(cmd, shell=True)
+
+
+def ensure_auditd_started():
```
Yingshun, line 30: Are these 2 functions related to 'libvirt'? I wonder if it's suitable for us to put auditd service and log related functions udner 'libvirt'. @dzhengfy  What's your opinion?

virttest/utils_libvirt/libvirt_audit.py
```diff
@@ -0,0 +1,38 @@
+"""
+libvirt audit related utility functions
```
dzhengfy, line 2: Is libvirt audit very significant feature to be wrapped up in a single file?

virttest/utils_libvirt/libvirt_audit.py
```diff
@@ -0,0 +1,38 @@
...
+    Configure audit log level"
+
+    :param libvirtd_config: libvirtd config object
+    """
+    libvirtd_config.audit_level = 1
+    libvirtd_config.audit_logging = 1
+    utils_libvirtd.Libvirtd('virtqemud').restart()
+
+
+def clean_up_audit_log_file():
```
dzhengfy, line 22: virttest/utils_libvirt/ the path is supposed to include those utilities using vm xml or virsh command.
So from above point, this function could be put into utils_audit.py (for example) under virttest.

virttest/utils_libvirt/libvirt_audit.py
dzhengfy: virttest/utils_libvirt/ the path is supposed to include those utilities using vm xml or virsh command.
So from above point, this function could be put into utils_audit.py (for example) under virttest.

virttest/utils_libvirt/libvirt_audit.py
chunfuwen: @dzhengfy "Is libvirt audit very significant feature to be wrapped up in a single file?" I wonder why you believe only  very significant feature  can be wrapped in single file ? and How you evaluate its significance?

In terms of audit  libvirt feature world, if you take a look at those cases against them in polarion, they are clearly separated with others

virttest/utils_libvirt/libvirt_audit.py
```diff
@@ -0,0 +1,38 @@
...
+libvirt audit related utility functions
+"""
+
+from avocado.utils import process
+from avocado.utils import service
+
+from virttest import utils_libvirtd
+
+
+def enable_audit_log(libvirtd_config):
```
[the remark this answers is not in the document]
chunfuwen: I automated 3 audit related cases, audit_level=1 is commonly necessary conditions since you need catch and check audit log information when validating cases. And audit_level=1 can be guarded to get full log.

virttest/utils_libvirt/libvirt_audit.py
chunfuwen: removed

virttest/utils_libvirt/libvirt_audit.py
chunfuwen: I automated 3 audit related cases, audit_level=1 is commonly necessary conditions since you need catch and check audit log information when validating cases. And audit_level=1 can be guarded to get full log.
But I can add one addition parameter to allow audit_level to be configured in parameter

virttest/utils_libvirt/libvirt_audit.py
chunfuwen: 1)personally I think current implementation for enable_audit_log and clean_up_audit_log_file are very simple because it only satisfy libvirt case's need(more straightforward think, it is just interpreting requirements from libvirt case). It is hard to predict they can satisfy the need other than libvirt test case. So I prefer to keep under utils_libvirt

virttest/utils_libvirt/libvirt_audit.py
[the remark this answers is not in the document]
chunfuwen: I automated 3 audit related cases, audit_level=1 is commonly necessary conditions since you need catch and check audit log information when validating cases. And audit_level=1 can be guarded to get full log.
But I can add one addition parameter to allow audit_level to be configured in parameter

virttest/utils_libvirt/libvirt_audit.py
chunfuwen: @dzhengfy "Is libvirt audit very significant feature to be wrapped up in a single file?" I wonder why you believe only very significant feature can be wrapped in single file ? and How you evaluate its significance?

In terms of audit libvirt feature world, if you take a look at those cases against them in polarion, they are clearly separated with others

virttest/utils_libvirt/libvirt_audit.py
chunfuwen: 1)personally I think current implementation for enable_audit_log and clean_up_audit_log_file are very simple because it only satisfy libvirt case's need(more straightforward think, it is just interpreting requirements from libvirt case). It is hard to predict they can satisfy the need other than libvirt test case. So I prefer to keep under utils_libvirt

chunfuwen:
this is not urgently needed, so close it firstly ,and can open it once needed

PR #3529, opened 2022-09-19
libvirt_numa: enhance the parsing

dzhengfy:
Sometimes we do not expect an exception is raised when the format is not qualified. Instead we need the function just bypass.

### DIFF
--- a/virttest/utils_libvirt/libvirt_numa.py
+++ b/virttest/utils_libvirt/libvirt_numa.py
@@ -83,12 +83,13 @@ def create_hmat_xml(vmxml, params):
     return vmxml
 
 
-def parse_numa_nodeset_to_str(numa_nodeset, node_list):
+def parse_numa_nodeset_to_str(numa_nodeset, node_list, ignore_error=False):
     """
     Parse numa nodeset to a string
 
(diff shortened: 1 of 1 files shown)


Yingshun (approved):

virttest/utils_libvirt/libvirt_numa.py
```diff
@@ -131,6 +132,8 @@ def _get_first_continuous_numa_node_index(node_list):
             numa_nodeset = '%s-%s,^%s' % (str(node_list[candidate_index]),
                                           str(node_list[candidate_index + 1]),
                                           str(node_list[candidate_index + 1]))
+    elif ignore_error:
+        LOG.error("Supported formats are not found. No parsing happens.")
```
chloerh, line 136: The function will return `numa_nodeset` as it was input when `ignore_error` is True, right?

dzhengfy: @chloerh , yes, you are right. That's what I want.

chloerh: OK.

PR #3532, opened 2022-09-26
libvirt.check_result:Replace mutable default args with Nones

chloerh:
Some of the original default args are []s which are mutable.This patch replaces them with Nones.


dzhengfy (approved):
LGTM

PR #3533, opened 2022-09-29
Allow destroying vms via monitor upon recreation

pevogam:
In some cases the vm's memory and its images are known to be clean of corruption and we can save up to 2 minutes per test (for thousands of tests) if we destroy the vm faster before recreating it anew.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -2943,7 +2943,6 @@ class VM(virt_vm.BaseVM):
         :raise PrivateBridgeError: If fail to bring the private bridge
         """
         error_context.context("creating '%s'" % self.name)
-        self.destroy(free_mac_addresses=False)
 
         if name is not None:
             self.name = name
@@ -4934,6 +4933,7 @@ class VM(virt_vm.BaseVM):
         """
         if self.is_alive():
             self.verify_status('paused')  # Throws exception if not
+            self.destroy(gracefully=False, free_mac_addresses=False)
         LOG.debug("Restoring VM %s from %s" % (self.name, path))
         # Rely on create() in incoming migration mode to do the 'right thing'
         self.create(name=self.name, params=self.params, root_dir=self.root_dir,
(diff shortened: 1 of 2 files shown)


virttest/qemu_vm.py
```diff
@@ -2839,7 +2839,7 @@ def _stop_daemons(self):
     def create(self, name=None, params=None, root_dir=None,
                timeout=120, migration_mode=None,
                migration_exec_cmd=None, migration_fd=None,
-               mac_source=None):
+               mac_source=None, safe_destroy=True):
```
luckyh, line 2842: @pevogam I can see the point of the new argument being introduced to `create`, however I'd prefer another approach, and that is to remove the invoking of `destroy` from `create` in order to make the implementation of `create` more proper. Then each VM launching workflow can fully control the way of if and how `destroy` would be called. Let me know your opinion, thanks!

luckyh:
Adding @YongxueHong to this thread, thanks.

virttest/qemu_vm.py
YongxueHong: Hi @pevogam 
I maybe prefer @luckyh's opinions, firstly, for adding `safe_destroy` here, I think it is not a better way to understand the definition of method `create`, I mean that the arguments should be related to the creating behaviors. And the job of creating a VM just focuses on creating, not including the destroy VM this step should be decoupled from the `create`.
I am glad to know your opinions., Thanks.

pevogam: I agree with both of you here, was just trying to avoid breaking the behavior for other people since AFAIK `vm.create()` is a ubiquitous line that is used throughout the project, not to mention possible external projects using the `vm` object too. But if you feel bold let's try it - I will push a change here that drops the "re-creation" functionality and update any lines I find at least in Avocado VT.

pevogam: Pushed and CI passed

luckyh (approved):
LGTM

pevogam:
Slightly tweaked this to accommodate https://github.com/avocado-framework/avocado-vt/pull/3564.

PR #3535, opened 2022-10-06
Make it possible to separately keep any vms before and after tests

pevogam:
We use the opportunity to also give these parameters clear names that also imply typical use case applications.

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1159,7 +1159,7 @@ def preprocess(test, params, env):
     # Destroy and remove VMs that are no longer needed in the environment or
     # leave them untouched if they have to be disregarded only for this test
     requested_vms = params.objects("vms")
-    keep_unrequested_vms = params.get_boolean("keep_env_vms", False)
+    keep_unrequested_vms = params.get_boolean("keep_unrequested_vms", False)
     for key in list(env.keys()):
         vm = env[key]
         if not isinstance(vm, virt_vm.BaseVM):
@@ -1699,7 +1699,7 @@ def postprocess(test, params, env):
         LOG.info("Sosreport for remote host: %s", sosreport_path)
     living_vms = [vm for vm in env.get_all_vms() if vm.is_alive()]
     # Close all monitor socket connections of living vm.
-    if not params.get_boolean("keep_env_vms", False):
+    if not params.get_boolean("keep_vms_after_test", False):
         for vm in living_vms:
             if hasattr(vm, "monitors"):
                 for m in vm.monitors:


virttest/env_process.py
```diff
@@ -1699,7 +1699,7 @@ def postprocess(test, params, env):
         LOG.info("Sosreport for remote host: %s", sosreport_path)
     living_vms = [vm for vm in env.get_all_vms() if vm.is_alive()]
     # Close all monitor socket connections of living vm.
-    if not params.get_boolean("keep_env_vms", False):
+    if not params.get_boolean("keep_vms_for_debugging", False):
```
luckyh, line 1702: Since debugging is not the only scenario for keeping VMs living after test finished, I'd suggest using `keep_vms_after_test` as the param name.

luckyh:
@YongxueHong @dzhengfy @chunfuwen would you like to review this PR as well? thanks!

virttest/env_process.py
pevogam: My original idea was to provide people with immediate idea of the major use case for the parameter because we have had cases in the past where developers are not seeing how a parameter is useful. But alright, I will rename this since it is slightly more general.

Keep in mind though that I haven't updated the `keep_unrequested_vms` parameter to `keep_vms_before_test` because the original code there only discards unrequested ones. Assuming you didn't comment on the line above I would accept this was intentional on your side.

luckyh: Okay, let me elaborate more on this point. To my understanding, `keep_vms_for_debugging` would function in a different manner compared with `keep_vms_after_test`, it's more like to promise that the VMs would still be alive after a job execution. As one execution could contain multiple tests, based on the current implementation of the VM management workflow through a single VT test lifecycle it'd be hard to make `keep_vms_for_debugging` functioning properly, since those VMs users wanted to keep can be killed in the later tests when `keep_unrequested_vms` was set in an unexpected value. In other words, users may expect that `keep_vms_for_debugging` could work independently but it's not true currently.

On the other hand, `keep_vms_after_test` would sounds like about not killing VMs at the end of a test execution. Whether VMs would be alive or not after a job execution depends on how users configure the `keep_unrequested_vms` param, and users can do whatever they want to the VMs surviving till the end.

> we have had cases in the past where developers are not seeing how a parameter is useful.

So my comment to this is we shall improve the document, which is a huge topic...

luckyh (approved):
LGTM

virttest/env_process.py
pevogam: Yes, the usual debugging use case requires the users to run a job with a single test in order to truly debug this test, something we have done quite often in the past. But what you say makes complete sense, I was just focusing on how the parameter can be useful and it definitely should not overpromise so good points there.

YongxueHong (approved):
LGTM

pevogam:
@YongxueHong Thanks for reviewing together with Xu Han, this pull request now has two approvals. Thanks!

YongxueHong:
> @YongxueHong Thanks for reviewing together with Xu Han, this pull request now has two approvals. Thanks!

Hi @pevogam 
Thanks for contributing, let's merge it.

PR #3536, opened 2022-10-10
utils_net.get_ip_address_by_interface:fix a regression bug

yanglei-rh:
A regression was introduced by ddc3e37, which results in unexcepted exceptions being raised instead of the correct one during some situations.

ID:2133384

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3462,7 +3462,7 @@ def get_ip_address_by_interface(ifname, ip_ver="ipv4", linklocal=False):
     :param ifname: interface name
     :param ip_ver: Host IP version, ipv4 or ipv6.
     :param linklocal: Whether ip address is local or remote
-    :raise NetError: When failed to fetch IP address (ioctl raised IOError.).
+    :raise NetError: When failed to fetch IP address.
     """
     if ip_ver == "ipv6":
         ver = netifaces.AF_INET6
@@ -3470,7 +3470,14 @@ def get_ip_address_by_interface(ifname, ip_ver="ipv4", linklocal=False):
     else:
         ver = netifaces.AF_INET
         linklocal_prefix = "169.254"
-    addr = netifaces.ifaddresses(ifname).get(ver)
+    try:
+        addr = netifaces.ifaddresses(ifname).get(ver)
+    # FIXME: The kind of exceptions caught should be more specific
+    except:
+        # TODO: NetError is a very general usage,it will be changed to a
+        # more friendly way in the future
+        raise NetError(
+            "Error while retrieving IP address from interface %s." % ifname)
 
     if addr is not None:
         try:


yanglei-rh:
Hello @luckyh @PaulYuuu Could you please help me review this patch when you have free time? Many thanks.

virttest/utils_net.py
```diff
@@ -3462,15 +3462,19 @@ def get_ip_address_by_interface(ifname, ip_ver="ipv4", linklocal=False):
...
         ver = netifaces.AF_INET6
         linklocal_prefix = "fe80"
     else:
         ver = netifaces.AF_INET
         linklocal_prefix = "169.254"
-    addr = netifaces.ifaddresses(ifname).get(ver)
+    try:
+        addr = netifaces.ifaddresses(ifname).get(ver)
+    except:
+        raise NetError(
```
YongxueHong, line 3476:
```suggestion
        raise IPAddrGetError(
```
I think we can raise the specific error type, such as `IPAddrGetError`.

yanglei-rh: @YongxueHong I'm trying to look at the whole python file, I don't think it's a pure renaming issue, if we change it we'll add new `Class` and other code to support it. So please help review it again,  maybe keeping the current code is a good idea.

yanglei-rh: It seems to me that the exception can be caught because the class is defined at the beginning and call in subsequent functions or classes. So if this name by changed, does it also need to create a class and call it when other function use it? Maybe I'm wrong, please help to point it out. Thanks a lot.
```
class NetError(Exception):

    def __init__(self, *args):
        Exception.__init__(self, *args)
```

YongxueHong: Hi @yanglei-rh 
Thanks for the explanation.
We raise the NetError here since somewhere will catch the `NetError`, right?
So we just focus make this function can raise such type `Exception` for the caller.
For the `IPAddrGetError` inherits to the `NetError`:
`class IPAddrSetError(NetError):`

So this update will not affect the current calling place, like:

yanglei-rh: Hello @YongxueHong Thanks for you explanation, I'll update the name later. Thanks again.

yanglei-rh: Hi @YongxueHong Updated, please help review it again, thanks a lot.

YongxueHong (approved):
LGTM.

virttest/utils_net.py
```diff
@@ -3462,15 +3462,19 @@ def get_ip_address_by_interface(ifname, ip_ver="ipv4", linklocal=False):
...
         linklocal_prefix = "fe80"
     else:
         ver = netifaces.AF_INET
         linklocal_prefix = "169.254"
-    addr = netifaces.ifaddresses(ifname).get(ver)
+    try:
+        addr = netifaces.ifaddresses(ifname).get(ver)
+    except:
+        raise IPAddrGetError(
+            "Error while retrieving IP address from interface %s." % ifname)
```
luckyh, line 3477: @yanglei-rh @YongxueHong `IPAddrGetError` would require `mac_addr` as the first argument for its `__init__` method, so this has to be updated.

https://github.com/avocado-framework/avocado-vt/blob/7689bcddf78eb5282a5980cd697af6e4ebbbd298/virttest/utils_net.py#L307-L309

virttest/utils_net.py
```diff
@@ -3462,15 +3462,19 @@ def get_ip_address_by_interface(ifname, ip_ver="ipv4", linklocal=False):
...
     if ip_ver == "ipv6":
         ver = netifaces.AF_INET6
         linklocal_prefix = "fe80"
     else:
         ver = netifaces.AF_INET
         linklocal_prefix = "169.254"
-    addr = netifaces.ifaddresses(ifname).get(ver)
+    try:
+        addr = netifaces.ifaddresses(ifname).get(ver)
+    except:
```
PaulYuuu, line 3475: I want to confirm which exceptions will be raised, to see if we really need to capture the whole Exception indeed.

virttest/utils_net.py
luckyh: And also, the error message indicated that it was for guest issues, therefore I'm not sure if it would make sense since our case is a host one, what do you think?

https://github.com/avocado-framework/avocado-vt/blob/7689bcddf78eb5282a5980cd697af6e4ebbbd298/virttest/utils_net.py#L315-L316

YongxueHong: Hi @luckyh 
Yeah, I forgot the usage of this Error.

YongxueHong: I re-checked the subclass of `NetError`, there is no other available error for this function.
Maybe we keep the original type.
Sorry for disturbing you.

yanglei-rh: Hello @luckyh @PaulYuuu Based the above comment, it should be changed `NetError`, right?

virttest/utils_net.py
yanglei-rh: Add this section as `TODO`. Please help me review it again, thanks a lot.

yanglei-rh:
Hello @luckyh @PaulYuuu @YongxueHong Please help me review it again, thanks a lot.

PaulYuuu (approved):

yanglei-rh:
Hello @luckyh @YongxueHong Could you please help me review it again? Thanks a lot.

YongxueHong (approved):
LGTM.

luckyh:
Hi @yanglei-rh , would you please try updating the commit by rebasing with the latest master to get rid off the CI failures? thanks!

yanglei-rh:
> Hi @yanglei-rh , would you please try updating the commit by rebasing with the latest master to get rid off the CI failures? thanks!

Done, and all CI checks passed.

PR #3539, opened 2022-10-11
utils_misc: Assert test entry point is in fact callable

bgartzi:
Introducing another check in `utils_misc.get_test_entrypoint_func` to make sure the `run` attribute of the test module is callable. The input module could have an attribute called run that is not callable (e.g. is holding a numerical value), so it looks like we should guard against that case.

This issue was elucidated during #3350 discussions and pylint check fixes. The pylint-disable comments introduced in the latter rely on the checks introduced on this PR, which has been split into this one since it didn't really match the scope of #3350.


luckyh (approved):
ACK

PR #3550, opened 2022-10-24
RHEL 9 definitions

richtja:
This adds definitions and kickstart file to support tests on RHEL 9.

### DIFF
--- a/virttest/shared/cfg/guest-os/Linux/RHEL/9.cfg
+++ b/virttest/shared/cfg/guest-os/Linux/RHEL/9.cfg
@@ -1,0 +1,44 @@
+- 9:
+    variants:
+        - aarch64:
+            vm_arch_name = aarch64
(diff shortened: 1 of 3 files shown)


clebergnu (commented):
Hi @richtja ,

A couple of questions:

1) The kickstart file seems almost identical to the CentOS-Stream 9 one from #3551, with one difference in the `echo` function that may or may not be on purpose.  Can you check that?
2) Depending on the answer of 1, can't we make both RHEL-9 and CentOS-stream 9 kickstart files the same?
3) The limits of installation based with an URL source mentioned in #3551 are not applicable here (because of `mem = 4096`) but I would be suprised if the `image_size` issue doesn't apply too.  Have you checked that?

If you decide to make the two kickstart files one, then maybe it's a good idea to combine this with the CentOS Stream 9 PR.

Thanks.

PR #3559, opened 2022-11-07
qdevices: change value in "cache-size" to integer

nickzhq:
ID: 2140497, 2140933

Signed-off-by: Houqi (Nick) Zuo <hzuo@redhat.com>

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -191,7 +191,7 @@ class DevContainer(object):
         return self.__qemu_ver
 
     def initialize_iothread_manager(self, params, guestcpuinfo,
-                                    cmdline_format_cfg=None):
+                                    cmdline_format_cfg={}):
         """Initialize iothread manager.
         :param params: vt params
         :param guestcpuinfo: Cpuinfo object that stores guest cpu info
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -892,6 +892,13 @@ class QBlockdevFormatQcow2(QBlockdevFormatNode):
     """ New a format qcow2 blockdev node. """
     TYPE = 'qcow2'
 
+    def _convert_blkdev_args(self, args):
+        for key, val in args.items():
+            # "cache-size" is from device( "driver": "qcow2" )
+            if key == "cache-size":
+                args[key] = int(val)
+        return super(QBlockdevFormatQcow2, self)._convert_blkdev_args(args)
+
 
 class QBlockdevFormatRaw(QBlockdevFormatNode):
     """ New a format raw blockdev node. """
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -1587,17 +1587,20 @@ class Memory(QObject):
     @staticmethod
     def _convert_memobj_args(args):
         """
-        Convert string to uint64,
-        and convert "size": "14336M"  to "size": 15032385536 (bytes)
+        Type convert, such as string to uint64( "size": "14336M"  to
+        "size": 15032385536 (bytes) )
 
         :param args: Dictionary with the qmp parameters.
         :type args: dict
         :return: Converted args.
         :rtype: dict
         """
-        if "size" in args:
-            args["size"] = int(utils_numeric.normalize_data_size(args["size"],
-                                                                 "B"))
+        for key, val in args.items():
+            if "size" == key:
+                args[key] = int(utils_numeric.normalize_data_size(val, "B"))
+            # "share" from object( "qom-type": "memory-backend-memfd" )
+            elif "share" == key:
+                args[key] = val == "yes"
 
 
 class Dimm(QDevice):
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -1160,9 +1160,10 @@ class QDevice(QCustomDevice):
                 command_dict[key] = str(val)
             # disable-legacy from device ("driver": "virtio-scsi-pci")
             # write-cache from device ("driver": "scsi-hd")
-            elif val in ('on', 'yes') and key not in expect_string_val:
+            elif val in ('on', 'yes', "true") and key not in expect_string_val:
                 command_dict[key] = True
-            elif val in ('off', 'no') and key not in expect_string_val:
+            elif val in ('off', 'no', "false") and \
+                    key not in expect_string_val:
                 command_dict[key] = False
             else:
                 command_dict[key] = val
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -1020,6 +1020,12 @@ class QBlockdevProtocolHTTPS(QBlockdevProtocol):
     """ New a protocol https blockdev node. """
     TYPE = 'https'
 
+    def _convert_blkdev_args(self, args):
+        for key, val in args.items():
+            if key == "timeout":
+                args[key] = int(val)
+        return super(QBlockdevProtocolHTTPS, self)._convert_blkdev_args(args)
+
 
 class QBlockdevProtocolFTP(QBlockdevProtocol):
     """ New a protocol ftp blockdev node. """


nickzhq:
(1/1) Host_RHEL.m9.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.1.0.x86_64.io-github-autotest-qemu.cache_sizes_test.q35: STARTED
 (1/1) Host_RHEL.m9.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.1.0.x86_64.io-github-autotest-qemu.cache_sizes_test.q35: PASS (412.67 s)

nickzhq:
@hellohellenmao @YongxueHong @luckyh  Please take a look, thanks!

virttest/qemu_devices/qdevices.py
```diff
@@ -842,6 +842,10 @@ def get_qid(self):
     def _cmdline_json(self):
         params = self.params.copy()
         out = "-%s " % self.type
+        for key, val in params.items():
+            # "cache-size" is from device( "driver": "qcow2" )
+            if key == "cache-size":
+                params[key] = int(val)
```
YongxueHong, line 848: Suggest letting the subclass `QBlockdevFormatQcow2` override this method `_cmdline_json` since this change relates to the `QBlockdevFormatQcow2`.
And I think there are also other blockdev nodes has such a change in the future, so we just focus on updating the corresponding subclass of the method instead of updating the parent class method.

nickzhq: > Suggest letting the subclass `QBlockdevFormatQcow2` override this method `_cmdline_json` since this change relates to the `QBlockdevFormatQcow2`. And I think there are also other blockdev nodes has such a change in the future, so we just focus on updating the corresponding subclass of the method instead of updating the parent class method.

Hello @YongxueHong , thanks for comment. Once we do the everything at subclass, there'll be a redundant codes. In addition, as I known, if "cache-size" is integer in qcow2 format, it must be integer in other format.  
What do you think?

YongxueHong: > > Suggest letting the subclass `QBlockdevFormatQcow2` override this method `_cmdline_json` since this change relates to the `QBlockdevFormatQcow2`. And I think there are also other blockdev nodes has such a change in the future, so we just focus on updating the corresponding subclass of the method instead of updating the parent class method.
> 
> Hello @YongxueHong , thanks for comment. Once we do the everything at subclass, there'll be a redundant codes. In addition, as I known, if "cache-size" is integer in qcow2 format, it must be integer in other format. What do you think?

Hi @nickzhq 
I don't think it is a redundant since we just make up for the base class method.
I mean that we can change the param dict then call the bass class method, like `super(QBlockdevFormatQcow2, self)._cmdline_json()` in the `QBlockdevFormatQcow2`
BTW, we can refer to other opinions @zhencliu @luckyh

nickzhq: > hi @nickzhq , I am afraid cache-size is a qcow2-specific attribute, what @YongxueHong means here is for a parent class QBlockdevNode, if we add cache-size there, it means all its children should have this attribute, but that is not the case, in your case, only qcow2 format node has the attribute, so we may better add it there

Thanks for explanation. The code is on changing.

nickzhq:
(1/1) Host_RHEL.m9.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.1.0.x86_64.io-github-autotest-qemu.cache_sizes_test.q35: STARTED
 (1/1) Host_RHEL.m9.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.1.0.x86_64.io-github-autotest-qemu.cache_sizes_test.q35: PASS (338.55 s)

virttest/qemu_devices/qdevices.py
```diff
@@ -892,6 +892,16 @@ class QBlockdevFormatQcow2(QBlockdevFormatNode):
...
 
+    def _cmdline_json(self):
+        params = self.params.copy()
+        out = "-%s " % self.type
+        for key, val in params.items():
+            # "cache-size" is from device( "driver": "qcow2" )
+            if key == "cache-size":
+                params[key] = int(val)
+        new_args = self._convert_blkdev_args(params)
+        return out + '\'' + json.dumps(new_args) + '\''
```
YongxueHong, line 903:
```suggestion
    def _convert_blkdev_args(args):
        for key, val in args.items():
            # "cache-size" is from device( "driver": "qcow2" )
            if key == "cache-size":
                args[key] = int(val)
        return super(QBlockdevFormatQcow2, self)._convert_blkdev_args(args)
```

nickzhq:
(1/1) Host_RHEL.m9.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.1.0.x86_64.io-github-autotest-qemu.cache_sizes_test.q35: STARTED
 (1/1) Host_RHEL.m9.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.1.0.x86_64.io-github-autotest-qemu.cache_sizes_test.q35: PASS (345.66 s)

nickzhq:
@YongxueHong @zhencliu @hellohellenmao  Please check again, thanks!

zhencliu:
hi @hellohellenmao , besides cache-size, there are still some 'int' options for qcow2 format node, see below:
{ 'struct': 'BlockdevOptionsQcow2',
   'base': 'BlockdevOptionsGenericCOWFormat',
   'data': { '*lazy-refcounts': 'bool',
             '*pass-discard-request': 'bool',
             '*pass-discard-snapshot': 'bool',
             '*pass-discard-other': 'bool',
             '*overlap-check': 'Qcow2OverlapChecks',
             '*cache-size': 'int',
             '*l2-cache-size': 'int',
             '*l2-cache-entry-size': 'int',
             '*refcount-cache-size': 'int',
             '*cache-clean-interval': 'int',
             '*encrypt': 'BlockdevQcow2Encryption',
             '*data-file': 'BlockdevRef' } }

such as l2-cache-size, do we have related test cases?

nickzhq:
Hello @YongxueHong , those commits has been divided into 3 commits.   
Thanks!

zhencliu:
hi @nickzhq and @YongxueHong , what about dividing the commits by device?
e.g. cache-size belongs to 'images' while the others belong to mem_devs,
then we make two commits:

headers:
Update the option values of devices in images
Update the option values of devices in mem_devs

in the body of commit, we can add more details, e.g.
 -object {"qom-type": "memory-backend-memfd", "shared": <bool>}
-blockdev {"driver":"qcow2", "cache-size: <int>}

there might be many commits if we fix one option in one commit

aliang123:
(1/1) Host_RHEL.m9.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.1.0.x86_64.io-github-autotest-qemu.qsd_live_snapshot.q35: PASS (84.67 s)

Acked-by: Aihua Liang<aliang@redhat.com>

hellohellenmao:
> have

@zhencliu No.

virttest/qemu_devices/qdevices.py
```diff
@@ -1579,17 +1588,20 @@ def _cmdline_json(self):
...
         """
-        if "size" in args:
-            args["size"] = int(utils_numeric.normalize_data_size(args["size"],
-                                                                 "B"))
+        for key, val in args.items():
+            if "size" == key:
+                args[key] = int(utils_numeric.normalize_data_size(val, "B"))
+            # "share" from object( "qom-type": "memory-backend-memfd" )
+            elif "share" == key:
+                args[key] = True if val == "yes" else False
```
YongxueHong, line 1604:
```suggestion
                args[key] = val == "yes"
```

virttest/qemu_devices/qdevices.py
```diff
@@ -1160,9 +1160,11 @@ def _cmdline_json(self):
                 command_dict[key] = str(val)
             # disable-legacy from device ("driver": "virtio-scsi-pci")
             # write-cache from device ("driver": "scsi-hd")
-            elif val in ('on', 'yes') and key not in expect_string_val:
+            # disable-modern from device ("driver": "virtio-scsi-pci")
+            elif val in ('on', 'yes', "true") and key not in expect_string_val:
                 command_dict[key] = True
-            elif val in ('off', 'no') and key not in expect_string_val:
+            elif val in ('off', 'no', "false") and \
+                    key not in expect_string_val:
```
YongxueHong, line 1167: I have a bit of confusion about the code and the description of the commit.
The commit said `qdevices: change type of "disable-modern" to integer ` but didn't see the type of `disable-modern` is changed to interger(`int`), or maybe I misunderstood your code.
Would you like to explain?
Thanks.

nickzhq: > I have a bit of confusion about the code and the description of the commit. The commit said `qdevices: change type of "disable-modern" to integer ` but didn't see the type of `disable-modern` is changed to interger(`int`), or maybe I misunderstood your code. Would you like to explain? Thanks.

`disable-modern` is "false" or "true" in string type and it should be converted to boolean.
There would be typo and let me change this typo.
Thanks!

nickzhq:
> hi @nickzhq and @YongxueHong , what about dividing the commits by device? e.g. cache-size belongs to 'images' while the others belong to mem_devs, then we make two commits:
> 
> headers: Update the option values of devices in images Update the option values of devices in mem_devs
> 
> in the body of commit, we can add more details, e.g. -object {"qom-type": "memory-backend-memfd", "shared": <bool>} -blockdev {"driver":"qcow2", "cache-size: <int>}
> 
> there might be many commits if we fix one option in one commit

Hello @zhencliu , I prefer keeping the those commits. But, we can add details into body of commits.  
Example:
  Header:
    qdevices: change value in "cache-size" to integer  
  Body of commit:  
    -blockdev {"driver":"qcow2", "cache-size: <int>} 
    
  What's your opinion? @YongxueHong 
  Thanks!

YongxueHong:
> > hi @nickzhq and @YongxueHong , what about dividing the commits by device? e.g. cache-size belongs to 'images' while the others belong to mem_devs, then we make two commits:
> > headers: Update the option values of devices in images Update the option values of devices in mem_devs
> > in the body of commit, we can add more details, e.g. -object {"qom-type": "memory-backend-memfd", "shared": } -blockdev {"driver":"qcow2", "cache-size: }
> > there might be many commits if we fix one option in one commit
> 
> Hello @zhencliu , I prefer keeping the those commits. But, we can add details into body of commits. Example: Header: qdevices: change value in "cache-size" to integer Body of commit: -blockdev {"driver":"qcow2", "cache-size: }
> 
> What's your opinion? @YongxueHong Thanks!

Hi @zhencliu  and @nickzhq 
Both of them are okay for me. The goal of commit is to track the changes and tell other reviewers the issues that you’re addressing and your motivations clearly, so it is okay that let the reviewers get the points for the changes.
BTW, here I just one suggestion that we can add the motivations, such as adding the keywords for `json command format` since the changes are from the requirement of `json format`

virttest/qemu_devices/qdevices.py
```diff
@@ -1160,9 +1160,11 @@ def _cmdline_json(self):
                 command_dict[key] = str(val)
             # disable-legacy from device ("driver": "virtio-scsi-pci")
             # write-cache from device ("driver": "scsi-hd")
-            elif val in ('on', 'yes') and key not in expect_string_val:
+            # disable-modern from device ("driver": "virtio-scsi-pci")
```
YongxueHong, line 1163: According to the other options `disable-legacy` and `write-cache`, we saw there is a variable `expect_string_val` to handle such behaviors that we don't change the value to the boolean type.
But for the options `disable-legacy` and `disable-modern`, the type of value seems like same from the raw command line, for example, `scsi=on,disable-legacy=off,disable-modern=on`.
Here we have different ways to handle the options `disable-legacy` and `disable-modern` in the json command line.
I don't know if it is a difference between `disable-legacy` and `disable-modern` to handle the value in the json format from the qemu product side.
If it is, I suggest we may remove the comment `# disable-modern from device ("driver": "virtio-scsi-pci")` here to avoid some mumbling.

nickzhq:
@YongxueHong @zhencliu  Please check again, thanks!

virttest/qemu_devices/qdevices.py
```diff
@@ -1153,9 +1161,10 @@ def _cmdline_json(self):
                 command_dict[key] = str(val)
             # disable-legacy from device ("driver": "virtio-scsi-pci")
             # write-cache from device ("driver": "scsi-hd")
-            elif val in ('on', 'yes') and key not in expect_string_val:
+            elif val in ('on', 'yes', "true") and key not in expect_string_val:
                 command_dict[key] = True
-            elif val in ('off', 'no') and key not in expect_string_val:
+            elif val in ('off', 'no', "false") and \
+                    key not in expect_string_val:
```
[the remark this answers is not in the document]
YongxueHong: Hi @nickzhq 
This commit `qdevices: change type of "disable-modern" to boolean` maybe not match the changes, I think we might update the commit.

hellohellenmao:
(1/1) Host_RHEL.m9.u2.ovmf.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.9.2.0.x86_64.io-github-autotest-qemu.cache_sizes_test.q35: PASS (298.74 s)

LGTM.

virttest/qemu_devices/qdevices.py
nickzhq: > According to the other options `disable-legacy` and `write-cache`, we saw there is a variable `expect_string_val` to handle such behaviors that we don't change the value to the boolean type. But for the options `disable-legacy` and `disable-modern`, the type of value seems like same from the raw command line, for example, `scsi=on,disable-legacy=off,disable-modern=on`. Here we have different ways to handle the options `disable-legacy` and `disable-modern` in the json command line. I don't know if it is a difference between `disable-legacy` and `disable-modern` to handle the value in the json format from the qemu product side. If it is, I suggest we may remove the comment `# disable-modern from device ("driver": "virtio-scsi-pci")` here to avoid some mumbling.

The qemu command line exists: `-device '{"id": "virtio_scsi_pci0", "driver": "virtio-scsi-pci", "bus": "pcie-root-port-2", "addr": "0x0", "disable-legacy": "on", "disable-modern": false}' `
@YongxueHong , the comment has been deleted. Thanks.

virttest/qemu_devices/qdevices.py
```diff
@@ -1579,17 +1587,20 @@ def _cmdline_json(self):
...
         """
-        if "size" in args:
-            args["size"] = int(utils_numeric.normalize_data_size(args["size"],
-                                                                 "B"))
+        for key, val in args.items():
+            if "size" == key:
+                args[key] = int(utils_numeric.normalize_data_size(val, "B"))
+            # "share" from object( "qom-type": "memory-backend-memfd" )
+            elif "share" == key:
+                args[key] = val == "yes"
```
zhencliu, line 1603: hi @nickzhq  @YongxueHong , do we need to check “on” and "true" here?

virttest/qemu_devices/qcontainer.py
```diff
@@ -191,7 +191,7 @@ def qemu_version(self):
         return self.__qemu_ver
 
     def initialize_iothread_manager(self, params, guestcpuinfo,
-                                    cmdline_format_cfg=None):
+                                    cmdline_format_cfg={}):
```
zhencliu, line 194: Although it's safe to use {} (read-only) as the default  here, I still suggest using None, and then we set it {} if cmdline_format_cfg is None.

Up to U, it's OK for me too.

virttest/qemu_devices/qdevices.py
nickzhq: > hi @nickzhq @YongxueHong , do we need to check “on” and "true" here?

Hello @zhencliu , for "share", we don't need.

virttest/qemu_devices/qcontainer.py
nickzhq: > Although it's safe to use {} (read-only) as the default here, I still suggest using None, and then we set it {} if cmdline_format_cfg is None.
> 
> Up to U, it's OK for me too.

Hi @zhencliu , if we use None at here, we have to check None and change None to {}. I prefer to set {} at here.
But, thanks your comment.

zhencliu (approved):

zhencliu:
hi @nickzhq  , please paste the results (the qemu cmdline snippets including the related options, e.g. -blockdev '{"node-name": "drive_image1", "driver": "qcow2", "read-only": false, "cache": {"direct": true, "no-flush": false}, "file": "file_image1", "cache-size": "0"}'

nickzhq:
> hi @nickzhq , please paste the results (the qemu cmdline snippets including the related options, e.g. -blockdev '{"node-name": "drive_image1", "driver": "qcow2", "read-only": false, "cache": {"direct": true, "no-flush": false}, "file": "file_image1", "cache-size": "0"}'

[stdlog]     -device '{"id": "virtio_scsi_pci0", "driver": "virtio-scsi-pci", "bus": "pcie-root-port-2", "addr": "0x0", "disable-legacy": "on", "disable-modern": false}'

YongxueHong:
Thanks all, let's merge it.

virttest/qemu_devices/qdevices.py
```diff
@@ -1579,17 +1593,20 @@ def _cmdline_json(self):
...
         :type args: dict
         :return: Converted args.
         :rtype: dict
         """
-        if "size" in args:
-            args["size"] = int(utils_numeric.normalize_data_size(args["size"],
-                                                                 "B"))
+        for key, val in args.items():
+            if "size" == key:
+                args[key] = int(utils_numeric.normalize_data_size(val, "B"))
```
mcasquer, line 1606: Perhaps this should be applied too for `align` option

PR #3564, opened 2022-11-17
Allow the option of creating a vm firstly after loading from file

pevogam:
Loading a running vm's memory from file would require it to be paused first but would also work just fine without any prior vm. In some cases the user has to create a vm and pause it just to be able to satisfy the requirements of previously running and paused vm thus leading to overhead and circumventing more direct use of this functionality.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -4932,7 +4932,8 @@ class VM(virt_vm.BaseVM):
         """
         Override BaseVM restore_from_file method
         """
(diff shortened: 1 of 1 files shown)


pevogam:
@luckyh, @YongxueHong This is a further improvement in the vm migration from tile method that allows us to create and load a vm from file at the same time, see the commit message for details.

luckyh (approved):
LGTM

YongxueHong (approved):
LGTM.

PR #3573, opened 2022-12-02
Remove duplicate definition of driver tag in Disk xml

chloerh:
'drivermetadata' should only include metadata_cache, the rest duplicates 'driver'. The name is more accurate with 'cache'.


nanli1 (approved):

virttest/libvirt_xml/devices/disk.py
```diff
@@ -636,77 +628,27 @@ def __init__(self, virsh_instance=base.base.virsh):
...
-
-            custom_disk.drivermetadata = new_one
+        max_size:
+            Int, attribute of MetadataCache max size
+        max_size_unit:
+            string, attribute of MetadataCache max size unit
         """
 
-        __slots__ = ('attrs', 'metadata_cache')
+        __slots__ = ('max_size', 'max_size_unit')
```
chunfuwen, line 641: <driver name='qemu' type='qcow2'>
driver itself can have attributes like above, how can we set this in MetadataCache class

chunfuwen: Moreover, from xml point view, metadata_cache is better to be defined as separate nested element in  driver tag

dzhengfy: If IIUC, the driver attributes will be set by using accessors.XMLElementDict('driver', self, parent_xpath='/',                                  tag_name='driver'),  the dict method, so we do not need to set them in  MetadataCache class. MetadataCache class is only for setting elements/attributes for  like below part:  
```
  <metadata_cache>
      <max_size unit='bytes'>1234</max_size>
    </metadata_cache>  
```

dzhengfy (commented):
LGTM

dzhengfy (approved):

virttest/libvirt_xml/devices/disk.py
chunfuwen: @dzhengfy 
driver_metadatacache represents complete entity with driver and metadatache, why not open API to set driver attribute?

chloerh: > Moreover, from xml point view, metadata_cache is better to be defined as separate nested element in driver tag

@chunfuwen It's better only if 'driver' tag hasn't been defined for almost a decade and hasn't been referenced hundreds of times. 
We cannot afford to make such a tremendous change of code which could bring in unpredictable risks just for a slight advantage that could be ignored for most cases. It's just uneconomic.

chunfuwen:
I have some confusion about about so called "
Remove duplicate definition of driver tag "
As per guideline , here is how previous implementation works?
```
            custom_disk = Disk(type_name='disk')
            driver_dict = {"name": "qemu","type": "qcow2"}
            new_one = custom_disk.new_drivermetadata(**{"attrs": driver_dict})
        Source of MetadataCache xml class
            metadata_cache_dict = {"max_size": "10", "max_size_unit": "bytes"}
            new_one.metadata_cache = custom_disk.DriverMetadata().new_metadatacache(**metadata_cache_dict)
            custom_disk.drivermetadata = new_one
```
1) Do we see duplicated python code here?
2) Do we see duplicate value defined in cfg?
3) Do we see duplicated attribute in finally generated metadata_cache Disk xml?

chloerh:
> I have some confusion about about so called " Remove duplicate definition of driver tag " As per guideline , here is how previous implementation works?
> 
> ```
>             custom_disk = Disk(type_name='disk')
>             driver_dict = {"name": "qemu","type": "qcow2"}
>             new_one = custom_disk.new_drivermetadata(**{"attrs": driver_dict})
>             metadata_cache_dict = {"max_size": "10", "max_size_unit": "bytes"}
>             new_one.metadata_cache = custom_disk.DriverMetadata().new_metadatacache(**metadata_cache_dict)
>             custom_disk.drivermetadata = new_one
> ```
> 
> 1. Do we see duplicated python code here?
> 2. Do we see duplicate value defined in cfg?
> 3. Do we see duplicated attribute in finally generated metadata_cache Disk xml?

@chunfuwen Do you have other questions of this change, besides the confusion of "duplication"? 
If "No", I'll try to explain the "duplication".
If "Yes", I'd like to resolve your other questions first, if it's ok with you. Thanks.

chunfuwen:
From my observation, it seems that this PR wants to improve current implementation.
I think a few question need clarification before we really take actions:
1)What it doesn't make sense in current implementation?
2)Why you think current change can improve that? and in what ways it improve?

By the way, I want to describe a little more details how previous implementation work?
After metamacache feature introduce into driver, we now have two typical use case in terms of driver
1)use case 1, <driver name='qemu' type='qcow2'>, this will be achieved by disk.driver = {} 
2)user case 2 :
```
  <driver name='qemu' type='qcow2'>
    <metadata_cache>
      <max_size unit='bytes'>1234</max_size>
    </metadata_cache>
  </driver
``
this will be achieved by custom_disk.drivermetadata, which was detailed in previous comment
3) We provide two different API to enable two different use cases, it is very similar to function overloading mechanism  in object-oriented

**Comment by chunfuwen** [comment]
@chloerh 
NO.
Please responds above mentioned questions directly.
Moreover, most importantly, we need keep compatibility with current implementation if you want something change.
That means we still need support below operation :
```
            custom_disk = Disk(type_name='disk')
            driver_dict = {"name": "qemu","type": "qcow2"}
            new_one = custom_disk.new_drivermetadata(**{"attrs": driver_dict})
            metadata_cache_dict = {"max_size": "10", "max_size_unit": "bytes"}
            new_one.metadata_cache = custom_disk.DriverMetadata().new_metadatacache(**metadata_cache_dict)
            custom_disk.drivermetadata = new_one
```

PR #3577, opened 2022-12-06
Enhance the creation of qemu machine dev

luckyh:
* qcontainer: refactor the routine of `machine_by_params`
* qdevices: introduce `QMachine` and apply it to virttest

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -292,6 +292,10 @@ class DevContainer(object):
         # -incoming defer
         if self.has_option('incoming defer'):
             self.caps.set_flag(Flags.INCOMING_DEFER)
+        # -machine memory-backend
+        machine_help = self.execute_qemu("-machine none,help")
+        if re.search(r"memory-backend=", machine_help, re.MULTILINE):
+            self.caps.set_flag(Flags.MACHINE_MEMORY_BACKEND)
         # -object sev-guest
         if self.has_object('sev-guest'):
             self.caps.set_flag(Flags.SEV_GUEST)
@@ -1061,7 +1065,7 @@ class DevContainer(object):
             pci_bridge_type = 'pci-bridge'
         pcie_root_port_params = params.get('pcie_root_port_params')
 
-        def pflash_handler(firmware_name, cmd):
+        def pflash_handler(firmware_name, machine_params):
             """
             Add pflash firmware for EFI boot
             """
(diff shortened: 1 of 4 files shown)


hereischen (approved):

luckyh:
Hi @YongxueHong @PaulYuuu @zhencliu would you mind to help review this PR? thanks!

virttest/qemu_vm.py
```diff
@@ -1517,9 +1513,9 @@ def sort_key(dev):
                 pcics.sort(key=sort_key, reverse=False)
             devices.insert(pcics)
 
-        def add_secure_guest_object(params):
+        def add_secure_guest(params):
```
PaulYuuu, line 1516: I would suggest adding some suffixes, `add_secure_guest` make me so confused, like adding a real guest.

zhencliu:
LGTM for the secure guest part, thanks for the update.

virttest/qemu_vm.py
luckyh: Yes, make sense, have updated the method name to `add_secure_guest_descriptor`.

luckyh:
Rebased to resolved the CI failures.

PaulYuuu (approved):

virttest/qemu_devices/qcontainer.py
```diff
@@ -1433,125 +1412,81 @@ def machine_riscv64_mmio(cmd=False):
...
+        avocado_machine = ''
+        invalid_machine = None
+        if not machine_type:
+            for m_name, m_desc in self.__machines_info.items():
+                if '(default)' in m_desc:
+                    machine_type = m_name
+                    break
             else:
+                if machine_type is None:
+                    machine_type = "<unspecified>"
```
YongxueHong, line 1453:
```suggestion
                machine_type = "<unspecified>"
```

YongxueHong (approved):
LGTM.

virttest/qemu_devices/qcontainer.py
luckyh: Okay, the reason of adding this check is that we may want to distinguish empty strings as the value from the case by not given the param at all, to give users the right hint at the following warning message. We could discuss if needed, thanks!

YongxueHong: > Okay, the reason of adding this check is that we may want to distinguish empty strings as the value from the case by not given the param at all, to give users the right hint at the following warning message. We could discuss if needed, thanks!

All right, It looks good to me.
Thanks.

YongxueHong:
Hi, all 
Thanks all, let's merge it.

PR #3580, opened 2022-12-08
Remove unused iscsi functions

qiankehan:
Remove the functions of setting target acls, set  initiator name of iscsiadm and restarting iscsid.

### DIFF
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -448,15 +448,6 @@ class IscsiTGT(_IscsiComm):
         if self.chap_user not in self.get_chap_accounts():
             LOG.error("Can't find account %s" % self.chap_user)
 
-    def delete_chap_account(self):
-        """
-        Delete the CHAP authentication account
-        """
-        if self.chap_user in self.get_chap_accounts():
-            cmd = "tgtadm --lld iscsi --op delete --mode account"
-            cmd += " --user %s" % self.chap_user
-            process.system(cmd)
-
     def get_target_account_info(self):
         """
         Get the target account information
@@ -627,45 +618,6 @@ class IscsiLIO(_IscsiComm):
                     else:
                         target = None
         return target
-
-    def set_chap_acls_target(self):
-        """
-        set CHAP(acls) authentication on a target.
-        it will require authentication
-        before an initiator is allowed to log in and access devices.
-
-        notice:
-            Individual ACL entries override common TPG Authentication,
-            which can be set by set_chap_auth_target().
-        """
-        # Enable ACL nodes
-        acls_cmd = "targetcli /iscsi/%s/tpg1/ " % self.target
-        attr_cmd = "set attribute generate_node_acls=0"
-        process.system(acls_cmd + attr_cmd)
-
-        # Create user and allow access
-        acls_cmd = ("targetcli /iscsi/%s/tpg1/acls/ create %s:client"
-                    % (self.target, self.target.split(":")[0]))
-        output = process.run(acls_cmd).stdout_text
-        if "Created Node ACL" not in output:
-            raise exceptions.TestFail("Failed to create ACL. (%s)" % output)
-
-        comm_cmd = ("targetcli /iscsi/%s/tpg1/acls/%s:client/"
-                    % (self.target, self.target.split(":")[0]))
-        # Set userid
-        userid_cmd = "%s set auth userid=%s" % (comm_cmd, self.chap_user)
-        output = process.run(userid_cmd).stdout_text
-        if self.chap_user not in output:
-            raise exceptions.TestFail("Failed to set user. (%s)" % output)
-
-        # Set password
-        passwd_cmd = "%s set auth password=%s" % (comm_cmd, self.chap_passwd)
-        output = process.run(passwd_cmd).stdout_text
-        if self.chap_passwd not in output:
-            raise exceptions.TestFail("Failed to set password. (%s)" % output)
-
-        # Save configuration
-        process.system("targetcli / saveconfig")
 
     def set_chap_auth_target(self):
         """
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -38,24 +38,6 @@ def get_image_filename(portal, target, lun=0, user=None, password=None):
     return uri.format(auth=auth, portal=portal, target=target, lun=lun)
 
 
-def restart_iscsid(reset_failed=True):
-    """
-    Restart iscsid service.
-    """
-    path.find_command('iscsid')
-    iscsid = service.Factory.create_service('iscsid')
-    if reset_failed:
-        iscsid.reset_failed()
-    if not iscsid.restart():
-        return False
-    else:
-        # Make sure exist connection is operational after recovery
-        process.run('iscsiadm -m node --rescan', timeout=30,
-                    verbose=False, ignore_status=True,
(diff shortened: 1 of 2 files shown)


qiankehan:
@chunfuwen @dzhengfy @lijinlijin Please help to review it. The RHEL8.8 virtual disk test run \#29 has passed with this PR

chunfuwen:
@qiankehan  can you give more details why remove above functions?
In terms of iscsi, there are multiple jobs ,which use it in various ways.
We need carefully every small step to change iscsi from previous my experience where it cost lots of time for debugging even though small change is made on iscsi.

qiankehan:
> @chunfuwen @dzhengfy @lijinlijin Please help to review it. The RHEL8.8 virtual disk test run `#29` has passed with this PR

1. 51723fcbfb4b6332ec6ff3f55317b493e6c58b13: From the search result of avocado-vt, tp-libvirt, tp-qemu and tp-spice, the function `set_chap_acls_target` and `delete_chap_account` are unused:
```
➜  Software ag '(set_chap_acls_target|delete_chap_account)' avocado-vt tp-libvirt tp-qemu tp-spice 
avocado-vt/selftests/unit/test_iscsi.py
141:    def setup_stubs_delete_chap_account(self, iscsi_obj):

avocado-vt/virttest/iscsi.py
451:    def delete_chap_account(self):
631:    def set_chap_acls_target(self):
```
2. After the former commit,  [targetcli ACLs](https://github.com/open-iscsi/targetcli-fb/blob/master/targetcli.8#L182) is not used. ACLs are to controll what iscsi initiator could connect the iscsi target. If they are not used, all initiator name could connect to the target. So we don't have to change the initiator name.
3. Since the former commit, we don't change the initiator name. We don't have to restart iscsid. `iscsiadm` will manage all the things of iscsi. And the iscsid could be started by the socket when iscsiadm connect to it[1]. 

[1]: https://github.com/open-iscsi/open-iscsi/commit/5d0e19fcc1cea77a72647cf96c5d3d773e8ee277

I have run the most of single host iscsi cases on iscsi-initiator-utils-6.2.1.4-4 qemu-kvm-6.2.0-27 libvirt-8.0.0-12, none of them fails due to iscsi:
```
for i in $(grep iscsi cases |grep -vE '(migrat|multi)');do avocado run --vt-type libvirt --vt-machine-type q35 $i;done
```
The failed cases are caused by:
- Fail to create qed image
- Fail to set sys_rawio capability

qiankehan:
[result.log](https://github.com/avocado-framework/avocado-vt/files/10264268/result.log)
The test result

qiankehan:
> restart_iscsid ACL initiatior Those functions are very natural features for iscsi, so even if we don't use them currently, I don't suggest we need remove them

What is the exact meaning of **natural**? Did you mean it is open-iscsi related?

In my opinion, to keep it or remove a function/class, just need to check if it is being or will be used.
I don't see any possibility that they will be used in tp-libvirt or tp-qemu. What's more, I looked up the commits and found they are never used:
```
➜  tp-libvirt git:(master) PAGER=cat; git log -1 --format="%h    %s"
daaeb66ae    Merge pull request #4764 from xiaodwan/add_kubevirt_support

➜  tp-libvirt git:(master) PAGER=cat; for i in $(git log --format='%h');do git show $i|grep -qiE "(set_chap_acls_target|delete_chap_account|restart_iscsid)"; if [ $? -eq 0 ];then git log $i -1 --format="%h    %s";fi;done

➜  tp-qemu git:(master) PAGER=cat; git log -1 --format="%h    %s"
13725995    Merge pull request #3606 from 6-dehan/new-api-cpustats

➜  tp-qemu git:(master) PAGER=cat; for i in $(git log --format='%h');do git show $i|grep -qiE "(set_chap_acls_target|delete_chap_account|restart_iscsid)"; if [ $? -eq 0 ];then git log $i -1 --format="%h    %s";fi;done

```

chunfuwen:
I just throw two cents here about my personal overview about this PR.
1)I don't see clear positive impact this PR will bring in
2)Look like this PR remove 3 methods to reduce maintenance effort, but if we look at virttest/iscsi.py commit history, we almost has no change for this file after it is merged
3)If we check tp-libvirt code only, there are 500+ places(egrep -nr 'iscsi' --include=*.py |wc -l) ,which is related to usage of iscsi. And iscsi.py are widely used in various ways. In order to make sure the change will not bring regression, lots of additional test efforts are needed,and plus maintainers's efforts for reviewing
4) In the past, there was one ever tiny change in iscsi.py, during debugging, even finding that avocado run will be hung under special condition, which we may believe it is not related iscsi.py.  So that actually happened.

qiankehan:
> I just throw two cents here about my personal overview about this PR. 1)I don't see clear positive impact this PR will bring in 2)Look like this PR remove 3 methods to reduce maintenance effort, but if we look at virttest/iscsi.py commit history, we almost has no change for this file after it is merged 3)If we check tp-libvirt code only, there are 500+ places(egrep -nr 'iscsi' --include=*.py |wc -l) ,which is related to usage of iscsi. And iscsi.py are widely used in various ways. In order to make sure the change will not bring regression, lots of additional test efforts are needed,and plus maintainers's efforts for reviewing 4) In the past, there was one ever tiny change in iscsi.py, during debugging, even finding that avocado run will be hung under special condition, which we may believe it is not related iscsi.py. So that actually happened.

1. the "clear positive impact" is **keep it more simple and clear**
2. Do you mean the **virttest/iscsi.py** is too old to change? I don't think it is a good idea for an active project
3. Sure. This PR is [widely tested](https://github.com/avocado-framework/avocado-vt/pull/3580#issuecomment-1358759626) as well. If any iscsi cases are missing, you can list them, and then I will run the tests.
4. Even though **iscsi.py** is too complex to change, we should not refuse to change. I know your concerns about the possible regressions to master branch. How about creating a develop branch and merging it to that branch if no questions on the codes and test themselfves? This develop branch could be used for code refactoring like this PR. When the develop branch pass the full tests, we can merge it to master.

chunfuwen:
@clebergnu could you give your thoughts about this PR?

virttest/iscsi.py
```diff
@@ -831,9 +737,6 @@ def export_target(self):
         # Save configuration
         process.system("targetcli / saveconfig")
 
-        # Restart iSCSI service
-        restart_iscsid()
```
chunfuwen, line 835: How can you make sure restart_iscsid() here is not necessary?

qiankehan: > How can you make sure restart_iscsid() here is not necessary?

The caller function **export_target** sets up the iscsi target, as well as the server side, while the iscsid is the initiator  , as well as the client side. Since they don't establish a connection before **export_target**, the client side iscsid doesn't need to restart. Just like you don't have to restart your browser for a new website.

chunfuwen (approved):
LGTM

dzhengfy:
@qiankehan I added some descriptions about your test results like you said below:
Tested with some cases in virtual disk/scsi_hostdev/backingchain/storage features. No regression.
` (1/1) type_specific.io-github-autotest-libvirt.virtual_disks.multivms.coldplug.vms_sharable_test.block_type: PASS (50.96 s)`

Based on above results, I am confident it is low risk to merge this one. Thanks.

dzhengfy (approved):
LGTM

PR #3561, opened 2022-11-11
iscsi.py: add authentication=1 to set_chap_auth_target function

meinaLi:
If the set_chap_auth_target function doesn't include authentication=1 attribute, this means it doesn't actually use chap function. This part has been more rigorously checked In new qemu version, so update it to fix some authentication failure.

Signed-off-by: Meina Li <meili@redhat.com>

PR #3582, opened 2022-12-12
Revert "iscsi.py: add authentication=1 to set_chap_auth_target function"

chunfuwen:
Reverts avocado-framework/avocado-vt#3561

There is some confused result after this PR is merged.
Moreover, even if we need authentication=1, we may not hard code the attribute, instead set this attribute under some condition, e.g. if enable_chap_authentication = 'yes', attr_cmd += "authentication=1


chunfuwen:
@dzhengfy 
@meinaLi 
please have a check

kylazhang (approved):

meinaLi (approved):

chunfuwen:
this is urgent need, let's merge it and trigger some test runs

PR #3585, opened 2022-12-13
fix travics ci failure due to loader module

chunfuwen:


chunfuwen:
this is not needed any more

Issue #3583, opened 2022-12-12
qemu loadvm error not detected


pevogam:
The current avocado-vt version on "master" does not detect error conditions from qemu's "loadvm" command, it just continues.

Easy way to reproduce this issue is to e.g. add an extra serial port to a VM and later on try without the config change:

```diff
 # You can assign more than one isa serial to guest with this parameter.
 # like:
 # serials = "S1 S2 S3"
-serials = "serial0"
+serials = "serial0 serial1"
 serial_type = "isa-serial"
```

This resulted in this error log:

[stdlog] 2022-12-05 13:16:46,775 avocado.virttest.qemu_monitor DEBUG| (monitor mis1.hmp1) Sending command 'loadvm on_customize'
[stdlog] 2022-12-05 13:16:46,775 avocado.virttest.qemu_monitor DEBUG| Send command: loadvm on_customize
[stdlog] 2022-12-05 13:16:53,288 avocado.virttest.qemu_monitor DEBUG| (monitor mis1.hmp1) Response to 'loadvm on_customize'
[stdlog] 2022-12-05 13:16:53,289 avocado.virttest.qemu_monitor DEBUG| (monitor mis1.hmp1) Unknown savevm section or instance 'serial' 1. Make sure that your current VM setup matches your saved VM setup, including any hotplugged devices
[stdlog] 2022-12-05 13:16:53,289 avocado.virttest.qemu_monitor DEBUG| (monitor mis1.hmp1) Error: Error -22 while loading VM state
[stdlog] 2022-12-05 13:16:53,289 avocado.virttest.qemu_monitor DEBUG| (monitor mis1.hmp1) Sending command 'cont'
[stdlog] 2022-12-05 13:16:53,290 avocado.virttest.qemu_monitor DEBUG| Send command: cont

Even if configuration changes of this kind might not be too frequent it is better if the error is detected instead of silently ignored.

pevogam:
Must hopefully be fixed with PR https://github.com/avocado-framework/avocado-vt/pull/3592.

pevogam:
Closing - fixed by https://github.com/avocado-framework/avocado-vt/pull/3592.

PR #3592, opened 2022-12-14
Raise an error if any errors are detected in monitor command's output

pevogam:
We raise a specialized exception for monitor errors in a manner similar to the one for QMP monitors.

This fixes issue #3583.

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -61,6 +61,10 @@ class MonitorProtocolError(MonitorError):
     pass
 
 
+class MonitorCmdError(MonitorError):
+    pass
+
+
 class MonitorNotSupportedError(MonitorError):
     pass
 
@@ -81,10 +85,22 @@ class MonitorNotSupportedMigCapError(MonitorNotSupportedError):
     pass
 
 
-class QMPCmdError(MonitorError):
+class HumanCmdError(MonitorCmdError):
+
+    def __init__(self, cmd, out):
+        MonitorError.__init__(self, cmd, out)
+        self.cmd = cmd
+        self.out = out
+
+    def __str__(self):
+        return ("Human monitor command %r failed    "
+                "error message: %r)" % (self.cmd, self.out))
+
+
+class QMPCmdError(MonitorCmdError):
 
     def __init__(self, cmd, qmp_args, data):
-        MonitorError.__init__(self, cmd, qmp_args, data)
+        MonitorCmdError.__init__(self, cmd, qmp_args, data)
         self.cmd = cmd
         self.qmp_args = qmp_args
         self.data = data
@@ -923,6 +939,8 @@ class HumanMonitor(Monitor):
             if s:
                 if o:
                     self._log_response(cmd, o, debug)
+                    if "Error: " in o:
+                        raise HumanCmdError(cmd, o)
                 return o
             else:
                 msg = ("Could not find (qemu) prompt after command '%s'. "


virttest/qemu_monitor.py
```diff
@@ -61,6 +61,18 @@ class MonitorProtocolError(MonitorError):
     pass
 
 
+class MonitorCmdError(MonitorError):
```
YongxueHong, line 64: Hi @pevogam 
Thanks for pointing out this mistake.
I have one question the `class MonitorCmdError` looks like includes the `class QMPCmdError` https://github.com/avocado-framework/avocado-vt/blob/f618e085b9203e9d2210252bedf1664049b83518/virttest/qemu_monitor.py#L84 by comparing the job what they handle.
Maybe we should find a way to distinguish them or make `QMPCmdError` inherit from this new class.
Or new `class HMPCmdError(MonitorError):` ?

pevogam: Hi @YongxueHong and thanks for reviewing!

> Maybe we should find a way to distinguish them or make QMPCmdError inherit from this new class.

This looks like the easier option without having to create an entire new exception hierarchy just for the human monitor. It also doesn't make sense to create new exception classes if they don't bring any additional utility and distinction in comparison to the old ones so this sounds like the better option for me.

> Or new class HMPCmdError(MonitorError): ?

This would end up branching out an entire HMP exception hierarchy and so far I can't see what we would gain from having identically behaving exceptions for QMP and HMP that are then just separated by name.

YongxueHong: > Hi @YongxueHong and thanks for reviewing!
> 
> > Maybe we should find a way to distinguish them or make QMPCmdError inherit from this new class.
> 
> This looks like the easier option without having to create an entire new exception hierarchy just for the human monitor. It also doesn't make sense to create new exception classes if they don't bring any additional utility and distinction in comparison to the old ones so this sounds like the better option for me.
> 

Hi @pevogam 
From my considering, how do we give the definition both of `MonitorCmdError` and `QMPCmdError` ?
I mean which scenario should call the `MonitorCmdError`  and which should call `QMPCmdError` ?
Yeah, for the `QMPCmdError`, I can get the purpose from the naming, but for the `MonitorCmdError`, I can not sure, if there are other monitor types involved in the future, like `XXXMonitor` then can we call the `MonitorCmdError` by default?

BTW there are two kinds of monitor inherited from `class Monitor(object)`, those are `class QMPMonitor(Monitor):` and `class HumanMonitor(Monitor):` so I think whether we can also refer to this way to design for the part of CmdError?

Thanks.
> > Or new class HMPCmdError(MonitorError): ?
> 
> This would end up branching out an entire HMP exception hierarchy and so far I can't see what we would gain from having identically behaving exceptions for QMP and HMP that are then just separated by name.

pevogam: I agree with you if separating `QMPCmdError` and `HumanCmdError` makes sense as semantics. At present the exceptions are identical in structure and we are only creating them as labels with no meaningful separation taking place. In short, we would define `QMPCmdError` and `HumanCmdError` classes only to follow the heuristics from separate monitor types, yet the classes themselves are both rather a shared `MonitorCmdError` with no distinction. So I would suggest `QMPCmdError` subclassing the `MonitorCmdError` or being dropped altogether and both monitor classes just using the `MonitorCmdError` type. The rest would be artificial separation of concerns. What do you think?

YongxueHong (commented):
Hi @pevogam 
My bad, I forget to submit the review. I thought it has been submitted.

pevogam:
Hi @YongxueHong I pushed the changes you requested. There are indeed semantic differences in the overall `QMPCmdError` construction vs the one we need for the `HumanCmdError` so I decided to go ahead with your suggestion even though I didn't quite get your arguments above. I hope we are on the same page now.

YongxueHong (approved):
LGTM.

luckyh (approved):

luckyh:
Thanks to all.

PR #3597, opened 2023-01-05
Add one simple method to check whether disk partition in vm can be normally used

chunfuwen:

### DIFF
--- a/virttest/utils_libvirt/libvirt_disk.py
+++ b/virttest/utils_libvirt/libvirt_disk.py
@@ -586,3 +586,30 @@ def fill_null_in_vm(vm, target, size_value=500):
             raise exceptions.TestError("Error happened when executing command:\n%s" % cmd)
     except Exception as e:
         raise exceptions.TestError(str(e))
+
+
+def check_virtual_disk_io(vm, partition):
+    """
+    Check if the disk partition in vm can be normally used.
+
+    :param vm: Vm instance
+    :param partition: the disk partition in vm to be checked.
+    :return: if the disk can be used, return True
+    """
+    session = None
+    try:
+        session = vm.wait_for_login()
+        cmd = ("fdisk -l /dev/{0} && mkfs.ext4 -F /dev/{0} && "
+               "mkdir -p test && mount /dev/{0} test && echo"
+               " teststring > test/testfile && umount test"
+               .format(partition))
+        status, output = session.cmd_status_output(cmd)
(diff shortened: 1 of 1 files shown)


chunfuwen:
@dzhengfy 
@Yingshun 
please take  a review.
This is  comment follow-up for PR:https://github.com/autotest/tp-libvirt/pull/4614 to separate one method under avocado-vt

dzhengfy (approved):
LGTM

Yingshun:
@nanli1 Could you please review this pr? Thanks!

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -586,3 +586,30 @@ def fill_null_in_vm(vm, target, size_value=500):
...
+    """
+    Check if the disk partition in vm can be normally used.
+
+    :param vm: Vm instance
+    :param partition: the disk partition in vm to be checked.
+    :return: if the disk can be used, return True
+    """
+    session = None
+    try:
+        session = vm.wait_for_login()
```
nanli1, line 601: Is that OK if check vm state and start vm before login,  because start vm is a mandatory step before this function ?

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -586,3 +586,30 @@ def fill_null_in_vm(vm, target, size_value=500):
             raise exceptions.TestError("Error happened when executing command:\n%s" % cmd)
     except Exception as e:
         raise exceptions.TestError(str(e))
+
+
+def check_virtual_disk_io(vm, partition):
```
nanli1, line 591: How about set a default value for partition, such as vdb or sdb

virttest/utils_libvirt/libvirt_disk.py
chunfuwen: Here vm.wait_for_login() happen because we need get one session to access VM internal. 
Checking VM state is more specific to test case,that often happen in your tp-libvirt test case code.

virttest/utils_libvirt/libvirt_disk.py
chunfuwen: Frankly speaking, I don't see benefit to set partition as one default value.
Without setting default value for partition, it help remind users that partition parameter need be given in any situation.

chunfuwen:
merge it since all reviewers approved

PR #3605, opened 2023-01-20
numa_hmat: updates initiator option

mcasquer:
numa_hmat: updates initiator option

According to qemu-7.2 the initiator value
is not required anymore in -numa node option.
It is still mandatory to provide it for -numa hmat-lb, so this updated the script in order to take a new
configuration value from the auto case.

ID: 2158423

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1977,7 +1977,7 @@ class VM(virt_vm.BaseVM):
                 if not numa_params.get('numa_hmat_lb'):
                     continue
                 nodeid = numa_params['numa_nodeid']
-                initiator = numa_params.get('numa_initiator', nodeid)
+                initiator = numa_params.get('numa_hmat_lb_initiator', nodeid)
                 for hmat_lb in numa_params.objects('numa_hmat_lb'):
                     devices.insert(
                         devices.numa_hmat_lb_define_by_params(
--- a/virttest/shared/cfg/base.cfg
+++ b/virttest/shared/cfg/base.cfg
@@ -249,6 +249,8 @@ smp = 1
 # Define initiator for numa node, it has to be a node with cpu, and if current
 # numa node has cpus the initiator has to be itself.
 # numa_initiator_node0 = 0
+# From qemu 7.2 version onwards the initiator value is not required anymore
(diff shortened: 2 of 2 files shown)


mcasquer:
@PaulYuuu @luckyh please could you review this PR? Thanks!

virttest/qemu_vm.py
```diff
@@ -1977,7 +1977,7 @@ def add_secure_guest_object(params):
                 if not numa_params.get('numa_hmat_lb'):
                     continue
                 nodeid = numa_params['numa_nodeid']
-                initiator = numa_params.get('numa_initiator', nodeid)
+                initiator = numa_params.get('numa_hmat_lb_initiator', nodeid)
```
PaulYuuu, line 1980: I'm curious about the impact of continuing to use the old naming, and we also need to consider the compatibility of older versions.

yanan-fu: @PaulYuuu  It need distinguish the `initiator` option for 
1. `-numa node` :  use param `numa_initiator` 
2. `-numa hmat-lb`:  use param `numa_hmat_lb_initiator`

version control is in test case level to decide which option be used or not for different device  :)

yanan-fu (approved):
Ack

virttest/qemu_vm.py
PaulYuuu: OK, then I would suggest also updating the cfg.
https://github.com/avocado-framework/avocado-vt/blob/f618e085b9203e9d2210252bedf1664049b83518/virttest/shared/cfg/base.cfg#L253-L255

yanan-fu: good catch ~

mcasquer: @PaulYuuu updated now, and numa hmat-lb block too, thanks! :grin:

PaulYuuu (approved):

luckyh (approved):

mcasquer:
@PaulYuuu @luckyh @yanan-fu since this PR has all your approvals, I guess it could be merged :relaxed:

luckyh:
> @PaulYuuu @luckyh @yanan-fu since this PR has all your approvals, I guess it could be merged relaxed

Sure, thanks.

PR #3622, opened 2023-02-15
fetch_attrs:Do not return invalid values 'None'

chloerh:
When the value we get with a certain key turns out to be None, we just ignore it and not return the data pair.


Yingshun (approved):

nanli1 (approved):

chloerh:
@Yingshun @nanli1 I've been thinking, should we just exclude all 'empty' value from being fetched? Like {} and []? What's your opinions?

PR #3632, opened 2023-02-22
libvirt: add helper function to find matches in Libvirt XML

smitterl:
Tested via https://github.com/autotest/tp-libvirt/pull/4770

XML attributes are unordered. Therefore, the existing function `check_dumpxml`
can't be used if attribute order changes.

Add new function that leverages the existing xmltreefile property to
match xpath expressions instead.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -3954,6 +3954,39 @@ def check_domuuid_compliant_with_rfc4122(dom_uuid_value):
     return dom_uuid_segments[2].startswith('4') and dom_uuid_segments[3][0] in '89ab'
 
 
+def check_xpaths(xml, xpaths, text=None):
+    """
+    Check if the xml has elements that match all xpaths and optionally
+    contain a given text.
+
+    :param xml: Libvirt XML instance
+    :param xpaths: list of xpath expressions as supported by etree.ElementTree
+    :param text: string to match the text node of matching elements
+                 per default this is not checked if not given
+    :return: None
+    :raises: TestFail if no element matches all xpaths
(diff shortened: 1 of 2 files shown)


smitterl:
Tested via https://github.com/autotest/tp-libvirt/pull/4770

virttest/utils_test/libvirt.py
```diff
@@ -3954,6 +3954,39 @@ def check_domuuid_compliant_with_rfc4122(dom_uuid_value):
     return dom_uuid_segments[2].startswith('4') and dom_uuid_segments[3][0] in '89ab'
 
 
+def check_xpaths(xml, xpaths, text=None):
```
chunfuwen, line 3957: @smitterl  thank your efforts in bring this.
Just a open discussion:
1)Before avocado-vt PR:3615, VMXML.new_from_dumpxml will output device xml attributes in the order where attribute is set when creating device xml.
2)After avocado-vt PR:3615, VMXML.new_from_dumpxml will output device xml attributes in the order strictly complying with virsh command output : virsh dumpxml(I assumed it is from yesterday's email chain discussion)
3)In tp-libvirt, we don't have branches or releases out, so tp-libvirt just use master branch to be compatibility with avocado-vt master branch(that is our currently libvirt CI strategy). Therefore,if  the change happen in avocado-vt master, we just change tp-libvirt to match the change in avocado-vt master branch. That means specific to input devices issue this time, independent checking order may not be necessary.
4) I think this method can be useful , considering  future use cases

chloerh (approved):

chunfuwen (approved):
LGTM
This is useful as a general helper

PR #3637, opened 2023-03-06
vt_utils/cpu.py: Add a new utility for the CPU

nickzhq:
Signed-off-by: Houqi (Nick) Zuo [hzuo@redhat.com](mailto:hzuo@redhat.com)

### DIFF
--- a/virttest/vt_utils/cpu.py
+++ b/virttest/vt_utils/cpu.py
@@ -1,0 +1,144 @@
+#
+# library for cpu related helper functions
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; specifically version 2 of the License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Red Hat (c) 2023 and Avocado contributors
+# Author: Houqi Zuo <hzuo@redhat.com>
+import re
+
+from avocado.utils import process
+
+
+def get_cpu_info():
+    """
+    Return information about the CPU information.
+
+    :return: cpu information
+    :rtype: list[dict]
+    """
+    cpu_info = list()
+    with open("/proc/cpuinfo") as fd:
+        processors = fd.read().strip().split("\n\n")
+    for processor in processors:
+        if processor != "":
+            info = dict(map(lambda x: [i.strip() for i in x.split(":", 1)],
+                            processor.split("\n")))
+            cpu_info.append(info)
+
+    return cpu_info
+
+
+def get_cpu_model_name():
+    """
+    Return physical cpu model.
+
+    :return: cpu model name
+    :rtype: string
+    :raises: An OSError will be raised if it's NOT available on the platform
+    """
+    cpu_model = process.run("lscpu").stdout_text
+    cpu_model_re = "(?m)^[mM]odel name:.*$"
+    cpu_model = re.search(cpu_model_re, cpu_model)
+    if not cpu_model:
+        raise OSError("The cpu model name was NOT found!")
+    cpu_model = cpu_model.groups()[0]
+    return cpu_model.split(":", 1)[-1].strip()
+
+
+def get_cpu_flags():
+    """
+    Return a list of the CPU flags.
+
+    :return: cpu flags
+    :rtype: list[string]
+    :raises: An OSError will be raised if it's NOT available on the platform
+    """
+    cpu_flags_re = "(?m)^flags\s+:\s+([\w\s]+)$"
+    with open("/proc/cpuinfo") as fd:
+        cpu_info = fd.read()
+    cpu_flags = re.search(cpu_flags_re, cpu_info)
+    if not cpu_flags:
+        raise OSError("The cpu flags were NOT found!")
+    cpu_flags = cpu_flags.groups()[0]
+    return re.split("\s+", cpu_flags.strip())
+
+
+def get_cpu_features():
+    """
+    Return a list of the CPU features.
+
+    :return: cpu features
+    :rtype: list[string]
+    :raises: An OSError will be raised if it's NOT available on the platform
+    """
+    cpu_features_re = "(?m)^[fF]eatures\s+:\s+([\w\s]+)$"
+    with open("/proc/cpuinfo") as fd:
+        cpu_info = fd.read()
+    cpu_features = re.search(cpu_features_re, cpu_info)
+    if not cpu_features:
+        raise OSError("The cpu features were NOT found!")
+    cpu_features = cpu_features.groups()[0]
+    return re.split("\s+", cpu_features.strip())
+
+
+def get_cpu_vendor_id():
+    """
+    Return the name of the CPU vendor ID.
+
+    :return: the name of the CPU vendor ID in string
+    :rtype: string
+    :raises: An OSError will be raised if it's NOT available on the platform
+    """
+    vendor_re = "(?m)^vendor_id\s+:\s+(\w+)$"
+    with open("/proc/cpuinfo") as fd:
+        cpu_info = fd.read()
+    vendor_id = re.search(vendor_re, cpu_info)
+    if not vendor_id:
+        raise OSError("The vendor id was NOT found!")
+    vendor_id = vendor_id.groups()[0]
+    return vendor_id
+
+
+def get_cpu_family():
+    """
+    Return the name of cpu family.
+
+    :return: the name of the cpu family in string
+    :rtype: string
+    :raises: An OSError will be raised if it's NOT available on the platform
+    """
+    cpu_family_re = "cpu family\s+:\s+(\w+)"
+    with open("/proc/cpuinfo") as fd:
+        cpu_info = fd.read()
+    cpu_family = re.search(cpu_family_re, cpu_info)
+    if not cpu_family:
+        raise OSError("The cpu family was NOT found!")
+    cpu_family = cpu_family.groups()[0]
+    return cpu_family
+
+
+def get_cpu_stepping():
+    """
+    Return the name of cpu stepping.
+
+    :return: the name of the cpu stepping in string
+    :rtype: string
+    :raises: An OSError will be raised if it's NOT available on the platform
+    """
+    stepping_re = "(?m)^stepping\s+:\s+(\w+)$"
+    with open("/proc/cpuinfo") as fd:
+        cpu_info = fd.read()
+    stepping = re.search(stepping_re, cpu_info)
+    if not stepping:
+        raise OSError("The cpu stepping was NOT found!")
+    stepping = stepping.groups()[0]
+    return stepping
--- a/spell.ignore
+++ b/spell.ignore
@@ -1177,6 +1177,7 @@ hotpluggable
 hotplugging
 hotunplug
 Hotunplug
+Houqi
 hpt
 Hpt
 html
@@ -1190,6 +1191,7 @@ hwaddr
 hwtype
 hyperv
 hypervisors
+hzuo
 IB
 ibm
 ic
@@ -2082,4 +2084,5 @@ zfgrepi
 zfile
 zgrep
 zgrepi
+Zuo
 zzz


nickzhq:
@YongxueHong @luckyh @yanan-fu @zhencliu  please take a look, thanks!

virttest/vt_utilities/cpu.py
```diff
@@ -0,0 +1,173 @@
...
+        Update the check pattern to a certain order and format
+        """
+        pattern_list = sorted(re.split(",", flags.strip()))
+        pattern = r"(\b%s\b)" % pattern_list[0]
+        for i in pattern_list[1:]:
+            pattern += r".+(\b%s\b)" % i
+        return pattern
+
+    if ARCH in ('ppc64', 'ppc64le'):
+        return []  # remove -cpu and leave it on qemu to decide
```
YongxueHong, line 117: We should update this part to adopt the current requirements of designing we wanted. This part of handling should be not decided by the virtualization testing layer.

nickzhq:
@YongxueHong @luckyh @yanan-fu @zhencliu please take a look, thanks!

virttest/vt_utilities/cpu.py
```diff
@@ -0,0 +1,174 @@
...
+                cpu_support_model.append(cpu_model)
+
+    return cpu_support_model
+
+
+def get_cpu_models_on_ppc64():
+    """
+    Get cpu model from ppc64/ppc64le cpuinfo
+    """
+    return []  # if platform is the ppc64/ppc64le
```
YongxueHong, line 138: Hi @maxujun 
Does the power platform support get the CPU mode on the host?
Thanks.

nickzhq: > Hi @maxujun Does the power platform support get the CPU mode on the host? Thanks.

Hello @maxujun , can you give some comments?
Thanks!

maxujun: Yes, support.

nickzhq: hello @zhencliu @luckyh @yanan-fu ,
do you guys have any comments? Thanks!

maxujun: Not support get cpu backward-compatibility mode list from host cpu info for power.only can get host self cpu model.

virttest/vt_utils/cpu.py
```diff
@@ -0,0 +1,176 @@
...
+    :return: cpu information
+    :rtype: dict
+    """
+    cmd = "lscpu | tee"
+    output = process.run(cmd, shell=True,
+                         ignore_status=True).stdout_text.splitlines()
+    return dict(map(lambda x: [i.strip() for i in x.split(":", 1)], output))
+
+
+def get_cpu_models():
```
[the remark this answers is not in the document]
zhencliu: It depends, we may introduce a lib function to get the cpu model name, i.e. Milan/Rome

snippet from lscpu output:
Model name:            AMD EPYC 7313 16-Core Processor

The use case here is feature owner may want to run tests on specific cpu, e.g. Milan (7313 here is Milan (codename)) 

cc @YongxueHong @nickzhq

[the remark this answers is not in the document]
nickzhq: Is it possible to have two independent functions in this module? One for `get_physical_cpu_model`, another for `get_cpu_codename`.

cc @YongxueHong @zhencliu

virttest/vt_utils/cpu.py
```diff
@@ -0,0 +1,77 @@
...
+def get_cpu_model():
+    """
+    Get physical cpu model.
+
+    :return: cpu model
+    :rtype: string
+    """
+    cpu_model = process.run("lscpu|grep \"^Model name:\"")
+    cpu_model = cpu_model.split(":", 1)[-1].lstrip()
+    return cpu_model
```
YongxueHong, line 43:
```
>>> from virttest.vt_utils import cpu
>>> cpu.get_cpu_model()
  File "/home/yhong/kar/workspace/avocado-vt/virttest/vt_utils/cpu.py", line 41, in get_cpu_model
    cpu_model = process.run("lscpu|grep \"^Model name:\"")
...
FileNotFoundError: [Errno 2] No such file or directory (lscpu|grep "^Model name:"): 'lscpu|grep'
```
Looks like it does not work.

nickzhq:
@YongxueHong @zhencliu @yafu-1  Please take a look, thanks!

nickzhq:
@YongxueHong @zhencliu @yafu-1 Please take a look, thanks!

virttest/vt_utils/cpu.py
```diff
@@ -0,0 +1,103 @@
...
+
+    :return: cpu information
+    :rtype: dict
+    """
+    cpu_info = dict()
+    with open("/proc/cpuinfo") as fd:
+        cpu_info_lines = fd.readlines()
+    for line in cpu_info_lines:
+        if line == "\n":
+            break
```
zhencliu, line 33: hi @nickzhq , I am sorry that I may make a wrong comment about cpu info in the afternoon. For this function, we may return a list of dicts, in which each dict contains the information of one processor. Why? Because these information may be used for some purposes, e.g. to get the count of sockets, with shell, we can do it in this way: grep 'physical id' /proc/cpuinfo | sort | uniq | wc -l, i.e. we need all cpus' information.

or we can depend on lscpu to get the count of sockets. cc @YongxueHong

virttest/vt_utils/cpu.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    Return information about the CPU information.
+
+    :return: cpu information
+    :rtype: dict
+    """
+    cpu_info = dict()
+    with open("/proc/cpuinfo") as fd:
+        cpu_info_lines = fd.readlines()
+    for line in cpu_info_lines:
```
zhencliu, line 31: for line in fd:

nickzhq:
@YongxueHong @zhencliu @yafu-1 Please take a look, thanks!

YongxueHong (approved):
LGTM.

nickzhq:
Hello @zhencliu @yanan-fu , do you guys have any comments? Thanks!

zhencliu (approved):

nickzhq:
@luckyh @yanan-fu Please take a look, thanks!

zhencliu:
hi @nickzhq , would you please add two functions:
get_cpu_family
get_cpu_stepping

Both are being used by sevctl command in future, cc @YongxueHong

virttest/vt_utils/cpu.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    Return the name of cpu family.
+
+    :return: the name of the cpu family in string
+    :rtype: string
+    """
+    cpu_family_re = "cpu family\s+:\s+(\w+)"
+    with open("/proc/cpuinfo") as fd:
+        cpu_info = fd.read()
+    cpu_family = re.findall(cpu_family_re, cpu_info)
```
zhencliu, line 116: we may also need to check if cpu_family is empty and return an OSError if it is, like what we did in other functions

virttest/vt_utils/cpu.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def get_cpu_family():
+    """
+    Return the name of cpu family.
+
+    :return: the name of the cpu family in string
+    :rtype: string
+    """
+    cpu_family_re = "cpu family\s+:\s+(\w+)"
```
zhencliu, line 113: do you think 'c' in 'cpu family' could be a capital letter?

Please also keep an eye on stepping

virttest/vt_utils/cpu.py
nickzhq: @zhencliu  After I made a test, the exception `OSError` should be added at here.

virttest/vt_utils/cpu.py
nickzhq: > do you think 'c' in 'cpu family' could be a capital letter?
> 
> Please also keep an eye on stepping

After I made check, the 'C' is NOT necessary at here since the x86_64 platform only has `cpu family`.

virttest/vt_utils/cpu.py
```diff
@@ -0,0 +1,137 @@
...
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; specifically version 2 of the License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Red Hat (c) 2023
```
YongxueHong, line 14:
```suggestion
# Copyright: Red Hat (c) 2023, The Avocado Authors.
```

nickzhq:
@YongxueHong @zhencliu Please take a look, thanks!

yanan-fu:
Suggest using the MULTILINE mode for regex.

nickzhq:
> Suggest using the MULTILINE mode for regex.

Done. Please check again, thanks! @yanan-fu

virttest/vt_utils/cpu.py
```diff
@@ -0,0 +1,137 @@
...
+
+def get_cpu_flags():
+    """
+    Return a list of the CPU flags.
+
+    :return: cpu flags
+    :rtype: list[string]
+    :raises: An OSError will be raised if it's NOT available on the platform
+    """
+    cpu_flags_re = "flags\s+:\s+([\w\s]+)\n"
```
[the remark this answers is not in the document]
nickzhq: `vmx flags` will be added. @yanan-fu

nickzhq:
@zhencliu @YongxueHong @yanan-fu Please take a look, thanks!

nickzhq:
@zhencliu @YongxueHong @yanan-fu Please take a look, thanks!

virttest/vt_utils/cpu.py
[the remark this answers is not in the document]
yanan-fu: I am not requesting for `vmx flags`.  I mean that the initial `get_cpu_flags` include both `flags` and `vmx flags`  when `(?m)` was not used before.

For `vmx flags`,  no usage AFAIK, and i think it is not suitable to be added in this file. @zhencliu @YongxueHong What's your opinion ?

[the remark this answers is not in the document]
zhencliu: Agreed, currently no need to add vmx flags. To get the flags of cpu, as yanan mentioned, we may need to add ^ to match the start of the string, or both flags and vms flags will be matched with re.M mode

YongxueHong:
LGTM

yanan-fu:
The function definition LGTM.

Just for the implementation about the `/proc/cpuinfo`,  there are several usage like:
```
results = re.findall(***)
...
results[0]
```
No union set with all the matched information,  It assume the info for different processor is same. It is ok.
Based on this,  how about return with the first matched info directly ?

nickzhq:
> The function definition LGTM.
> 
> Just for the implementation about the `/proc/cpuinfo`, there are several usage like:
> 
> ```
> results = re.findall(***)
> ...
> results[0]
> ```
> 
> No union set with all the matched information, It assume the info for different processor is same. It is ok. Based on this, how about return with the first matched info directly ?

In my opinion, using re.findall or re.search() is okay to me. No matter the function we use, there does NOT have a significant reason to change.
@zhencliu @YongxueHong  what's your opinions?  Thanks!
cc @yanan-fu

zhencliu:
> > The function definition LGTM.
> > Just for the implementation about the `/proc/cpuinfo`, there are several usage like:
> > ```
> > results = re.findall(***)
> > ...
> > results[0]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > No union set with all the matched information, It assume the info for different processor is same. It is ok. Based on this, how about return with the first matched info directly ?
> 
> In my opinion, using re.findall or re.search() is okay to me. No matter the function we use, there does NOT have a significant reason to change. @zhencliu @YongxueHong what's your opinions? Thanks! cc @yanan-fu

It's OK for me for the current implementation, thanks.

YongxueHong:
> n my opinion, using re.findall or re.search() is okay to me. No matter the function we use, there does NOT have a significant reason to change.
> @zhencliu @YongxueHong what's your opinions? Thanks!
> cc @yanan-fu

Hi @nickzhq  and @yanan-fu 
Both of them are okay for me, but I prefer @yanan-fu 's suggestion since it is not ambiguous.
With the background, we know there is and only one result of the output, so we just get only one,
but if we use the `findall` to get the result, it will bring a misunderstanding that we can get multiple results here just get the first one as we expect. BTW, there is a difference of the time complexity between the `re.findall` and `re.search()` under the lots of data invoking, the `search` is faster than `findall`.
In short, it will bring a different understanding of code with different implementations sometimes. For clean code, we suggest using the appropriate code.

virttest/vt_utils/cpu.py
```diff
@@ -0,0 +1,144 @@
...
+    :rtype: list[string]
+    :raises: An OSError will be raised if it's NOT available on the platform
+    """
+    cpu_flags_re = "(?m)^flags\s+:\s+([\w\s]+)$"
+    with open("/proc/cpuinfo") as fd:
+        cpu_info = fd.read()
+    cpu_flag = re.search(cpu_flags_re, cpu_info)
+    if not cpu_flag:
+        raise OSError("The cpu flags were NOT found!")
+    cpu_flag = cpu_flag.groups()[0]
```
yanan-fu, line 71: `cpu_flags` ?

nickzhq: > `cpu_flags` ?

Done.

yanan-fu (approved):
Ack

nickzhq:
Hello @YongxueHong , can you help to merge this one? thanks!

YongxueHong:
Thanks all, let's merge this patch.

virttest/vt_utils/cpu.py
```diff
@@ -0,0 +1,144 @@
...
+    :rtype: string
+    :raises: An OSError will be raised if it's NOT available on the platform
+    """
+    cpu_model = process.run("lscpu").stdout_text
+    cpu_model_re = "(?m)^[mM]odel name:.*$"
+    cpu_model = re.search(cpu_model_re, cpu_model)
+    if not cpu_model:
+        raise OSError("The cpu model name was NOT found!")
+    cpu_model = cpu_model.groups()[0]
+    return cpu_model.split(":", 1)[-1].strip()
```
YongxueHong, line 54: Hi @nickzhq 
I hit an error when calling this function.
The following is the detail:
```
  File "/home/yhong/kar/workspace/avocado-vt/virttest/vt_utils/cpu.py", line 57, in get_cpu_model_name
    cpu_model = cpu_model.groups()[0]
IndexError: tuple index out of range
```
Could help to check it?
Thanks in advance.

nickzhq: > get_cpu_model_name

This bug will be fixed at patch#4030 ( https://github.com/avocado-framework/avocado-vt/pull/4030 ). Thanks!

Houqi

PR #3639, opened 2023-03-08
utils_stress: Don't compile app if available

smitterl:
Test code should not download and compile code.

Assume that the program is available on the guest
or that it has been made available via a package
manager.

Skip compilation if the assumption is met.

### DIFF
--- a/virttest/utils_stress.py
+++ b/virttest/utils_stress.py
@@ -210,6 +210,10 @@ def install_stressapptest(vm):
     :param vm: the vm to be installed with stressapptest
     """
     session = vm.wait_for_login(timeout=360)
+    if utils_package.package_install("stressapptest", session, timeout=300):
+        session.close()
+        return
+
     name = ["git", "gcc", "gcc-c++", "make"]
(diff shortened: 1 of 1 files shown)


virttest/utils_stress.py
```diff
@@ -210,6 +210,10 @@ def install_stressapptest(vm):
     :param vm: the vm to be installed with stressapptest
     """
     session = vm.wait_for_login(timeout=360)
+    if utils_package.package_install("stressapptest", session, timeout=300):
```
chunfuwen, line 213: Maybe this line code can be put before line 222, which is clustered code for stressapptest install

smitterl: > Maybe this line code can be put before line 222, which is clustered code for stressapptest install

I'm not sure if I understand you correctly. If you suggest this line be added before line 222 then as a result calling `install_stressapptest` would try to install git, gcc and other tools even if the stressapptest program is available. This doesn't make sense.

chunfuwen (approved):
LGTM

dzhengfy (approved):
LGTM

PR #3635, opened 2023-03-01
verify_kernel_crash:Expand the scope of regularity checking

zhenyzha:
According to the comment left by Linus Torvalds in the development 
community "We're trying to get rid of BUG_ON() because it kills the 
machine. Not replace it with another bogus thing that kills a machine."

Therefore, in future kernel development, developers will not 
use BUG_ON() unless necessary.
On the other hand, the expansion of the detection range is a 
double-edged sword. 
I don't want to classify all end traces as kernel crashes.
So only widen the search if console has a complete "cut here" 
"end trace", remove the "BUG:" keyword.

ID: 2174339
Signed-off-by: zhenyzha [zhenyzha@redhat.com](mailto:zhenyzha@redhat.com)

PR #3644, opened 2023-03-15
VMKernelCrashError:Add a new exception type

zhenyzha:
Add a new exception type, which is used to describe the
call trace of warning type, and this type of call trace will
not cause VM dead. But it will cause some codes to crash inside
the kernel, so a new function is needed to describe it.

ID: 2178511
Signed-off-by: zhenyzha [zhenyzha@redhat.com](mailto:zhenyzha@redhat.com)

### DIFF
--- a/virttest/virt_vm.py
+++ b/virttest/virt_vm.py
@@ -185,6 +185,16 @@ class VMDeadError(VMError):
         return (msg)
 
 
+class VMKernelCrashError(VMError):
+
+    def __init__(self, kernel_crash):
+        VMError.__init__(self, kernel_crash)
+        LOG.debug(kernel_crash)
+
+    def __str__(self):
+        return ("VM hit a kernel crash, see debug/serial log for details")
+
+
 class VMDeadKernelCrashError(VMError):
 
     def __init__(self, kernel_crash):
@@ -988,8 +998,9 @@ class BaseVM(object):
         """
         Find kernel crash message on the VM serial console.
 
-        :raise: VMDeadKernelCrashError, in case a kernel crash message was
-                found.
+        :raise VMKernelCrashError: if a kernel crash message was found.
+        :raise VMDeadKernelCrashError: if a kernel crash message was found
+                                       and the VM is dead.
         """
         panic_re = [r"BUG:.*---\[ end trace .* \]---"]
         panic_re.append(r"----------\[ cut here.*\[ end trace .* \]---")
@@ -1002,7 +1013,9 @@ class BaseVM(object):
                 return
             match = re.search(panic_re, data, re.DOTALL | re.MULTILINE | re.I)
             if match:
-                raise VMDeadKernelCrashError(match.group(0))
+                if self.is_dead():
+                    raise VMDeadKernelCrashError(match.group(0))
+                raise VMKernelCrashError(match.group(0))
 
     @session_handler
     def verify_dmesg(self, dmesg_log_file=None, connect_uri=None):


zhenyzha:
@PaulYuuu @YongxueHong Could you pls help review when you are free? thanks.

```
 (1/5) Host_RHEL.m9.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.numa_opts.nodes.0.arm64-pci: STARTED
 (1/5) Host_RHEL.m9.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.numa_opts.nodes.0.arm64-pci: PASS (47.31 s)
 (2/5) Host_RHEL.m9.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.numa_opts.nodes.1.arm64-pci: STARTED
 (2/5) Host_RHEL.m9.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.numa_opts.nodes.1.arm64-pci: PASS (41.31 s)
 (3/5) Host_RHEL.m9.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.numa_opts.nodes.2.arm64-pci: STARTED
 (3/5) Host_RHEL.m9.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.numa_opts.nodes.2.arm64-pci: ERROR: VM hit a kernel crash, see debug/serial log for details (44.85 s)
 (4/5) Host_RHEL.m9.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.numa_opts.nodes.3.arm64-pci: STARTED
 (4/5) Host_RHEL.m9.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.numa_opts.nodes.3.arm64-pci: PASS (44.64 s)
 (5/5) Host_RHEL.m9.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.numa_opts.nodes.64.arm64-pci: STARTED
 (5/5) Host_RHEL.m9.u3.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.numa_opts.nodes.64.arm64-pci: ERROR: VM hit a kernel crash, see debug/serial log for details (172.27 s)
```

virttest/virt_vm.py
```diff
@@ -1002,7 +1012,7 @@ def verify_kernel_crash(self):
                 return
             match = re.search(panic_re, data, re.DOTALL | re.MULTILINE | re.I)
             if match:
-                raise VMDeadKernelCrashError(match.group(0))
+                raise VMKernelCrashError(match.group(0))
```
PaulYuuu, line 1015: This change will cause some test case checkpoints to be failed, so with this change, we need to update test case part, not only tp-qemu, but also tp-libvirt.

zhenyzha: About tp-qemu and tp-libvirt cases failed.
I'm not sure if this step should be modified at our discretion.
This is more like a product issue causing automation cases to fail, 
and ultimately whether to keep this checkpoint may be up to the feature owner to decide.

zhenyzha: According to the offline discussion, I will send a depend patch 
to update the lower layer call about the exception.
thanks for reminding :)

harihare: VMDeadKernelCrashError handles both Kernel trace and crash. Then why the change required?

zhenyzha: > VMDeadKernelCrashError handles both Kernel trace and crash. Then why the change required?

Please refer to the patch for the background: https://github.com/avocado-framework/avocado-vt/pull/3635

zhenyzha: Thanks a lot, I think your suggestion is very beneficial. 
After matching we can refine exception types. 
I'll commit the changes later.

zhenyzha:
@PaulYuuu @YongxueHong Could you pls help review again when you are free? thanks.

virttest/virt_vm.py
```diff
@@ -1002,7 +1012,10 @@ def verify_kernel_crash(self):
                 return
             match = re.search(panic_re, data, re.DOTALL | re.MULTILINE | re.I)
             if match:
-                raise VMDeadKernelCrashError(match.group(0))
+                if "BUG" in match.group(0):
```
YongxueHong, line 1015: I don't think that we can judge the VM as dead according to ` if "BUG" in match.group(0)`.
I think we check the status of the VM and then raise the corresponding error.

zhenyzha: Thanks for the reminder, it has been corrected.

virttest/virt_vm.py
```diff
@@ -1002,7 +1012,10 @@ def verify_kernel_crash(self):
                 return
             match = re.search(panic_re, data, re.DOTALL | re.MULTILINE | re.I)
             if match:
-                raise VMDeadKernelCrashError(match.group(0))
+                if self.is_dead():
+                    raise VMDeadKernelCrashError(match.group(0))
+                else:
+                    raise VMKernelCrashError(match.group(0))
```
YongxueHong, line 1018:
```suggestion
                raise VMKernelCrashError(match.group(0))
```

virttest/virt_vm.py
```diff
@@ -1002,7 +1012,10 @@ def verify_kernel_crash(self):
                 return
             match = re.search(panic_re, data, re.DOTALL | re.MULTILINE | re.I)
             if match:
-                raise VMDeadKernelCrashError(match.group(0))
+                if self.is_dead():
```
PaulYuuu, line 1015: `is_dead` can only check the qemu process, not the guest status. when the guest crashes, if qemu can report `guest-panicked` status, then we can use it, but if not, there is not a good idea the check the guest status.

YongxueHong: > `is_dead` can only check the qemu process, not the guest status. when the guest crashes, if qemu can report `guest-panicked` status, then we can use it, but if not, there is not a good idea the check the guest status.

Hi @PaulYuuu 
You's concern makes sense. 
I think we should give the definition of the `VMDead` of `VMDeadKernelCrashError`.
What's the meaning of `VMDead`? Does it relate to the VM class?
In another way, how do we define the VM's dead status here?
CC @zhenyzha

virttest/virt_vm.py
zhenyzha: Corrected

virttest/virt_vm.py
```diff
@@ -988,8 +998,8 @@ def verify_kernel_crash(self):
         """
         Find kernel crash message on the VM serial console.
 
-        :raise: VMDeadKernelCrashError, in case a kernel crash message was
-                found.
+        :raise: VMKernelCrashError or VMDeadKernelCrashError, in case
+                a kernel crash message was found.
```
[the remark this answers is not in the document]
zhenyzha: Corrected

virttest/virt_vm.py
zhenyzha: Hello guys,

For me, it's relatively accurate to use it for checking in `virt_vm`.
We can refer to `verify_alive(self)`, its judgment is also through `self.is_dead()`
https://github.com/avocado-framework/avocado-vt/blob/master/virttest/virt_vm.py#L699

If we really want to explicitly define `VMDead` from the abstraction layer, 
I think overall refactoring may be needed here.
Of course, if there is a better way of judging, I am also happy to correct it.

virttest/virt_vm.py
```diff
@@ -185,6 +185,16 @@ def __str__(self):
...
+
+    def __init__(self, kernel_crash):
+        VMError.__init__(self, kernel_crash)
+        LOG.debug(kernel_crash)
+
+    def __str__(self):
+        return ("VM hit a kernel crash, see debug/serial log for details")
+
+
 class VMDeadKernelCrashError(VMError):
```
PaulYuuu, line 198:
```suggestion
class VMDeadKernelCrashError(VMKernelCrashError):
```

virttest/virt_vm.py
PaulYuuu: I think you should check `verify_alive` in qemu_vm.py and libvirt_vm.py, this method in virt_vm.py never be used for tp-qemu and tp-libvirt.

zhenyzha:
After discussion, we think that the verification method of the verify_kernel_crash function itself is inaccurate. On the other hand, the definition of VMDead is different in tp-libvirt and tp-qemu. So finally decided to close this patch. 
If we want to make changes in the future, we may need to refactor this function as a whole from the definition.

PR #3652, opened 2023-03-28
ks: Support install specific packages in the "%packages" step

PaulYuuu:
1. Add a new parameter to install pkgs in "%packages" phase
2. Split instlock to 2 functions
    1. Install pkgs in "%post" phase, supports specific nvr, url install
    2. Lock pkgs you want

ID: 2182236

### DIFF
--- a/virttest/shared/unattended/RHEL-8-devel.ks
+++ b/virttest/shared/unattended/RHEL-8-devel.ks
@@ -57,6 +57,7 @@ dmidecode
 alsa-utils
 sg3_utils
 -gnome-initial-setup
+KVM_TEST_PACKAGES_PKGS
 %end
 
 %post
@@ -108,15 +109,20 @@ EOF
 cat >> '/home/test/.bashrc' << EOF
 alias shutdown='sudo shutdown'
 EOF
-# Install and lock packages specified via 'kickstart_instlock_pkgs' parameter
-install_and_lock()
+# Install packages specified via 'kickstart_post_pkgs' parameter
+install_pkgs()
 {
-    for PKG in KVM_TEST_PKGS; do
+    for PKG in KVM_TEST_POST_PKGS; do
         ECHO "dnf install $PKG -y --nogpgcheck"
         dnf install $PKG -y --nogpgcheck
         if [ $? -ne 0 ]; then
             ECHO "$PKG installation failed."
         fi
+}
+# Lock packages specified via 'kickstart_lock_pkgs' parameter
+lock_pkgs()
+{
+    for PKG in KVM_TEST_LOCK_PKGS; do
         ECHO "dnf versionlock add $PKG"
         dnf versionlock add $PKG
         if [ $? -ne 0 ]; then
@@ -124,6 +130,7 @@ install_and_lock()
         fi
     done
 }
-install_and_lock
+install_pkgs
+lock_pkgs
 ECHO 'Post set up finished'
 %end
--- a/virttest/tests/unattended_install.py
+++ b/virttest/tests/unattended_install.py
@@ -433,12 +433,30 @@ class UnattendedInstallConfig(object):
             content = "\n".join(lines)
             contents = re.sub(dummy_repos_re, content, contents)
 
-        dummy_pkgs_re = r'\bKVM_TEST_PKGS\b'
-        if re.search(dummy_pkgs_re, contents):
-            # Extra packages to be installed and locked
+        # NOTE: This is an experimental feature to install extra packages
+        # during %packages step, even it supports kickstart syntax rules, we
+        # should only use it for add packages.
+        dummy_packages_pkgs_re = r'\bKVM_TEST_PACKAGES_PKGS\b'
+        if re.search(dummy_packages_pkgs_re, contents):
+            # Extra packages to be installed at "%packages" step
             # Use space as a separator for multiple pkgs
-            pkgs = self.params.get("kickstart_instlock_pkgs", "")
-            contents = re.sub(dummy_pkgs_re, pkgs, contents)
+            pkgs = self.params.get("kickstart_packages_pkgs", "").split()
+            content = "\n".join(pkgs)
+            contents = re.sub(dummy_packages_pkgs_re, content, contents)
+
+        dummy_post_pkgs_re = r'\bKVM_TEST_POST_PKGS\b'
+        if re.search(dummy_post_pkgs_re, contents):
+            # Extra packages to be installed at "%post" step
+            # Use space as a separator for multiple pkgs
+            pkgs = self.params.get("kickstart_post_pkgs", "")
+            contents = re.sub(dummy_post_pkgs_re, pkgs, contents)
+
+        dummy_lock_pkgs_re = r'\bKVM_TEST_LOCK_PKGS\b'
+        if re.search(dummy_lock_pkgs_re, contents):
+            # Packages to be locked
+            # Use space as a separator for multiple pkgs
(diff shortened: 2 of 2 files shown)


PaulYuuu:
Hello @YongxueHong @yanan-fu, can you help to review?

yanan-fu:
Hi @PaulYuuu  I am thinking,  may be, this requirement can be combined with the `KVM_TEST_PKGS`.

For `KVM_TEST_PKGS`, the initial purpose with patch https://github.com/avocado-framework/avocado-vt/pull/3566 is to install some special pkgs (can work together with `kickstart_extra_repo` or not). IMO, both `%packages` and `%post` can meet it.
The difference is `KVM_TEST_PKGS` be locked to ensure it can not be changed by os update or something else.  
Does `KVM_REQUIRED_PKGS` has the same requirement for this usage ?

yanan-fu:
cc @lkotek to aware of this.

lkotek:
@yanan-fu Thanks for letting me know. I agree on what you wrote. The only important difference is the package locking used in #3566. If it is not required for this use case I can also imagine we can add some option to make package locking mechanism to be optional only.

lkotek:
> @yanan-fu @lkotek, sure, we can use KVM_TEST_PKGS. But we want to combine, the parameter `kickstart_instlock_pkgs` I think it must update without `lock`, this is disruptive. Generic speaking, I think installing pkg in `%package` or `%post` should be the same result, if you think this patch can be an enhancement for #3566, then let's consider changing the parameter name `kickstart_instlock_pkgs`. BTW, if we never upgrade the guest system, then it doesn't matter if we lock it.

@PaulYuuu I see your point. In that case I am fine with the approach you introduced using `KVM_REQUIRED_PKGS`. For #3566, locking was quite important, so let's keep these two things separated as these can serve its purpose the best way. Maybe, it also helps to keep implementation simpler than via combining them :slightly_smiling_face:

PaulYuuu:
> @PaulYuuu I see your point. In that case I am fine with the approach you introduced using `KVM_REQUIRED_PKGS`. For #3566, locking was quite important, so let's keep these two things separated as these can serve its purpose the best way. Maybe, it also helps to keep implementation simpler than via combining them 🙂

Thanks @lkotek. And these two parameters are used in different steps(`%packages` and `%post`), so the new parameter may help us to control more simply pkg list, @yanan-fu what do you think?

yanan-fu:
I means use one param to control pkg install (as both `%post` and `%packages` can meet the existing requirements with https://github.com/avocado-framework/avocado-vt/pull/3566 ), and another param to control pkg lock, do it before system upgrade step in post is enough.  We do not drop pkg lock step as it is important for the pre requirement.

My concern is,  it is not clear to understand the difference with `KVM_REQUIRED_PKGS` and `KVM_TEST_PKGS` if user do not know the background. Or use more readable name to distinguish them.

What do you think about this @PaulYuuu  ?

PaulYuuu:
> I means use one param to control pkg install (as both `%post` and `%packages` can meet the existing requirements with #3566 ), and another param to control pkg lock, do it before system upgrade step in post is enough. We do not drop pkg lock step as it is important for the pre requirement.

So as I mentioned above, if we want to use one parameter, `kickstart_instlock_pkgs` to update to a more common name. For package installation, better to install them in `%packages` step, that's the point of this step.
> 
> My concern is, it is not clear to understand the difference with `KVM_REQUIRED_PKGS` and `KVM_TEST_PKGS` if user do not know the background. Or use more readable name to distinguish them.

Yes, they are all to install pkgs but in different steps. In my understanding, `REQUIRED` means it is needed for the whole testing, `TEST` is for some specific test cases, or to test it?
> 
> What do you think about this @PaulYuuu ?

PaulYuuu:
OK, @yanan-fu @lkotek, after an offline discussion, let's split the parameter, install pkgs in `%packages`, lock pkgs in `%post`, and they can be different.

PaulYuuu:
But let me point out 1 issue, with this change, we cannot install specific version rpm or install from URL.
```
dnf install pkg-0.1.2.rpm
dnf install https://example.com/pkg-0.1.2.rpm
```

yanan-fu:
> But let me point out 1 issue, with this change, we cannot install specific version rpm or install from URL.
> 
> ```
> dnf install pkg-0.1.2.rpm
> dnf install https://example.com/pkg-0.1.2.rpm
> ```

As far as i know, it is possible to pin a package version in  `%packages`, such as `pkg-0.1.2`.
Otherwise it can not meet the requirement with https://github.com/avocado-framework/avocado-vt/pull/3566.
Anyway, I will verify this scenario first.

yanan-fu:
> As far as i know, it is possible to pin a package version in `%packages`, such as `pkg-0.1.2`. Otherwise it can not meet the requirement with #3566. Anyway, I will verify this scenario first.

I am afraid i am wrong:
https://access.redhat.com/solutions/1149893

PaulYuuu:
> > As far as i know, it is possible to pin a package version in `%packages`, such as `pkg-0.1.2`. Otherwise it can not meet the requirement with #3566. Anyway, I will verify this scenario first.
> 
> I am afraid i am wrong: [access.redhat.com/solutions/1149893](https://access.redhat.com/solutions/1149893)

Yes, this is why I want to use 2 parameters :(

yanan-fu:
@PaulYuuu  Is it must for the new requirement with this patch to install pkg in the `%packages` step ?   I am thinking `%post` step can support both pkg name and pkg nvr.

(The specified nvr is must for the requirements with https://github.com/avocado-framework/avocado-vt/pull/3566)

yanan-fu:
> Hello @yanan-fu, finally, I use `KVM_TEST` pref.ix, even the name of `KVM_TEST_PACKAGES_PKGS` is strange, but it also makes sense, how do you think?

The new name is make sense to me.

cc @lkotek for this change ~
A summary:
The requirements are different:
1.  Install pkgs in `%packages` step from the distro, it does not support full nvr.
2.  Install pkgs with full nvr in `%post` steps
3.  Lock pkgs with full nvr in `%post` steps

Based on the requirements above, we adjust the name with:
1. `KVM_TEST_PACKAGES_PKGS`:  the pkgs will be installed in `%packages` step
2. `KVM_TEST_POST_PKGS`:  the pkgs be installed in `%post` step
3. `KVM_TEST_LOCK_PKGS`: the pkgs need be locked, only supported by `%post` step

@lkotek  Do you have any other concern for this change ? Thanks!

lkotek:
@yanan-fu I am completely fine with the change, the name serves better its purpose. We just need to update our internal kickstart profiles in KAR repo which currently uses the old name.

yanan-fu (approved):
Ack

yanan-fu:
@YongxueHong  Could you please help review ？ Thanks

virttest/shared/unattended/RHEL-8-devel.ks
```diff
@@ -109,21 +110,26 @@ cat >> '/home/test/.bashrc' << EOF
 alias shutdown='sudo shutdown'
 EOF
 # Install and lock packages specified via 'kickstart_instlock_pkgs' parameter
```
YongxueHong, line 112:
```suggestion
# Install and lock packages specified via 'kickstart_post_pkgs' and 'kickstart_lock_pkgs' parameter
```

PaulYuuu: Thank you, divided it into 2 notes.

YongxueHong (approved):
LGTM.

PR #3656, opened 2023-04-06
qemu_vm: Let target VM remove tmp files after a local migration

zhencliu:
When we perform a live migration on a single node, never remove the related resources of the source VM, because the target VM is still using these resources after migration, so let the target VM do it.

ID: 2184877

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -4860,6 +4860,12 @@ class VM(virt_vm.BaseVM):
                         if os.path.isfile(save2):
                             os.remove(save2)
 
+            # FIXME: Never remove the source VM's files when doing a local
+            # migration because these files are still being used by the
+            # target VM, let the target VM do the cleanup work
+            if local and self.devices.temporary_image_snapshots:
+                self.devices.temporary_image_snapshots.clear()
+
             # Switch self <-> clone
             temp = self.clone(copy_state=True)
             self.destroy(gracefully=False)      # self is the source dead vm


zhencliu:
hi @PaulYuuu @YongxueHong , would you please have a look? IMO, we can simply clear the set if it's a local migration, because when clone is created, the related files(vars file and snapshot file) will be added into the temporary_image_snapshots even if they are existing files (vars file).

I performed some migration tests and didn't find issues, also I ran multi_vms_file_transfer (image_snapshot=yes) successfully, but I didn't find a case with both image_snapshot=yes and local migration, the only one (local migration+snapshot) is Aihua's new case https://github.com/autotest/tp-qemu/pull/3673, it also worked well.

cc @luckyh

virttest/qemu_vm.py
```diff
@@ -4860,6 +4860,12 @@ def _set_migrate_capability(vm, capability, value, is_src_vm=True):
                         if os.path.isfile(save2):
                             os.remove(save2)
 
+            # FIXME: Never remove the source VM's files when doing a local
+            # migration because these files are still being used by the
+            # target VM, let the target VM do the cleanup work
+            if local and self.devices.temporary_image_snapshots:
+                self.devices.temporary_image_snapshots.clear()
```
PaulYuuu, line 4867: The dst VM will generate the **temporary_image_snapshots** but not copy from the src VM, even if they are the same, not sure if there is a potential problem. Ack this time.

PaulYuuu (approved):

virttest/qemu_vm.py
```diff
@@ -4860,6 +4860,12 @@ def _set_migrate_capability(vm, capability, value, is_src_vm=True):
                         if os.path.isfile(save2):
                             os.remove(save2)
 
+            # FIXME: Never remove the source VM's files when doing a local
+            # migration because these files are still being used by the
+            # target VM, let the target VM do the cleanup work
```
YongxueHong, line 4865: I think there is a situation in which the target VM still can not clean up the `temporary_image_snapshots` for migrating the VM in the `exec` way.
Since this way, the clone will not do the `create` action so the `temporary_image_snapshots` is still empty before being destroyed.

YongxueHong: > I think there is a situation in which the target VM still can not clean up the `temporary_image_snapshots` for migrating the VM in the `exec` way. Since this way, the clone will not do the `create` action so the `temporary_image_snapshots` is still empty before being destroyed.

Hi @zhencliu 
My bad, it will still be created.

YongxueHong (approved):
LGTM.

YongxueHong:
Hi @aliang123 
Could you help to review it? 
Thanks.

aliang123:
(1/1) Host_RHEL.m9.u3.ovmf.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.9.3.0.x86_64.io-github-autotest-qemu.migration_with_json_backing.q35: PASS (56.44 s)

Acked-by: Aihua Liang <aliang@redhat.com>

YongxueHong:
Thanks all. Let's merge it.

PR #3668, opened 2023-04-20
utils_net.get_default_gateway:support ipv6

chloerh:


virttest/utils_net.py
```diff
@@ -3886,7 +3886,7 @@ def get_host_iface():
     return [_.strip() for _ in re.findall("(.*):", host_iface_info)]
 
 
-def get_default_gateway(iface_name=False, session=None):
+def get_default_gateway(iface_name=False, session=None, ip_ver='ipv4'):
```
Yingshun, line 3889: could you please update docstring?

virttest/utils_net.py
```diff
@@ -3895,16 +3895,21 @@ def get_default_gateway(iface_name=False, session=None):
...
     else:
-        cmd = "ip route | awk '/default/ { print $3 }'"
+        cmd = "%s route | awk '/default/ { print $3 }'" % ip_cmd
     try:
         if session:
             output = session.cmd_output(cmd).strip()
             LOG.debug("Guest default gateway is %s" % output)
         else:
             output = process.run(cmd, shell=True).stdout_text.rstrip()
+            LOG.debug("Host default gateway is %s" % output)
```
Yingshun, line 3912:
```suggestion
            LOG.debug("Host default gateway is %s", output)
```

virttest/utils_net.py
```diff
@@ -3895,16 +3895,21 @@ def get_default_gateway(iface_name=False, session=None):
     :return: A string of the host's or guest's default gateway or interface.
     :rtype: string
     """
+    if ip_ver == 'ipv4':
+        ip_cmd = 'ip'
+    elif ip_ver == 'ipv6':
+        ip_cmd = 'ip -6'
```
Yingshun, line 3901: I wonder if it's better to add an 'else' here to avoid 'undefine' error when 'ip_ver' is neither 'ipv4' nor 'ipv6'. what do you think?

virttest/utils_net.py
chloerh: Updated

virttest/utils_net.py
chloerh: Updated

virttest/utils_net.py
chloerh: I guess, which type of exception do you prefer?

Yingshun: I think it's ok to set a default value or raise an exception. or something like below:
```suggestion
    if ip_ver == 'ipv6':
        ip_cmd = 'ip -6'
    else:
        ip_cmd = 'ip'
```
its up to you.

Yingshun (approved):

yalzhang (approved):

PR #3685, opened 2023-05-16
hotfix: syntax in the RHEL-8-devel.ks file

PaulYuuu:
In the install_pkgs function on RHEL-8-devel.ks, missing "done" in the "for" loop, this will block the guest installation, fix it.


PaulYuuu:
Hello @YongxueHong , please help to merge it, it's a hotfix and needs to fix asap, thanks in advance.

PR #3689, opened 2023-05-18
utils_net:get_linux_iface_info:update log output

chloerh:


virttest/utils_net.py
```diff
@@ -3737,8 +3737,8 @@ def get_linux_iface_info(iface='', mac=None, session=None):
 
     try:
         ip_output_str = run_func(ip_cmd).strip()
+        LOG.debug(f'Interface info {iface}:\n{ip_output_str}')
```
dzhengfy, line 3740: Could you explain the benefit for the change?

chloerh: We met error here, I'd like to check the output string.

dzhengfy (approved):
LGTM

PR #3693, opened 2023-05-22
Sevctl service

zhencliu:


YongxueHong:
> hi @YongxueHong , would you please have a look? Thanks.
> 
> The dir structure under vt_utils about secure guest may change, but IMO, we can review code first

Sure.

PR #3694, opened 2023-05-23
libvirt_monitor: Update to check more items for domjobinfo

cliping:


virttest/utils_libvirt/libvirt_monitor.py
```diff
@@ -8,30 +8,107 @@
...
     """
-    LOG.info("Check domjobinfo")
-    if remote:
+    def check_domjobinfo_items(expected_domjobinfo):
+        """
+        Check the items of domjobinfo output
+
+        :param expected_domjobinfo: the expected domjobinfo
+        """
+        def compare_value(value_type, line):
```
chunfuwen, line 23: compare_value and check_items container long codes. Could they be moved outside check_domjobinfo_output. 
If those function are exclusively for check_domjobinfo_output's use, we can use _ prefix to define them as inner methods

cliping: Updated, thanks.

chunfuwen (approved):
lgtm

PR #3698, opened 2023-06-12
test_setup: enable customize hugepage mount path

dzhengfy:
This is to enable the capability to customize the hugepage mount path.

### DIFF
--- a/virttest/test_setup/__init__.py
+++ b/virttest/test_setup/__init__.py
@@ -279,7 +279,7 @@ class HugePageConfig(object):
         self.max_vms = int(params.get("max_vms", 0))
         self.qemu_overhead = int(params.get("hugepages_qemu_overhead", 128))
         self.deallocate = params.get("hugepages_deallocate", "yes") == "yes"
-        self.hugepage_path = '/mnt/kvm_hugepage'
+        self.hugepage_path = params.get("hugepage_path", "/mnt/kvm_hugepage")
         self.kernel_hp_file = params.get("kernel_hp_file", "/proc/sys/vm/nr_hugepages")
         if not os.path.exists(self.kernel_hp_file):
             raise exceptions.TestSkipError("Hugepage config file is not found: "


nanli1 (approved):

cliping (approved):
LGTM

Yingshun (approved):

virttest/test_setup/__init__.py
```diff
@@ -279,7 +279,7 @@ def __init__(self, params):
         self.max_vms = int(params.get("max_vms", 0))
         self.qemu_overhead = int(params.get("hugepages_qemu_overhead", 128))
         self.deallocate = params.get("hugepages_deallocate", "yes") == "yes"
-        self.hugepage_path = '/mnt/kvm_hugepage'
+        self.hugepage_path = params.get("hugepage_path", "/mnt/kvm_hugepage")
```
zhenyzha, line 282: @dzhengfy  The change caused regression:
https://github.com/autotest/tp-qemu/blob/master/qemu/tests/cfg/hugepage_mem_stress.cfg#L19C13-L19C26
Causes cases hugepage_mem_stress.non_existent_mem_path.arm64-pci: FAIL: HugePages leaked. (106.03 s)

The added interface has the same name as https://github.com/avocado-framework/avocado-vt/blob/master/virttest/qemu_vm.py#L1107C49-L1107C49, resulting in the failure to set the negative scene.

Specific test scenarios:
1. Mount this hugepage to /mnt/kvm_hugepage
mount -t hugetlbfs -o pagesize=2048K none /mnt/kvm_hugepage
2. HugePages didn't leak when using another path /mnt/tmp
/usr/libexec/qemu-kvm \
......
-object '{"size": 8589934592, "mem-path": "/mnt/kvm_hugepage", "id": "mem-machine_mem", "qom-type": "memory-backend-file"}' 

Cc: @YongxueHong @yanan-fu @luckyh

luckyh: Thanks @zhenyzha for addressing the root cause of this issue! If that's the case, I'd suggest renaming the param name on vt side to `vt_env_hugepage_path` where just adding the `vt_env_` prefix as namespace to the current one. @dzhengfy @Yingshun does it also sound good to you?

Yingshun: I agree to rename the parameter but i'm not sure if it's a proper one which starts with 'vt_env_'.  Can we get something similar? like 'hugepage**s**_path'?

luckyh: > I agree to rename the parameter but i'm not sure if it's a proper one which starts with 'vt_env_'. Can we get something similar? like 'hugepage**s**_path'?

@Yingshun ok, let me clarify on my point. The issue is, from time to time, we could see the naming collision of params because of its flat hierarchy, in a typical case (would be like the one we just encountered) among the vt framework and test providers.

Well, it would be hard for test script developers to be aware of whether a param name has already been or will be occupied from the framework side, therefore we would usually leave the name with prefix/namespace omitted for the convenient of test script development and set explicit additional prefix/namespace to the one on vt side. And it is one of the reasons I'd prefer using `vt_env_hugepage_path`.

Apart from that, regarding the settings of hugepages, AFAICS we could introduce the path option to multiple concepts even within the scope of vt, for example, a new one for vm-level beside the one for system-level we introduced by this PR (if I understood the implementation correctly), hence we have to differentiate them via its prefix.

I'd suggest as so based on the above points, however, the name `vt_env_` is not something I'd insist on, we could think a better one as you wish. But I guess it's not that rational to omit the prefix.

Let me know your opinion.

Thanks!

dzhengfy: @luckyh I agree to change to another name begining with `vt_` as it is required by functions in avocado-vt.   But what is for with `env_` here? I am also ok with it , but just curious about it. Or maybe we simply use `vt_hugepage_path`? Any concern with it?

luckyh: > But what is for with `env_` here?

@dzhengfy let's suppose that there would be other facilities consuming the hugepage resources at the VT side, if we'd like to make theirs path option to be configurable independently as well then we might have to introduce a sub-namepace for distinguishing them. `env_` was just the one I'd like to purpose for the hugepage setup procedure because, IMO, the procedure was introduced for manipulating the hugepages of the entire scope of host environment. However, we could change the name if you think `env_` is not suitable in this case, my point is that we'd better to explicitly add the sub-namespace into the param name.

(Btw, from my perspective, it seems that we should not introduce such a param/variable to the hugepage setup procedure, and instead we should let each facility introduce that. For example, the hugetlb entrypoint/mountpoint will be consumed by VMs then the VM manager would be in charge of doing that thing and therefore `vm_hugepage_entrypoint` or `vm_hugepage_mountpoint` may be a better name we could consider to use.)

dzhengfy: I like `vm_hugepage_mountpoint`. 
I submitted a fix https://github.com/avocado-framework/avocado-vt/pull/3741

PR #3700, opened 2023-06-13
libvirt_network: Add port to pattern string

cliping:
Sometimes we need to use port to get expected network connection number.


cliping:
Test result in https://github.com/autotest/tp-libvirt/pull/4948.

dzhengfy (approved):
LGTM

PR #3707, opened 2023-06-16
check_guest_xml_by_xpaths:Fix issue of ignoring mot-matched text

chloerh:
The elements of `matches` do not equal to True which cause the expression to be False at any time which means it's ignoring non-matched text.


meinaLi (approved):

dzhengfy (approved):
LGTM

PR #3709, opened 2023-06-19
test_setup: Refactor network proxies setup

bgartzi:
This is the first commit of a series that will try to tidy up the env_process module's preprocess and postprocess methods.

The first refactored element is the network proxies setup. This commit adds the NetworkProxies setuper. The latter will take care of the needed actions to prepare the test env according to the configuration params. It also registers it in the setup_manager already defined in the env_process module.

ID: 2211405


bgartzi:
Hi @YongxueHong,

I'm adding the first patch among several other ones refactoring the `env_process` `pre/postprocess` functions. This one, as discussed is about the proxies setup step.

I'd appreciate it a lot if you could drop some comments/review.

Thanks a lot!

cc @luckyh

YongxueHong (approved):
LGTM.

PR #3714, opened 2023-07-07
Fixup VMIPAddressMissingError

cliping:
Sometimes vm failed to get ip because address cache is NULL, so
 need to update address cache before getting ip.

### DIFF
--- a/virttest/migration.py
+++ b/virttest/migration.py
@@ -132,6 +132,10 @@ class MigrationTest(object):
             func = exceptions.TestFail
             uri_backup = vm.connect_uri
             vm.connect_uri = uri
+            vm.cleanup_serial_console()
+            vm.create_serial_console()
+            utils_net.update_mac_ip_address(vm, timeout=120)
+            vm.cleanup_serial_console()
             server_session = test_setup.remote_session(params)
             # after migration VM would take some time to respond and to
             # avoid the race of framework querying IP address before VM


cliping:
Before:
 (1/1) type_specific.io-github-autotest-libvirt.migration.migrate_graphics.listen_type_socket.p2p: ERROR: No ipv4 DHCP lease for MAC 52:54:00:17:27:3a (174.50 s)

After:
 (1/1) type_specific.io-github-autotest-libvirt.migration.migrate_graphics.listen_type_socket.p2p: PASS (241.94 s)

virttest/migration.py
```diff
@@ -132,6 +132,10 @@ def ping_vm(self, vm, params, uri=None, ping_count=10,
             func = exceptions.TestFail
             uri_backup = vm.connect_uri
             vm.connect_uri = uri
+            vm.cleanup_serial_console()
```
chunfuwen, line 135: Do line 135 and 136 are necessary?

cliping: They are necessary. The existing serial console may be connected to the source vm. So when we ping vm on dest, we need to clean the console first. Then we create a new serial console for vm on dest.

dzhengfy (approved):
LGTM

chloerh (approved):

chunfuwen (approved):
lgtm

PR #3717, opened 2023-07-12
add log_file to console xml

nanli1:
log_file needs to be added

for https://github.com/autotest/tp-libvirt/pull/4966

### DIFF
--- a/virttest/libvirt_xml/devices/console.py
+++ b/virttest/libvirt_xml/devices/console.py
@@ -11,7 +11,7 @@ from virttest.libvirt_xml.devices.character import CharacterBase
 class Console(CharacterBase):
 
     __slots__ = ('protocol_type', 'target_port', 'target_type', 'sources',
-                 'alias')
+                 'alias', 'log')
 
     def __init__(self, type_name='pty', virsh_instance=base.virsh):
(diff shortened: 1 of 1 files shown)


nanli1:
@dzhengfy  Hi Dzheng ,Could you help review

virttest/libvirt_xml/devices/console.py
```diff
@@ -26,6 +26,8 @@ def __init__(self, type_name='pty', virsh_instance=base.virsh):
                                  has_subclass=True)
         accessors.XMLElementDict('alias', self, parent_xpath='/',
                                  tag_name='alias')
+        accessors.XMLElementDict('log_file', self, parent_xpath='/',
```
dzhengfy, line 29: According to the log definition in libvirt.org, I think it is better to use `log` instead of `log_file`, because it also include other attributes, like `append`. What do you think?

`<log file="/var/log/libvirt/qemu/guestname-serial0.log" append="off"/>`

nanli1: Hi dzheng, Thanks your reminder  ,Done

dzhengfy (approved):

dzhengfy:
@nanli1 tp-libvirt seems need update accordingly

PR #3721, opened 2023-07-13
Reduce the libvirt logging verbosity (not needed for qemu vms)

pevogam:
We do this by importing these modules conditionally within the modules of shared qemu and livbirt vm usage.

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -34,12 +34,10 @@ from virttest import cpu
 from virttest import storage
 from virttest import utils_libguestfs
 from virttest import qemu_storage
-from virttest import utils_libvirtd
 from virttest import data_dir
 from virttest import utils_net
 from virttest import nfs
 from virttest import libvirt_vm
-from virttest import virsh
 from virttest import utils_test
 from virttest import utils_iptables
 from virttest import utils_package
@@ -53,6 +51,12 @@ from virttest.staging import service
 from virttest.test_setup.core import SetupManager
 from virttest.test_setup.os_posix import UlimitConfig
 from virttest.test_setup.networking import NetworkProxies
+
+
+# lazy imports for dependencies that are not needed in all modes of use
+from virttest._wrappers import lazy_import
+utils_libvirtd = lazy_import("virttest.utils_libvirtd")
+virsh = lazy_import("virttest.virsh")
 
 
 try:
--- a/virttest/_wrappers.py
+++ b/virttest/_wrappers.py
@@ -6,6 +6,7 @@
 
 import sys
 import importlib
+import importlib.util
 from importlib.machinery import (SourceFileLoader, SOURCE_SUFFIXES,
                                  SourcelessFileLoader, BYTECODE_SUFFIXES,
                                  ExtensionFileLoader, EXTENSION_SUFFIXES)
@@ -70,3 +71,20 @@ def load_source(name, path):
     """
     spec = importlib.util.spec_from_file_location(name, path)
     return _load_from_spec(spec)
+
+
+def lazy_import(name):
+    """ Allows for an early import that is initialized upon use (lazy import).
+
+    :param name: Name of the module that is going to be imported
+    :type name: String
+    """
+    spec = importlib.util.find_spec(name)
+    if spec is None:
+        raise ImportError(f"Could not import module {name}")
+    loader = importlib.util.LazyLoader(spec.loader)
+    spec.loader = loader
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[name] = module
+    loader.exec_module(module)
+    return module
--- a/virttest/utils_libvirtd.py
+++ b/virttest/utils_libvirtd.py
@@ -56,6 +56,7 @@ class Libvirtd(object):
         self.daemons = []
         self.service_list = []
 
+        # we only import this module conditionally to make this warning always applicable
         if LIBVIRTD is None:
             LOG.warning("Libvirtd service is not available in host, "
                         "utils_libvirtd module will not function normally")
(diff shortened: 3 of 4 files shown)


pevogam:
@luckyh @PaulYuuu I have been wanting to reduce these constant warnings we see in our Qemu VM runs for very long now, let me know what you think about this suggestion of conditional and localized imports in shared usage places.

pevogam:
(I don't quite get why the signed-off static check failed since I have signed the commit but will push again after some changes are requested if any)

virttest/env_process.py
```diff
@@ -1135,6 +1133,7 @@ def preprocess(test, params, env):
         migrate_setup.migrate_pre_setup(dest_uri, params)
         # Map hostname and IP address of the hosts to avoid virsh
         # to error out of resolving
+        from virttest import virsh
```
luckyh, line 1136: Seems that we could leverage the [lazy importing](https://docs.python.org/3.10/library/importlib.html?highlight=lazy_import#implementing-lazy-imports) feature to get rid of adding an explicit import statement in every place referred to that module. I just tried with the following example and it worked as expected.

So what do you think?

(p.s. looks that the [`_wrappers`](https://github.com/avocado-framework/avocado-vt/blob/master/virttest/_wrappers.py) could be the module the `lazy_import` function to be introduced to.)

```python
# /tmp/test/main.py 
import importlib.util
import sys


def lazy_import(name):
    spec = importlib.util.find_spec(name)
    loader = importlib.util.LazyLoader(spec.loader)
    spec.loader = loader
    module = importlib.util.module_from_spec(spec)
    sys.modules[name] = module
    loader.exec_module(module)
    return module


virsh = lazy_import("virsh")


if __name__ == "__main__":
    if len(sys.argv) > 1 and sys.argv[1] == 'libvirt':
        virsh.foo()
```
```python
# /tmp/test/virsh.py 
print("imported virsh")


def foo():
    print("calling virsh.foo")
```
```
$ python3 /tmp/test/main.py
(no output)

$ python3 /tmp/test/main.py libvirt
imported virsh
calling virsh.foo
```

luckyh (commented):
Hi @pevogam , I'd really like this proposal, thanks! However, just have a small suggestion regarding the code changes.

pevogam:
@luckyh I followed your request but oddly ended up with `AttributeError: 'NoneType' object has no attribute 'loader'` so I will have to look a bit deeper in the suggested implementation and test it myself, it seems I have to supply different (complete?) namespace for the lazily imported modules.

virttest/env_process.py
```diff
@@ -55,6 +53,12 @@
 from virttest.test_setup.networking import NetworkProxies
 
 
+# lazy imports for dependencies that are not needed in all modes of use
+from virttest._wrappers import lazy_import
+utils_libvirtd = lazy_import("utils_libvirtd")
+virsh = lazy_import("virsh")
```
luckyh, line 59: > @luckyh I followed your request but oddly ended up with `AttributeError: 'NoneType' object has no attribute 'loader'` so I will have to look a bit deeper in the suggested implementation and test it myself, it seems I have to supply different (complete?) namespace for the lazily imported modules.

@pevogam after checking the code changes, I guess that the issue was caused by intending on doing implicit relative imports. So would you please try with the following tweak to see if it could fix the issue?

```suggestion
utils_libvirtd = lazy_import("virttest.utils_libvirtd")
virsh = lazy_import("virttest.virsh")
```

Btw, we can apply such a check to get a better hint to the import issue you just saw, thanks!

```python
def lazy_import(name):
    spec = importlib.util.find_spec(name)
    if spec is None:
        raise ImportError(...)
    ...
```

pevogam: Ah indeed makes complete sense, pushed with the additional validation which is also a good idea.

luckyh (approved):
LGTM

luckyh:
Hi @dzhengfy @chloerh @Yingshun , would you mind to check this works for you case? thanks in advance!

PaulYuuu (approved):

pevogam:
Hi all, there are currently two approvals for this PR so I guess we can merge it. Regarding verbosity of OpenVSwitch I guess we will simply debug a problem in https://github.com/avocado-framework/avocado-vt/discussions/3720#discussioncomment-6702543.

luckyh:
> Hi all, there are currently two approvals for this PR so I guess we can merge it. Regarding verbosity of OpenVSwitch I guess we will simply debug a problem in [#3720 (reply in thread)](https://github.com/avocado-framework/avocado-vt/discussions/3720#discussioncomment-6702543).

Yeah, sure.

luckyh:
Thanks to all.

luckyh:
> Hi @luckyh, unfortunately this doesn't seem to work in all cases since using `__spec__` counts as module attribute retrieval ...

Hi @pevogam, I haven't dived into the details of how python manipulates the module importing, but it seems that the issue was related to the routine the `lazy_import` handled a second call with one certain module. Would you mind to try with the following fix to see if it can resolve your cases as well?

```patch
diff --git a/virttest/_wrappers.py b/virttest/_wrappers.py
index 7ca031f8..bb2fbfcb 100644
--- a/virttest/_wrappers.py
+++ b/virttest/_wrappers.py
@@ -79,6 +79,8 @@ def lazy_import(name):
     :param name: Name of the module that is going to be imported
     :type name: String
     """
+    if name in sys.modules:
+        return sys.modules[name]
     spec = importlib.util.find_spec(name)
     if spec is None:
         raise ImportError(f"Could not import module {name}")
```

pevogam:
Hi @luckyh, ok this indeed fixes the issue. I guess we will proceed with the cleanup then despite the large diff above and I will create a pull request with it next.

pevogam:
Feel free to check #3783.

PR #3722, opened 2023-07-18
virsh: align same usage for emulatorpin

dzhengfy:
Usually the options value should be started with --. The old codes added '--' in this utility function which is not suitable. We expect the client provides like '--config' to invoke this function as other functions do.


Yingshun (approved):
LGTM

dzhengfy:
I will submit other tp-libvirt PRs to align with this changes

chloerh (approved):

PR #3723, opened 2023-07-19
nodedev_xml: extend MdevXML with attr elements

smitterl:
Add support for attribute elements of the mediated device capability xml, e.g.:

```
 <attr name='assign_adapter' value='3'/>
 <attr name='assign_domain' value='43'/>
```

### DIFF
--- a/virttest/libvirt_xml/nodedev_xml.py
+++ b/virttest/libvirt_xml/nodedev_xml.py
@@ -152,15 +152,41 @@ class MdevXML(CAPXML):
     """
     class for capability whose type is mdev
     """
-    __slots__ = ('type_id', 'uuid')
+    __slots__ = ('type_id', 'uuid', 'attrs')
 
     def __init__(self, virsh_instance=base.virsh):
         accessors.XMLAttribute('type_id', self, parent_xpath='/',
                                tag_name='type', attribute='id')
         accessors.XMLElementText('uuid', self, parent_xpath='/',
                                  tag_name='uuid')
+        accessors.XMLElementList(property_name='attrs',
+                                 libvirtxml=self,
+                                 parent_xpath='/',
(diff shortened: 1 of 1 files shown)


smitterl:
Tested via https://github.com/autotest/tp-libvirt/pull/5040

cliping (approved):
LGTM

dzhengfy:
@chloerh Could you help review it as you also have many experience on this?

virttest/libvirt_xml/nodedev_xml.py
```diff
@@ -152,16 +152,66 @@ class MdevXML(CAPXML):
...
+        Convert an attr tag xml element to an object of AttrXML.
+        """
+        if tag != 'attr':
+            return None    # Don't convert this item
+        newone = AttrXML(virsh_instance=libvirtxml.virsh)
+        newone.xmltreefile = new_treefile
+        return newone
+
+
+class AttrXML(base.LibvirtXMLBase):
```
chloerh, line 197: According to the xml of <attr>, seems no need to define this class, we could simply define the accessor with 'has_subclass = False' and with shorter marshal functions, if <attr> doesn't have sub-element. What do you think?

chloerh: Like this:
```
accessors.XMLElementList('vcpupins', self, parent_xpath='/',
                                 marshal_from=self.marshal_from_vcpupins,
                                 marshal_to=self.marshal_to_vcpupins)
```

smitterl: Thank you for the suggestion. It works well.

chloerh: Thanks for taking my suggestion, btw there's inspekt issue needs to be resovle.

chloerh (approved):

PR #3725, opened 2023-07-21
v2v: add function for vpx without username

vwu-vera:
PR#5048


vwu-vera:
@xiaodwan Please help to review. Thanks.:)

virttest/utils_v2v.py
```diff
@@ -70,6 +70,20 @@ def get_uri(self, hostname, vpx_dc=None, esx_ip=None, username=None):
         self.vpx_dc = vpx_dc
         self.esx_ip = esx_ip
         self.username = username if username else 'root'
+        self.vpx_no_username = False
+        return uri_func()
+
+    def get_uri_without_username(self, hostname, vpx_dc=None, esx_ip=None):
```
xiaodwan, line 76: I think keeping the logic in one function is better.
There is no need to add a new function.

virttest/utils_v2v.py
```diff
@@ -182,6 +205,7 @@ def get_cmd_options(self, params):
         self.input_file = self.params.get('input_file')
         self.new_name = self.params.get('new_name')
         self.username = self.params.get('username', 'root')
+        self.vpx_no_username = self.params.get('vpx_no_username')
```
xiaodwan, line 208: Params_get can read the variants in cfg file directly. 
The old code style is not good and looks bad.

virttest/utils_v2v.py
vwu-vera: I considered keeping the logic in one function at first. 
However, this original func is for public use. If adding the parameters to control the conditions in one func, all calls in other code scripts need to be modified accordingly.
In this consideration, making the func to be isolated only for this case is safer and better.

xiaodwan (approved):

vwu-vera:
@chunfuwen Please help to review this PR. Thanks a lot. :)

virttest/utils_v2v.py
```diff
@@ -70,6 +70,20 @@ def get_uri(self, hostname, vpx_dc=None, esx_ip=None, username=None):
...
         self.esx_ip = esx_ip
         self.username = username if username else 'root'
+        self.vpx_no_username = False
+        return uri_func()
+
+    def get_uri_without_username(self, hostname, vpx_dc=None, esx_ip=None):
+        """
+        Uri dispatcher.
+
+        :param hostname: String with host name.
```
chunfuwen, line 80: For other parameters e.g vpx_dc,esx_ip=, it also need add doc string

vwu-vera: Done. Thanks for the suggestion.

chunfuwen (approved):
lgtm

PR #3731, opened 2023-08-04
Add wait in create_scsi_disk method

chunfuwen:
Add wait in create_scsi_disk method

In some extreme situation, loading scsi_debug module may not be returned smoothly, and it leads to failure in finding scsi device

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -2878,6 +2878,15 @@ def create_scsi_disk(scsi_option, scsi_size="2048"):
             linux_modules.unload_module("scsi_debug")
         linux_modules.load_module("scsi_debug dev_size_mb=%s %s" %
                                   (scsi_size, scsi_option))
+
+        def _get_scsi_disk():
+            """attempt to get scsi disk multiple times until getting real one"""
+            scsi_disk = process.run("lsscsi|grep scsi_debug|awk '{print $6}'",
+                                    shell=True).stdout_text.strip()
+            return 'dev' in scsi_disk
+
+        utils_misc.wait_for(lambda: _get_scsi_disk(),
+                            timeout=10, ignore_errors=True)
         # Get the scsi device name
         result = process.run("lsscsi|grep scsi_debug|awk '{print $6}'",
                              shell=True)


chunfuwen:
2023-08-04 02:51:49,183 process          L0658 INFO | Running '/sbin/modprobe scsi_debug dev_size_mb=100 ptype=1'
2023-08-04 02:51:49,234 process          L0740 INFO | Command '/sbin/modprobe scsi_debug dev_size_mb=100 ptype=1' finished with 0 after 0.048040036s
2023-08-04 02:51:49,234 process          L0658 INFO | Running 'lsscsi|grep scsi_debug|awk '{print $6}''
2023-08-04 02:51:49,239 process          L0470 DEBUG| [stdout] -
2023-08-04 02:51:49,239 process          L0740 INFO | Command 'lsscsi|grep scsi_debug|awk '{print $6}'' finished with 0 after 0.003242141s
2023-08-04 02:51:50,241 process          L0658 INFO | Running 'lsscsi|grep scsi_debug|awk '{print $6}''
2023-08-04 02:51:50,246 process          L0470 DEBUG| [stdout] /dev/st0

chunfuwen:
WARN 1-type_specific.io-github-autotest-libvirt.scsi_device_manipulate.hotplug.scsi_hostdev.positive.tap_library -> TestWarn: Test passed but there were warnings during execution. Check the log for details.

virttest/utils_test/libvirt.py
```diff
@@ -2878,6 +2878,15 @@ def create_scsi_disk(scsi_option, scsi_size="2048"):
...
         linux_modules.load_module("scsi_debug dev_size_mb=%s %s" %
                                   (scsi_size, scsi_option))
+
+        def _get_scsi_disk():
+            """attempt to get scsi disk multiple times until getting real one"""
+            scsi_disk = process.run("lsscsi|grep scsi_debug|awk '{print $6}'",
+                                    shell=True).stdout_text.strip()
+            return 'dev' in scsi_disk
+
+        utils_misc.wait_for(lambda: _get_scsi_disk(),
```
dzhengfy, line 2888: If timeout is met, should we handle the error?  If that, we only need to invoke L 2891 when no timeout?

chunfuwen: utils_misc.wait_for () itself has limitation that it can not return the late iteration result. So actually here it works similar to time.sleep(), but maybe a little smarter return earlier if condition is met.
So line 2891 still is in need

chunfuwen: @dzhengfy 
1)here utils_misc.wait_for() just work similar to time.sleep,and let's lsssci output have 'dev' in their output
2)nothing else logic need be changed, so we don't need handle the error since the following code will handle that before

dzhengfy (approved):
LGTM

virttest/utils_test/libvirt.py
dzhengfy: I just feel there is little redundent codes, but acceptable.

nanli1 (approved):

PR #3732, opened 2023-08-04
I1101 E721 are newly enforced python code style

chunfuwen:
I1101 E721 are newly enforced python code style
This cause multiple checking failure in currect codes, so disable them

PR #3734, opened 2023-08-04
Fix static checker

smitterl:
Alternative solution for https://github.com/avocado-framework/avocado-vt/pull/3732
Ignore I1101 but fix E721

Fixes for E721 applied as indicated by https://github.com/PyCQA/pycodestyle/blob/1c2147c703e3b934675b270391fab03cb30d73a2/CHANGES.txt#L12

### DIFF
--- a/scripts/regression.py
+++ b/scripts/regression.py
@@ -235,7 +235,7 @@ Please check sysinfo directory in autotest result to get more details.
 
         for line in range(len(s1)):
             tmp = []
-            if type(s1[line]) != list:
+            if not isinstance(s1[line], list):
                 tmp = s1[line]
             else:
                 if len(s1[line][0]) < 2:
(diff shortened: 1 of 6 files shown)


scripts/regression.py
```diff
@@ -235,7 +235,7 @@ def getTtestPvalue(self, fs_dict1, fs_dict2, paired=None, ratio=None):
 
         for line in range(len(s1)):
             tmp = []
-            if type(s1[line]) != list:
+            if not isinstance(s1[line]), list):
```
Yingshun, line 238:
```suggestion
            if not isinstance(s1[line], list):
```

smitterl: Thanks!

Yingshun (approved):

dzhengfy (approved):
LGTM

PR #3737, opened 2023-08-10
Multi hosts migration cvm resmgr

zhencliu:


virttest/vt_agent/managers_nopool_imp/cvm_resmgr/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+class LaunchSecurity(object):
+    _instance = None
+
+    @classmethod
+    def dispatch(cls):
+        return cls._instance
+
+    @classmethod
+    def startup(cls, config):
```
zhencliu, line 13: 1. job starts: from serialized object, loading into memory; detect the current env for resource refreshment - initializing step
2. startup: maintain mgr itself states
3. resource can be passed from one case to another(continuous), from job to job, but a case only live for 'some time', so when to do cleanup work for the resource.
4. resource lifecycle management

PR #3739, opened 2023-08-14
env_process: Refactor BridgeConfig into a Setuper

bgartzi:
Writing the code to setup/cleanup a bridge into a `Setuper` subclass. Then, registering it in the `setup_manager` from `virttest.env_process`, instead of calling the old set/cleanup code.

ID: 2228738

### DIFF
--- a/virttest/test_setup/networking.py
+++ b/virttest/test_setup/networking.py
@@ -4,6 +4,7 @@ import re
 import urllib.request
 
 from virttest.test_setup.core import Setuper
+from virttest.test_setup import PrivateBridgeConfig, PrivateOvsBridgeConfig
 
 
 class NetworkProxies(Setuper):
@@ -26,3 +27,48 @@ class NetworkProxies(Setuper):
             process in which such installation has taken place.
         """
         pass
+
+
+class BridgeConfig(Setuper):
+
+    def setup(self):
+        setup_pb = False
+        ovs_pb = False
+        for nic in self.params.get('nics', "").split():
+            nic_params = self.params.object_params(nic)
+            if nic_params.get('netdst') == 'private':
+                setup_pb = True
+                params_pb = nic_params
+                self.params['netdst_%s' % nic] = nic_params.get("priv_brname", 'atbr0')
+                if nic_params.get("priv_br_type") == "openvswitch":
+                    ovs_pb = True
+
+        if setup_pb:
+            if ovs_pb:
+                brcfg = PrivateOvsBridgeConfig(params_pb)
+            else:
+                brcfg = PrivateBridgeConfig(params_pb)
+            brcfg.setup()
+
+    def cleanup(self):
+        setup_pb = False
+        ovs_pb = False
+        for nic in self.params.get('nics', "").split():
+            nic_params = self.params.object_params(nic)
+            if nic_params.get('netdst') == 'private':
+                setup_pb = True
+                params_pb = nic_params
+                break
+        else:
+            setup_pb = self.params.get("netdst") == 'private'
+            params_pb = self.params
+
+        if params_pb.get("priv_br_type") == "openvswitch":
+            ovs_pb = True
+
+        if setup_pb:
+            if ovs_pb:
+                brcfg = PrivateOvsBridgeConfig(params_pb)
+            else:
+                brcfg = PrivateBridgeConfig(params_pb)
+            brcfg.cleanup()
--- a/selftests/unit/test_setup_networking.py
+++ b/selftests/unit/test_setup_networking.py
@@ -1,7 +1,8 @@
 import unittest
 from unittest.mock import Mock, patch
 
-from virttest.test_setup.networking import NetworkProxies
+from virttest.utils_params import Params
+from virttest.test_setup.networking import NetworkProxies, BridgeConfig
 
 
 class TestProxySetuper(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


bgartzi:
Hi @YongxueHong,

I'm adding the patch refactoring the bridge set/cleanup. Could you check this patch and share your thoughts please?

Thanks a lot!

virttest/test_setup/networking.py
```diff
@@ -26,3 +27,35 @@ def cleanup(self):
...
+class BridgeConfig(Setuper):
+
+    def __init__(self, test, params, env):
+        setup_pb = False
+        ovs_pb = False
+        for nic in params.get('nics', '').split():
+            nic_params = params.object_params(nic)
+            if nic_params.get('netdst') == 'private':
+                setup_pb = True
+                params_pb = nic_params
```
YongxueHong, line 41: Just one comment for assigning the `params_pb`, I compared this part between the setup and cleanup process, and found it is a little different.
In the setup process, the `params_pb` will be overridden by `nic_params` for each loop, but in the cleanup process, it just is assigned once if `netdst` is `private`. 
However, I didn't know why it handled this part in a different way before.
To avoid the potential regression, I suggest keeping the current code implementation during the refactoring process, We could handle the issues in the future or the next stage if it is really needed to do.
Please let me know your idea.
Thanks.

bgartzi: It seems that the logic you mention was introduced in the first implementation of private bridge setup/cleanup available (55df394376c3fad608e0bc52dacb6ecf6d3f4a20). This logic asymmetry haven't been updated since then.

### TL;DR
IMO, the behavior of the previous logic could bring some trouble that the one proposed here would not.

### First case: multiple bridge configurations

Let's suppose the case in which a test-case tries to set up more than 1 bridge (this is not currently supported by any of the versions of the code, as only one bridge is setup by test-case, but anyway). The configuration would look something similar to:
```yaml
nics = "nic1 nic2"
netdst_nic1 = private
netdst_nic2 = private
priv_br_type_nic2 = openvswitch
<nic1 bridge params>
<nic2 bridge params>
```
Then, the actual proposal would set up the bridge based on the configuration of `nic1` and would clean up based on the same configuration.

However, the previous version of the code would setup based on the configuration of `nic2`, setting a OVS bridge in this case. Later on, during the cleanup procedure, if would try to clean the `nic1` config, though.  That would probably raise an error and what's more, wouldn't clean the OVS bridge for the next test case to come.

Anyway, that's a weird situation, as AFAIK, there isn't any test-case in which more than 1 private network bridge has to be set up. I could also confirm that with @yanglei-rh before.

### Second case: one and only bridge configuration

Let's suppose the case that was currently supported. In that case (i.e. only one private bridge in the configuration), there shouldn't be any difference from one to the other version.
In the previous version:

- Setup would get the last `nic` which `<nic>_netdst == 'private'` in the list. That is: the only one available.
- Cleanup would get the first one: That is, the same as before, as there's only one configured private bridge.

In the current proposal both setup and cleanup would act on the same bridge, as both steps are based on the same configuration parsing logic.

In that case, there shouldn't be any issue, AFAIK. LMK if you could suspect any weird/obscure/hidden behavior that could mess this up.

### Third case: Dynamically updated bridge configuration
 
This is one of the weird/obscure/hidden behavior that comes to my mind: A test case sets up its own private bridge, updates the general environment parameter dict and waits for the `postprocess` function to clean that up.

In that case, both of the code versions would be vulnerable to cleaning up only one of the bridges that were configured during the execution.

In the previous code, the last one in the list of `nics` would be cleaned up and that would depend on how the test case appended the nic id to the configuration.

The actual proposal wouldn't depend on that, as it would try to clean the same bridge as it set up before. However, it wouldn't clean the one set up by the test case, as it couldn (but not necessarily) with the previous version.

### Forth case: the test case sets and cleans up the bridge

I looked through the main upstream providers: tp-libvirt and tp-qemu. It seems that there is one test case setting its own private bridge in the latter. `tp-qemu/qemu/tests/timedrift_adjust_time.py` doesn't configure a bridge in its configuration file. Instead, it sets its own configuration on its parameters, sets the bridge and it cleans it up afterwards.

At the time (10th of November of 2014, according to when that part of the test-case was added https://github.com/autotest/tp-qemu/commit/1d60fc6bd98d00fe9013e174676447b940618a27)  it was possible to set up private bridges by just adding a couple of parameters to the test case configuration. In fact, even setting up OVS bridges in `env_process` was supported since July of that same year (see 4d0e1be3c366160d89af2761a9ee04d1b379d8fa). Interestingly, the test case follows a similar setup/cleanup pattern to the one I'm trying to add in this PR.

AFAIK, there are not bugs tracking issues happened by this.

### Getting back to the point
If we really are concerned by possible regressions introduced by these changes, I could stick as similar as possible to the previous code logic/structure. However, due to the reasons mentioned above, I think the new proposal could bring an even safer behavior due to how it handles multiple-bridge configs compared to the previous version.

Let me know your thoughts Yongxue and sorry for the long message. Research might have gone too far

bgartzi: Hi Yongxue,

First of all, I'm sorry for the late reply to this. I thought I did at least leave a "I'll be working on it" comment.

I'm sharing a version that keeps the `BridgeSetuper` code closer to the original code from the former `env_process.py` file as you suggested, I hope that helps.

However, I noticed that the actual shared version of the code is not able to properly handle cases in which the bridge parameters are not global to the cartesian dict, but unique for the nic being configured. That is, in other words, if configuration looks somehow like:
```
{...
    'nics': 'bridge',
    'netdst_bridge': 'private',
...}
```
the bridge will be in fact set up but the cleanup methods won't be called at all. That was already present on the previous `env_process` code, so I guess it's not a problem at all for any avocado-vt user configuring bridges. I guess it's not a priority, but I think it's a bug that could bring a headache to someone trying to configure it that way in the future.

Let me know what do you think.

Thanks!

YongxueHong: Hi @bgartzi 
I understand your concern and I agree with the current solution totally.
Actually, there is a potential issue for the multiple NICs with private bridge, not only for the pre-process but also for the post-process. We have not met this scenario, but we should leave it as the next job after the refactoring or as you mentioned: `it's a bug that could bring a headache to someone trying to configure it that way in the future.` and then fix it :)
Let me know what is your opinion.
Thanks.

bgartzi: Hi @YongxueHong,

I agree with you. I will work on that on a separated issue. I will ping you when that's ready.

Thanks a lot!

YongxueHong:
Hi @bgartzi 
Would you like to help to run the regression test with this patch for the different types of `netdst`? Thanks.

bgartzi:
I just rebased the source branch on top of the latest master branch.

YongxueHong:
Hi @leidwang 
Would you like to help review it? 
Thanks in advance.

yanglei-rh:
> Hi @leidwang Would you like to help review it? Thanks in advance.

Hi @YongxueHong Looks like you'd like remind me to help review it, right?

yanglei-rh:
LGTM
I used this patch to test full virtual network testing on switch, ovs common and ovs only, everything are works fine.

YongxueHong:
Thanks @bgartzi @yanglei-rh 
Let's merge it.

PR #3742, opened 2023-08-16
Add a parameter for TLS connection

cliping:
Some scenarios don't need 'ip_address' in cert. So add a parameter to
 determine whether the 'ip_address' needs to be set.


chunfuwen (approved):
lgtm

PR #3743, opened 2023-08-16
Add a function for the disk shrink generated event check

peixiu:
Before, we have not check for shrink disk event after shrink action was executed. Now, added it in, it's necessary to us to anlyse the issue and also more reasonable.

ID: 1200

### DIFF
--- a/virttest/utils_windows/drive.py
+++ b/virttest/utils_windows/drive.py
@@ -88,6 +88,19 @@ def shrink_volume(session, vol_id, size):
     session.cmd(shrink_cmd.format(script_path))
 
 
+def check_shrink_completion_event(session, vol_id):
+    """
+    Check the disk extend or shrink finished in windows guest.
+
+    :param session: Session object.
+    :param vol_id: Drive letter.
+    """
+    cmd = 'wevtutil qe Application /q:"*[System/EventID=258]" /f:text'
+    status, output = session.cmd_status_output(cmd)
+    msg = "The storage optimizer successfully completed shrink on Windows (%s:)" % vol_id
+    return status == 0 and msg in output
+
+
 def get_disk_props_by_serial_number(session, serial_number, props):
     """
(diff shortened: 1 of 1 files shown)


qingwangrh:
LGTM

peixiu:
@YongxueHong Could you help to review this patch? thanks~

virttest/utils_windows/drive.py
```diff
@@ -88,6 +88,19 @@ def shrink_volume(session, vol_id, size):
     session.cmd(shrink_cmd.format(script_path))
 
 
+def check_shrink_completion_event(session, vol_id):
```
YongxueHong, line 91: Hi @peixiu 
Sorry for replying so late.
From my understanding of https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/wevtutil  and the purpose of this module `Windows drive utilities`.
```
{qe | query-events} <Path> [/lf:<Logfile>] [/sq:<Structquery>] [/q:<Query>] [/bm:<Bookmark>] [/sbm:<Savebm>] [/rd:<Direction>] [/f:<Format>] [/l:<Locale>] [/c:<Count>] [/e:<Element>]  
Reads events from an event log, from a log file, or using a structured query. By default, you provide a log name for <Path>. However, if you use the /lf option, then <Path> must be a path to a log file. If you use the /sq parameter, <Path> must be a path to a file that contains a structured query.
```
I think the function `check_shrink_completion_event` is close to the case level, not a framework API.
My idea is that we could create a function to warp the drive of querying event logs here, like 
`def query_event_logs(session)` or `def get_event_logs(session)`, and then we could use the returning info from the previous function to complete our checking in the test case.
I have another suggestion that we could put this code into the provider of tp-qemu since it looks like a tp-qemu's function and it doesn't relate to the Windows utils drive tool so it doesn't fit here.
Please let me know your opinion. Thanks.

peixiu:
Move the function to case level, thanks all~

PR #3618, opened 2023-02-09
Add missing attributes to the test status

richtja:
In avocado #5586 we added missing tributes for result plugins. This is a fix for avocado-vt runner.

Reference: https://github.com/avocado-framework/avocado/pull/5586
Signed-off-by: Jan Richter <jarichte@redhat.com>

PR #3745, opened 2023-08-17
vt_runner: Add missing whiteboard to the test status

PaulYuuu:
After switched to nrunner, whiteboard is missing from the result, so it is always empty to show. This commit add `WhiteboardMessage` to catch it.

ID: 2232529


PaulYuuu:
Hello @richtja @luckyh, would you please help to review? I referred to this: #3618, but not sure it's the correct place to fix this.

luckyh (approved):
LGTM

clebergnu (approved):
LGTM, thanks!

Issue #3748, opened 2023-08-21
'ImageFont' object has no attribute 'getsize'


luckyh:
Got this error with `Pillow==10.0.0` when spawning VM screendumps due to `getsize` has been removed in Pillow 10.

```
# Pillow==9.5.0 

>>> from PIL import ImageFont
>>> font = ImageFont.load_default()
>>> font.getsize('1')
<stdin>:1: DeprecationWarning: getsize is deprecated and will be removed in Pillow 10 (2023-07-01). Use getbbox or getlength instead.
(6, 11)
```

```
# Pillow==10.0.0 

>>> from PIL import ImageFont
>>> font = ImageFont.load_default()
>>> font.getsize('1')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'ImageFont' object has no attribute 'getsize'
>>> font.getbbox('1')
(0, 0, 6, 11)
```

PR #3749, opened 2023-08-21
Fix an issue of ppm_utils

luckyh:
We could see the following error popped up when enabling screendumps

> 'ImageFont' object has no attribute 'getsize'

so let's fix it.

Issue: https://github.com/avocado-framework/avocado-vt/issues/3748

### DIFF
--- a/virttest/ppm_utils.py
+++ b/virttest/ppm_utils.py
@@ -424,7 +424,10 @@ def add_timestamp(image, timestamp, margin=2):
     font = ImageFont.load_default()
     watermark = time.strftime('%c', time.localtime(timestamp))
     # bar height = text height + top margin + bottom margin
-    bar_height = font.getsize(watermark)[1] + 2 * margin
+    if hasattr(font, 'getbbox'):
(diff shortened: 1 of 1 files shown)


luckyh:
@YongxueHong @yanan-fu @PaulYuuu would you mind to help review this PR? thanks!

YongxueHong (approved):
LGTM.

PaulYuuu (approved):
Refer to source, as `getbbox` will insert `left` and `right` in the return value, so this fix LGTM.
```python
    def getbbox(self, text, *args, **kwargs):
        """
        Returns bounding box (in pixels) of given text.

        .. versionadded:: 9.2.0

        :param text: Text to render.
        :param mode: Used by some graphics drivers to indicate what mode the
                     driver prefers; if empty, the renderer may return either
                     mode. Note that the mode is always a string, to simplify
                     C-level implementations.

        :return: ``(left, top, right, bottom)`` bounding box
        """
        _string_length_check(text)
        width, height = self.font.getsize(text)
        return 0, 0, width, height
```

PR #3752, opened 2023-08-25
virsh: enable page size unit definition

dzhengfy:

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -4686,12 +4686,13 @@ def domstats(domains="", options="", **dargs):
     return command(cmd, **dargs)
 
 
-def freepages(cellno=None, pagesize=None, options="", **dargs):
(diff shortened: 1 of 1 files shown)


virttest/virsh.py
```diff
@@ -4701,6 +4702,8 @@ def freepages(cellno=None, pagesize=None, options="", **dargs):
         cmd += " --cellno %s" % cellno
     if pagesize is not None:
         cmd += " --pagesize %s" % pagesize
+    if pagesize and sizeunit:
```
chloerh, line 4705: Could we put it this way?
```
    if pagesize is not None:
        cmd += " --pagesize %s" % pagesize
        if sizeunit:
             cmd += sizeunit

dzhengfy: Yours is better. Updated.

dzhengfy:
I have checked the provider cases and no impact.

PR #3760, opened 2023-09-04
test_setup: add new param

dzhengfy:
Sometimes less pages are allocated, but it is acceptable, so no need to raise an error

### DIFF
--- a/virttest/test_setup/__init__.py
+++ b/virttest/test_setup/__init__.py
@@ -515,13 +515,16 @@ class HugePageConfig(object):
         obj = kernel_interface.SysFS(node_page_path)
         return obj.sys_fs_value
 
-    def set_node_num_huge_pages(self, num, node, pagesize):
+    def set_node_num_huge_pages(self, num, node, pagesize, ignore_error=False):
         """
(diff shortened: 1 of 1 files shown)


chunfuwen (approved):
lgtm

virttest/test_setup/__init__.py
```diff
@@ -515,22 +515,31 @@ def get_node_num_huge_pages(self, node, pagesize, type="total"):
...
-            raise ValueError("Cannot set %s hugepages on node %s, please check"
-                             " if the node has enough memory" % (num, node))
+            if not ignore_error:
+                raise ValueError("Cannot set %s hugepages on node %s, please check"
+                                 " if the node has enough memory" % (num, node))
+            else:
+                LOG.warning("Only %s pages are allocated instead "
+                            "of %s pages", obj.sys_fs_value, int(num))
+                return False
+        return True
```
chloerh, line 542: Would you consider merging the error/waring message for these 2 scenarios?

dzhengfy: updated

chloerh (approved):

PR #3762, opened 2023-09-06
Add sev qmp support

zixi-chen:
Add qmp cmd about sev platform, capabilities, launch measurement
and attestation report.

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -3635,3 +3635,45 @@ class QMPMonitor(Monitor):
         cmd = "query-sgx-capabilities"
         self.verify_supported_cmd(cmd)
         return self.cmd(cmd)
+
+    def query_sev(self):
+        """
+        Query SEV information. (Since 2.12)
+
+        :return: Information about SEV
+        """
+        cmd = "query-sev"
+        self.verify_supported_cmd(cmd)
+        return self.cmd(cmd)
+
+    def query_sev_capabilities(self):
+        """
+        Query SEV capabilities. (Since 2.12)
+
+        :return: SEV capability objects.
+        """
+        cmd = "query-sev-capabilities"
+        self.verify_supported_cmd(cmd)
+        return self.cmd(cmd)
+
+    def query_sev_launch_measure(self):
+        """
+        Query the SEV guest launch information. (Since 2.12)
+
+        :return: The SEV launch measure info for the guest.
+        """
+        cmd = "query-sev-launch-measure"
+        self.verify_supported_cmd(cmd)
+        return self.cmd(cmd)
+
+    def query_sev_attestation_report(self, mnonce):
+        """
+        Query SEV attestation report. (Since 6.1)
+
+        :param mnonce: A random 16 bytes value encoded in base64
+        :type mnonce: str
+        :return: The SEV attestation report objects.
+        """
+        cmd = "query-sev-attestation-report"
+        self.verify_supported_cmd(cmd)
+        return self.cmd(cmd, {"mnonce": mnonce})


zixi-chen:
@XueqiangWei @zhencliu Could you please help to review this patch?

virttest/qemu_monitor.py
```diff
@@ -3635,3 +3636,46 @@ def query_sgx_capabilities(self):
...
+    def query_sev_attestation_report(self):
+        """
+        Query SEV attestation report. (Since 6.1)
+        Note that it needs to generate a random 16 bytes value encoded
+        in base64 to query attestation report.
+
+        :return: The SEV attestation report objects.
+        """
+        cmd = "query-sev-attestation-report"
+        mnonce = base64.b64encode(os.urandom(16)).decode()
```
zhencliu, line 3679: as mnonce will be included in the report, do we want to set the mnonce outside?
e.g. set it in test case, call this function, verify if the mnonce is included, or if the data from the report is returned in an expected way

zhencliu:
add @YongxueHong

virttest/qemu_monitor.py
zixi-chen: Thanks Zhenchao! Yes we can set mnonce outside, my concern is if anyone who not familiar about this function, they may not generate a mnonce with the right encoding method, so I put it in this function. By the way, could we make this an optional input, if no input, then we generate  mnonce inside it.

virttest/qemu_monitor.py
```diff
@@ -3635,3 +3636,46 @@ def query_sgx_capabilities(self):
...
+        self.verify_supported_cmd(cmd)
+        return self.cmd(cmd, {"arguments": {"mnonce": mnonce}})
```
YongxueHong, line 3681:
```suggestion
    def query_sev_attestation_report(self, mnonce):
        """
        Query SEV attestation report. (Since 6.1)
 
        :param mnonce: A random 16 bytes value encoded in base64
        :type mnonce: str
        :return: The SEV attestation report objects.
        """
        cmd = "query-sev-attestation-report"
        self.verify_supported_cmd(cmd)
        return self.cmd(cmd, {"arguments": {"mnonce": mnonce}})
```
Here, we focus on warping the qmp command `query-sev-attestation-report` according to https://github.com/qemu/qemu/blob/c152379422a204109f34ca2b43ecc538c7d738ae/qapi/misc-target.json#L215, not better to include the upper-level logic, we could let the caller handle the generator of the mnonce and then pass it to the function.
BTW, we could not provide the default value for it, beside the qmp will generate a default value if we did not pass the `mnonce` argument, like https://github.com/qemu/qemu/blob/c152379422a204109f34ca2b43ecc538c7d738ae/qapi/block.json#L328-L334 and https://github.com/avocado-framework/avocado-vt/blob/7c8692a2e2fa21307c742af549f082bb2b4c07b1/virttest/qemu_monitor.py#L2842-L2871

virttest/qemu_monitor.py
zhencliu: If no input, currently, it means this option is not required by the qmp cmd, e.g. speed for block commit, qemu will use a default value, in that case, we can cover the scenario that using default values.

No worry. As a public interface, we need to write the doc string carefully to tell users how to call it, if they use it in a wrong way, then they should get an error or something likely, which is expected

zixi-chen:
Thanks @zhencliu and @YongxueHong I updated the mnonce function, the mnonce will be generated in the case. Could you please review again?

Test log:
[stdlog] 2023-09-07 05:46:41,242 avocado.virttest.qemu_monitor DEBUG| (monitor vm3.qmpmonitor1) Sending command 'query-sev-attestation-report'
[stdlog] 2023-09-07 05:46:41,242 avocado.virttest.qemu_monitor DEBUG| Send command: {"execute": "query-sev-attestation-report", "arguments": {"mnonce": "xmB1aepyedclcsn8t8Eivw=="}, "id": "gaDWs8eD"}
[stdlog] 2023-09-07 05:46:41,247 avocado.virttest.qemu_monitor DEBUG| (monitor vm3.qmpmonitor1) Response to 'query-sev-attestation-report' (re-formatted)
[stdlog] 2023-09-07 05:46:41,247 avocado.virttest.qemu_monitor DEBUG| (monitor vm3.qmpmonitor1)    data: xmB1aepyedclcsn8t8EivxwJnIX5YyDp9mTEUpU2bMqkWqW6hharrYgz+KIlg47kAwAAAAIQAAACAAAAAAAAAGQn2SQr+aeJ7rwpyjDwYlv/mcshOIVgPhFORoHaa6SppjuSiNEV8IB14DrWVmVZMQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAMNwmomgvqugwHROqKQlQb3OUZPiikFokeHy8csmxNetuNx92KxFE4hin2XIfDHeUwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==

zhencliu (approved):

YongxueHong (approved):

YongxueHong (approved):
LGTM

PR #3765, opened 2023-09-08
qemu_vm: includes memory preallocation for hugepages

mcasquer:
qemu_vm: includes memory preallocation for hugepages

Using hugetlb memory without preallocation is a dangerous practice.
As confirmed, setting preallocation 'true' by default for all the
memory backed by hugepages.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1090,6 +1090,8 @@ class VM(virt_vm.BaseVM):
                 name = "machine_mem"
                 backend_options = {}
                 backend_options["size_mem"] = "%sM" % params["mem"]
+                if params.get("vm_mem_prealloc"):
+                    backend_options["prealloc_mem"] = params.get("vm_mem_prealloc")
                 if params.get("vm_mem_backend"):
                     backend_options["backend_mem"] = params.get("vm_mem_backend")
                 if params.get("vm_mem_backend") == "memory-backend-file":
@@ -1100,6 +1102,7 @@ class VM(virt_vm.BaseVM):
                 if params.get("hugepage_path"):
                     backend_options["backend_mem"] = "memory-backend-file"
                     backend_options["mem-path_mem"] = params["hugepage_path"]
+                    backend_options["prealloc_mem"] = params.get("vm_mem_prealloc", "yes")
                 backend_options["share_mem"] = params.get("vm_mem_share")
                 backend_param = Params(backend_options)
                 dev = devices.memory_object_define_by_params(backend_param,


virttest/qemu_vm.py
```diff
@@ -1100,6 +1100,7 @@ def add_memorys(devices, params):
                 if params.get("hugepage_path"):
                     backend_options["backend_mem"] = "memory-backend-file"
                     backend_options["mem-path_mem"] = params["hugepage_path"]
+                    backend_options["prealloc_mem"] = params.get("prealloc_mem", "yes")
```
PaulYuuu, line 1103: `prealloc` is a common option for memory objects, not only for hugepage.
```
  prealloc-context=<link<thread-context>> - Context to use for creating CPU threads for preallocation
  prealloc-threads=<int> - Number of CPU threads to use for prealloc
  prealloc=<bool>        - Preallocate memory
```

You can set default to yes for hugepage, but also need to consider the `vm_mem`.

luckyh, line 1103:
```suggestion
                    backend_options["prealloc_mem"] = params.get("vm_mem_prealloc", "yes")
```
to get a consistent naming convention for the VM-level memory object.

mcasquer: @PaulYuuu sorry I don't get your point, what's exactly your suggestion? Is it related with the [comment below](https://github.com/avocado-framework/avocado-vt/pull/3765#discussion_r1320928240)?

mcasquer: @luckyh 
Just one question, should this naming convention apply to other places, i.e. tp-qemu, where this param is being requested as `prealloc_mem` ?

PaulYuuu: Yes @mcasquer, `memory-backend-ram`, `memory-backend-file` and `memory-backend-memfd` are support `prealloc`, so you can use `vm_mem_prealloc` to set it. But for `hugepage`, I think directly set `backend_options["prealloc_mem"] = "yes"` is okay as suggested  by the developer. Or you can also make if as configurable: `backend_options["prealloc_mem"] = params.get("vm_mem_prealloc", "yes")
`

mcasquer: > Yes @mcasquer, `memory-backend-ram`, `memory-backend-file` and `memory-backend-memfd` are support `prealloc`, so you can use `vm_mem_prealloc` to set it. But for `hugepage`, I think directly set `backend_options["prealloc_mem"] = "yes"` is okay as suggested by the developer. Or you can also make if as configurable: `backend_options["prealloc_mem"] = params.get("vm_mem_prealloc", "yes") `

@PaulYuuu got it, thanks for the explanation :grinning:

luckyh: > @luckyh Just one question, should this naming convention apply to other places, i.e. tp-qemu, where this param is being requested as `prealloc_mem` ?

@mcasquer not really, that's because we want to distinguish the VM-level param (`vm_mem_prealloc`) from the device-level param (`prealloc_mem`) at the current stage and let `prealloc_mem` be only handled by [here](https://github.com/avocado-framework/avocado-vt/blob/7d5d9b0851764e03bdeb6393b9b9e00eb51066e0/virttest/qemu_vm.py#L1061).

mcasquer: @luckyh got it, clear from my side

mcasquer:
@PaulYuuu @luckyh could this PR be merged? Thanks!

virttest/qemu_vm.py
```diff
@@ -1100,6 +1100,7 @@ def add_memorys(devices, params):
                 if params.get("hugepage_path"):
                     backend_options["backend_mem"] = "memory-backend-file"
                     backend_options["mem-path_mem"] = params["hugepage_path"]
+                    backend_options["prealloc_mem"] = params.get("vm_prealloc_mem", "yes")
```
luckyh, line 1103:
```suggestion
                    backend_options["prealloc_mem"] = params.get("vm_mem_prealloc", "yes")
```

PaulYuuu:
> @PaulYuuu @luckyh could this PR be merged? Thanks!

I'm afraid not, when I want to set this option in `memory-backend-ram` and `memory-backend-memfd`, what can I do?

The situation this topic leads to is: for VM mem backend, we only have a few configurable options, but we support many options, in qcontainer.py, we have `attrs = qdevices.Memory.__attributes__[params["backend"]][:]`

@luckyh @mcasquer, Do you think it is worth referring to other memory objects?

virttest/qemu_vm.py
luckyh, line 1103: As what @PaulYuuu pointed out, we'd better support setting the `prealloc` option for non-hugepage case as well, within this context.

luckyh:
@zhenyzha  would you mind to help review this PR? thanks!

virttest/qemu_vm.py
mcasquer: Sorry for the typo, updated now!

mcasquer:
> 

@PaulYuuu IMHO, this should be only done for hugepages backed memory. For the others it's not important if you preallocate the memory at vm boot up. Anyways, IIUC, and you want to specify this option for a `memory-backend-ram` or `memory-backend-memfd` object, the `prealloc_$mem_id` could be included in the test cfg file as it's already implemented.

virttest/qemu_vm.py
mcasquer: @luckyh It's not needed, only in a few already-implemented cases prealloc is used for others than `memory-backend-file`

PaulYuuu: Even though we have test cases covering prealloc, the VM default memory backend also supports all attrs.

My concern is this: in the future, we want to support another option, add support to VT, then support another option, and then add support again, this cycle continues.

The history of vm_mem is at the first, only support `memory-backend-ram`, and then supports custom `vm_mem_backend` and hugepage, and now we want to support a new option.

zhenyzha: @mcasquer I understand the original intention of this patch.
But other backends that the maintainer considers for support also need to be considered.
Because the current framework does not support setting preallocate for all memory-backends.
For example, if we want to test `prealloc` on `memory-backend-memfd` in the future, then we need to issue a similar patch to support it.
So it makes sense to support it in all memory backends this time.

mcasquer: @PaulYuuu @zhenyzha I got the idea now. In that case we can do it now for all the memory-backend options.

mcasquer: I guess the `backend_options["prealloc_mem"]` has to be extracted from the `hugepage_path` condition

mcasquer:
@PaulYuuu @luckyh @zhenyzha could you review again this PR? Thanks!

virttest/qemu_vm.py
```diff
@@ -1090,6 +1090,7 @@ def add_memorys(devices, params):
                 name = "machine_mem"
                 backend_options = {}
                 backend_options["size_mem"] = "%sM" % params["mem"]
+                backend_options["prealloc_mem"] = params.get("vm_mem_prealloc")
```
luckyh, line 1093:
```suggestion
                if params.get("vm_mem_prealloc"):
                    backend_options["prealloc_mem"] = params.get("vm_mem_prealloc")
```

mcasquer: Updated, thanks! :grinning:

PaulYuuu (approved):

luckyh (approved):

mcasquer:
Could this PR be merged? Thanks!

mcasquer:
This is a kindly reminder, could this PR be merged? Thanks!

luckyh:
> This is a kindly reminder, could this PR be merged? Thanks! @luckyh @PaulYuuu

@zhenyzha does the patch look good to you as well?

zhenyzha:
The patch works as expected 
LGTM.Acked-by: Zhenyu Zhang <zhenyzha@redhat.com>

git am 3765.patch
Applying: qemu_vm: includes memory preallocation for hugepages

```
 -object '{"size": 31138512896, "prealloc": true, "mem-path": "/mnt/kvm_hugepage", "id": "mem-machine_mem", "qom-type": "memory-backend-file"}'

 (1/2) Host_RHEL.m9.u4.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.aarch64.page_64k.io-github-autotest-qemu.boot_vm_in_hugepage.2M.arm64-pci: STARTED
 (1/2) Host_RHEL.m9.u4.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.aarch64.page_64k.io-github-autotest-qemu.boot_vm_in_hugepage.2M.arm64-pci: PASS (345.50 s)
 (2/2) Host_RHEL.m9.u4.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.aarch64.page_64k.io-github-autotest-qemu.boot_vm_in_hugepage.2M.specify_hp_file.arm64-pci: STARTED
 (2/2) Host_RHEL.m9.u4.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.aarch64.page_64k.io-github-autotest-qemu.boot_vm_in_hugepage.2M.specify_hp_file.arm64-pci: PASS (336.96 s)
```

PR #3775, opened 2023-10-07
test_setup: fix incorrect format

dzhengfy:

### DIFF
--- a/virttest/test_setup/__init__.py
+++ b/virttest/test_setup/__init__.py
@@ -532,9 +532,10 @@ class HugePageConfig(object):
         obj.sys_fs_value = num
         # If node has some used hugepage, result will be larger than expected.
         if obj.sys_fs_value < int(num):
-            msg = ("%s (expecting %s) hugepages is set on node %s, "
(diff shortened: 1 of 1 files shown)


virttest/test_setup/__init__.py
```diff
@@ -533,8 +533,9 @@ def set_node_num_huge_pages(self, num, node, pagesize, ignore_error=False):
         # If node has some used hugepage, result will be larger than expected.
         if obj.sys_fs_value < int(num):
             msg = "%s (expecting %s) hugepages is set on node %s, "
-            "please check if the node "
-            "has enough memory" % (obj.sys_fs_value, num, node)
+            msg += "please check if the node has enough memory"
```
chloerh, line 536: How about put '\' at the end?

dzhengfy: Agree. Updated.

chloerh (approved):
There're trailing spaces that need to be removed

dzhengfy:
> There're trailing spaces that need to be removed

Done

PR #3778, opened 2023-10-12
filesystem.py: add boot order attribute for filesystem device

meinaLi:
Add boot order attribute for filesystem device.

### DIFF
--- a/virttest/libvirt_xml/devices/filesystem.py
+++ b/virttest/libvirt_xml/devices/filesystem.py
@@ -10,7 +10,7 @@ from virttest.libvirt_xml import accessors
 
 class Filesystem(base.TypedDeviceBase):
 
-    __slots__ = ('accessmode', 'source', 'target', 'driver', 'binary', 'alias')
+    __slots__ = ('accessmode', 'source', 'target', 'driver', 'binary', 'alias', 'boot')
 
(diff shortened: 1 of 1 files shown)


cliping (approved):
LGTM

virttest/libvirt_xml/devices/filesystem.py
```diff
@@ -29,6 +29,8 @@ def __init__(self, type_name='mount', virsh_instance=base.base.virsh):
         accessors.XMLElementDict('alias', self,
                                  parent_xpath='/',
                                  tag_name='alias')
+        accessors.XMLAttribute('boot', self, parent_xpath='/',
```
chloerh, line 32: We could consider XMLElementDict if there are new attributes to 'boot' element in the future. 
And we could keep the current implement if the chance of adding new attribute is really small.

meinaLi: I think we can keep it in current status because it is unlikely that there will be a new attribute to 'boot' element.  If there are indeed changes later, we can update them together.

chloerh: Of course.

chloerh (approved):

PR #3783, opened 2023-10-22
Complete the libvirt verbosity fix

pevogam:
This performs additional lazy imports necessary to avoid libvirt related messages for qemu-only usage and fixes the way it is done for some of the cases.

### DIFF
--- a/virttest/_wrappers.py
+++ b/virttest/_wrappers.py
@@ -79,6 +79,8 @@ def lazy_import(name):
     :param name: Name of the module that is going to be imported
     :type name: String
     """
+    if name in sys.modules:
+        return sys.modules[name]
     spec = importlib.util.find_spec(name)
     if spec is None:
         raise ImportError(f"Could not import module {name}")
--- a/virttest/cpu.py
+++ b/virttest/cpu.py
@@ -30,12 +30,16 @@ from avocado.utils.software_manager import manager
 from avocado.utils import process
 from avocado.utils import path
 from avocado.core import exceptions
-from virttest import virsh
 from virttest import utils_misc
-from virttest import libvirt_xml
-from virttest.libvirt_xml.xcepts import LibvirtXMLNotFoundError
 from virttest import libvirt_version
 from virttest import data_dir
+
+
+# lazy imports for dependencies that are not needed in all modes of use
+# (as the cpu module is generic to qemu vm use only, libvirt is optional)
+from virttest._wrappers import lazy_import
+virsh = lazy_import("virttest.virsh")
+libvirt_xml = lazy_import("virttest.libvirt_xml")
 
 
 LOG = logging.getLogger('avocado.' + __name__)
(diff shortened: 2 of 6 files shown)


pevogam:
This pull request is a follow up from https://github.com/avocado-framework/avocado-vt/pull/3721 where integration tests found more locations of the most recent master branch that import the libvirt modules in a noisy way.

virttest/_wrappers.py
```diff
@@ -79,6 +79,8 @@ def lazy_import(name):
     :param name: Name of the module that is going to be imported
     :type name: String
     """
+    if name in sys.modules:
+        return sys.modules[name]
```
luckyh, line 83: Suggest splitting the fix into a seperated commit.

virttest/cpu.py
```diff
@@ -125,7 +129,7 @@ def get_cpu_xmldata(vm, options=""):
     cpu_xmldata['mtype'] = vm_xml.os.machine
     try:
         cpu_xmldata['current_vcpu'] = int(vm_xml.current_vcpu)
-    except LibvirtXMLNotFoundError:
+    except libvirt_xml.xcepts.LibvirtXMLNotFoundError:
```
luckyh, line 132: Just curious that as the `libvirt_xml.base` module has already been patched with the lazy importing fix of `virsh` do we really need `libvirt_xml.xcepts` to be imported in a lazy manner?

pevogam: I think I originally had `xcepts` lazily imported too but after the `libvirt_xml` was fixed here I unified both `libvirt_xml` and `libvirt_xml.xcepts` under the same lazy import. The current diff (at the time of speaking) has been integration tested and no libvirt warnings are emitted meaning that we don't need extra changes to make it work. Note however that some modules might have been imported this way simply because they contain `libvirt` functionality and as such might always become verbose too if somebody imports the virsh or other modules there. In that regard I considered taking this a step further from just tracing the warning and patching up the various locations it comes from right now and decided to isolate libvirt imports semantically for more future proof changes here.

virttest/_wrappers.py
pevogam: Ok, makes sense.

virttest/cpu.py
luckyh: ok, understand your concern now, thanks!

luckyh (approved):
LGTM

luckyh:
Hi @dzhengfy @chloerh @Yingshun , would you mind to check this works for you case? thanks in advance!

dzhengfy:
> Hi @dzhengfy @chloerh @Yingshun , would you mind to check this works for you case? thanks in advance!

We will test with the pr accordingly and will give feedback. Thanks

dzhengfy:
> > Hi @dzhengfy @chloerh @Yingshun , would you mind to check this works for you case? thanks in advance!
> 
> We will test with the pr accordingly and will give feedback. Thanks

We tested gating and migration and cpu jobs with it, the result seems ok. So I approved.

dzhengfy (approved):

luckyh:
Thanks to all.

PR #3784, opened 2023-10-23
Fix typo of parameter ignore_status comments for func check_guest_xml_by_xpaths

liang-cong-red-hat:
Comments of available parameters for ignore_status are mismatched, so correct it.


chloerh (approved):

dzhengfy (approved):
Thanks for fixing it.

PR #3789, opened 2023-10-30
Update event timeout for copy storage migration

cliping:
It takes longer to get the desired event in copy storage migration. So update event timeout.

### DIFF
--- a/virttest/migration.py
+++ b/virttest/migration.py
@@ -412,6 +412,7 @@ class MigrationTest(object):
             after_event = one_func.get('after_event')
             before_event = one_func.get('before_event')
             func = one_func.get('func')
+            wait_for_after_event_timeout = one_func.get('wait_for_after_event_timeout', '30')
             if after_event and not virsh_event_session:
                 raise exceptions.TestError("virsh session for collecting domain "
                                            "events is not provided")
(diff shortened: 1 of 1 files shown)


cliping:
Test result in https://github.com/autotest/tp-libvirt/pull/5234.

nanli1 (approved):

virttest/migration.py
```diff
@@ -408,16 +408,19 @@ def _run_complex_func(vm, one_func, virsh_event_session=None):
             after_event = one_func.get('after_event')
             before_event = one_func.get('before_event')
             func = one_func.get('func')
+            wait_for_event_timeout = one_func.get('wait_for_event_timeout')
             if after_event and not virsh_event_session:
                 raise exceptions.TestError("virsh session for collecting domain "
                                            "events is not provided")
 
+            if not wait_for_event_timeout:
+                wait_for_event_timeout = '30'
```
Yingshun, line 417: Can you plz explain these 2 lines?

chloerh: Seems equal with `wait_for_event_timeout = one_func.get('wait_for_event_timeout', '30')`

cliping: We can use the get() method of a dict object and return a default value if a specific key doesn't exist. If the key exists and it's None, it'll be returned.

For example,
>>> one_func = {'before_pause': 'yes', 'need_sleep_time': '5', 'wait_for_event_timeout': None}
>>> one_func.get('wait_for_event_timeout', '30')
>>> one_func.get('non_exist', '30')
'30'

Yingshun: Ok, so will you merge these 2 lines to 'wait_for_event_timeout = one_func.get('wait_for_event_timeout')' to make it simple?
BTW， is this 'timeout' value paired with 'after_event'? If yes, can we highlight 'after_event' in it's name?

cliping: Updated, thanks.

Yingshun (approved):

chloerh (approved):

PR #3798, opened 2023-11-20
Change vm.destroy gracefulness for performance when discarding vms

pevogam:
During environment setup we discard vms that are not going to participate in a test but doing so gracefully results in a large timeout (60 seconds by default) for each one and slows down the test and overall test run significantly. As the discarded vm will either be fully reconstructed or restored from a clean state when it is needed again by a test, graceful shutdown is not needed and ridding ourselves of it will improve performance for free.

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1156,6 +1156,7 @@ def preprocess(test, params, env):
     # leave them untouched if they have to be disregarded only for this test
     requested_vms = params.objects("vms")
     keep_unrequested_vms = params.get_boolean("keep_unrequested_vms", False)
+    kill_unrequested_vms_gracefully = params.get_boolean("kill_unrequested_vms_gracefully", True)
     for key in list(env.keys()):
         vm = env[key]
         if not isinstance(vm, virt_vm.BaseVM):
@@ -1165,7 +1166,7 @@ def preprocess(test, params, env):
                 LOG.debug("The vm %s is registered in the env and disregarded "
                           "in the current test", vm.name)
             else:
-                vm.destroy()
+                vm.destroy(gracefully=kill_unrequested_vms_gracefully)
                 del env[key]
 
     global KVM_MODULE_HANDLERS


pevogam:
@luckyh Hopefully this is an easy diff and it seems reasonable to you.

luckyh:
Hi @pevogam, thanks for bringing this up! I just have a concern to the following point.

> As the discarded vm will either be fully reconstructed or restored from a clean state when it is needed again by a test ...

However, the guest os images might continue to be used for the rest tests. As unclean shutdowns may trigger the guest encountering into the rescue mode (especially for Windows guests), without any protection being taken (e.g. using snapshots) those tests may fail to run into the end which would be not expected by the users.

So I'd suggest that, if we exactly want such a tweak to be applied, we'd better either

- make the behavior configurable by a param and tell the users about the risk

or

- do ungraceful shutdowns only if there is any protection been activated to the os images

Let me know your thoughts.

pevogam:
> Hi @pevogam, thanks for bringing this up! I just have a concern to the following point.
> 
> > As the discarded vm will either be fully reconstructed or restored from a clean state when it is needed again by a test ...
> 
> However, the guest os images might continue to be used for the rest tests. As unclean shutdowns may trigger the guest encountering into the rescue mode (especially for Windows guests), without any protection being taken (e.g. using snapshots) those tests may fail to run into the end which would be not expected by the users.
> 
> So I'd suggest that, if we exactly want such a tweak to be applied, we'd better either
> 
>     * make the behavior configurable by a param and tell the users about the risk

Indeed I guess you are right about the images, I originally thought the vms are always reconstructed on top of black images since anything else might not isolate tests well enough but I was wrong here. So I pushed changes making this configurable, let me know what you think.

virttest/env_process.py
```diff
@@ -1156,6 +1156,7 @@ def preprocess(test, params, env):
     # leave them untouched if they have to be disregarded only for this test
     requested_vms = params.objects("vms")
     keep_unrequested_vms = params.get_boolean("keep_unrequested_vms", False)
+    protect_unrequested_vms = params.get_boolean("protect_unrequested_vms", True)
```
luckyh, line 1159: However I just kind of worry on if it would promise too much by using the param name `protect_unrequested_vms`, because in fact we don't intent to "protect" these vms (as we will invoke a destroy call afterward) but rather merely shutdown them in a safer manner. Therefore, I'd prefer the param name to be `gracefully_destroy_unrequested_vms`.

pevogam: Hmmm I thought about this as well but I find the whole `gracefully_destroy_...` way too long for a parameter name. I think we cannot typically add all details of the context for a parameter name into the parameter name itself and a typical VT user needing some configuration will look up the context a parameter is used in instead of expecting to understand all of its details from its name. In this sense one could easily see the context where `protect_unrequested_vm` is used and understand more about what protection is meant.

luckyh: > In this sense one could easily see the context where protect_unrequested_vm is used and understand more about what protection is meant.

Well, people would usually have subjective understanding to things and each could hold different opinion against others. I can feel a very similar taste as, for instance, the issue of the `close_hooks` attribute being argued at the PR 3782. What I really concern about is that, if someone in the future would exactly ask for the param to comply with a true 'protect_unrequested_vm' implementation, shall we tell him/her that "well, you cannot just understand the param name as what it is literally meant", just like what I did on that PR?

IMO, having a long param name won't be a real issue compared with picking ambiguous one which cannot explain well about the truth. There were already too much debt needed to be paid within the VT modules so let's be careful with new ones.

luckyh: > ... this would need a separate description attribute to accompany each parameter.

The description associated with each param would be something really help, and I also had some idea about the enhancement at that place. Btw, we do require developers to add descriptions or at least some notes into the corresponding cfg files when introducing new params.

https://github.com/avocado-framework/avocado-vt/blob/065fa0af97a58a06da11a37e3a6423d2689d45b5/virttest/shared/cfg/base.cfg#L8-L9

However, I don't think having a description for the newly introduced param would resolved all the potential issues. My main concern is that, to my understanding the `protect_unrequested_vms` just tell another story which let me could think about a completely different thing rather than the actual one introduced by this PR. So the issue there is, if we are going use *B* to name the A feature (tentatively, no matter whether it a good choice or not), then what word should we use to name the B feature as *B* has already been occupied? If there is no word like *B'* we'd probably have to use, let's say, *C* instead which would be yet another workaround (that's one of the reasons how the debt grew in the past).

> one could look up `*_unrequested_vms parameters` for more context instead of considering `*_unrequested_vms_for_env_process_before_each_test` as part of the parameter name to avoid this

True, but only if we'd never add any other procedures for manipulating the unrequested vms in the afterward stages, otherwise we might need to have something like `*_unrequested_vms_for_env_process_after_each_test` to be distinguished with the former.

From my perspective, the reasons that made the param name longer and longer can be summarized into the following points:

1. cartesian conf does not support isolating params by namespaces.
2. with 1, prefixes/suffixes are required to represent the namespace info for avoiding naming collisions.
3. nested conditions are transformed into multiple boolean params instead of a single enum, therefore a low-hierarchy param may need an additional prefix/suffix as the namespace info, for avoiding naming collisions as well.
   (in our case,
      `keep_unrequested_vms=y/n`
      `gracefully_destroy_unrequested_vms=y/n`
    vs.
      `treat_unrequested_vms=keep/destroy/gracefully_destroy/...`)
4. other tree-like concepts request something similar as 3.

We have been aware of the issue for several years, but unlucky, it's not a high prioritize one in the todo list. So still wait for a proper time slot for having the improvement.

> then how about `shutdown_unrequested_vms` to be the new parameter added to the already existing `keep_unrequested_vms` we use in similar context?

In fact, we could have such a param and that's because *shutdown* could be assigned with a different meaning than *keep*/*destroy*, where the former would just intend to shutdown the vms but keep the definitions while the latter would keep/clear everything. Then the param name could also be `shutdown_unrequested_vms_on_keep` I guess.

pevogam: Considering
```
kill_vm_before_test = no
paused_after_start_vm = no
# Some postprocessor params
kill_vm = no
kill_vm_gracefully = no
kill_unresponsive_vms = no
# Wait time before kill vm
kill_timeout = 60
```
then how about `kill_unrequested_vms_gracefully`? I prefer if the parameter key would be more in line with pre-existing keys. Everywhere in VT where we use `kill_vm` and add the "gracefully" or not we destroy the vm in one of the specified ways.

luckyh: > then how about kill_unrequested_vms_gracefully? I prefer if the parameter key would be more in line with pre-existing keys. Everywhere in VT where we use kill_vm and add the "gracefully" or not we destroy the vm in one of the specified ways.

Yeah, that sounds good me and I agree with you on your point above, thanks!

PaulYuuu (approved):

luckyh:
Thanks all.

PR #3799, opened 2023-11-20
Support new nic driver igb on the framework

yanglei-rh:
Upstream add an new nic model igb since qemu-8.0.0 which allow testing applications' SR-IOV features on a machine without a specific hardware.In order to cover more nic model,so QE sent a new patch to support this driver.

ID: 1543
Signed-off-by: Lei Yang leiyang@redhat.com

### DIFF
--- a/virttest/shared/cfg/guest-hw.cfg
+++ b/virttest/shared/cfg/guest-hw.cfg
@@ -23,6 +23,10 @@ variants:
         only q35
         required_qemu = [2.6.0, )
         nic_model = e1000e
+    - igb:
+        only q35
+        required_qemu = [8.0.0, )
+        nic_model = igb
     - virtio_net:
         nic_model = virtio
         # Assign the queue number of nic device


yanglei-rh:
Hello @PaulYuuu @YongxueHong Could you please help me review this patch? Thanks in advance.

virttest/shared/cfg/guest-hw.cfg
```diff
@@ -23,6 +23,10 @@ variants:
         only q35
         required_qemu = [2.6.0, )
         nic_model = e1000e
+    - igb:
+        only q35
+        required_qemu = [8.1.0, )
```
PaulYuuu, line 28: igb device was introduced from qemu-8.0.0. https://www.mail-archive.com/qemu-devel@nongnu.org/msg947415.html

yanglei-rh: > igb device was introduced from qemu-8.0.0. https://www.mail-archive.com/qemu-devel@nongnu.org/msg947415.html

Downstream qemu 8.0.0 only included a part of code. According to the internal bug it is supported starting from 8.1.0-4 .

PaulYuuu: This is an upstream project and should not only be considered downstream.

yanglei-rh: > This is an upstream project and should not only be considered downstream.

Changed to qemu-8.0.0, please help review it again.

PaulYuuu (approved):
Even igb can be used on other platforms if build from source, but I keep an open mind here. Ack.

YongxueHong:
Hi @yanglei-rh 
Please update the part of the commit message about the internal test requirement info.
We are contributing to the upstream and public projects, and I am sure you are doing that from both the upstream and downstream sides at the same time.
So I suggest that we could update the commit message from the upstream perspective.
Thanks.

yanglei-rh:
> Hi @yanglei-rh Please update the part of the commit message about the internal test requirement info. We are contributing to the upstream and public projects, and I am sure you are doing that from both the upstream and downstream sides at the same time. So I suggest that we could update the commit message from the upstream perspective. Thanks.

@YongxueHong Done, please help review it again, thanks a lot.

PR #3802, opened 2023-11-23
add support for address in memory target

nanli1:
Need address attr in Memory.target class

For https://github.com/autotest/tp-libvirt/pull/5293

### DIFF
--- a/virttest/libvirt_xml/devices/memory.py
+++ b/virttest/libvirt_xml/devices/memory.py
@@ -65,7 +65,8 @@ class Memory(base.UntypedDeviceBase):
         __slots__ = ('size', 'size_unit', 'node', 'label',
                      'requested_size', 'requested_unit',
                      'current_size', 'current_unit',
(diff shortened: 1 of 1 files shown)


virttest/libvirt_xml/devices/memory.py
```diff
@@ -113,6 +114,11 @@ def __init__(self, virsh_instance=base.base.virsh):
                                          'virsh_instance': virsh_instance})
             accessors.XMLElementBool('readonly', self, parent_xpath='/',
                                      tag_name='readonly')
+            accessors.XMLElementNest('address', self, parent_xpath='/',
+                                     tag_name='address', subclass=Memory().Address,
```
chloerh, line 118:
```suggestion
                                     tag_name='address', subclass=Memory.Address,
```
Would this way also work? Therefore we might not need to instantiate `Memory` class.

nanli1: Yes ,  it works. Thanks haijiao : D

PR #3803, opened 2023-11-23
cpu: Add missed cpu model flags

nanliu-r:
ID: 1567

This CPU model also require the three cpu flags to boot up forcibly.


nanliu-r:
Test PASS when searching the latest CPU model on the host to test existed cpu flags.
@YongxueHong @PaulYuuu Hi, could you help review this? Thanks.

YongxueHong:
CC @fbq815 Please help review if this PR also works for the s390 arch.
Thanks.

PaulYuuu (approved):

fbq815:
seems we don't have these cpu models on s390x, could skip this time

YongxueHong:
> seems we don't have these cpu models on s390x, could skip this time

Got it, thanks.

PR #3808, opened 2023-12-04
Support to specify virtio-win's version range

leidwang:
Add a checkpoint here to check the virtio-win version, if the version is in the specified range, test will continue, if not, cancel the test.

ID:1594

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1255,6 +1255,27 @@ def preprocess(test, params, env):
         version_info["vm_bootloader_version"] = str(vm_bootloader_ver)
         LOG.debug("vm bootloader version: %s", vm_bootloader_ver)
 
+    # Checking required virtio-win version, if not satisfied, cancel test
+    if params.get("required_virtio_win") or params.get("required_virtio_win_prewhql"):
+        if params.get("cdrom_virtio"):
+            cdrom_virtio = params["cdrom_virtio"]
+            cdrom_virtio_path = os.path.basename(utils_misc.get_path(data_dir.get_data_dir(), cdrom_virtio))
+            virtio_win_range = params.get("required_virtio_win_prewhql") if re.search(
+                'prewhql', cdrom_virtio_path) else params.get("required_virtio_win")
+            if virtio_win_range:
+                LOG.info("Checking required virtio-win version: %s" % virtio_win_range)
+                match = re.search("virtio-win-(?:prewhql-)?(\d+\.\d+(?:\.\d+)?-\d+)", cdrom_virtio_path)
+                if match.group(1) is None:
+                    test.error("Can not get virtio-win version from \"cdrom_virtio\": %s" % cdrom_virtio)
+                cdrom_virtio_version = re.sub('-', '.', match.group(1))
+                if cdrom_virtio_version not in VersionInterval(virtio_win_range):
+                    test.cancel("Got virtio-win version:%s, which is not in %s" %
+                                (cdrom_virtio_version, virtio_win_range))
+            else:
+                test.error("The limitation for virtio-win is not suitable for the cdrom_virtio")
+        else:
+            LOG.warning("required_virtio_win(prewhql) can not take effect without cdrom_virtio")
+
     # Get the Libvirt version
     if vm_type == "libvirt":
         libvirt_ver_cmd = params.get("libvirt_ver_cmd", "libvirtd -V|awk -F' ' '{print $3}'")
--- a/virttest/shared/cfg/virtio-win.cfg
+++ b/virttest/shared/cfg/virtio-win.cfg
@@ -34,6 +34,10 @@
 #
 # So you have to map the paths of your cd containing the drivers on the config
 # variables. More details below.
+#
+# The parameter required_virtio_win is used to specify a range of virtio-win iso.
+# required_virtio_win_prewhql = [0.1.180, 0.1.280]
+# required_virtio_win = [1.9.22.3, 1.9.36.0)
 
 unattended_install.cdrom, whql.support_vm_install:
     Windows..virtio_blk, Windows..virtio_scsi, Windows..virtio_net:


virttest/env_process.py
```diff
@@ -1240,6 +1240,22 @@ def preprocess(test, params, env):
             test.cancel("Got host qemu version:%s, which is not in %s" %
                         (host_qemu, required_qemu))
 
+    # Checking required virtio driver version, if not satisfied, cancel test
+    if params.get("required_virtio") and params.get("cdrom_virtio"):
+        required_virtio = params.get("required_virtio")
```
vivianQizhu, line 1245: Then we have to give a default range here to make sure it won't affect general tests.

virttest/env_process.py
```diff
@@ -1240,6 +1240,22 @@ def preprocess(test, params, env):
             test.cancel("Got host qemu version:%s, which is not in %s" %
                         (host_qemu, required_qemu))
 
+    # Checking required virtio driver version, if not satisfied, cancel test
+    if params.get("required_virtio") and params.get("cdrom_virtio"):
```
vivianQizhu, line 1244: Oh here is the limitation, please ignore below comment then. But `required_virtio` is confusing, has to be `required_virtio_win`, as this refers to package name, not same to `cdrom`.

vivianQizhu:
@xiagao Would you please help review? Thanks.

virttest/env_process.py
```diff
@@ -1240,6 +1240,22 @@ def preprocess(test, params, env):
             test.cancel("Got host qemu version:%s, which is not in %s" %
                         (host_qemu, required_qemu))
 
+    # Checking required virtio driver version, if not satisfied, cancel test
+    if params.get("required_virtio_win") and params.get("cdrom_virtio"):
```
vivianQizhu, line 1244: What if `required_virtio_win` defined but `cdrom_virtio` is missing, should we raise up error?

leidwang: Yes, it is a nice reminder for user.Will update it right now.
We can raise up an error like "`required_virtio_win` can not take effect without  `cdrom_virtio`".

leidwang:
Hi @luckyh @yanan-fu 

As we discussed in the meeting, feel free to add your comments here.Thanks.

virttest/env_process.py
```diff
@@ -1240,6 +1240,25 @@ def preprocess(test, params, env):
             test.cancel("Got host qemu version:%s, which is not in %s" %
                         (host_qemu, required_qemu))
 
+    # Checking required virtio-win version, if not satisfied, cancel test
+    if params.get("required_virtio_win"):
+        if params.get("cdrom_virtio"):
+            required_virtio_win = params.get("required_virtio_win")
```
vivianQizhu, line 1246: We don't provide a default `required_virtio_win` so here could be `params['']`, as `params.get()` has more trasactions which is not necessary here.

virttest/env_process.py
```diff
@@ -1240,6 +1240,25 @@ def preprocess(test, params, env):
...
+            LOG.info("Test requires virtio-win version: %s" % required_virtio_win)
```
vivianQizhu, line 1247:
```suggestion
            LOG.info("Checking required virtio-win version: %s" % required_virtio_win)
```

virttest/env_process.py
```diff
@@ -1240,6 +1240,25 @@ def preprocess(test, params, env):
...
+            cdrom_virtio = params["cdrom_virtio"]
+            if re.search('prewhql', cdrom_virtio):
+                match = re.search("\d+.\d+-\d+", cdrom_virtio)
+            else:
+                match = re.search("\d+.\d+\d+-\d+", cdrom_virtio)
+            if match is None:
+                test.cancel("Can not get virtio-win version")
```
vivianQizhu, line 1254: 1. I would suggest here use `error` instead of `cancel` as here must be something wrong, not only a simple precondition mismatch (where cancel suite for).
2. Print the cdrom_virtio detail here would be more friendly for debugging.
```suggestion
                test.error("Can not get virtio-win version from \"cdrom_virtio\": %s" % cdrom_virtio)
```

virttest/env_process.py
```diff
@@ -1240,6 +1240,25 @@ def preprocess(test, params, env):
...
+            cdrom_virtio_version = re.sub('-', '.', match.group(0))
```
vivianQizhu, line 1255: Could you add comment with the example of the format to define `required_virtio_win`, maybe also including other related information like how virtio_win version strings usually look like, maybe within "base.cfg" or somewhere?

leidwang: How about add a annotation here like that:
`# Example value of required_virtio_win: [0.1.180, 0.1.280] or [1.9.8.6,1.9.35.0)`

vivianQizhu: @leidwang I suppose a better location would be at .cfg configuration file: `# required_virtio_win = [0.1.180, 0.1.280]`, which aligns to most of param definition examples given in this project.

virttest/shared/cfg/base.cfg
```diff
@@ -1087,3 +1087,7 @@ uuid_dimm = ""
 
 # Set image_extent_size_hint, displayed in qemu-img info extend_size_hint field.
 #image_extent_size_hint = 1G #To set extent_size_hint to 1G
+
+
+# The parameter required_virtio_win is used to specify a range of virtio-win iso.
```
vivianQizhu, line 1092:
```suggestion
# Set required_virtio_win, specify a range of virtio-win iso version (optional)
```

vivianQizhu: @leidwang Sorry, after a second consideration, maybe better move to Windows.cfg, or where cdrom_virtio primarily defined.

leidwang: Okay, yes, it would be better than base.cfg.Thanks!

virttest/utils_misc.py
```diff
@@ -4429,3 +4429,21 @@ def _md5(fd):
...
+    """
+    LOG.info("Get %s real file path" % file_path)
+    if os.path.islink(file_path):
+        real_path = os.path.realpath(file_path)
+        if os.path.islink(real_path):
+            get_real_file_path(real_path)
+        else:
+            return real_path
+    else:
+        return file_path
```
vivianQizhu, line 4449:
```suggestion
def get_real_path(org_path):
    """
    Check if the file/directory is a real path, get the real
    path if it is not

    param org_path: the orginial path
    """

    LOG.debug("Get real path of %s" % org_path)
    if os.path.islink(org_path):
        real_path = os.path.realpath(org_path)
        return get_real_path(real_path)
    else:
        return org_path
```

vivianQizhu: @leidwang Ah, `os.path.realpath` is recursive already, I suppose this function was not needed at all then. Would you please elaborate on the actual scenario which I probably omitted?

leidwang: Yes, you are right.Thanks for the guidance.

https://docs.python.org/3/library/os.path.html#os.path.realpath

virttest/shared/cfg/base.cfg
leidwang: Hi @vivianQizhu 
I checked the contents of windows.cfg and it seems to be used to save some configurations of the Windows VM. But base.cfg will save some examples of different parameters. Based on this, maybe base.cfg is a better place.What do you think?
FYI:https://github.com/avocado-framework/avocado-vt/blob/master/virttest/shared/cfg/guest-os/Windows.cfg

vivianQizhu: @leidwang Or maybe here: https://github.com/avocado-framework/avocado-vt/blob/master/virttest/shared/cfg/virtio-win.cfg#L52

virttest/env_process.py
```diff
@@ -1244,16 +1244,26 @@ def preprocess(test, params, env):
...
-        version_info["vm_bootloader_version"] = str(vm_bootloader_ver)
-        LOG.debug("vm bootloader version: %s", vm_bootloader_ver)
+    # Checking required virtio-win version, if not satisfied, cancel test
+    if params.get("required_virtio_win"):
+        if params.get("cdrom_virtio"):
+            required_virtio_win = params["required_virtio_win"]
+            LOG.info("Checking required virtio-win version: %s" % required_virtio_win)
+            cdrom_virtio = params["cdrom_virtio"]
+            cdrom_virtio_path = os.path.join(data_dir.get_data_dir(), cdrom_virtio)
+            cdrom_virtio_path = utils_misc.get_real_file_path(cdrom_virtio_path)
```
vivianQizhu, line 1254: Then this should work directly.
```suggestion
            cdrom_virtio_path = os.path.realpath(cdrom_virtio_path)
```

leidwang: Yes,but we need check if it is a link file.If so, then get the realpath, so the `get_real_file_path` still has meaning.

leidwang: Oh yes, we can use `os.path.realpath `directly. If so, we no longer need `get_real_file_path`.

vivianQizhu:
@yanan-fu Would you please help review this one? Many thanks.

yanan-fu:
Sorry for missing needinfo.
Firstly, @leidwang   could you please make the commit clean enough ? Thanks

leidwang:
> Sorry for missing needinfo. Firstly, @leidwang could you please make the commit clean enough ? Thanks

Okay, removed the redundant commit, thanks  @yanan-fu @vivianQizhu

virttest/env_process.py
```diff
@@ -1255,6 +1255,27 @@ def preprocess(test, params, env):
...
         LOG.debug("vm bootloader version: %s", vm_bootloader_ver)
 
+    # Checking required virtio-win version, if not satisfied, cancel test
+    if params.get("required_virtio_win"):
+        if params.get("cdrom_virtio"):
+            required_virtio_win = params["required_virtio_win"]
+            LOG.info("Checking required virtio-win version: %s" % required_virtio_win)
+            cdrom_virtio = params["cdrom_virtio"]
+            cdrom_virtio_path = os.path.join(data_dir.get_data_dir(), cdrom_virtio)
+            cdrom_virtio_path = os.path.realpath(cdrom_virtio_path)
```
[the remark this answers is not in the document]
leidwang: Hi @yanan-fu 

We have some softlink for the virtio-win iso, so we need to get the real path first.That's why we need `os.path.realpath` here.

I will update the code to the following, thanks for your suggestions.

cdrom_virtio_path = os.path.realpath(utils_misc.get_path(data_dir.get_data_dir(), cdrom_virtio))

virttest/env_process.py
```diff
@@ -1255,6 +1255,27 @@ def preprocess(test, params, env):
...
+            if re.search('prewhql', cdrom_virtio_path):
+                match = re.search("\d+.\d+-\d+", cdrom_virtio_path)
+            else:
+                match = re.search("\d+.\d+\d+-\d+", cdrom_virtio_path)
```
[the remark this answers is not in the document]
leidwang: Okay, thanks a lot!

virttest/env_process.py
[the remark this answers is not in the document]
yanan-fu: utils_misc.get_path will do that,  you can check the details of the function, thanks!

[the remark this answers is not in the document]
leidwang: Oh, got it. thank you for your patience.

virttest/env_process.py
```diff
@@ -1255,6 +1255,25 @@ def preprocess(test, params, env):
         version_info["vm_bootloader_version"] = str(vm_bootloader_ver)
         LOG.debug("vm bootloader version: %s", vm_bootloader_ver)
 
+    # Checking required virtio-win version, if not satisfied, cancel test
+    if params.get("required_virtio_win") or params.get("required_virtio_win_prewhql"):
+        if params.get("cdrom_virtio"):
+            cdrom_virtio = params["cdrom_virtio"]
+            virtio_win_range = params.get("required_virtio_win_prewhql") if re.search(
+                'prewhql', cdrom_virtio) else params.get("required_virtio_win")
```
yanan-fu, line 1263: Suggest checking this after get actual iso name.

leidwang: Okay, it will be better.Thanks!

virttest/env_process.py
```diff
@@ -1255,6 +1255,25 @@ def preprocess(test, params, env):
...
+            if virtio_win_range:
+                LOG.info("Checking required virtio-win version: %s" % virtio_win_range)
+                cdrom_virtio_path = os.path.basename(utils_misc.get_path(data_dir.get_data_dir(), cdrom_virtio))
+                match = re.search("virtio-win-(?:prewhql-)?(\d+\.\d+(?:\.\d+)?-\d+)", cdrom_virtio_path)
+                if match.group(1) is None:
+                    test.error("Can not get virtio-win version from \"cdrom_virtio\": %s" % cdrom_virtio)
+                cdrom_virtio_version = re.sub('-', '.', match.group(1))
+                if cdrom_virtio_version not in VersionInterval(virtio_win_range):
+                    test.cancel("Got virtio-win version:%s, which is not in %s" %
+                                (cdrom_virtio_version, virtio_win_range))
```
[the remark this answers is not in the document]
leidwang: Hi @yanan-fu 

1.The current code logic is that if there is no available virtio_win_range for the current cdrom_virtio, this checkpoint will be skipped because the limitations of the virtio_win version are not suitable for the test's cdrom_virtio. 
 I would prefer to report a error if the configuration is wrong, because it is an error in the cfg, not a failure.What do you think?

2.If the cdrom_virtio not in the range, test will cancel.

Thanks a lot.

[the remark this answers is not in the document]
yanan-fu: > Hi @yanan-fu
> 
> 1.The current code logic is that if there is no available virtio_win_range for the current cdrom_virtio, this checkpoint will be skipped because the limitations of the virtio_win version are not suitable for the test's cdrom_virtio. I would prefer to report a error if the configuration is wrong, because it is an error in the cfg, not a failure.What do you think?

Yes, the new added L1274~L1275 is what i expect,  raise error when cdrom type and version type does not match,  instead of continue test.

test.error is suitable.        


> 
> 2.If the cdrom_virtio not in the range, test will cancel.
> 
> Thanks a lot.

yanan-fu (approved):
Ack

vivianQizhu (approved):

PR #3811, opened 2023-12-07
utils_libvirt/libvirt_disk: add function to get disk by serial

smitterl:
Use the SERIAL to identify a disk and return its path.

Suggested-by: Chunfu Wen <chwen@redhat.com>

Used in https://github.com/autotest/tp-libvirt/pull/5323

### DIFF
--- a/virttest/utils_libvirt/libvirt_disk.py
+++ b/virttest/utils_libvirt/libvirt_disk.py
@@ -24,6 +24,31 @@ from virttest.libvirt_xml.devices.disk import Disk
 LOG = logging.getLogger('avocado.' + __name__)
 
 
+def get_disk_by_serial(session, serial):
+    """
+    Gets the disk by its serial value
+
+    :param session: VM session
+    :param serial: The expected serial value
+    :return path: absolute path of disk
+                  e.g. /dev/sda
+    :raises TestError: If the disk cannot be found.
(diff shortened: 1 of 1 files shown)


virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -24,6 +24,37 @@
 LOG = logging.getLogger('avocado.' + __name__)
 
 
+def get_new_disk(session):
+    """
+    Identifies the single disk below /dev/
+    that's not mounted currently.
+
+    :param session: VM session
+    :param test: Test instance
```
chunfuwen, line 33: here there is no test parameter, others LGTM

smitterl: Thanks for spotting it, fixed.

smitterl:
@chunfuwen There was a problem with my implementation. At some point the 'new_disk' might also have been mounted during the tests. So, here I changed the patch to identifying with the serial as proposed by you https://github.com/autotest/tp-libvirt/pull/5323#discussion_r1416578287

Please can you review again?

smitterl:
Using the serial doesn't work for LUNS:
"error: unsupported configuration: scsi-block 'lun' devices do not support the serial property"

PR #3814, opened 2023-12-12
utils_libvirt/libvirt_disk: add function to get disk by serial

smitterl:
Use the SERIAL to identify a disk and return its path.

Suggested-by: Chunfu Wen <chwen@redhat.com>

### DIFF
--- a/virttest/utils_libvirt/libvirt_disk.py
+++ b/virttest/utils_libvirt/libvirt_disk.py
@@ -102,6 +102,31 @@ def get_non_root_disk_name(session):
         return name.strip(), mpoint.strip()
 
 
+def get_disk_by_serial(session, serial):
+    """
+    Gets the disk by its serial value
+
+    :param session: VM session
+    :param serial: The expected serial value
+    :return path: absolute path of disk
+                  e.g. /dev/sda
+    :raises TestError: If the disk cannot be found.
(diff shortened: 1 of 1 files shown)


smitterl:
@chunfuwen Please help review. As mentioned on https://github.com/avocado-framework/avocado-vt/pull/3811#issuecomment-1850699674 unfortunately we can't set serial on LUNs but its implementation might be useful on other occasions.

chunfuwen (approved):
lgtm

chunfuwen:
@smitterl ,could you upload one simple test result?

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -102,6 +102,31 @@ def get_non_root_disk_name(session):
         return name.strip(), mpoint.strip()
 
 
+def get_disk_by_serial(session, serial):
```
dzhengfy, line 105: We already have same function in utils_disk.py.

`def get_disk_by_serial(serial_str, session=None):`

smitterl: Thanks for the pointer. I'll close this one.

smitterl:
Closing this as there is already another implementation with the same API.

PR #3815, opened 2023-12-12
utils_libvirt/libvirt_disk: add function to get disk not on /

smitterl:
Use "/" mount information to identify a disk and return its name.

### DIFF
--- a/virttest/utils_libvirt/libvirt_disk.py
+++ b/virttest/utils_libvirt/libvirt_disk.py
@@ -24,6 +24,84 @@ from virttest.libvirt_xml.devices.disk import Disk
 LOG = logging.getLogger('avocado.' + __name__)
 
 
+def get_non_root_disk_name(session):
+    """
+    Returns the first disk name under /dev whose device doesn't have any
+    partitions or volumes or itself mounting "/".
+    Assuming that the system is installed on a single disk, this
+    can be used to identify an added disk.
+    It also includes read-only devices as such, e.g. sr0.
+
+    It iterates over output with the following structure in mind:
+    vda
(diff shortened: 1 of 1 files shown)


chunfuwen (approved):
lgtm

virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -24,6 +24,83 @@
...
+            continue
+        name, mpoint = entry.groups()
+        is_disk = disk_pattern.match(name)
+        if root_mounted:
+            if is_disk:
+                root_disk = line
+                break
+            else:
+                idx = idx - 1
+                continue
```
nanli1, line 81: Just confirm, We will never run into this `if root_mounted` branch if we set  ` root_mounted = False` in L63 and not pass it in `def get_non_root_disk_name(session)` right?

smitterl: @nanli1 That is correct this branch will not be entered if root_mounted is never set to True in L. 83. This means there is no mount point for "/" (L. 82), therefore the idx will just go up until len(lines) and the TestError is raised  (L. 94).

smitterl:
In the last push I modified the following lines to remove whitespace from the return values
```
        name, mpoint = entry_pattern.match(non_root_disks[0]).groups()
        return name.strip(), mpoint.strip()
```

This was necessary because the libvirt_disk.linux_disk_check was failing because the name was returned as 'vdb             ' with additional whitespace. This is fixed now and doesn't invalidate previous test results as they assumed this was already the case.

nanli1 (approved):

PR #3818, opened 2023-12-22
utils_net: get default gateway for specific interface

smitterl:
Add optional parameter in order to retrieve the default gateway for a specific interface.

On a system with two interfaces, the routes might look like

```
default via 10.172.161.254 dev enc1 proto dhcp src 10.172.160.70 metric 100
default via 192.168.126.1 dev enc2 proto dhcp src 192.168.126.48 metric 102
```

Filtering only by default will return two ips

```
10.172.161.254
192.168.126.1
```

The change allows us to select the interface and thus retrieve only

```
192.168.126.1
```

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -3893,7 +3893,7 @@ def get_host_iface():
 
 
 def get_default_gateway(iface_name=False, session=None, ip_ver='ipv4',
-                        force_dhcp=False):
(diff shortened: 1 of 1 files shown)


smitterl:
@chloerh please help review

yalzhang (approved):

chloerh (approved):

PR #3821, opened 2024-01-04
Reduce logging verbosity of ovs initialization

luckyh:
When ovs management tool was not installed on the system, an error message would be emitted in some certain contexts due to there always be a corresponding check been executed no matter whether the tool was exactly required upon the users scenarios.

We had better to refactor this, but at the moment let us firstly workaround the issue by reducing the logging verbosity as such an error message did not make too much sense for the users.


pevogam (approved):
LGTM and thanks for this

YongxueHong (approved):

PR #3823, opened 2024-01-08
Replacing readfp function as read_file function!

Anushree-Mathur:
In python3.12 readfp function has been removed and it should be replaced as read_file function. readfp was deprecated since python3.2 but got removed in python3.12.
**For reference :-**
1) https://docs.python.org/3.12/whatsnew/3.12.html 
2) https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.readfp
Signed-off-by: Anushree Mathur <anushree.mathur@linux.vnet.ibm.com>

### DIFF
--- a/virttest/utils_config.py
+++ b/virttest/utils_config.py
@@ -1,6 +1,7 @@
 import ast
 import logging
 import os.path
+import sys
 from six import StringIO
 
 try:
@@ -101,7 +102,12 @@ class SectionlessConfig(object):
         self.parser.optionxform = str
         self.backup_content = open(path, 'r').read()
         read_fp = StringIO('[root]\n' + self.backup_content)
-        self.parser.readfp(read_fp)
+        #readfp function has been deprecated since python3.2 
+        #and removed in python3.12
+        if sys.version_info >= (3, 12):
+            self.parser.read_file(read_fp)
+        else:
+            self.parser.readfp(read_fp)
 
     def __sync_file(self):
         out_file = open(self.path, 'w')


bskjois (approved):
LGTM

sacsant (commented):
LGTM

virttest/utils_config.py
```diff
@@ -101,7 +102,12 @@ def __init__(self, path):
         self.parser.optionxform = str
         self.backup_content = open(path, 'r').read()
         read_fp = StringIO('[root]\n' + self.backup_content)
-        self.parser.readfp(read_fp)
+        #readfp function has been deprecated since python3.2 
+        #and removed in python3.12
+        if sys.version_info >= (3, 12):
+            self.parser.read_file(read_fp)
+        else:
+            self.parser.readfp(read_fp)
```
luckyh, line 110: Hi @Anushree-Mathur thank you for fixing the issue! Just a minor comment regarding the code change

```suggestion
        self.parser.read_file(read_fp)
```

as `read_file` was introduced since python 3.2 and at the moment avocado-vt only support the version > 3.7

Anushree-Mathur: @luckyh Thank you for the review! It was only deprecated since python3.2! I tried the same code for python versions such as python 3.7 3.8 till 3.12. But this problem only came for Python3.12 version as this readfp has been removed since python3.12. That is why I mentioned this condition in the code itself. Hope i am able to clarify my point!
Could you please let me know if you agree with it!

luckyh: I see. However, the reason I suggested as so is that the else clause will become a dead code after dropping the support of the python versions prior to 3.12 in the future, which would request another PR for the removal, while the suggested approach works for both the cases.

What do you think?

Anushree-Mathur:
@dzhengfy @Yingshun Could you please review and merge this PR!

PraveenPenguin:
@Anushree-Mathur can you wrap both the commits in single commit

PR #3831, opened 2024-01-18
Add discard_granularity attribute in blockio section

chunfuwen:
From function point, it is already supported, just update related parameter description


dzhengfy (approved):
LGTM

PR #3833, opened 2024-01-26
Drop Python 3.7 support

luckyh:
Since Python 3.7 has been EOL'd and recently the support has also been dropped by the latest Avocado.

Reference: https://github.com/avocado-framework/avocado/pull/5855


luckyh:
ok, I didn't notice that there is already an existing PR fixing the issue, so let's close this one in favor of https://github.com/avocado-framework/avocado-vt/pull/3832, thanks.

PR #3864, opened 2024-03-08
Add storage and image management support

zhencliu:


virttest/vt_resmgr/resources/resource.py
```diff
@@ -0,0 +1,135 @@
...
+            "bind": self.bind,
+            "unbind": self.unbind,
+            "allocate": self.allocate,
+            "release": self.release,
+            "sync": self.sync,
+        }
+
+    @classmethod
+    def resource_type(cls):
+        raise cls._RESOURCE_TYPE
```
nickzhq, line 31: raise => return?

PR #2929, opened 2021-02-24
qdevices: Remove "--daemon" option on QSwtpmDev

PaulYuuu:
`aexpect.run_bg` will also run the shell command in the background, so
if swtpm running in the daemon mode, then "run_bg" will return an idle
process and the swtpm process is not controlled.

ID: 1932183
Signed-off-by: Yihuang Yu <yihyu@redhat.com>

PR #3866, opened 2024-03-11
Introduce nettype passt to qemu vm

luckyh:
passt is designed as a replacement for slirp (--netdev user), a new
approach implemented usermode networking which improves on security
and performance.
    
This patch enables the functionality to allow qemu VMs to use passt
as the network backends for nic devices.

### DIFF
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -2539,7 +2539,18 @@ class QNetdev(QCustomDevice):
                     value = True
                 elif value in ("off", "no", "false"):
                     value = False
-            new_args[key] = value
+
+            subs = key.split(".")
+            curr = new_args
+            for subk in subs[:-1]:
+                try:
+                    int(subk)
+                    subv = list()
+                except ValueError:
+                    subv = dict()
+                curr.setdefault(subk, subv)
+                curr = curr[subk]
+            curr[subs[-1]] = value
 
         return new_args
 
--- a/virttest/qemu_devices/qdevices.py
+++ b/virttest/qemu_devices/qdevices.py
@@ -10,13 +10,17 @@ import json
 import logging
 import os
 import re
+import shutil
+import signal
 import time
 import traceback
 from collections import OrderedDict
 from pathlib import Path
+from uuid import uuid4
 
 import aexpect
 import six
+from avocado.utils import process
 from six.moves import xrange
 
 from virttest import qemu_monitor, utils_logfile, utils_misc, utils_numeric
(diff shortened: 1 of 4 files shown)


yanglei-rh:
LGTM and it can test pass on my side.

yanglei-rh (approved):

virttest/qemu_devices/qdevices.py
```diff
@@ -2352,6 +2356,127 @@ def __eq__(self, other):
...
+                cmd += f" --tcp-ports {spec}"
+            elif proto.lower() in ("udp", "u"):
+                cmd += f" --udp-ports {spec}"
+            else:
+                raise ValueError(f"unsupported protocol {proto}")
+        return cmd
+
+    def start_daemon(self):
+        """Start the passt daemon."""
+        cmd = "%s --one-off --socket %s" % (
```
PaulYuuu, line 2445: I would recommend adding `--foreground` by default, do not let `QDaemonDev` manage a background process, I think it's outside the scope of its design. And with this, `process.pid_exists(pid)` may not a must, can just use the original one.

But let's listen to other peoples' voices.

virttest/qemu_devices/qdevices.py
```diff
@@ -2416,7 +2541,20 @@ def _convert_netdev_args(args):
                     value = True
                 elif value in ("off", "no", "false"):
                     value = False
-            new_args[key] = value
+                elif key == "reconnect":
+                    value = int(value)
```
PaulYuuu, line 2545: Can this be covered in `elif isinstance(value, str) and value.isdigit():`?

yanglei-rh:
Hello @luckyh This patch can test pass from my side, and looks like there are also no other issues, can it be merge?

luckyh:
@yanglei-rh according to the PR review workflow of VT we just need another approval for this one before getting ready for merge, so let's wait for @PaulYuuu 's feedback, thanks!

virttest/qemu_devices/qdevices.py
PaulYuuu: Put https://github.com/avocado-framework/avocado-vt/pull/2929 as a reference. In qdevices.py, we already use aexpect.run_bg, any process will run in backaground, not need to background again.

@yanglei-rh, feel free leave your comment.

PaulYuuu: And a unit test like this:
```python
>>> passt1 = client.run_bg("passt")
>>> passt2 = client.run_bg("passt -f")
>>> passt1.is_alive()
False
>>> passt2.is_alive()
True
```

yanglei-rh: Hi @PaulYuuu Sorry for the late reply, I agree with your point.

virttest/qemu_devices/qdevices.py
```diff
@@ -2352,6 +2356,127 @@ def __eq__(self, other):
...
+        :param binary: The binary of virtiofs daemon.
+        :type binary: str
+        :param sock_path: The socket path of of virtiofs daemon.
+        :type sock_path: str
+        :param source: The source of virtiofs daemon.
+        :type source: str
+        :param extra_options: The external options of virtiofs daemon.
+        :type extra_options: str
+        :param enable_debug_mode: Enable debug mode of virtiofs daemon.
+        :type enable_debug_mode: bool
```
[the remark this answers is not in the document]
luckyh: Good catch! have updated, thanks!

virttest/qemu_devices/qdevices.py
yanglei-rh: Hi @PaulYuuu Thanks for pointing it out
Hi @luckyh Maybe we can directly to use `aexpect.run_bg` instead of call `QDaemonDev`?

luckyh: As daemonized processes are always configured with extra settings [1], I'd suggest keeping the daemon mode in favor of problem detection, in case the user made SELinux enforced.

What do you think @PaulYuuu ?

[1] https://passt.top/passt/tree/util.c#n449

PaulYuuu (approved):

YongxueHong (approved):
LGTM.

PR #3868, opened 2024-03-12
qemu_storage: add compression_type for create and convert

hellohellenmao:
Signed-off-by: Tingting Mao <timao@redhat.com>

ID: 1953

### DIFF
--- a/virttest/qemu_storage.py
+++ b/virttest/qemu_storage.py
@@ -594,6 +594,7 @@ class QemuImg(storage.QemuImg):
                 "extent_size_hint",
                 ("qcow2", "raw", "luks"),
             ),
+            "image_compression_type": (None, "compression_type", ("qcow2",)),
         }
         image_format = params.get("image_format", "qcow2")
         options = []
@@ -840,6 +841,9 @@ class QemuImg(storage.QemuImg):
                preallocated(optional)
                    if preallocation when create image, allowed values: off,
                    metadata. Default is "off"
+               image_compression_type(optional)
+                   if image_compression_type when create image, allowed
+                   values: zlib, zstd
 
         :return: tuple (path to the image created, process.CmdResult object
                  containing the result of the creation command).
@@ -976,6 +980,8 @@ class QemuImg(storage.QemuImg):
                 the convert target image tag
             compressed
                 indicates that target image must be compressed
+            image_compression_type
+                indicates the type of the compression is zlib or zstd
             sparse_size
                 indicate the consecutive number of bytes contains zeros to
                 create sparse image during conversion
--- a/virttest/shared/cfg/base.cfg
+++ b/virttest/shared/cfg/base.cfg
@@ -1088,6 +1088,10 @@ uuid_dimm = ""
 # Set image_extent_size_hint, displayed in qemu-img info extend_size_hint field.
 #image_extent_size_hint = 1G #To set extent_size_hint to 1G
 
+# Set the compression_type when creating or converting to a qcow2 image
+# Supported types: zlib, zstd
+#image_compression_type = zstd
+
 #
 #VM bootloader params
 #


virttest/qemu_storage.py
```diff
@@ -840,6 +841,9 @@ def create(self, params, ignore_errors=False):
                preallocated(optional)
                    if preallocation when create image, allowed values: off,
                    metadata. Default is "off"
+                compression_type(optional)
+                    if compression_type when create image, allowed values: zlib,
+                   zstd. Default is "zlib"
```
YongxueHong, line 846:
```suggestion
                   zstd. 
```
The default `"zlib"` comes from the qemu layer, not the framework. To avoid misunderstanding, I suggest we should not expose any default product behaviors in the framework.
BTW, the previous similar docstring is not appropriate, like `preallocated`
Thanks.

virttest/qemu_storage.py
```diff
@@ -594,6 +594,7 @@ def _parse_options(self, params):
                 "extent_size_hint",
                 ("qcow2", "raw", "luks"),
             ),
+            "compression_type": (None, "compression_type", ("qcow2",)),
```
YongxueHong, line 597:
```suggestion
            "image_compression_type": (None, "compression_type", ("qcow2",)),
```
Suggest adding the prefix `image_` to dedicate this parameter to the image since I am afraid that other components also have the same attribute.

CC @zhencliu

zhencliu:
> > hi @hellohellenmao , could you possibly paste the updated qemu-img command lines(with compress enabled) here?
> 
> Sure @zhencliu #qemu-img convert -c -f qcow2 -O qcow2 -o compression_type=zstd /home/kvm_autotest_root/images/rhel940-64-virtio-scsi-ovmf.qcow2 /home/kvm_autotest_root/images/image1_to_qcow2.qcow2 or #qemu-img create -f qcow2 -o compression_type=zstd test.qcow2 5G

Thanks, Tingting. I am curious about the qemu-img create command, we also create a compressed image(zlib) even if compression_type is not set, right? If that is the case, for case 3799, we always create a compressed image? Can we disable compression?

hellohellenmao:
> > > hi @hellohellenmao , could you possibly paste the updated qemu-img command lines(with compress enabled) here?
> > 
> > 
> > Sure @zhencliu #qemu-img convert -c -f qcow2 -O qcow2 -o compression_type=zstd /home/kvm_autotest_root/images/rhel940-64-virtio-scsi-ovmf.qcow2 /home/kvm_autotest_root/images/image1_to_qcow2.qcow2 or #qemu-img create -f qcow2 -o compression_type=zstd test.qcow2 5G
> 
> Thanks, Tingting. I am curious about the qemu-img create command, we also create a compressed image(zlib) even if compression_type is not set, right? If that is the case, for case 3799, we always create a compressed image? Can we disable compression?

Good question. Actually the answer here is no. The image created without `-o compression_type` is not compressed. And the dumped info of the image with "compression-type": "zlib", which just mean "that if any compressed clusters are on the image (maybe written in the future), they will use zlib compression".

zhencliu:
> > > > hi @hellohellenmao , could you possibly paste the updated qemu-img command lines(with compress enabled) here?
> > > 
> > > 
> > > Sure @zhencliu #qemu-img convert -c -f qcow2 -O qcow2 -o compression_type=zstd /home/kvm_autotest_root/images/rhel940-64-virtio-scsi-ovmf.qcow2 /home/kvm_autotest_root/images/image1_to_qcow2.qcow2 or #qemu-img create -f qcow2 -o compression_type=zstd test.qcow2 5G
> > 
> > 
> > Thanks, Tingting. I am curious about the qemu-img create command, we also create a compressed image(zlib) even if compression_type is not set, right? If that is the case, for case 3799, we always create a compressed image? Can we disable compression?
> 
> Good question. Actually the answer here is no. The image created without `-o compression_type` is not compressed. And the dumped info of the image with "compression-type": "zlib", which just mean "that if any compressed clusters are on the image (maybe written in the future), they will use zlib compression".

Thanks for your reply, I see your point, then we need to verify a compressed image in another way, like 'qemu-img map', because 'qemu-img info' cannot tell the truth

virttest/qemu_storage.py
```diff
@@ -594,6 +594,7 @@ def _parse_options(self, params):
                 "extent_size_hint",
                 ("qcow2", "raw", "luks"),
             ),
+            "image_compression_type": (None, "compression_type", ("qcow2",)),
```
zhencliu, line 597: Sorry, I should add this comment yesterday, I just forgot it.
What about adding image_compression_type  in shared/cfg/base.cfg, you can refer to image_extent_size_hint. You might add some comments about this param, like it's used for a qcow2 image only, and the supported types are zlib and zstd currently

virttest/shared/cfg/base.cfg
```diff
@@ -1088,6 +1088,9 @@ uuid_dimm = ""
 # Set image_extent_size_hint, displayed in qemu-img info extend_size_hint field.
 #image_extent_size_hint = 1G #To set extent_size_hint to 1G
 
+#image_compression_type, which is used for a qcow2 image(create or convert) only
```
zhencliu, line 1091: Would you mind formatting the comments, e.g.  (please polish the wording to make it more clear from your perspective)
# Set the compression type when creating or converting a qcow2 image
# Supported types: zlib, zstd
#image_compression_type = zstd

The last line just gives an example, without a space between '#' and 'image', so it's easy for someone to uncomment the line and make the param effective.

zhencliu (approved):

virttest/qemu_storage.py
```diff
@@ -840,6 +841,9 @@ def create(self, params, ignore_errors=False):
                preallocated(optional)
                    if preallocation when create image, allowed values: off,
                    metadata. Default is "off"
+                image_compression_type(optional)
```
YongxueHong, line 844:
```suggestion
               image_compression_type(optional)
```

PR #3872, opened 2024-03-15
utils_misc: Add a new func for get host_bridge on a host

xiaodwan:
Add a new function to get host bridges on a host

### DIFF
--- a/virttest/utils_sys.py
+++ b/virttest/utils_sys.py
@@ -9,6 +9,8 @@ import re
 
 from avocado.utils import process
 
+from virttest.utils_misc import cmd_status_output
+
 LOG = logging.getLogger("avocado." + __name__)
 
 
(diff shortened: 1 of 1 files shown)


xiaodwan:
@dzhengfy @chunfuwen Please help to review this PR, thanks.

virttest/utils_misc.py
```diff
@@ -634,6 +634,23 @@ def get_full_pci_id(pci_id):
         return None
 
 
+def get_host_bridge_id(session=None):
```
dzhengfy, line 637: could you help move this function to utils_sys.py? We avoid adding new function into utils_misc.py.

dzhengfy (commented):
Codes LGTM

virttest/utils_misc.py
chloerh: +1

dzhengfy (approved):
LGTM

PR #3874, opened 2024-03-19
Fix failures to start libvirtd

cliping:
On RHEL 8, SELinux is preventing systemd from create access on the sock_file libvirt-sock. So we can generate a local policy module to allow this access.


cliping:
Error log:

Mar 18 12:23:56 ... setroubleshoot[271675]: SELinux is preventing systemd from create access on the sock_file libvirt-sock. For complete SELinux messages run: sealert -l 90ce38b1-f248-46ed-86fd-abec321b69c2
Mar 18 12:23:56 ... setroubleshoot[271675]: SELinux is preventing systemd from create access on the sock_file libvirt-sock.#012#012*****  Plugin catchall (100. confidence) suggests   **************************#012#012If you believe that systemd should be allowed create access on the libvirt-sock sock_file by default.#012Then you should report this as a bug.#012You can generate a local policy module to allow this access.#012Do#012allow this access for now by executing:#012# ausearch -c 'systemd' --raw | audit2allow -M my-systemd#012# semodule -X 300 -i my-systemd.pp#012
Mar 18 12:23:56 ... systemd-logind[1385]: Session 849 logged out. Waiting for processes to exit.

smitterl:
The libvirt-socket should have the right permissions. If not, there should be something wrong with the setup, IMO. If the test code alters the permissions for the test it doesn't seem to test in the actual security environment, as if we had setenforce off. But I might be missing something.

PR #3878, opened 2024-03-27
libvirt_disk: return instead of exception

dzhengfy:

### DIFF
--- a/virttest/utils_libvirt/libvirt_disk.py
+++ b/virttest/utils_libvirt/libvirt_disk.py
@@ -37,7 +37,7 @@ def get_non_root_disk_name(session):
     return get_non_root_disk_names(session)[0]
 
 
-def get_non_root_disk_names(session):
+def get_non_root_disk_names(session, ignore_status=False):
     """
     Returns the disk names under /dev whose device doesn't have any
     partitions or volumes or itself mounting "/".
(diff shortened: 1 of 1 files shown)


virttest/utils_libvirt/libvirt_disk.py
```diff
@@ -66,8 +67,6 @@ def get_non_root_disk_names(session):
     s, o = utils_misc.cmd_status_output(
         cmd, ignore_status=False, shell=True, verbose=True, session=session
     )
-    if s:
```
smitterl, line 69: I think you shouldn't remove these lines but do
```
if s and not ignore_status:
    raise ....
```
to maintain existing behavior.

smitterl (changes requested):
@dzhengfy Thanks for this. I think it needs one small change for backward compatibility, rest LGTM.

virttest/utils_libvirt/libvirt_disk.py
chloerh: +1

dzhengfy: ah... yes,  I just realize ignore_status does not take effect in  utils_misc.cmd_status_output() when session is not None.

Updated.

smitterl (approved):
lgtm

chloerh (approved):

PR #3884, opened 2024-03-29
add snapshot_status function

nanli1:
For https://github.com/autotest/tp-libvirt/pull/5501

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -3329,6 +3329,36 @@ def snapshot_list(name, options=None, **dargs):
     return ret
 
 
+def snapshot_status(name, options=None, **dargs):
+    """
+    Get list of snapshots status of domain.
+
+    :param name: name of domain
+    :param options: options of snapshot_list
+    :param dargs: standardized virsh function API keywords
+    :return: dict of snapshot names and snapshot status, e.g.:
+    get {'s1': 'running', 's2': 'running'} for below result
+    # virsh snapshot-list avocado-vt-vm1
+     Name   Creation Time               State
+    ---------------------------------------------
+     s1     2024-04-01 08:50:53 -0400   running
+     s2     2024-04-01 08:50:53 -0400   running
+    """
+    dargs["ignore_status"] = False
+    status_dict = {}
+
+    cmd = "snapshot-list %s" % name
+    if options is not None:
+        cmd += " %s" % options
+    sc_output = command(cmd, **dargs)
+    slist = re.findall("(\S*) *\d+.* \d*:\d*:\d* .\d* *(\S*)", sc_output.stdout_text)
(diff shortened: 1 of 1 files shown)


virttest/virsh.py
```diff
@@ -3329,6 +3329,33 @@ def snapshot_list(name, options=None, **dargs):
     return ret
 
 
+def snapshot_status(name, options=None, **dargs):
```
Yingshun, line 3332: I wonder if we should focus on virsh utilities in virsh.py and move other related functions to utils_libvirt. 

BTW, can you plz give an example what kind of 'status' the snapshots could be? or it should be 'snapshot_state'?

virttest/virsh.py
```diff
@@ -3329,6 +3329,33 @@ def snapshot_list(name, options=None, **dargs):
...
+    # CmdResult is handled here, force ignore_status
+    dargs["ignore_status"] = True
+    status_dict = {}
+
+    cmd = "snapshot-list %s" % name
+    if options is not None:
+        cmd += " %s" % options
+    sc_output = command(cmd, **dargs)
+    if sc_output.exit_status != 0:
+        raise process.CmdError(cmd, sc_output, "Failed to get list of snapshots")
```
Yingshun, line 3350: Can it be covered by setting 'dargs["ignore_status"] = False' ?

virttest/virsh.py
Yingshun, line 3332: Instead of defining a new function, can you call libvirt_misc.convert_to_dict() to get the information you want?

virttest/virsh.py
nanli1: Thanks yingshun ,You are right , updated

virttest/virsh.py
nanli1: hi yingshun ,Added the example

nanli1: Hi yingshun , Do you mean call the  libvirt_misc.convert_to_dict() after getting response from `def snapshot_list`, but it's weird , the  `def snapshot_list` only return snap name list such as ['s1', 's2', 's3'], this may due to all the snap list content are not oftenly used.

Yingshun: Ok, then you can reuse virsh.command to execute snapshot-list subcmd if you dont want to enhance the existing snapshot_list function I think.

Yingshun: Ok, then you can reuse virsh.command to execute snapshot-list subcmd if you dont want to enhance the existing snapshot_list function I think.

Yingshun: @nanli1 I don't mean just adding an example in this comment, it's also about the name and the location, can you plz recheck it?

PR #3889, opened 2024-04-10
Update for block device in vm xml

cliping:


virttest/utils_test/libvirt.py
```diff
@@ -2846,9 +2846,24 @@ def set_vm_disk(vm, params, tmp_dir=None, test=None):
...
         Otherwise, use current source file with update params.
         """
-        if disk_src_name:
-            blk_source = disk_src_name
-        disk_params_src = {"source_file": blk_source}
+        if disk_type == "block":
+            src_lv_name = params.get("src_lv_name")
+            src_vg_name = params.get("src_vg_name")
+            slice_offset = params.get("slice_offset")
+            slice_size = params.get("slice_size")
```
Yingshun, line 2853: I think it's better to have default values , or check if they have proper values before setting them. what do you think?

virttest/utils_test/libvirt.py
```diff
@@ -2846,9 +2846,24 @@ def set_vm_disk(vm, params, tmp_dir=None, test=None):
         If disk_src_name is given, replace current source file
         Otherwise, use current source file with update params.
         """
-        if disk_src_name:
-            blk_source = disk_src_name
-        disk_params_src = {"source_file": blk_source}
+        if disk_type == "block":
```
Yingshun, line 2849: Your code does not match the comment above -- 'If disk_src_name is given, replace current source file
￼        Otherwise, use current source file with update params.'

What about adding a new 'elif' for this type?

Yingshun (commented):
Should it be a new elif

nanli1 (commented):
any result

cliping:
Thanks Yingshun and nanli1, I moved this part of code to https://github.com/autotest/tp-libvirt/pull/5563. So close this PR.

PR #3897, opened 2024-04-19
qemu_vm: Fix the issue of iterating a dict

zhencliu:
ID: 2281


zhencliu:
hi @luckyh and @YongxueHong , I just found a bug when doing the tdx patch, it's a tiny mistake that I must miss it before and it cannot appear when the dict is empty, would you please help to have a look, thanks.

YongxueHong (approved):

luckyh (approved):

PR #3902, opened 2024-04-23
env_process: Fix the issue of iterating an integer

zhencliu:
A recent avocado commit(241c9cdf62f0162e24c37913ab37053971563ee4) enhanced cpu.get_family to return an integer for a AMD cpu, which can cause error, but get_family was declared to return a *name*, instead of an integer, which is confused. Added this commit as a workaround to bypass the error.

We expect a better solution for cpu.get_family later

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1036,7 +1036,7 @@ def preprocess(test, params, env):
     except Exception:
         LOG.warning("Could not get host cpu family")
     migration_setup = params.get("migration_setup", "no") == "yes"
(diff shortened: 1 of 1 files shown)


zhencliu:
Pulled the latest avocado, and ran a test with a AMD cpu (cpu family	: 25), it passed.

Hi @luckyh and @YongxueHong , would you please help to review the patch? Thanks.

YongxueHong (approved):
LGTM.

luckyh (approved):

PR #3905, opened 2024-04-25
net/if.py: refactor module

nickzhq:
Refactor the utils_net.py:
1. Split net/if module from utils_net module.

### DIFF
--- a/virttest/vt_utils/net/drivers/bridge.py
+++ b/virttest/vt_utils/net/drivers/bridge.py
@@ -1,0 +1,68 @@
+# Library for the bridge address related functions.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; specifically version 2 of the License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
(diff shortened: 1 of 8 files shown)


nickzhq:
@zhencliu @luckyh @YongxueHong  Please take a review, thanks!

nickzhq:
@YongxueHong Please take a review, thanks!

chloerh:
Hi, I found there are 2 `gen_ipv4_addr` functions in virttest and traced back to this pr. I was wondering that are we going to keep both of the functions? Thanks.

YongxueHong:
> Hi, I found there are 2 `gen_ipv4_addr` functions in virttest and traced back to this pr. I was wondering that are we going to keep both of the functions? Thanks.

Hi @chloerh 
Yeah, indeed, it looks like redandunt, but there are lots of calling places need to update, so we separate this task, firstly we will keep both at the same time for a while, and than plan to update the existing functionality with the new module within proper time in the future.  We prefer to use the this module's API for your case. 
Thanks.

chloerh:
@YongxueHong  Thanks for your clarification!
Currently, it seems these modules are not being used by tp-qemu or tp-libvirt. And the code in utils_net keeps changing since then.
So it would be really helpful to be notified when the switch happens. Thanks in advance.

YongxueHong:
> @YongxueHong Thanks for your clarification! Currently, it seems these modules are not being used by tp-qemu or tp-libvirt. And the code in utils_net keeps changing since then. So it would be really helpful to be notified when the switch happens. Thanks in advance.

Hi @chloerh 
Thanks for the kind reminder. We will double-check the `utils_net` with the latest code to keep the consistency when we replace it.

PR #3910, opened 2024-05-13
Replace readfp with read_file

Yingshun:
ConfigParser.readfp was removed in the latest python(3.12), so updating to use read_file().


Yingshun:
@brucepatricklee   Can you plz help confirm if it works? thx!

chloerh (approved):

PR #3913, opened 2024-05-21
cpu: adapt help text parsing to support modern qemus

luckyh:
Starting from qemu v9.0.0, the help text of cpu models associated with all the hardware architectures was unified to the pattern below.

    $ qemu -cpu '?'
    Available CPUs:
      <model>   <description>
      ...

As a result, we have to modify the corresponding parseing function to cater for the change, since the arch names were no longer showed in front of the model names.

This patch is going to introduce a new parser for handling the new pattern of cpu help text, meanwhile keep the old one for supporting legacy qemus.

ID: 2369

### DIFF
--- a/virttest/cpu.py
+++ b/virttest/cpu.py
@@ -1050,6 +1050,42 @@ def get_host_cpu_models():
     return cpu_support_model
 
 
+def parse_qemu_cpu_models_modern(help_text):
+    """
+    Get all cpu models from qemu -cpu help text.
+
+    Note: not completely supported for qemu version prior to v9.0.0
+          due to not all the hardware architectures printed cpu model
+          descriptions with using the pattern below.
+
+          Available CPUs:
+          ...
+
+    :param help_text: Text produced by <qemu> -cpu '?'.
+    :return: List of cpu models.
+    """
+    header_words = ("Available", "CPUs:")
+
+    model_list = []
+    search = False
+    for line in help_text.splitlines():
+        columns = re.split(r"\s+", line)
+        if tuple(columns) == header_words:
+            search = True
+            continue
+        if search:
+            if len(columns) == 1 and not columns[0]:
+                continue
+            elif columns[0] in ("", "x86", "PowerPC", "s390"):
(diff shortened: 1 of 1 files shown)


virttest/cpu.py
```diff
@@ -1050,6 +1050,42 @@ def _make_up_pattern(flags):
     return cpu_support_model
 
 
+def parse_qemu_cpu_models_modern(help_text):
```
zhencliu, line 1053: hi @luckyh , do we expect the users to use this function in future to get the cpu model list? This is a public function which can cover both the old and new version of qemu

virttest/cpu.py
```diff
@@ -1050,6 +1050,42 @@ def _make_up_pattern(flags):
...
+    """
+    Get all cpu models from qemu -cpu help text.
+
+    Note: not completely supported for qemu version prior to v9.0.0
```
aliang123, line 1057: Hi, @luckyh the behavior change commit is actually on top of qemu9.0-rc4, so I think QEMU versions that are rebased before rc4 won't be affected by it. This means that the legacy versions may include some qemu9.0 version. Not clear about the rebase rule, I think we'd better check it with yfu first.

zhencliu (approved):

virttest/cpu.py
luckyh: @zhencliu I believe the old one (`extract_qemu_cpu_models`) would be still the best choice for users in a long time, since it automatically invoked the new one within its implementation therefore no check will be required when users intend to call the functionality.

virttest/cpu.py
luckyh: @aliang123 good point. However, whether the function is going to be invoked did not depend on any check on the qemu versions, instead, the only prerequisite is that the given `help_text` must contain the `"Available CPUs:"` pattern so that the model names could be identified and extracted afterwards. This is just a note for users who would like to call the new function directly and to expect it working properly in their use cases.

luckyh:
Thanks all.

PR #3921, opened 2024-06-03
env_process: Refactor procedure that adds migrate_vms to vms

bgartzi:
Refactor the procedure that made the "vms" parameter aware of the migrate_vms into a `Setuper` subclass.
That step will be managed `env_process::_setup_manager` from now on.

ID: 2429

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1003,6 +1003,12 @@ def preprocess(test, params, env):
     if callable(preprocess_vm_off_hook):
         preprocess_vm_off_hook(test, params, env)  # pylint: disable=E1102
 
+    # Add migrate_vms to vms
+    migrate_vms = params.objects("migrate_vms")
+    if migrate_vms:
+        vms = list(set(params.objects("vms") + migrate_vms))
+        params["vms"] = " ".join(vms)
+
     # Check if code coverage for qemu is enabled and
     # if coverage reset is enabled too, reset coverage report
     gcov_qemu = params.get("gcov_qemu", "no") == "yes"
@@ -1101,12 +1107,6 @@ def preprocess(test, params, env):
     vm_type = params.get("vm_type")
 
     base_dir = data_dir.get_data_dir()
-
-    # Add migrate_vms to vms
-    migrate_vms = params.objects("migrate_vms")
-    if migrate_vms:
-        vms = list(set(params.objects("vms") + migrate_vms))
-        params["vms"] = " ".join(vms)
 
     # Permit iptables to permit 49152-49216 ports to libvirt for
     # migration and if arch is ppc with power8 then switch off smt


bgartzi:
Hi @YongxueHong! I'm sharing yet another patch refactoring the `env_process` `preprocess` function for review, could you have a look?
Thanks!

virttest/env_process.py
```diff
@@ -1096,18 +1097,13 @@ def preprocess(test, params, env):
...
 
     vm_type = params.get("vm_type")
 
     base_dir = data_dir.get_data_dir()
 
-    # Add migrate_vms to vms
-    migrate_vms = params.objects("migrate_vms")
-    if migrate_vms:
-        vms = list(set(params.objects("vms") + migrate_vms))
-        params["vms"] = " ".join(vms)
```
YongxueHong, line 1109: Let's skip this part since it should be part of the VM process: https://github.com/avocado-framework/avocado-vt/blob/b411ac28f5abda64de16cea97ff906e05013b8f0/virttest/env_process.py#L883-L886
Actually,  this part does not relate to the environment of the host setupper, just add the parameters `migrate_vms` to the global `vms`.
As for the previous setuppers, we can see that mainly set up the host environments, such as Networking, Bridge, and so on.
So I prefer to skip this part. What about your opinions? @luckyh 
Please let me know your thoughts @bgartzi Thanks.

bgartzi: Hi @YongxueHong!

Yeah, I get your point. I was considering these kind of pieces of code as part of the refactoring. The reason for that was that changing the order they are ran could break the order in which each of these steps are executed. `Setuper`'s `self.params` is just a reference, so updating it's keys and values affect directly to the dictionary being used inside the `preprocess` or `postprocess` function. 

Should we run this piece of code in the beginning of the `preprocess` function instead?
Should we treat these kind of steps (tweaking test parameter config) in a different manner?

Thanks for the feedback and let me know your thoughts. Let's also see what @luckyh thinks about it.

luckyh: > Should we run this piece of code in the beginning of the preprocess function instead?

@bgartzi @YongxueHong it looks that the piece of code is relocatable so I'd like to prefer this option.

> Should we treat these kind of steps (tweaking test parameter config) in a different manner?

That's a good point. From my perspective, we did require a mechanism for resolving params (e.g. merge, concatenate, etc.) before running into the `preprocess` function. However, this would also require some new abstracts to be introduced, therefore I'd suggest we could skip it for this time.

What do you think?

bgartzi: I agree. I'm relocating it to... right after the `preprocess_vm_off_hook` call. Setup steps between that and the `_setup_manager` initialization don't seem to rely on `params['vms']`, so it looks safe.

I'm pushing the patch now.

YongxueHong (approved):
LGTM.

PR #3925, opened 2024-06-07
Fix VM creation checking issue

qingwangrh:
Sometimes, the VM may not step in dead immediately, so the first check may not detect the VM is dead.
It should check the VM status over time.

ID:2384

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -3863,9 +3863,15 @@ class VM(virt_vm.BaseVM):
                         self, m_name, m_params, timeout
                     )
                 except qemu_monitor.MonitorConnectError as detail:
-                    LOG.error(detail)
-                    self.destroy()
-                    raise
+                    if self.process.is_alive():
+                        LOG.error(detail)
+                        self.destroy()
+                        raise
+                    else:
+                        stat = self.process.get_status()
+                        output = self.process.get_output().strip()
+                        self.destroy()
+                        raise virt_vm.VMCreateError(qemu_command, stat, output)
 
                 # Add this monitor to the list
                 self.monitors.append(monitor)


qingwangrh:
Pass test with this fix

qingwangrh:
@luckyh @YongxueHong Please help to review. thanks.

virttest/qemu_vm.py
```diff
@@ -3839,18 +3839,22 @@ def create(
                 self.destroy()
                 raise virt_vm.VMStartError(self.name, err)
 
+            def _check_vm_alive():
+                if not self.process.is_alive():
```
luckyh, line 3843: Looks we just need to put the condition check into the `wait_for` instead of the entire if clause.

```python
if not wait_for(self.process.is_alive, 5):
    ...
```

virttest/qemu_vm.py
```diff
@@ -3862,13 +3866,13 @@ def create(
...
+                    # Add this monitor to the list
+                    self.monitors.append(monitor)
                 except qemu_monitor.MonitorConnectError as detail:
-                    LOG.error(detail)
-                    self.destroy()
-                    raise
-
-                # Add this monitor to the list
-                self.monitors.append(monitor)
+                    if _check_vm_alive():
```
luckyh, line 3872: Would you please elaborate more about the reason for having this change?

qingwangrh: Sometimes The VM runs for a while before it finally fails. The first detection will get the VM status alive. The VM will be dead in seconds.  The following operation will report the monitoring issue. It hides the real root cause. 

The first detection should be in duration and must be detected again after monitoring access timeout.
the monitor issue is available when VM is alive, but the monitor fails.

virttest/qemu_vm.py
qingwangrh: If it just checks one time, it is ok. but the same checking needs to be done when monitoring access timeout.

qingwangrh: If we want to check at one place, it may consider to add check time, it could be like this :

if utils_misc.wait_for(lambda:not self.process.is_alive(), 8, 3):

What do you think ?

luckyh (approved):
LGTM

virttest/qemu_vm.py
```diff
@@ -3862,13 +3862,17 @@ def create(
...
-
-                # Add this monitor to the list
-                self.monitors.append(monitor)
+                    if self.process.is_alive():
+                        LOG.error(detail)
+                        self.destroy()
+                        raise
+                    else:
+                        status = self.process.get_status()
+                        output = self.process.get_output().strip()
```
YongxueHong, line 3874:
```suggestion
                        output = self.process.get_output().strip()
                        self.destroy()
```
I think we also need to destroy the VM here.

qingwangrh: updated

virttest/qemu_vm.py
```diff
@@ -3862,13 +3862,17 @@ def create(
                     monitor = qemu_monitor.wait_for_create_monitor(
                         self, m_name, m_params, timeout
                     )
+                    # Add this monitor to the list
+                    self.monitors.append(monitor)
```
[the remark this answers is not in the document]
qingwangrh: The previous code have 2 issues:
1. self.monitors is list , it should store all available monitors instead last one
2. the variable reference scope has potential risk, Local variable 'monitor' might be referenced before assignment

YongxueHong (approved):
LGTM.

PR #3931, opened 2024-06-17
utils_net:Remove color of "ip" output

chloerh:
The color charaters are hard to parse and affecting many tests. Add -c=never to remove color of ip command's output to fix the problem.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -1434,7 +1434,7 @@ def get_net_if(runner=local_runner, state=".*", qdisc=".*", optional=".*"):
     :param optional: optional match for interface find
     :return: List of network interfaces.
     """
-    cmd = "ip link"
+    cmd = "ip -c=never link"
     # As the runner converts stdout to unicode on Python2,
     # it has to be converted to string for struct.pack().
     result = str(runner(cmd))
@@ -1476,7 +1476,7 @@ def get_remote_host_net_ifs(session, state=None):
     """
     phy_interfaces = []
     vir_interfaces = []
-    cmd = "ip link"
+    cmd = "ip -c=never link"
     if not state:
         state = ".*"
     cmd_output = session.cmd_status_output(cmd)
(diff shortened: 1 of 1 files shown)


yanqzhan (approved):
LGTM. Network job cases can pass with it.

yanglei-rh:
Hi @chloerh  This change caused many automation errors for virtual network feature. Because after executing command with` -c=never,` when judging whether the command is executed correctly, it always gets non-`0`, which means that the command execution failed, thus causing tests failures. So could you please help re-write this to avoide cause regression issues on the virtual network side? Thanks.

chloerh:
> Hi @chloerh This change caused many automation errors for virtual network feature. Because after executing command with` -c=never,` when judging whether the command is executed correctly, it always gets non-`0`, which means that the command execution failed, thus causing tests failures. So could you please help re-write this to avoide cause regression issues on the virtual network side? Thanks.

Thanks for pointing this out, I'll remove `-c=never` for commands that don't need to have their output to be parsed first.

chloerh:
@yanglei-rh Hi, could you help provide some specific commands that I get non-0 return? I have removed a few options https://github.com/avocado-framework/avocado-vt/pull/3956

yanglei-rh:
> @yanglei-rh Hi, could you help provide some specific commands that I get non-0 return? I have removed a few options #3956

Hi @chloerh , I hit this issues with command ` ip -c=never addr add 192.168.10.11/24 dev eth0`, it always prompted me that is not recognized as an internal or external command.

chloerh:
> > @yanglei-rh Hi, could you help provide some specific commands that I get non-0 return? I have removed a few options #3956
> 
> Hi @chloerh , I hit this issues with command ` ip -c=never addr add 192.168.10.11/24 dev eth0`, it always prompted me that is not recognized as an internal or external command.

Thanks! I'll check this part.

fbq815:
@chloerh I've found out this change is blocking some tests cause the iproute2 version is too old to support "-c=never", my guest have the iprout2 version of "ip utility, iproute2-ss170501" maybe we need some internel workaround of this change

PR #3938, opened 2024-07-01
Enable virt_use_nfs when selinux is enforcing

xiaodwan:


dzhengfy (approved):
It is reasonable as this is in libvirt.py and only test libvirt cases. LGTM.

PR #3940, opened 2024-07-03
libvirt.py -- Fix create_disk_xml() sometimes creates inaccessible file

Slancaster1:
### The Issue
create_disk_xml() uses a disk.Disk() object to create a disk XML file. The disk.Disk() object, deletes the underlying XML when the object is garbage collected. The create_disk_xml() function only returns the XML filename, not the entire disk.Disk() object. As a result, the disk.Disk() object is garbage collected, and the underlying XML file is deleted.

This will result in any access to the XML file failing after a garbage collection cycle has been run.

### The Solution
Copy the contents of the XML file into a more permanent file.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -28,6 +28,7 @@ import shutil
 import string
 import sys
 import time
+import uuid
 
 import aexpect
 import six
@@ -1881,7 +1882,6 @@ def create_disk_xml(params):
         diskxml.xmltreefile.write()
     except Exception as detail:
         LOG.error("Fail to create disk XML:\n%s", detail)
-    LOG.debug("Disk XML %s:\n%s", diskxml.xml, str(diskxml))
 
     # Wait for file completed
     def file_exists():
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1892,7 +1892,13 @@ def file_exists():
 
     # Wait for file write over with '</disk>' keyword at the file end
     wait_for_file_over("</disk>", diskxml.xml)
-    return diskxml.xml
+
+    fd, path = tempfile.mkstemp(prefix="tmp_xml", suffix=".xml")
```
dzhengfy, line 1896: The function tempfile.mkstemp()  is insecurity, so do not recommend to use it. 
You might use data_dir.get_tmp_dir(), then you need not to delete the temp file after the test, either.

dzhengfy (commented):
Thanks for your fixing.

virttest/utils_test/libvirt.py
Slancaster1: Done!

virttest/utils_test/libvirt.py
```diff
@@ -27,7 +27,9 @@
 import shutil
 import string
 import sys
+import tempfile
```
dzhengfy, line 30: Could you remove this?

Slancaster1: Of course!

dzhengfy (approved):
LGTM

chloerh (approved):

chloerh:
"signed-off" missing in commit msg.

Slancaster1:
{quote}
"signed-off" missing in commit msg.
{quote}

Oops :facepalm: Fixed it!

PR #3941, opened 2024-07-04
add function check_guest_machine_type and check_audit_log

nanli1:
need to add function

For https://github.com/autotest/tp-libvirt/pull/5657

### DIFF
--- a/virttest/utils_libvirt/libvirt_vmxml.py
+++ b/virttest/utils_libvirt/libvirt_vmxml.py
@@ -57,6 +57,30 @@ def check_guest_xml(vm_name, pat_in_dumpxml, option="", status_error=False):
         raise exceptions.TestFail(msg)
 
 
+def check_guest_machine_type(vmxml, expected_version="9.4.0"):
+    """
+    Check guest machine type version
+
+    :param vmxml: the guest xml.
+    :param expected_version: expected guest machine version, eg
+    if <type arch='x86_64' machine='pc-q35-rhel8.2.0'>hvm</type> exists,
+    which means expected_version is 8.2.0
+    :return: True or False, the flag of checking successfully or not.
+    """
+    actual_version = re.sub(r"[a-zA-Z]", "", vmxml.os.machine.split("-")[-1])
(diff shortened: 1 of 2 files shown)


virttest/utils_libvirt/libvirt_vmxml.py
```diff
@@ -57,6 +57,26 @@ def check_guest_xml(vm_name, pat_in_dumpxml, option="", status_error=False):
...
+    """
+    Check guest machine type version
+
+    :param vmxml: the guest xml.
+    :param expected_version: expected guest machine version, eg
+    if <type arch='x86_64' machine='pc-q35-rhel8.2.0'>hvm</type> exists,
+    which means expected_version is 8.2.0
+    :return: True or False, the flag of checking successfully or not.
+    """
+    actual_version = int("".join(re.findall(r"\d+", vmxml.os.machine.split("-")[-1])))
```
dzhengfy, line 70: When expected_version="9.10.0" and actual_version= "10.0.0". The function will break.

So perhaps we have to compare the values one by one.

virttest/utils_libvirt/libvirt_vmxml.py
```diff
@@ -57,6 +57,26 @@ def check_guest_xml(vm_name, pat_in_dumpxml, option="", status_error=False):
...
+    expected_version = int(expected_version.replace(".", ""))
+
+    if actual_version < expected_version:
+        LOG.debug("Guest config machine should be >= rhel{}".format(expected_version))
+        return True
```
dzhengfy, line 75: I think we should return False in this condition.

virttest/utils_libvirt/libvirt_vmxml.py
nanli1: Thanks your reminder, Updated

virttest/utils_libvirt/libvirt_vmxml.py
nanli1: Thanks , Updated

dzhengfy (approved):
LGTM

PR #3942, opened 2024-07-05
libvirt_vm.py: Fix Invalid Escape Sequence Warnings

Slancaster1:
### The Issue
According to the python 3.12 release notes, a SyntaxWarning will now be raised on invalid escape sequences

https://docs.python.org/dev/whatsnew/3.12.html#other-language-changes

This will be upgraded to a "SyntaxError" in the future

This results in some regex strings, declared without the 'r' prefix generating warnings

### Note
I have only altered the regexs relevant to a few tests. There are many more regexs in the code that need this done.

### The Fix
Start marking regexs with the 'r' prefix


Slancaster1:
Is it a good idea to attempt to do this for all regexs in the code base?

dzhengfy (approved):
LGTM

PR #3943, opened 2024-07-08
env_process: Refactor running-as-root check into a Setuper 

bgartzi:
Moving the logic that checked that the process was run as root into a `Setuper`. Functional changes aren't intended with this patch.

ID: 2430

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -54,7 +54,10 @@ from virttest.test_setup.networking import (
     NetworkProxies,
 )
 from virttest.test_setup.os_posix import UlimitConfig
-from virttest.test_setup.requirement_checks import CheckInstalledCMDs
+from virttest.test_setup.requirement_checks import (
+    CheckInstalledCMDs,
+    CheckRunningAsRoot,
+)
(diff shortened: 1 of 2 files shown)


bgartzi:
Hi @YongxueHong, yet another `env_process`-`preprocess` refactoring patch.
Could you review it please?
Thanks a lot!

virttest/env_process.py
```diff
@@ -1087,13 +1090,9 @@ def preprocess(test, params, env):
                     pvr,
                     remote_pvr,
                 )
-    # First, let's verify if this test does require root or not. If it
-    # does and the test suite is running as a regular user, we shall just
-    # throw a TestSkipError exception, which will skip the test.
```
YongxueHong, line 1092: Just one comment, I suggest adding the above comments to the `class CheckRunningAsRoot(Setuper):` to keep the history.

bgartzi: I removed it as I considered the called function name `verify_running_as_root` clear enough. Anyway, let me push that right now.

YongxueHong: Yeah, you're right, I stand on your side.

PR #3946, opened 2024-07-10
Add conditional nat rule deletion for nftables

quanwenli:
- check if iptables or nftables is used
- if using nftables, use nft command to add the rules
- clean the nftables

id: 2592

### DIFF
--- a/virttest/test_setup/__init__.py
+++ b/virttest/test_setup/__init__.py
@@ -818,7 +818,11 @@ class PrivateBridgeConfig(object):
             u_port = params.get("guest_port_unattended_install", "13323")
             if u_port not in ports:
                 ports.append(u_port)
-            self.iptables_rules = self._assemble_iptables_rules(ports)
+            self.firewall_tool = self._check_firewall_tool()
+            if self.firewall_tool == "iptables":
+                self.iptables_rules = self._assemble_iptables_rules(ports)
+            else:
+                self.iptables_rules = self._assemble_nftables_rules(ports)
             self.physical_nic = params.get("physical_nic")
             if self.physical_nic:
                 if self.physical_nic.split(":", 1)[0] == "shell":
@@ -833,6 +837,14 @@ class PrivateBridgeConfig(object):
             if params.get("bridge_force_create", "no") == "yes":
                 self.force_create = True
         self.bridge_manager = utils_net.Bridge()
+
+    def _check_firewall_tool(self):
+        cmd = "find /usr/lib/modules/ -name iptable_filter.ko.xz"
+        result = process.system_output(cmd)
+        if result:
+            return "iptables"
+        else:
+            return "nft"
 
     def _assemble_iptables_rules(self, port_list):
         rules = []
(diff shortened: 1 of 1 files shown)


quanwenli:
@YongxueHong could. you help review it ? thanks a lot.

yanglei-rh (approved):
LGTM, and test pass on my side.

YongxueHong:
Hi @PaulYuuu , @fbq815 , @zhenyzha 
Could you please review this pull request to see if it affects your platform?
Thank you in advance.

zhenyzha:
> Hi @PaulYuuu , @fbq815 , @zhenyzha Could you please review this pull request to see if it affects your platform? Thank you in advance.

Yes, it works on ARM
```
# find /usr/lib/modules/ -name iptable_filter.ko.xz
.../kernel/net/ipv4/netfilter/iptable_filter.ko.xz
.../kernel/net/ipv4/netfilter/iptable_filter.ko.xz
```

virttest/test_setup/__init__.py
```diff
@@ -818,7 +818,11 @@ def __init__(self, params=None):
             u_port = params.get("guest_port_unattended_install", "13323")
             if u_port not in ports:
                 ports.append(u_port)
-            self.iptables_rules = self._assemble_iptables_rules(ports)
+            self.firewall_tool = self._check_firewall_tool()
+            if self.firewall_tool == "iptables":
+                self.iptables_rules = self._assemble_iptables_rules(ports)
+            else:
+                self.iptables_rules = self._assemble_nftables_rules(ports)
```
YongxueHong, line 825: Hi @quanwenli 
What's the policy for selecting the rules?
In other words, How can I determine the default rules when both `iptables` and `nftables` exist on the system?

From my researching, choosing between `iptables` and `nftables` depends on your specific requirements, the complexity of your network configurations, and your performance needs. For new projects or when upgrading existing systems, `nftables` is generally the preferred choice.

These are just my personal thoughts. I would like to hear your perspective. Thanks.

virttest/test_setup/__init__.py
```diff
@@ -834,6 +838,14 @@ def __init__(self, params=None):
...
         self.bridge_manager = utils_net.Bridge()
 
+    def _check_firewall_tool(self):
+        cmd = "find /usr/lib/modules/ -name iptable_filter.ko.xz"
+        result = process.system_output(cmd)
+        if result:
+            return "iptables"
+        else:
+            return "nft"
+
```
PaulYuuu, line 848: Even if iptable_filter can be found, that doesn't mean Firewalld uses it.

To check the Firewalld's backend, is to check its config file.
```
$ grep "^FirewallBackend" /etc/firewalld/firewalld.conf
FirewallBackend=nftables
```

I would suggest using firewall-cmd to add/del rules, firewall-cmd is the front, with this we don't need to care about the backend, and it can handle common use cases.

virttest/test_setup/__init__.py
quanwenli: Iptables was the initial tool used, and then nftables was introduced. For a while, both existed at the same time. In the latest systems, iptables has been retired. I create this pull request since it block the test in latest system. I think we can keep using iptables even if both iptables and nftables are available, as it would enhance the stability of the code.

For perf testing, I don’t care which tool is used since it will be flushed during the test case execution.

virttest/test_setup/__init__.py
quanwenli: The firewalld utility can only handle simple firewall use cases and cannot set up complex firewalls. sometimes It cannot provide equivalent iptables/nft commands. Also, I'm not sure if all systems have firewalld packages available. using firewalld instead may affect the stability of the framework.

virttest/test_setup/__init__.py
YongxueHong: For individual test cases, choosing tools may depend on the test developer. Some developers may not care about the tool used, while others may have strong preferences. Therefore, we should not set `iptables` as the default mandatory tool. Instead, it is better to base this decision on specific conditions.

Hi, @leidwang 
What do you think about choosing the tool to add the rules?
CC @fbq815 @zhenyzha 
Thanks.

yanglei-rh: Hi @YongxueHong I agree with @quanwenli point, I don't care which tools are used in testing, just it can support to add rules is ok for me. The iptables is deprecate from rhel 9,however the rhel 9 still an find `iptable_filter.ko.xz`, I think we should re-write choosing tool's rule.

yanglei-rh: Hi @YongxueHong @quanwenli Let me correct my points, from my side if using iptables on rhel 9 it will hit automation bug since host dmesg will print like this `Warning: Unmaintained driver is detected: nft_compat` error messges. Therefore, based on this, which tools will be used is important to me.  So @quanwenli , could you please help re-write this function `_check_firewall_tool(self):` ? I think based on the host version to choosing tools is better than find a iptables's file.

quanwenli: Firstly, according to the article at https://access.redhat.com/solutions/6739041, iptables is still supported in RHEL9, and the warning message is considered acceptable.

Secondly, as detailed in the 'Migrating from iptables to nftables' link, it's clear that:
  - firewalld should be used for simple firewall configurations.
  -nftables should be employed for more complex and performance-critical setups, such as those required for an entire network.

Finally, at begining design , I considered  using the host version to determine support for nftables. However, it's unclear from which version nftables should be supported. so, I think from module perspective, choosing modules  is more acceptable and accurate

fbq815: @YongxueHong since iptables is deprecated on 10 I agree to use nftables on 10 and later

quanwenli:
It has workaround with installed latest kernel-modules-extra. and vt maintenance will automate this according to the meeting.
So close it. Thanks all

PR #3956, opened 2024-07-23
utils_net:Remove unnecessary -c=never of ip command

chloerh:
The option -c=never is unnecessary if the output doesn't need to be parsed. It might cause error, therefore remove it.


yanglei-rh (approved):
LGTM and test pass.

Yingshun (approved):

PR #3969, opened 2024-08-06
guest_agent.py: Add new command 'guest-network-get-route'

6-dehan:
Add a new guest agent command 'gunetwork-get-routeats' to get route info of Guest.

ID: 2588
Signed-off-by: demeng <demeng@redhat.com>

### DIFF
--- a/virttest/guest_agent.py
+++ b/virttest/guest_agent.py
@@ -1047,3 +1047,14 @@ class QemuAgent(Monitor):
         cmd = "guest-get-cpustats"
         self.check_has_command(cmd)
         return self.cmd(cmd)
+
+    def get_network_route(self):
+        """
+        Get the network route of the guest by guest agent operation
+
+        :return: a list of network route info such as 'iface' and 'gateway'
+        'metric' and 'refcnt'...
+        """
+        cmd = "guest-network-get-route"
+        self.check_has_command(cmd)
+        return self.cmd(cmd)


virttest/guest_agent.py
```diff
@@ -1047,3 +1047,14 @@ def get_cpustats(self):
...
+    def get_network_route(self):
+        """
+        Get the network route of the guest by guest agent operation
+
+        :return: a list of network route info such as 'iface' and 'gateway'
+        'metric' and 'refcnt'...
+        """
+        cmd = "guest-network-get-route"
+        self.check_has_command(cmd)
+        return self.cmd(cmd)
```
leidwang, line 1060: LGTM.
Just sharing my thoughts on this function. It seems like there are a lot of functions that only execute one command, maybe we can consider writing a function and pass the command directly to that function, then we can simplify the code. After that, we do not need to write a new function for a new command. I am okay with the current code, we can try to improve it in the future.Thanks.

```suggestion
    def get_guest_info(self, cmd):
        """
         ...
        :return: a list of guest info such as 'iface' and 'gateway'
        'metric' and 'refcnt'...
        """
        self.check_has_command(cmd)
        return self.cmd(cmd)
```

leidwang (approved):

virttest/guest_agent.py
6-dehan: every single API has its independent description and I don't think it could be simplified into one function. 
And we have to provide a hint that the API's function is to upstream user.

6-dehan:
@vivianQizhu please help to review and merge this patch. thanks

leidwang:
Hi @vivianQizhu Would you please help review this MR when you have time.Thanks!

vivianQizhu (approved):

PR #3972, opened 2024-08-18
utils-stress:Update script to take vms parameter

misanjumn:
### utils-stress:Update script to take vms parameter

This patch will allow us to send vms as paramter to VMStressEvents
Through this change, we can specify in which vms we want to run the stress_events
If vms parameter is not specified, then default vms from env will be considered
Thus, there will not be any error in the dependant files

Signed-off-by: Misbah Anjum N [misanjum@linux.vnet.ibm.com](mailto:misanjum@linux.vnet.ibm.com)

### DIFF
--- a/virttest/utils_stress.py
+++ b/virttest/utils_stress.py
@@ -16,7 +16,7 @@ LOG = logging.getLogger("avocado." + __name__)
 
 
 class VMStressEvents:
-    def __init__(self, params, env):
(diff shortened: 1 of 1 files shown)


PraveenPenguin (approved):
LGTM

chunfuwen (approved):
lgtm

PR #3974, opened 2024-08-20
x86_cpu_models: update cpu models category

nanliu-r:
ID: 3974

Need to add versioned CPU models into category

### DIFF
--- a/virttest/cpu.py
+++ b/virttest/cpu.py
@@ -45,9 +45,20 @@ ARCH = platform.machine()
 
 CPU_TYPES = {
     "AuthenticAMD": [
+        "EPYC-Genoa-v1",
         "EPYC-Genoa",
+        "EPYC-Milan-v2",
+        "EPYC-Milan-v1",
         "EPYC-Milan",
+        "EPYC-Rome-v4",
+        "EPYC-Rome-v3",
+        "EPYC-Rome-v2",
+        "EPYC-Rome-v1",
         "EPYC-Rome",
+        "EPYC-v4",
+        "EPYC-v3",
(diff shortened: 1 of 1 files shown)


nanliu-r:
For AMD, will add the versioned CPU models for EPYC and newer.

For Intel, will add the versioned CPU models for Icelake and newer, except CPU model "Snowridge" which has been deprecated since rhel10.

-     Based on fix: https://bugzilla.redhat.com/show_bug.cgi?id=1942914#c20


Will consider add more versioned CPU models gradually in future instead of adding them all once in case there is any bad regression issue.

 
FEEDBACK from Developer

And normally CPU features cannot change because if they do, things are going to break upon migration, e.g. a feature may suddenly appear or disappear. When we need to change features, we introduce a new version of a CPU model, see e.g Skylake-Server-noTSX-IBRS, Skylake-Server-v5 and so on. We need to guarantee the guest ABI compatibility.

- https://qemu-project.gitlab.io/qemu/system/qemu-cpu-models.html

nanliu-r:
Test PASS on following hardware products.
AMD:

1. EPYC
2. EPYC-Rome
3. EPYC-Milan
4. EPYC-Genoa

Intel:

1. Icelaker Server
2. Cooperlake
3. Sapphire Rapids

Test scenarios: x86_cpu_model,x86_cpu_flags,cpu_device_hotpluggable

@PaulYuuu @luckyh @yanan-fu Please help review this patch, thanks!

PaulYuuu (approved):

yanan-fu (approved):

PR #3975, opened 2024-08-20
libvirt_xml/devices/address: add optional zpci subelement

smitterl:
On s390x, there can be a zpci subelement which is used to determine properties inside the guest: uid corresponds to the domain and fid corresponds to the slot number.

### DIFF
--- a/virttest/libvirt_xml/devices/address.py
+++ b/virttest/libvirt_xml/devices/address.py
@@ -15,6 +15,14 @@ class Address(base.TypedDeviceBase):
     def __init__(self, type_name, virsh_instance=base.base.virsh):
         # Blindly accept any/all attributes as simple dictionary
         accessors.XMLElementDict("attrs", self, parent_xpath="/", tag_name="address")
+        accessors.XMLElementNest(
+            "zpci_attrs",
+            self,
+            parent_xpath="/",
+            tag_name="zpci",
+            subclass=self.Zpci,
+            subclass_dargs={"virsh_instance": virsh_instance},
+        )
         super(self.__class__, self).__init__(
             device_tag="address", type_name=type_name, virsh_instance=virsh_instance
         )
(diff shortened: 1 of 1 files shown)


smitterl:
Tested manually:

```
>>> from virttest.libvirt_xml.devices import address as a
>>> addr = a.Address.new_from_dict({"type_name": "pci", "domain": "0x0000", "bus":"0x00", "slot":"0x00", "function":"0x0"})
>>> str(addr)
'<address type="pci" domain="0x0000" bus="0x00" slot="0x00" function="0x0" />'
>>> zpci = addr.Zpci.new_from_dict({"uid":"0x01", "fid":"0x0"})
>>> addr.set_zpci_attrs(zpci)
>>> str(addr)
'<address type="pci" domain="0x0000" bus="0x00" slot="0x00" function="0x0"><zpci uid="0x01" fid="0x0" /></address>'
```

smitterl:
Used in https://github.com/autotest/tp-libvirt/pull/5853

chunfuwen (approved):
lgtm

virttest/libvirt_xml/devices/address.py
```diff
@@ -46,3 +54,33 @@ def new_from_element(cls, element, virsh_instance=base.base.virsh):
...
+                self,
+                parent_xpath="/",
+                tag_name="zpci",
+                attribute="fid",
+            )
+            super(self.__class__, self).__init__(virsh_instance=virsh_instance)
+            self.xml = "<zpci/>"
+
+        @classmethod
+        def new_from_dict(cls, attributes, virsh_instance=base.base.virsh):
```
chloerh, line 82: I'm thinking we could have a base function of this kind : new_from_dict() for all xml classes, which could be implemented really simple by:
```
inst = class()
inst.setup_attrs(**attrs)
return inst
```
This way, we could use this one-line function to create xml instance without having to create a new new_from_dict function for each class

chloerh: maybe new_from_attrs() to avoid conflict

smitterl: @chloerh That base class, shouldn't that be dealt with in another PR? IIUC, it will require other classes to be modified, too. Do I understand this correctly?

chloerh: Yes. I forgot to update my comment. Actually I found that there is a function new_from_dict for all device classes. Anyway, this pr is ok for me.

chloerh (approved):

PR #3983, opened 2024-08-28
utils_sys.py: Add function for finding process PIDS

Slancaster1:
Added small utility function get_pids_for(\<list\>). This function reads the output of 'ps aux' and gets the pids for any process in \<list\>.

### DIFF
--- a/virttest/utils_sys.py
+++ b/virttest/utils_sys.py
@@ -72,3 +72,33 @@ def get_host_bridge_id(session=None):
 
     host_bridges = re.findall(hostbridge_regex, output)
     return host_bridges if host_bridges else []
+
+
+def get_pids_for(process_names, sort_pids=True, session=None):
+    """
+    Given a list of names, retrieve the PIDs for
+    matching processes. Sort of equivalent
+    to: 'ps aux | grep name'
+
+    :param process_names: List of process names to look for
+    """
+
+    status, ps_cmd = cmd_status_output("ps aux", shell=True, session=session)
+    if status != 0 or not ps_cmd:
+        return []
+
+    ps_output = ps_cmd.split("\n")
+    relevant_procs = [
+        proc
+        for proc in ps_output
+        for wanted_name in process_names
+        if wanted_name in proc
+    ]
+
+    relevant_procs = [line.split() for line in relevant_procs]
+    relevant_pids = [int(proc[1]) for proc in relevant_procs]
+
+    if sort_pids:
+        relevant_pids.sort()
+
+    return relevant_pids


Slancaster1:
@dzhengfy I do not know why the static checks fail

dzhengfy:
> @dzhengfy I do not know why the static checks fail

From the error message, you can see it is ** Running signedoff-check... error. There is incorrect sign-off message in your commit with  `root committed 12 hours ago` now. But you should give like` Signed-off-by: Slancaster1 <slancast@redhat.com>`

virttest/utils_sys.py
```diff
@@ -72,3 +72,26 @@ def get_host_bridge_id(session=None):
...
     return host_bridges if host_bridges else []
+
+
+def get_pids_for(process_names, session=None):
+    """
+
+    Given a list of names, retrieve the
+    PIDs for matching processes
+    Sort of equivalent to: 'ps aux | grep name'
+    :param process_names: List of process names to look for
```
dzhengfy, line 83: Please add one more blank line above this line which is to apply to docstring style. Thanks.
And also add a docstring for return value.

virttest/utils_sys.py
```diff
@@ -72,3 +72,26 @@ def get_host_bridge_id(session=None):
...
+
+    """
+
+    status, ps_cmd = cmd_status_output("ps aux", shell=True, session=session)
+    if status != 0 or not ps_cmd:
+        return []
+
+    ps_output = ps_cmd.split("\n")
+    relevant_procs = [x for x in ps_output for n in process_names if n in x]
```
dzhengfy, line 92: Could you please replace x, n with more meaning words for better readability?

virttest/utils_sys.py
```diff
@@ -72,3 +72,26 @@ def get_host_bridge_id(session=None):
...
+    relevant_procs = [x.split() for x in relevant_procs]
```
dzhengfy, line 93: same as above

virttest/utils_sys.py
```diff
@@ -72,3 +72,26 @@ def get_host_bridge_id(session=None):
...
+    relevant_pids = [int(x[1]) for x in relevant_procs]
```
dzhengfy, line 94: same as above

virttest/utils_sys.py
```diff
@@ -72,3 +72,31 @@ def get_host_bridge_id(session=None):
...
+    relevant_procs = [
+        proc
+        for proc in ps_output
+        for wanted_name in process_names
+        if wanted_name in proc
+    ]
+
+    relevant_procs = [line.split() for line in relevant_procs]
+    relevant_pids = [int(proc[1]) for proc in relevant_procs]
+    relevant_pids.sort()
```
dzhengfy, line 100: If we sort the pids forcefully,  those pids could not match those process names in their original order.
The order might be useful for later use. So I suggest we'd better provide an option for sorting. 
What do you think of?

dzhengfy (commented):
Others LGTM

dzhengfy (approved):

Yingshun (approved):

PR #3984, opened 2024-08-29
Replacing deprecated 'warn' function with 'warning'

bskjois:
The warn() is deprecated in all the supported versions of python and is now removed in python3.13.

Hence we should be using warning() which is functionally identical to warn().

Referrence:
https://docs.python.org/3.12/library/logging.html#logging.Logger.warning https://docs.python.org/3.13/whatsnew/3.13.html#logging

### DIFF
--- a/avocado_vt/plugins/vt_joblock.py
+++ b/avocado_vt/plugins/vt_joblock.py
@@ -108,7 +108,7 @@ class VTJobLock(Pre, Post):
                     try:
                         os.unlink(path)
                     except OSError:
-                        self.log.warn("Unable to remove stale lock: %s", path)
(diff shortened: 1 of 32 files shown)


luckyh:
Hi @bskjois , thank you for introducing this enhancement!

Before we are going to merge the PR, would you mind to fix the style issues caught by the static checks? thanks in advance!

bskjois:
Hello @luckyh,
Thanks for the review and comment. I have fixed the style issues. Please take a look.

Thank you!

luckyh (approved):

luckyh:
Thanks!

PR #4002, opened 2024-09-18
Obtaining the ip address of the guest and update address cache

TasmiyaNalatwad:
Few of the times it is seen that the ip address is not being fetch and the avocado runs were failed with error "ERROR: Failures occurred while postprocess:\n\n: Guest virt-tests-vm1 dmesg verification failed: Login timeout expired (output: 'exceeded 240 s timeout, last failure: No ipv4 DHCP lease for MAC aa:bb:cc:dd:ee:ff') "

To handle this error the patch has been sent. The patch helps in obtaining ip address of the guest using "virsh-net-dhcp-leases default" command. If the guest mac address is found in the command output, the mac ipv4 address is obatined and updated in the address.cache


TasmiyaNalatwad:
@luckyh @lmr @yanglei-rh Requesting to please review this PR and let me know the valuable inputs. 

Thank you!

luckyh:
> Requesting to please review this PR and let me know the valuable inputs.

Forwarding this to the maintainers who have better experience on libvirt use cases than me.

@dzhengfy @chloerh @Yingshun would you mind to help review this PR? thanks in advance!

yanglei-rh:
Hi @TasmiyaNalatwad Thanks for help sync this MR first, I am not a libvirt tester, so @luckyh mentioned user are more suitable candidate.

TasmiyaNalatwad:
Hi @chloerh Would like to know if you've had a chance to run the tests for the patch. Please let me know if there's any feedback or input you can share. Will work on it.

Thanks!

virttest/utils_net.py
```diff
@@ -4889,3 +4890,21 @@ def check_class_rules(ifname, rule_id, bandwidth, expect_none=False):
...
+
+
+def obtain_guest_ip_from_dhcp_leases(mac):
+    """
+    Obtaining the guest ip address from virsh-net-dhcp-leases command
+    :param: Mac address of the guest
+    :return: return ip-address if found for given mac in the
+             virsh-net-dhcp-leases default table, else return None
+    """
+    output = virsh.net_dhcp_leases("default")
```
chloerh, line 4902: This function will get dhcp lease from network "default" only, which might not be the default setting for test vms, right?

virttest/libvirt_vm.py
```diff
@@ -378,6 +379,15 @@ def backup_xml(self, active=False):
             LOG.error("Failed to backup xml file:\n%s", detail)
             return ""
 
+    def _get_address(self, index=0, ip_version="ipv4", session=None, timeout=60.0):
+        try:
+            return super()._get_address(index, ip_version, session, timeout)
+        except virt_vm.VMIPAddressMissingError:
+            mac = self.get_mac_address(index).lower()
+            ipaddr = utils_net.obtain_guest_ip_from_dhcp_leases(mac)
```
chloerh, line 387: I have another concern: we have this scenario: 
modifying interface type from network to bridge, but keep the mac address. Seems the dhcp-lease will not be deleted for a while, but the ip address of interface must change. There's a chance we're getting the 'old' ip address, right?

chloerh:
> Hi @chloerh Would like to know if you've had a chance to run the tests for the patch. Please let me know if there's any feedback or input you can share. Will work on it.
> 
> Thanks!

Hi @TasmiyaNalatwad  I have finished the run, but met some error which I'm **not** sure was brought by this patch. I have left 2 comments, Please check. Thanks.

TasmiyaNalatwad:
Thanks @chloerh Will look into the comments.

TasmiyaNalatwad:
Hi @chloerh Thanks a lot for the reviews. After thinking that it is a valid scenario where `virsh net-dhcp-leases` may take some time to load the new ip when there is change in interfaces from bridge to network or vice-versa we will have a risk of having older data.
Hence why I have removed `"virsh-net-dhcp-leases default"` and i am using `"virsh domifaddr"` . The code is for just libvirt users as i am defining it inside libvirt_vm.py

Every time the ip address is obtained, it will be purely from `virsh domifaddr vm-name --full --source arp` which wont be having issues with lease time or older data from cache.

Requesting you to please review the PR

virttest/utils_net.py
```diff
@@ -4889,3 +4890,21 @@ def check_class_rules(ifname, rule_id, bandwidth, expect_none=False):
...
+             virsh-net-dhcp-leases default table, else return None
+    """
+    output = virsh.domifaddr(vm_name, "--full --source arp")
+    lines = output.stdout.splitlines()
+    for line in lines:
+        if mac in line:
+            parts = line.split()
+            for part in parts:
+                if "/" in part:
+                    return part.split("/")[0]
```
luckyh, line 4909: One minor issue is that, do you consider to support handling ipv6 addresses within this change? otherwise, I'd require to introduce a check against the `VM._get_address`, that is, only calling this function when `ip_version` is `"ipv4"`.

luckyh:
Cc @nickzhq

virttest/utils_net.py
TasmiyaNalatwad: @luckyh Thank you for your inputs.
We generally rely on IPv4 addresses and do not consider IPv6 in our current workflow. Additionally, I have reviewed the code and observed that IPv4 is explicitly used throughout, with no instances where IPv6 is being requested.

@luckyh @nickzhq Please let me know how you think on this.

Thank you

luckyh: Well, then I will request either such a change like the following

```suggestion
def obtain_guest_ip_from_domifaddr(vm_name, mac, ip_version="ipv4"):
    """
    Obtaining the guest ip address from virsh-net-dhcp-leases command
    :param: Mac address of the guest
    :return: return ip-address if found for given mac in the
             virsh-net-dhcp-leases default table, else return None
    """
    if ip_version != "ipv4":
        # no support for ipv6 at the moment
        return None
    output = virsh.domifaddr(vm_name, "--full --source arp")
    lines = output.stdout.splitlines()
    for line in lines:
        if mac in line:
            parts = line.split()
            for part in parts:
                if "/" in part:
                    return part.split("/")[0]
```

or a modification to the caller, e.g.

```diff
@@ -2,7 +2,9 @@
         try:
             return super()._get_address(index, ip_version, session, timeout)
         except virt_vm.VMIPAddressMissingError:
-            mac = self.get_mac_address(index).lower()
-            ipaddr = utils_net.obtain_guest_ip_from_domifaddr(self.name, mac)
-            self.address_cache[mac] = ipaddr
-            return ipaddr
+            if ip_version == "ipv4":
+                mac = self.get_mac_address(index).lower()
+                ipaddr = utils_net.obtain_guest_ip_from_domifaddr(self.name, mac)
+                self.address_cache[mac] = ipaddr
+                return ipaddr
+        return None
```
Otherwise, it's still buggy.

TasmiyaNalatwad:
> Hello @TasmiyaNalatwad , does this patch you created impact the tp-qemu since I did NOT know requirement background exactly?

Hi @nickzhq  This is just for libvirt users, the code changes are also in libvirt_vm.py file to make it specific for libvirt users and tp-qemu will remain untouched here.

Thank you

luckyh (approved):
LGTM

TasmiyaNalatwad:
Thanks @luckyh for reviews and Approval.

@chloerh @nickzhq Would you please review and help me finalise the patch for merging?

Thank you so much

nickzhq:
It's okay for me.

hariharants:
As we have required approvals, May I request to merge the patch

luckyh:
> As we have required approvals, May I request to merge the patch

let's wait for @chloerh for confirming if the above comments have been addressed.

Thanks,

PR #4013, opened 2024-10-17
workaround bad file descriptor issue happened in postprocess 

chunfuwen:
In recent tp-libvirt testing, lots of case false warning failure happened due to ungraceful handling of postprocess cleanup in env_process.py.
Here it is workaround solution, just output error message instead of raising exception

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -1785,7 +1785,7 @@ def postprocess(test, params, env):
             LOG.error(details)
 
     if err:
-        raise RuntimeError("Failures occurred while postprocess:\n%s" % err)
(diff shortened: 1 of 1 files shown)


nanli1 (approved):
Agree this update, useful

chloerh (commented):
Could we set a condition to filter the `file descriptor` issues instead of ignoring all?

cliping:
The bad file descriptor issue in postprocess has been fixed in https://github.com/avocado-framework/avocado-vt/pull/4001. And it did not appear in 103lts branch. Maybe we can close this PR.

PR #4016, opened 2024-10-25
Return the local varriable to delay the desctruction of the xml file

liang-cong-red-hat:
When we use migration option --xml or --persistent-xml, we need provide a existing xml file, but in _update_xmlfile_path_in_migrate_options() vmxml object is a local variable, it will be released after the _update_xmlfile_path_in_migrate_options()  is done, so the xml file with vmxml object is released at the same time.

That will cause the migration fail. So return the local variable would delay the destruction after the migration.


dzhengfy (approved):
LGTM

PR #4024, opened 2024-11-12
Change default memory-backend for s390x

maxujun:
Change default memory-backend to memory-backend-file


PaulYuuu:
@maxujun @fbq815, any background on this change?

fbq815:
@PaulYuuu based on libvirt usage, memory-backend-file is set by default and I've add other details in jira task, we could discuss there

PaulYuuu:
> @PaulYuuu based on libvirt usage, memory-backend-file is set by default and I've add other details in jira task, we could discuss there

From which libvirt version memory-backend-file is the default? Or always the default?

fbq815:
@maxujun @PaulYuuu sorry guys, I think we could close this patch, please check detail in jira ticket

PR #4030, opened 2024-11-21
vt_utils/cpu.py: fix bug

nickzhq:
changing from re.match.groups() to re.match.group().

### DIFF
--- a/virttest/vt_utils/cpu.py
+++ b/virttest/vt_utils/cpu.py
@@ -54,7 +54,7 @@ def get_cpu_model_name():
     cpu_model = re.search(cpu_model_re, cpu_model)
     if not cpu_model:
         raise OSError("The cpu model name was NOT found!")
-    cpu_model = cpu_model.groups()[0]
+    cpu_model = cpu_model.group()
     return cpu_model.split(":", 1)[-1].strip()
 
 


nickzhq:
@YongxueHong  Please take a look, thanks!

virttest/vt_utils/cpu.py
```diff
@@ -54,7 +54,7 @@ def get_cpu_model_name():
     cpu_model = re.search(cpu_model_re, cpu_model)
     if not cpu_model:
         raise OSError("The cpu model name was NOT found!")
-    cpu_model = cpu_model.groups()[0]
+    cpu_model = cpu_model.group()
```
YongxueHong, line 57: Hi @nickzhq 
It looks like it still doesn't work well since the `cpu_model` is the `tuple` data type.
```
>>> from virttest.vt_utils import cpu
>>> cpu.get_cpu_model_name()
  File "/home/yhong/kar/workspace/avocado-vt/virttest/vt_utils/cpu.py", line 58, in get_cpu_model_name
    return cpu_model.split(":", 1)[-1].strip()
AttributeError: 'tuple' object has no attribute 'split'
>>> 
```

nickzhq: > Hi @nickzhq It looks like it still doesn't work well since the `cpu_model` is the `tuple` data type.
> 
> ```
> >>> from virttest.vt_utils import cpu
> >>> cpu.get_cpu_model_name()
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/home/yhong/kar/workspace/avocado-vt/virttest/vt_utils/cpu.py", line 58, in get_cpu_model_name
>     return cpu_model.split(":", 1)[-1].strip()
> AttributeError: 'tuple' object has no attribute 'split'
> >>> 
> ```

Can you show me the machine you used? This code works for me at both my laptop and remote server.
The following are return: 
Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
AMD EPYC 7313 16-Core Processor

Thanks!

YongxueHong: Hi @nickzhq 
Sorry, it's my fault, your code works well.
Thanks.

PR #4035, opened 2024-12-03
env_process: Refactor kvm module reload/restore steps

bgartzi:
KVM modules are reloaded and restored before and after running tests. That was done directly in the preprocess and postprocess functions in virttest.env_process. Write a Setuper subclass that implements those in setup/cleanup methods and register the setuper in the env_process setup_manager.

This is a patch from a larger patch series refactoring the env_process preprocess and postprocess functions. In each of these patches, a pre/post process step is identified and replaced with a Setuper subclass so the following can finally be met:
    - Only cleanup steps of successful setup steps are run to avoid possible environment corruption or hard to read errors.
    - Running setup/cleanup steps symmetrically during env pre/post process.
    - Reduce explicit pre/post process function code length.

ID: 2934

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -22,7 +22,6 @@ from avocado.utils import process as a_process
 from six.moves import xrange
 
 from virttest import (
-    arch,
     cpu,
     data_dir,
     error_context,
@@ -31,7 +30,6 @@ from virttest import (
     qemu_storage,
     storage,
     test_setup,
-    utils_kernel_module,
     utils_libguestfs,
     utils_logfile,
     utils_misc,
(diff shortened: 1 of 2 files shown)


bgartzi:
A test failed at first [0]. After checking it, I thought it could have been a cirrus issue, I'm not sure though. A flaky test perhaps? Anyway, after pushing the patch again all the checks were able to pass.

@YongxueHong would you mind having a look? Thanks!

0. https://cirrus-ci.com/task/5848486225117184

smitterl (approved):

YongxueHong (approved):
LGTM.

PR #4037, opened 2024-12-05
Move function get_vm_ip from tp-libvirt to avocado-vt

chloerh:
Moving from https://github.com/autotest/tp-libvirt/blob/master/provider/virtual_network/network_base.py#L22-L55


smitterl:
@chloerh Please link the PRs in tp-libvirt that remove the function from network base and use this new import path.

PR #4039, opened 2024-12-10
env_process: Tested package version check refactoring

bgartzi:
Another bunch of patches regarding the `env_process` `preprocess` and `postprocess` functionr refactoring patch series. In this one, those steps related to extracting, logging and checking required package versions are refactored.

I usually send 1 patch per PR. However, I'm sending a bunch of them together this time. This is due to the `version_info`, a variable that lived inside the `preprocess` function and that most of the version-check related steps used. I'm doing so not to modify the public API of the `env_process` submodule.

First `version_info` is made a global variable of the submodule, so both steps inside the `preprocess` and those being refactored can access it. Once all have been written into `Setuper`s, then the `version_info` variable is also migrated to the `virttest.test_setup.requirement_checks` submodule.

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -55,7 +55,6 @@ from virttest.test_setup.networking import (
 )
 from virttest.test_setup.os_posix import UlimitConfig
 from virttest.test_setup.ppc import SwitchSMTOff
-from virttest.test_setup import requirement_checks
 from virttest.test_setup.requirement_checks import (
     CheckInstalledCMDs,
     CheckKernelVersion,
@@ -64,6 +63,7 @@ from virttest.test_setup.requirement_checks import (
     CheckRunningAsRoot,
     CheckVirtioWinVersion,
     LogBootloaderVersion,
+    LogVersionInfo,
 )
 from virttest.test_setup.storage import StorageConfig
 from virttest.test_setup.verify import VerifyHostDMesg
@@ -1022,14 +1022,12 @@ def preprocess(test, params, env):
     _setup_manager.register(LogBootloaderVersion)
     _setup_manager.register(CheckVirtioWinVersion)
     _setup_manager.register(CheckLibvirtVersion)
+    _setup_manager.register(LogVersionInfo)
     _setup_manager.do_setup()
 
     vm_type = params.get("vm_type")
 
     base_dir = data_dir.get_data_dir()
-
-    # Write it as a keyval
-    test.write_test_keyval(requirement_checks.version_info)
 
     libvirtd_inst = None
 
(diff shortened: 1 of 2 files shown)


bgartzi:
Hey @YongxueHong, this one is larger than usual. As mentioned, it was so as not to change  the public `env_process` API between different MRs.

However, I pushed a new version (which results in the same code) that doesn't really change the API of `env_process`. Now, the main reason to push all of these together is just that all of them look into different package versions and make sure they meet the test demands.

If you'd prefer, I could split the PR into a bunch of them instead as usual. Just let me know.

YongxueHong:
> Hey @YongxueHong, this one is larger than usual. As mentioned, it was so as not to change the public `env_process` API between different MRs.
> 
> However, I pushed a new version (which results in the same code) that doesn't really change the API of `env_process`. Now, the main reason to push all of these together is just that all of them look into different package versions and make sure they meet the test demands.
> 
> If you'd prefer, I could split the PR into a bunch of them instead as usual. Just let me know.

Hi @bgartzi 
It's clear enough for me, let's keep them. Thanks.

virttest/env_process.py
```diff
@@ -1033,7 +1034,6 @@ def preprocess(test, params, env):
 
     base_dir = data_dir.get_data_dir()
 
-    version_info = {}
```
YongxueHong, line 1036: Hi @bgartzi 
In my opinion, we might introduce a new interface for user, in this case, we can get the `version_info` everywhere as we make it as an attribute of the module `requirement_checks`. Therefore, the user could get the info  by calling `requirement_checks.version_info`. 
Once we decide to do that, we need to update the related cases, there are some cases that also store the version by test.write_test_keyval, for example:
https://github.com/autotest/tp-qemu/blob/fc98779efd3988ad9622f802cd4fd5dea93676b3/qemu/tests/performance.py#L297-L299
```
    test.write_test_keyval({"kvm-userspace-ver": kvm_ver})
    test.write_test_keyval({"host-kernel-ver": host_ver})
    test.write_test_keyval({"guest-kernel-ver": guest_ver})
```
Base on the current solution, we should store the above info to `requirement_checks.version_info`, please correct me if I misunderstood.

If you just aim to refactor the setuper and have to move the `version_info` into a global variable, I think we can let the each setuper write the version info by calling `self.test.write_test_keyval`, since we get the test object inside the setuper class.
For example:
class CheckKernelVersion(Setuper):
    def setup(self):
        # Get the KVM kernel module version
        ...
        LOG.debug("KVM version: %s" % kvm_version)
        version_info["kvm_version"] = str(kvm_version)
        test.write_test_keyval(version_info)
        ...

YongxueHong (approved):
LGTM.

PR #4042, opened 2024-12-28
Use raw image format by default for block devices

pevogam:
Not providing a default behavior leads to use of variable without a value and thus cryptic errors regarding the simple fact that we have to provide an image format. It is also not recommended in any conditionals as it leads to undefined behaviors. Finally, in the case of multiple images we may have "image_format=" to use empty string which should be handled similarly to None (lack of any image_format parameter) or generally as the comment suggests to a meaningful default behavior.

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -2563,9 +2563,11 @@ class DevContainer(object):
                 format_cls = qdevices.QBlockdevFormatLuks
             elif imgfmt == "nvme":
                 format_cls = qdevices.QBlockdevFormatRaw
-            elif imgfmt is None:
-                # use RAW type as the default
+            elif imgfmt is None or imgfmt == "":
+                # use RAW type as the default (None or empty string)
                 format_cls = qdevices.QBlockdevFormatRaw
+            else:
+                raise ValueError(f"Unsupported image format: {imgfmt}")
 
             protocol_node = protocol_cls(name)
             devices.append(protocol_node)


pevogam:
@luckyh @YongxueHong I hope this is an improvement, it definitely helped us.

virttest/qemu_devices/qcontainer.py
```diff
@@ -2563,8 +2563,8 @@ def define_hbas(
                 format_cls = qdevices.QBlockdevFormatLuks
             elif imgfmt == "nvme":
                 format_cls = qdevices.QBlockdevFormatRaw
-            elif imgfmt is None:
-                # use RAW type as the default
+            else:
+                # use RAW type as the default (None, empty string, or else)
```
YongxueHong, line 2567: Hi @pevogam 
Thank you for pointing this out. Based on your commit message, it seems we didn’t account for the empty string case.
To address this, we could add a condition like `imgfmt == ""` or use `elif not imgfmt` to handle the both cases.
For other values of `imgfmt`, we should avoid providing a default type since the user has already specified the format. Instead, we can handle unsupported formats by raising an error in the else block:
`raise ValueError(f"Unsupported image format: {imgfmt}.")`
Let me know your thoughts on this approach. Thank you!

pevogam: > For other values of imgfmt, we should avoid providing a default type since the user has already specified the format. 

Interesting I thought we specify all possible formats in the above conditionals? Can you give an example of such a format that we don't have to handle above but is also not the default raw format?

YongxueHong: > > For other values of imgfmt, we should avoid providing a default type since the user has already specified the format.
> 
> Interesting I thought we specify all possible formats in the above conditionals? Can you give an example of such a format that we don't have to handle above but is also not the default raw format?

We've handled almost all formats, but there's an edge case where the user might mistakenly specify an unsupported image format. In such cases, we can't use the default image format, as it wouldn't make sense and could introduce a logical error. This could potentially cause issues.
On the other hand, based on my understanding, we don't have sufficient evidence to justify using the raw format as the default for the 'else' image. Instead, we can ensure that it properly handles None and empty cases, since the user likely doesn't care about the image format. In such cases, we could provide a default image format.
Thanks.

pevogam: Ok, I see and what you say makes sense. Pushed with the changes.

YongxueHong (approved):

PR #4044, opened 2024-12-29
Add the virtiofs element of openfiles

iccaszhulili:


Yingshun (dismissed):

virttest/libvirt_xml/devices/filesystem.py
```diff
@@ -77,6 +78,7 @@ class Binary(base.base.LibvirtXMLBase):
             "lock_posix",
             "flock",
             "thread_pool_size",
+            "open_files",
```
Yingshun, line 81: Is it better to rename it to 'openfiles_max'? The naming format needs to be consistent with other variables I think. What do you think?

iccaszhulili: @Yingshun Updated. Thanks.

Yingshun (approved):

chloerh (approved):

PR #4045, opened 2024-12-30
QEMU_VM: Add nic_extra_param_types support

heywji:
Add nic_extra_param_types params support for nic_extra_params to optimize virtio-net QEMU cmdline type handling.

ID: 3269

Signed-off-by: wji <wji@redhat.com>

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -740,11 +740,24 @@ class VM(virt_vm.BaseVM):
                     dev.parent_bus = pci_bus
                     dev.set_param("addr", pci_addr)
                 if nic_extra_params:
+                    nic_extra_params_type = dict(
+                        eval(self.params.get("nic_extra_params_type", {}))
+                    )
                     nic_extra_params = (
                         _.split("=", 1) for _ in nic_extra_params.split(",") if _
                     )
                     for key, val in nic_extra_params:
-                        dev.set_param(key, val)
+                        value_type = nic_extra_params_type.get(key)
+                        if value_type:
+                            try:
+                                converted_val = eval(value_type)(val)
+                                dev.set_param(key, converted_val)
+                            except ValueError as e:
+                                raise ValueError(
+                                    f"Failed to convert {key}: {val} to {value_type}: {e}"
+                                )
+                        else:
+                            dev.set_param(key, val)
                 dev.set_param("bootindex", bootindex)
                 if "aarch64" in params.get("vm_arch_name", arch.ARCH):
                     if "rombar" in devices.execute_qemu("-device %s,?" % model):


heywji:
@leidwang Hi,

Could you please help review this netkvm qemu_vm improvement?

I have tested **positive** and **negative** logic and **kept the original logic** that defaults to str. All of that works great for me.

heywji:
@vivianQizhu Hi, I'm calling you to review this patch as well. Thanks a lot.

virttest/qemu_vm.py
```diff
@@ -740,11 +740,24 @@ def add_nic(
...
+                    )
                     nic_extra_params = (
                         _.split("=", 1) for _ in nic_extra_params.split(",") if _
                     )
                     for key, val in nic_extra_params:
-                        dev.set_param(key, val)
+                        value_type = nic_extra_params_type.get(key)
+                        if value_type:
+                            try:
+                                converted_val = value_type(val)
```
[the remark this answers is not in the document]
leidwang:
```
>>> nic_extra_params_type = {"a":"int"}
>>> value_type = nic_extra_params_type.get("a")
>>> value_type
'int'
>>> val = "10"
>>> converted_val = value_type(val)
TypeError: 'str' object is not callable
>>> converted_val = eval(value_type)(val)
>>> type(converted_val)
<class 'int'>
```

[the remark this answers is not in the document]
heywji: Hi @leidwang
We have a split opinion on line 1. Currently, I am using it in this way:

```
>>> nic_extra_params_type = {"a":int}
>>> value_type = nic_extra_params_type.get("a")
>>> value_type
<class 'int'>
>>> val = "10"
>>> converted_val = value_type(val)
>>> type(converted_val)
<class 'int'>
>>> 
```

virttest/qemu_vm.py
```diff
@@ -740,11 +740,24 @@ def add_nic(
                     dev.parent_bus = pci_bus
                     dev.set_param("addr", pci_addr)
                 if nic_extra_params:
+                    nic_extra_params_type = dict(
+                        eval(self.params.get("nic_extra_params_type", {}))
```
heywji, line 744: Add some details here: Setting 'str' as the default value for line 760 is not straightforward, as we cannot predict how the user will utilize the key(for the val is 'str', yes). So, perhaps we should take a look at this carefully.

virttest/qemu_vm.py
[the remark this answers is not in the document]
leidwang: `nic_extra_params_type = {"a":int}` 
"int" is not a number, it's a string, I think we need add "" for it.

virttest/qemu_vm.py
leidwang: I think we already discussed this issue before you sent the MR.If we do not set params's type, it will be a default value(str).

virttest/qemu_vm.py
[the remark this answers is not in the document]
heywji: After discussing with @leidwang offline, we want to let @vivianQizhu decide.

In my opinion, Any data can be put into a dict of Python class type because all memory is mapped at the System Bottom. I don't think there is a need for" to set the class type.

Leidong's opinion is that he saw some "int"(highlight: "") in our avocado-vt framework and tp-qemu project. We should make the style the same.

We are just waiting for your professional opinion. @vivianQizhu

heywji:
Hi @leidwang,

I accepted your advice about this patch after talking with @vivianQizhu offline.
I will update the code and test it. I will let you review it again when everything is ready.

Thank you~

heywji:
@leidwang, call you to review this patch again. Thanks in advance.

leidwang (approved):
LGTM.

virttest/qemu_vm.py
vivianQizhu: This logic is wrong, you could have multiple extra params and they probably have different value type, so you should have it in below `for` loop.

vivianQizhu: @heywji Please check the above comment.

heywji: @vivianQizhu Hi Qianqian, let me introduce my codes more here. Line743-744 `nic_extra_params_type` is used to get the value of the `nic_extra_params_type` variable from our tp-qemu cfg file. Usually, only two logics exist in which the user sets the `nic_extra_params_type` variable or not. Line 751 of conditional judgment will follow the logic of the master branch if users do not set the `nic_extra_params_type` variable. Line 753-754 will handle users' different value types if users set the `nic_extra_params_type` variable.

vivianQizhu: @heywji No, consider you have 3 extra params, one string, one bool, and one int, current implementation can not handle that.

vivianQizhu:
@PaulYuuu Nice patch for https://github.com/avocado-framework/avocado-vt/pull/4056, but I don't find how it handle extra_params, it only handles named parameters, I suppose this PR won't be conflict with that one, as extra_params can only be string now, we need to extend it.

virttest/qemu_vm.py
vivianQizhu: Oh my bad, you have a dictionary defined in cfg, surprisingly. I don't like it honestly, but it works and nothing more to complain about.

vivianQizhu:
@PaulYuuu Correct myselft, I see how it handles all params, with `_update_args_type_from_qemu()`.
@heywji Could you try with https://github.com/avocado-framework/avocado-vt/pull/4056 and see if it fixes your issue? If so I guess we can close this one.

heywji:
@vivianQizhu @PaulYuuu OK, I will try it.

PaulYuuu:
> @PaulYuuu Correct myself, I see how it handles all params, with `_update_args_type_from_qemu()`.
Yes, it should cover all parameters, also it supports both nic_extra_params and netdev_extra_params. The PR has not yet collected all correct parameter types, it should require all feature owners for review.

heywji:
Hi all,
I closed this patch first since I have confirmed that #4056 can solve the same problem with a better method.

For the next step, I will mainly focus on my tp-qemu, which adds a new test case.
netkvm: Add a case for packet loss when slowly reusing memory buffers [#4247](https://github.com/autotest/tp-qemu/pull/4247)

Best,
Wecom

PR #4054, opened 2025-01-17
env_process: Refactor huge pages setup/cleanup steps

bgartzi:
Creating a new Setuper subclass for setting and cleaning huge pages up. Removing the original code from virttest.env_process and replacing it instead with the new HugePagesSetup class being registered in the setup_manager.

_pre_hugepages_surp and _post_hugepages_surp were left in env_process. Their goal is to provide a mechanism in env_process to raise a TestFail in case pages were leaked during a test. If that mechanism was refactored into the setuper, the TestFail would be masked by just an Error due to the way setup_manager handles postprocess exceptions. Changing the way SetupManager handles that requires bigger discussion on how the test infrastructure should handle test status reports, which is a way broader topic that what this patch aims to be.

This is a patch from a larger patch series refactoring the env_process preprocess and postprocess functions. In each of these patches, a pre/post process step is identified and replaced with a Setuper subclass so the following can finally be met:
    - Only cleanup steps of successful setup steps are run to avoid possible environment corruption or hard to read errors.
    - Running setup/cleanup steps symmetrically during env pre/post process.
    - Reduce explicit pre/post process function code length.

### DIFF
--- a/virttest/test_setup/memory.py
+++ b/virttest/test_setup/memory.py
@@ -1,15 +1,21 @@
-from virttest import env_process, test_setup, utils_libvirtd
+from virttest import test_setup, utils_libvirtd
 from virttest.test_setup.core import Setuper
 
 
 class HugePagesSetup(Setuper):
+    def __init__(self, test, params, env):
+        super().__init__(test, params, env)
+        # default num of surplus hugepages, in order to compare the values
+        # before and after the test when 'setup_hugepages = yes'
+        self._pre_hugepages_surp = 0
+
     def setup(self):
         # If guest is configured to be backed by hugepages, setup hugepages in host
         if self.params.get("hugepage") == "yes":
             self.params["setup_hugepages"] = "yes"
         if self.params.get("setup_hugepages") == "yes":
             h = test_setup.HugePageConfig(self.params)
-            env_process._pre_hugepages_surp = h.ext_hugepages_surp
+            self._pre_hugepages_surp = h.ext_hugepages_surp
             suggest_mem = h.setup()
             if suggest_mem is not None:
                 self.params["mem"] = suggest_mem
@@ -24,4 +30,7 @@ class HugePagesSetup(Setuper):
             h.cleanup()
             if self.params.get("vm_type") == "libvirt":
                 utils_libvirtd.Libvirtd().restart()
-            env_process._post_hugepages_surp = h.ext_hugepages_surp
+            post_hugepages_surp = h.ext_hugepages_surp
+            if post_hugepages_surp > self._pre_hugepages_surp:
+                leak_num = post_hugepages_surp - self._pre_hugepages_surp
+                raise test_setup.HugePagesLeakError("%d huge pages leaked!" % leak_num)
--- a/virttest/test_setup/memory.py
+++ b/virttest/test_setup/memory.py
@@ -1,0 +1,27 @@
+from virttest import env_process, test_setup, utils_libvirtd
+from virttest.test_setup.core import Setuper
+
+
+class HugePagesSetup(Setuper):
+    def setup(self):
+        # If guest is configured to be backed by hugepages, setup hugepages in host
+        if self.params.get("hugepage") == "yes":
+            self.params["setup_hugepages"] = "yes"
+        if self.params.get("setup_hugepages") == "yes":
+            h = test_setup.HugePageConfig(self.params)
+            env_process._pre_hugepages_surp = h.ext_hugepages_surp
+            suggest_mem = h.setup()
+            if suggest_mem is not None:
+                self.params["mem"] = suggest_mem
+            if not self.params.get("hugepage_path"):
+                self.params["hugepage_path"] = h.hugepage_path
+            if self.params.get("vm_type") == "libvirt":
+                utils_libvirtd.Libvirtd().restart()
+
+    def cleanup(self):
+        if self.params.get("setup_hugepages") == "yes":
+            h = test_setup.HugePageConfig(self.params)
+            h.cleanup()
+            if self.params.get("vm_type") == "libvirt":
+                utils_libvirtd.Libvirtd().restart()
+            env_process._post_hugepages_surp = h.ext_hugepages_surp
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -92,11 +92,6 @@ _vm_info_thread_termination_event = None
 
 _setup_manager = SetupManager()
 
-# default num of surplus hugepage, order to compare the values before and after
-# the test when 'setup_hugepages = yes'
-_pre_hugepages_surp = 0
-_post_hugepages_surp = 0
-
 #: Hooks to use for own customization stages of the virtual machines with
 #: test, params. and env as supplied arguments
 preprocess_vm_off_hook = None
@@ -1576,9 +1571,6 @@ def postprocess(test, params, env):
 
     if err:
         raise RuntimeError("Failures occurred while postprocess:\n%s" % err)
-    elif _post_hugepages_surp > _pre_hugepages_surp:
-        leak_num = _post_hugepages_surp - _pre_hugepages_surp
-        raise exceptions.TestFail("%d huge pages leaked!" % leak_num)
 
 
 def postprocess_on_error(test, params, env):
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -46,6 +46,7 @@ from virttest.test_setup.core import SetupManager
 from virttest.test_setup.gcov import ResetQemuGCov
 from virttest.test_setup.kernel import ReloadKVMModules
 from virttest.test_setup.libvirt_setup import LibvirtdDebugLogConfig
+from virttest.test_setup.memory import HugePagesSetup
 from virttest.test_setup.migration import MigrationEnvSetup
 from virttest.test_setup.networking import (
     BridgeConfig,
@@ -1023,6 +1024,7 @@ def preprocess(test, params, env):
     _setup_manager.register(CheckVirtioWinVersion)
     _setup_manager.register(CheckLibvirtVersion)
     _setup_manager.register(LogVersionInfo)
+    _setup_manager.register(HugePagesSetup)
     _setup_manager.do_setup()
 
     vm_type = params.get("vm_type")
@@ -1030,24 +1032,6 @@ def preprocess(test, params, env):
     base_dir = data_dir.get_data_dir()
 
     libvirtd_inst = None
-
-    # If guest is configured to be backed by hugepages, setup hugepages in host
-    if params.get("hugepage") == "yes":
-        params["setup_hugepages"] = "yes"
-
-    if params.get("setup_hugepages") == "yes":
-        global _pre_hugepages_surp
-        h = test_setup.HugePageConfig(params)
-        _pre_hugepages_surp = h.ext_hugepages_surp
-        suggest_mem = h.setup()
-        if suggest_mem is not None:
-            params["mem"] = suggest_mem
-        if not params.get("hugepage_path"):
-            params["hugepage_path"] = h.hugepage_path
-        if vm_type == "libvirt":
-            if libvirtd_inst is None:
-                libvirtd_inst = utils_libvirtd.Libvirtd()
-            libvirtd_inst.restart()
 
     if params.get("setup_thp") == "yes":
         thp = test_setup.TransparentHugePageConfig(test, params, env)
@@ -1533,21 +1517,6 @@ def postprocess(test, params, env):
 
     libvirtd_inst = None
     vm_type = params.get("vm_type")
-
-    if params.get("setup_hugepages") == "yes":
-        global _post_hugepages_surp
-        try:
-            h = test_setup.HugePageConfig(params)
-            h.cleanup()
-            if vm_type == "libvirt":
-                if libvirtd_inst is None:
-                    libvirtd_inst = utils_libvirtd.Libvirtd()
-                libvirtd_inst.restart()
-        except Exception as details:
-            err += "\nHP cleanup: %s" % str(details).replace("\\n", "\n  ")
-            LOG.error(details)
-        else:
-            _post_hugepages_surp = h.ext_hugepages_surp
 
     if params.get("setup_thp") == "yes":
         try:
--- a/virttest/test_setup/__init__.py
+++ b/virttest/test_setup/__init__.py
@@ -76,6 +76,14 @@ class THPKhugepagedError(THPError):
     pass
 
 
+class HugePagesLeakError(Exception):
+    """
+    Thrown when huge pages are leaked after cleanup.
+    """
+
+    pass
+
+
 class PolkitConfigError(Exception):
 
     """


YongxueHong:
Hi @PaulYuuu 
Could you help to review it? Thanks.

virttest/test_setup/memory.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class HugePagesSetup(Setuper):
+    def setup(self):
+        # If guest is configured to be backed by hugepages, setup hugepages in host
+        if self.params.get("hugepage") == "yes":
+            self.params["setup_hugepages"] = "yes"
+        if self.params.get("setup_hugepages") == "yes":
+            h = test_setup.HugePageConfig(self.params)
+            env_process._pre_hugepages_surp = h.ext_hugepages_surp
```
PaulYuuu, line 12: `_pre_hugepages_surp` and `_post_hugepages_surp` are for hugepage leak check, with this Setuper, I would suggest dropping them. by returning a variable after `do_cleanup`. so `leak_num = _post_hugepages_surp - _pre_hugepages_surp` can short to `leak_num = <new_var>`.

bgartzi: If I understood correctly, you are proposing to make the `cleanup` method of this `Setuper` return the `leak_num` value?
That would also involve updating the `SetupManager` behavior to meet the demands of `HugePagesSetup`. If that's the case, we would permit every `Setuper` return a value, which goes against the current implementation. We would then have to update the `SetupManager` `do_cleanup` logic to handle that.

In my opinion, if we were to do that, we would have to think of a _protocol_ of some sort to make this implementable by each `Setuper` instead of adding specific `Setuper` logic into the rather general `SetupManager`. Could something like adding a `post_cleanup_check` function into the core `Setuper` and calling it after the `cleanup` method has been called from `SetupManager.do_cleanup` be the answer to that issue?

I also thought on other approaches, as implementing a core `Singleton` abstraction, so we would be able to reach `Setuper` instances instead of classes from within `env_process` so we could call extra functions on demand after the cleanup would have terminated. However, this approach sounds too complex for a workaround, and it could introduce further issues, as `Setuper` instances "_surviving_" from one test case run to the next one.

PaulYuuu: > If I understood correctly, you are proposing to make the `cleanup` method of this `Setuper` return the `leak_num` value?
No, the workaround is the Setuper will calculate `_post_hugepages_surp - _pre_hugepages_surp` and set env_process._hugepage_leaks(take this name as example).

The complex solution you mentioned is that Setuper can return something. I agree we can do this, but not now, the implementation can closely combine env_process and Setuper.

PaulYuuu: No, IIRC, if `hugepages_deallocate` is not set, we will never hit the error, `HugePages_Surp` only changes when leaked or using THP, which is unrelated to setup/teardown.  And yes we should move `self.ext_hugepages_surp = utils_memory.get_num_huge_pages_surp()` out from the `if self.deallocate` in cleanup.

Without the Setuper, cannot handle huge page leak checks in the cleanup in the current pre/post process context, pre and post will init a separate HugePageConfig class, so we cannot directly check `ext_hugepages_surp` cross in 2 HugePageConfig.

For now, as we have each setuper, then check leak is possible, and we can do it, even in this PR or an individual one.

bgartzi: Following each discussion thread:

## **Deallocation + leak check**
#### What should we do if deallocation doesn't happen? Should we check for leaks?
I see there isn't any test setting `hugepages_deallocte`. I see that it is set to `hugepages_deallocate = yes` by default here: https://github.com/avocado-framework/avocado-vt/blob/3fdae9b00ad61943c7a9039ee5167d71e335f395/virttest/shared/cfg/guest-hw.cfg#L220-L223

Based on 5b4db50a3fa6b3220ee90e5157b261cb6a5d5ccc, I see the deallocate option was added due to some small-memory PPC systems. So, it could be set to `no` for that reason.

I'm not a memory or hugepages expert, so I'm not sure if I understand this properly. But, based on what I understand by allocation/deallocation, here are my question:
- If we _allocate_ or mount huge pages, does the `utils_memory.get_num_huge_pages_surp` output value change? In other words, does the surp huge pages value change if we mount/unmount hugepages?
- If that happens, I would expect the same to happen if we deallocate/unmount hugepages. Right?
- If both of the previous are true: We measure `ext_hugepages_surp` before `setup`. So if we measure it after `setup` but before `cleanup`, the obtained value would be different. If we would run the leak check (AKA what triggers the TestFail) between `setup` and `cleanup` I'd expect the test to fail. Is that assumption correct?
- Even if the previous one is true: if we don't `deallocate`, we wouldn't run `cleanup` at all, so I'd expect the `utils_memory.get_num_huge_pages_surp` value to be the same we would obtain right after setup. That is, different to pre-`setup`, where we measured `ext_hugepages_surp` for the last time. If that's correct, wouldn't getting `self.ext_hugepages_surp = utils_memory.get_num_huge_pages_surp()` out from the `if self.deallocate` clause cause all tests not deallocating (few, as far as I know) to fail?

Again, this logical reasoning is based on my limited understanding on the topic, so please, correct me.

## **Leak check failure trigger**
#### Where should we trigger the failure/error if we detect a leak?
@YongxueHong, as you suggest, I would refactor the piece of code running the check into `HugePagesSetup.cleanup`. However, that brings me some concerns about the raised error type:
- The check raises a `TestFail`. That sets the test status to "Failure".
- If we write the check into `HugePagesSetup.cleanup`, the `SetupManager` will gather the raised exception (together with any other sort of exception) and will raise a `RuntimeError`. This would set the test status to "Error".
- If we change this, we would change the behavior of all tests related to huge pages and how those register their result status.
- That's my main concern: in summary, a test error not being the same as a test failure.

I believe we should limit ourselves to just postprocessing the test environment here, not declaring the result of the test during cleanup. However, I also understand that if we only can check `ext_hugepages_surp` only after huge pages `umount`, test cases don't have a way of checking that by themself either. If the spec of the huge pages tests declare that they should fail if there is a leak, then I guess we can't decide whether an error will suffice or not. So long story short: I'm not sure about masking the `TestFail` the leak check raises under a `RuntimeError`.

I'd like to hear your thoughts about this.

bgartzi: > > Following each discussion thread:
> > ## **Deallocation + leak check**
> > #### What should we do if deallocation doesn't happen? Should we check for leaks?
> > I see there isn't any test setting `hugepages_deallocte`. I see that it is set to `hugepages_deallocate = yes` by default here:
> > https://github.com/avocado-framework/avocado-vt/blob/3fdae9b00ad61943c7a9039ee5167d71e335f395/virttest/shared/cfg/guest-hw.cfg#L220-L223
> > 
> > Based on [5b4db50](https://github.com/avocado-framework/avocado-vt/commit/5b4db50a3fa6b3220ee90e5157b261cb6a5d5ccc), I see the deallocate option was added due to some small-memory PPC systems. So, it could be set to `no` for that reason.
> > I'm not a memory or hugepages expert, so I'm not sure if I understand this properly. But, based on what I understand by allocation/deallocation, here are my question:
> > 
> > * If we _allocate_ or mount huge pages, does the `utils_memory.get_num_huge_pages_surp` output value change? In other words, does the surp huge pages value change if we mount/unmount hugepages?
> > * If that happens, I would expect the same to happen if we deallocate/unmount hugepages. Right?
> > * If both of the previous are true: We measure `ext_hugepages_surp` before `setup`. So if we measure it after `setup` but before `cleanup`, the obtained value would be different. If we would run the leak check (AKA what triggers the TestFail) between `setup` and `cleanup` I'd expect the test to fail. Is that assumption correct?
> > * Even if the previous one is true: if we don't `deallocate`, we wouldn't run `cleanup` at all, so I'd expect the `utils_memory.get_num_huge_pages_surp` value to be the same we would obtain right after setup. That is, different to pre-`setup`, where we measured `ext_hugepages_surp` for the last time. If that's correct, wouldn't getting `self.ext_hugepages_surp = utils_memory.get_num_huge_pages_surp()` out from the `if self.deallocate` clause cause all tests not deallocating (few, as far as I know) to fail?
> > 
> > Again, this logical reasoning is based on my limited understanding on the topic, so please, correct me.
> > ## **Leak check failure trigger**
> > #### Where should we trigger the failure/error if we detect a leak?
> > @YongxueHong, as you suggest, I would refactor the piece of code running the check into `HugePagesSetup.cleanup`. However, that brings me some concerns about the raised error type:
> > 
> > * The check raises a `TestFail`. That sets the test status to "Failure".
> > * If we write the check into `HugePagesSetup.cleanup`, the `SetupManager` will gather the raised exception (together with any other sort of exception) and will raise a `RuntimeError`. This would set the test status to "Error".
> 
> Hi @bgartzi From my understanding of the env_proces, I think the purpose of raising the `RuntimeError` at the end of the post process is to collect all the errors and not stop to perform the next `cleanup` if there is one failed cleanup, so we can not raise each error type and in the meanwhile make sure all the `cleanup` can be executed. In case we want to add another checkpoint to raise a TestFail later, how do we guarantee that all checkpoints will be executed? So I think we can let the `RuntimeError`(a unified error) cover the errors from each `cleanup`.
> 
Yes, that's how `cleanup` works. There's one way that comes to my mind, see:
```patch
diff --git a/virttest/test_setup/core.py b/virttest/test_setup/core.py
--- a/virttest/test_setup/core.py
+++ b/virttest/test_setup/core.py
@@ -2,6 +2,7 @@ import logging
 from abc import ABCMeta, abstractmethod
 
 import six
 
 LOG = logging.getLogger("avocado." + __name__)
 
@@ -94,11 +95,32 @@ class SetupManager(object):
 
         :return: Errors occurred in cleanup procedures.
         """
+        cancels = []
+        fails = []
         errors = []
+        warns = []
         while self.__setupers:
             try:
                 self.__setupers.pop().cleanup()
+            except exceptions.TestCancel as err:
+                LOG.error(str(err))
+                cancels.append(str(err))
+            except exceptions.TestFail as err:
+                LOG.error(str(err))
+                fails.append(str(err))
+            except exceptions.TestWarn as err:
+                LOG.warning(str(err))
+                warns.append(str(err))
             except Exception as err:
                 LOG.error(str(err))
                 errors.append(str(err))
+        if cancels:
+            msg = "\n".join(cancels)
+            raise exception.TestCancel(msg)
+        elif fails:
+            msg = "\n".join(fails)
+            raise exception.TestFail(msg)
+        elif not errors and warns:
+            msg = "\n".join(warns)
+            raise exception.TestWarn(msg)
         return errors
```
However, that implies some sort of error/fail/warning sort of hierarchy, and that we support postprocess steps to raise `TestFails` as well as other sort of errors. However, we already agreed that that is not a behavior we want to support. And if we did, I think that would be worth of a whole new separate patch.

As you mention, yes, I also agree we can just merge all errors into the `RuntimeError` that the `Setuper`'s `cleanup` raises.

The problem is that if we do mask the leak detection under the `RuntimeError`, the behavior users (hugepages tests) would expect from avocado-vt will change. If you, as avocado-vt maintainers agree on changing that behavior, I'm also okay with that. Which based on the comments below, I think is the case.

> > * If we change this, we would change the behavior of all tests related to huge pages and how those register their result status.
> > * That's my main concern: in summary, a test error not being the same as a test failure.
> > 
> > I believe we should limit ourselves to just postprocessing the test environment here, not declaring the result of the test during cleanup. However, I also understand that if we only can check `ext_hugepages_surp` only after huge pages `umount`, test cases don't have a way of checking that by themself either. If the spec of the huge pages tests declare that they should fail if there is a leak, then I guess we can't decide whether an error will suffice or not. So long story short: I'm not sure about masking the `TestFail` the leak check raises under a `RuntimeError`.
> > I'd like to hear your thoughts about this.
> 
> I think the key point is to understand the main responsibility of the `post-process`, even the `pre-process`.
> 
> I think the preprocess is mainly responsible for preparing the environment for the test case, for example, setting it up, installing the package, and so on. Instead, the `post process` works for cleaning up and uninstalling the package to keep the environment pure for the next test case. (We can not maintain and guide others in developing it if we do not define both the preprocess and postprocess.)
> 
> And if some errors occurred during the post process, we can not think it belongs to the TestFail or TestError, since those errors do not affect the result of the test case from the test case perspective, it just did some teardown steps finally.
> 
> Therefore, If this checkpoint is one step for the test case, I prefer to let the test case handle it, rather than in the post-process, otherwise, it should follow up the rule of the post-process which is to raise all the errors with a unified error type(Whatever the type of the error). Sum it up: Let the test case raise the `TestError`, `TestFail`, and so on, instead in the pre-process and post-process.
> 
Yes, I also agree that would be the ideal situation. However, based on my understanding, hugepage tests need the `cleanup` to be executed, that is they need to wait until hugepages are released until they check whether a leak happened or not. That's why the `TestFail` in this case was written in `postprocess` instead of in the test case itself, I think.

> Please let me know your thoughts. Thanks.
Looking forward to know about yours. I think we're converging into an agreement. Thanks a lot!

YongxueHong: Hi @MiriamDeng @fbq815 @zhenyzha 
How about your opinions from the feature owner's perspective?  
Could we move the checking huge page leak to the `cleanup` or keep this checking in the post process?
Please let me know your views. Thanks.
CC @mcasquer

bgartzi: BTW, I just added two commits that move the huge pages leak out of the `virttest.env_process.postprocess`. Hope that serves as a draft. I can squash the commits into one once we all agree this is the solution that we want.

fbq815: I think it's fine it we move the checking hugepage leak to the cleanup

zhenyzha: I think there is no problem moving it to the cleanup stage, but I'm not sure if this will have an impact on the existing cases of tp-qemu. Our current practice is to configure hugepages before preprocess_vm(). like this: https://github.com/autotest/tp-qemu/blob/master/qemu/tests/hugepage_specify_node.py#L70-L76

zhenyzha: Thanks for the explanation. So this patch is acceptable to me :)

bgartzi: @YongxueHong, I don't think it's going to impact that specific test case.
The test case is using `virttest.test_setup.HugePageConfig`. That class is not changed by this patch. This patch introduces `virttest.test_setup.memory.HugePagesSetup` and adds the huge page leak check in its' cleanup method, not  in `HugePageConfig`'s. `...memory.HugePagesSetup` (being added in this PR) uses `HugePageConfig` itself in its `setup` and `cleanup` methods. As the `HugePagesSetup` class is new, tp-qemu tests aren't using it, so this shouldn't affect any test at all (apart from the testfail -> error thing we already discussed).

I think this is a matter of almost similar names between the setuper, and an "almost-setuper" class that already existed for huge pages setup (`HugePageConfig`), which is the one test cases are using directly.

bgartzi: Missed the deletion :sweat_smile:. Also removing mine. BTW, do you think I should find a better name for the setuper class to avoid these kind of confusions in the future?

mcasquer: IMO this patch looks good to me as well :)
Ps: I will trigger some hugepages tests later

mcasquer:
Memory huge_page test runs **passed** with RHEL 10 and Win2025 guests
```
 (01/18) Host_RHEL.m10.u0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.10.0.x86_64.io-github-autotest-qemu.unattended_install.cdrom.extra_cdrom_ks.default_install.aio_threads.q35: STARTED
 (01/18) Host_RHEL.m10.u0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.10.0.x86_64.io-github-autotest-qemu.unattended_install.cdrom.extra_cdrom_ks.default_install.aio_threads.q35: PASS (955.79 s)
 (02/18) Host_RHEL.m10.u0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.10.0.x86_64.io-github-autotest-qemu.system_reset_bootable.q35: STARTED
 (02/18) Host_RHEL.m10.u0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.10.0.x86_64.io-github-autotest-qemu.system_reset_bootable.q35: PASS (385.09 s)
...
 (18/18) Host_RHEL.m10.u0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.10.0.x86_64.io-github-autotest-qemu.hugepage_mem_stress.non_existent_mem_path.q35: STARTED
 (18/18) Host_RHEL.m10.u0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.10.0.x86_64.io-github-autotest-qemu.hugepage_mem_stress.non_existent_mem_path.q35:PASS (135.92 s)
RESULTS    : PASS 15 | ERROR 0 | FAIL 0 | SKIP 3 | WARN 0 | INTERRUPT 0 | CANCEL 0
```

```
 (01/18) Host_RHEL.m9.u6.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2025.x86_64.io-github-autotest-qemu.unattended_install.cdrom.extra_cdrom_ks.default_install.aio_threads.q35: STARTED
 (01/18) Host_RHEL.m9.u6.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2025.x86_64.io-github-autotest-qemu.unattended_install.cdrom.extra_cdrom_ks.default_install.aio_threads.q35: PASS (5306.26 s)
 (02/18) Host_RHEL.m9.u6.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2025.x86_64.io-github-autotest-qemu.disable_win_update.q35: STARTED
 (02/18) Host_RHEL.m9.u6.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2025.x86_64.io-github-autotest-qemu.disable_win_update.q35: PASS (499.09 s)
...
 (18/18) Host_RHEL.m9.u6.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2025.x86_64.io-github-autotest-qemu.migrate.tcp.with_filter_off.with_post_copy.q35: STARTED
 (18/18) Host_RHEL.m9.u6.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2025.x86_64.io-github-autotest-qemu.migrate.tcp.with_filter_off.with_post_copy.q35: PASS (274.00 s)
RESULTS    : PASS 15 | ERROR 0 | FAIL 0 | SKIP 3 | WARN 0 | INTERRUPT 0 | CANCEL 0
```

mcasquer (approved):
LGTM

YongxueHong (approved):
LGTM.

YongxueHong:
Hi @dzhengfy @chunfuwen 
I assume that this change may affect the libvirt test.
Could you help to review it from the libvirt side? Thanks a lot.

virttest/test_setup/memory.py
```diff
@@ -0,0 +1,36 @@
...
+            if self.params.get("vm_type") == "libvirt":
+                utils_libvirtd.Libvirtd().restart()
+
+    def cleanup(self):
+        if self.params.get("setup_hugepages") == "yes":
+            h = test_setup.HugePageConfig(self.params)
+            h.cleanup()
+            if self.params.get("vm_type") == "libvirt":
+                utils_libvirtd.Libvirtd().restart()
+            post_hugepages_surp = h.ext_hugepages_surp
```
chunfuwen, line 33: Maybe move post_hugepages_surp into __init__(self, test, params, env):  to easily to be tracked

bgartzi: It would be initialized to 0 and its value wouldn't change until `cleanup` is called. Do you consider there's some benefit from tracking it?

chunfuwen: just put variables into __init__  to make it more visible, but it doesn't impact function. 
LGTM

chunfuwen (approved):
lgtm

PR #4061, opened 2025-02-07
Get the active interface associated with gw

yalzhang:
The original get_default_gateway_json() in utils_net.py is incomplete and can not work well when iface_name=True. Update it.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -4112,9 +4112,15 @@ def get_default_gateway_json(
     elif len(default_route_list) == 1:
         default_route = default_route_list[0]
         if "gateway" in default_route:
-            return default_route["gateway"]
+            gw = default_route["gateway"]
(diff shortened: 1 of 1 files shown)


yalzhang:
before the fix, when iface_name=true, it still get the gateway not the iface name. The case below will fail:
virtual_network.connectivity_check.ethernet_interface.managed_no.negative_test.managed_yes.non_root_user
Error while retrieving IP address from interface 10.73.105.254.
after the fix, it works well
(1/1) type_specific.io-github-autotest-libvirt.virtual_network.connectivity_check.ethernet_interface.managed_no.negative_test.managed_yes.non_root_user: PASS (44.23 s)

yalzhang:
@chloerh  Please help to review, Thank you!

yalzhang:
duplicate

PR #4062, opened 2025-02-07
netperf_stress: Support running netperf clients on systems with more than 256 CPUs

bgartzi:
Netperf client couldn't start on systems with more than 256 cpus. This impacts the ability to run `netperf_stress` tests and possible some other more in those systems.

This patch tries to alleviate the issue by modifying the number of maximum cpus supported by netperf prior to the binary compilation.

To do so, the patch is divided in two commits:
1. To compile netperf, first the source code needs to be decompressed. Then, the code is compiled into a binary. Both commands were run through the same remote session. The first commit in this patch splits the pre-compilation and compilation steps into two different steps. That way, we could modify netperf's source code on-the-go if needed.
2. The second commit adds a mechanism to detect whether the system running netperf has a number of CPUS bigger than the value defined in netperf's `MAXCPUS` or not. First, it finds the number of CPUs the system is running on. Then, it finds the source code file in which `MAXCPUS` is defined and its value. If the value found is smaller than the number of CPUs in the system, it updates it's value by issuing a `sed` command. If it is bigger, the value is left as-is. This happens prior to compilation.

ID: 3365

### DIFF
--- a/virttest/utils_netperf.py
+++ b/virttest/utils_netperf.py
@@ -158,6 +158,63 @@ class NetperfPackage(remote_old.Remote_Package):
             self.session.cmd(clean_cmd, ignore_all_errors=True)
         self._release_session()
 
+    def _get_current_max_cpus(self):
+        """
+        Finds the file in the netperf source code that defines the MAXCPUS macro
+        and its value.
+        :return: Tuple containing the path to the file containing
+                 and the value assigned to the macro.
+        """
+        try:
+            max_cpus_fp, res_cpus = self.session.cmd_output(
+                'grep -orP "^#define *MAXCPUS *\K[0-9]+" %s' % self.netperf_dir
+            ).split(":", maxsplit=1)
+            max_cpus = int(res_cpus)
+        except aexpect.ShellError as err:
+            raise NetperfPackageError(
+                "Failed to find MAXCPUS value and file under '%s', see: %s"
+                % (self.netperf_dir, err)
+            )
+        except ValueError:
+            raise NetperfPackageError(
+                "Invalid MAXCPUS value found in file '%s': %s" % (max_cpus_fp, res_cpus)
+            )
+        LOG.debug(
+            "Found MAXCPUS is defined in %s with value %d" % (max_cpus_fp, max_cpus)
+        )
+        return max_cpus_fp, max_cpus
+
+    def _mod_max_cpus(self):
+        """
+        netperf is hardcoded to support up to 256 CPUS.
+        Modify the value from the source code previous to compilation to the
+        desired value.
+        """
+        # Obtain host/guest/vm/target netperf running sytem's number of CPUs
+        try:
+            n_cpus = int(
+                self.session.cmd_output('lscpu | grep -oP "^CPU\(s\)\: *\K[0-9]+"')
+            )
+            LOG.debug("Found %d cpus in the system running netperf" % n_cpus)
+        except aexpect.ShellError as err:
+            LOG.warning(
+                "Couldn't retrieve the number of cpus in the system: see %s" % err
+            )
+            n_cpus = 0
+        except ValueError:
+            LOG.warning("Couldn't parse the number of cpus in the system")
+            n_cpus = 0
+        max_cpus_file_path, current_max_cpus = self._get_current_max_cpus()
+        if current_max_cpus >= n_cpus:
+            LOG.debug("Bypassing netperf's MAXCPUS value modification")
+            return
+        LOG.info("Increasing netperf's MAXCPUS to %d" % n_cpus)
+        sed_cmd = 'sed -i "s/^\(#define *MAXCPUS *\)[0-9]* /\\1%d/g" %s' % (
+            n_cpus,
+            max_cpus_file_path,
+        )
+        self.session.cmd_output(sed_cmd, timeout=5)
+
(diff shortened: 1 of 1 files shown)


bgartzi:
@rh-jugraham could you have a look these patches and confirm that it would help fix your issue?

rh-jugraham:
@bgartzi This does address the issue!! 

To test it out, since there are currently very few machines with a CPU count > 256, I tested to ensure this change didn't negatively affect the tests when the machine already had fewer than 256 CPUs - all tests passed. Then, I made a slight modification to have the code set MAXCPUS to a smaller value, which in turn caused the tests to fail as expected - this showed that MAXCPUS was successfully modified and it had the expected effect on the tests.

```
max_cpus_file_path, current_max_cpus = self._get_current_max_cpus()
        if current_max_cpus >= n_cpus:
            LOG.debug("Bypassing netperf's MAXCPUS value modification")
            # return
        n_cpus = 32     # test that changing MAXCPUS affects the test
        LOG.info("Increasing netperf's MAXCPUS to %d" % n_cpus)
        sed_cmd = 'sed -i "s/^\(#define *MAXCPUS *\)[0-9]* /\\1%d/g" %s' % (
            n_cpus,
            max_cpus_file_path,
        )
```

rh-jugraham:
@yiqianwei @yanglei-rh hello! Was wondering if this PR could be revisited?

yanglei-rh:
Just review it from code level and based on the basic netperf stress from virtual network, this change has not impact for my feature.

yanglei-rh (approved):

yanglei-rh:
@bgartzi @rh-jugraham  Sorry for late reply your coment,after review the avocado-vt's reminder email was filtered to Trash.

yiqianwei:
Test results:
For x86_64 host with 448 cpus
 (1/1) Host_RHEL.m9.u6.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.10.0.x86_64.io-github-autotest-qemu.netperf_stress_test.TCP_STREAM.host2guest.q35: PASS (443.60 s)

check log:
[stdlog] 2025-06-27 05:39:50,887 avocado.virttest.utils_netperf utils_netperf    L0198 DEBUG| Found 448 cpus in the system running netperf
[stdlog] 2025-06-27 05:39:50,887 aexpect.client client           L1364 DEBUG| [10.26.0.68] Sending command: grep -orP "^#define *MAXCPUS *\K[0-9]+" /var/tmp/netperf-2.7.1
[stdlog] 2025-06-27 05:39:51,072 avocado.virttest.utils_netperf utils_netperf    L0182 DEBUG| Found MAXCPUS is defined in /var/tmp/netperf-2.7.1/src/netlib.h with value 256
[stdlog] 2025-06-27 05:39:51,073 avocado.virttest.utils_netperf utils_netperf    L0211 INFO | Increasing netperf's MAXCPUS to 448
[stdlog] 2025-06-27 05:39:51,073 aexpect.client client           L1364 DEBUG| [10.26.0.68] Sending command: sed -i "s/^\(#define *MAXCPUS *\)[0-9]* /\1448/g" /var/tmp/netperf-2.7.1/src/netlib.h

ACK

virttest/utils_netperf.py
```diff
@@ -158,26 +158,91 @@ def env_cleanup(self, clean_all=True):
...
+        self.env_cleanup(clean_all=False)
+        try:
+            self.session.cmd("echo $?", timeout=60)
+            self.session.cmd(pre_setup_cmd, timeout=600)
+        except aexpect.ShellError as e:
+            raise NetperfPackageError("Compilation setup failed: %s" % e)
+        try:
+            self._mod_max_cpus()
+        except aexpect.ShellError as err:
+            raise NetperfPackageError("Max CPU modification failed: %s" % err)
```
YongxueHong, line 233:
```suggestion
            LOG.error("Max CPU modification failed: %s" % err)
```
Hi @bgartzi 
I have a question about the `pack_compile` procedure. I've noticed that the test case will fail if it's unable to modify the max_cpus setting.

Is this CPU configuration a mandatory requirement for the tool to run correctly? I want to confirm because this step could be blocking valid test runs on platforms where this setting isn't critical.

If the feature owners confirm it's a required step, that's fine. I just want to ensure we aren't causing unnecessary failures.

What are your thoughts on this? Let me know if you see any issues with this approach.
Thanks.
CC @rh-jugraham @yanglei-rh @yiqianwei @fbq815 @zhenyzha

YongxueHong: Hi @bgartzi 
Thanks for the explanation—your solution works for me.
My only consideration is that calling `_get_current_max_cpus` could introduce new NetperfPackageError exceptions, though this is likely a rare edge case.

For example:
```
raise NetperfPackageError(
    "Failed to find MAXCPUS value and file under '%s', see: %s"
    % (self.netperf_dir, err)
)
```
and 
 ```
 raise NetperfPackageError(
      "Invalid MAXCPUS value found in file '%s': %s" % (max_cpus_fp, res_cpus)
  )
```

BTW, since the original issue was on ARM, I'm asking @fbq815 to review this change from the perspective of the s390x platform.
@fbq815, could you please take a look?
Thanks.

YongxueHong (approved):
LGTM.

PR #4070, opened 2025-03-04
utils_test: created function to update default kernel

rh-jugraham:
Created function to update default kernel

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -251,6 +251,109 @@ def update_boot_option(
             session.close()
 
 
+def update_vm_default_kernel(
+    vm,
+    kernel_version,
+    reboot,
+    guest_arch_name,
+    timeout="",
+    serial_login=False,
+):
+    """
+    Update guest default kernel.
+
+    :param vm: The VM object.
+    :param kernel_version: The full kernel version that should be set as the new default.
+                           Can use the entire kernel path or just the full kernel name.
+                           get_available_kernel_paths(session, pattern) is a useful helper function.
+    :param reboot: Whether to reboot the VM and update the kernel used or not.
+    :param guest_arch_name: Guest architecture, e.g. x86_64, s390x, aarch64.
+    :param timeout: Timeout for login and reboot.
+    :param serial_login: Login guest via serial session.
+    :raise NotImplementedError: Raised if Windows guest, not Linux.
+    :raise exceptions.TestError: Raised if failed to update the guest kernel.
+    """
+    if vm.params.get("os_type") == "windows":
+        raise NotImplementedError(
+            "update_vm_default_kernel() is supported only for Linux guest"
+        )
+
+    timeout = int(timeout) if timeout else int(vm.params.get("login_timeout"))
+    session = vm.wait_for_login(
+        timeout=timeout, serial=serial_login, restart_network=True
+    )
+    try:
+        if not utils_package.package_install("grubby", session=session):
+            raise exceptions.TestError("Failed to install grubby package")
+
+        cmd = "grubby --default-kernel"
+        status, output = session.cmd_status_output(cmd)
+        if status != 0:
+            LOG.error(output)
+        else:
+            if kernel_version in output:
+                LOG.info(
+                    "%s is already the default kernel version (%s)"
+                    % (kernel_version, output)
+                )
+                return
+
+        msg = "Set default guest kernel"
+        cmd = 'grubby --set-default="%s"' % kernel_version
+        __run_cmd_and_handle_error(
+            msg, cmd, session, "Failed to set default guest kernel"
+        )
+
+        if guest_arch_name == "s390x":
+            msg = "Update boot media with zipl"
+            cmd = "zipl"
+            __run_cmd_and_handle_error(
+                msg, cmd, session, "Failed to update boot media with zipl"
+            )
+
+        if reboot:
+            LOG.info("Rebooting guest ...")
+            session = vm.reboot(session=session, timeout=timeout, serial=serial_login)
+
+    finally:
+        if session:
+            session.close()
+
+
+def get_available_kernel_paths(session, pattern):
+    """
+    Get the full kernel paths given an identifying pattern
+
+    :params session: the vm session
+    :params pattern: regex pattern to identify specific kernel
+    :returns: list of kernel paths
+    :raises exceptions.TestError: raised if no matching kernels could be found
+    """
+    try:
+        if not utils_package.package_install("grubby", session=session):
+            raise exceptions.TestError("Failed to install grubby package")
+        cmd = "grubby --info=ALL"
+        lines = session.cmd_output(cmd).splitlines()
+
+        kernel_paths = []
+        for line in lines:
+            if line.startswith("kernel=") and re.match(pattern, line):
+                kernel = line.split("=")[1].strip('"')
+                kernel_paths.append(kernel)
+        LOG.info("kernel path(s) found: %s" % kernel_paths)
+
+        if not kernel_paths:
+            LOG.debug("Full %s output: %s" % (cmd, lines))
+            raise exceptions.TestError(
+                "No kernels with %s pattern could be found" % pattern
+            )
+        return kernel_paths
+    except:
+        raise exceptions.TestError(
+            "No kernels with %s pattern could be found" % pattern
+        )
+
+
 def stop_windows_service(session, service, timeout=120):
     """
     Stop a Windows service using sc.


rh-jugraham:
@smitterl As per discussion. Review would be appreciated! 

Used update_boot_option() function right above in the file as inspiration.

virttest/utils_test/__init__.py
```diff
@@ -251,6 +251,71 @@ def update_boot_option(
             session.close()
 
 
+def update_default_kernel(
```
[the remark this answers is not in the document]
dzhengfy: Usually we do not add new functions to __init__.py any more. This file is already too big.
Could you please move this function into utils_sys.py?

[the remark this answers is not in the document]
dzhengfy: I know there is update_boot_option() in __init__.py, too. Actually I prefer to move it out of the file, too, but we can do it later in another pr.

dzhengfy:
@smitterl , I feel if update_boot_option()  can be moved to another file, that will be better. So I do not suggest we add new function here. Not sure about your opinion. Welcome your feedback. Thanks

virttest/utils_test/__init__.py
```diff
@@ -251,6 +251,71 @@ def update_boot_option(
...
+            LOG.error(output)
+        else:
+            if kernel_version in output:
+                LOG.info(
+                    "%s is already the default kernel version (%s)"
+                    % (kernel_version, output)
+                )
+                return
+
+        if not utils_package.package_install("grubby", session=session):
```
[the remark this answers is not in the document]
smitterl: +1

smitterl:
> @smitterl , I feel if update_boot_option() can be moved to another file, that will be better. So I do not suggest we add new function here. Not sure about your opinion. Welcome your feedback. Thanks

I proposed to add it here because it's semantically related to the boot parameter update function but I wouldn't insist on it.

Alternatively, we can do both:
# add to utils_sys.py
# make it available in __init__.py 
```python
from utils_sys import update_default_kernel
```
IIUC, this makes it also available in this file. However, it's not strictly necessary, so whatever you prefer is fine with me.

virttest/utils_test/__init__.py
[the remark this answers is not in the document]
smitterl: Maybe for reuse of existing sessions in test code we could expect session as parameter instead.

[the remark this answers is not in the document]
rh-jugraham: @smitterl Since rebooting will result in a new session, would this imply also returning the session (either the updated one or, if no reboot, then the passed in one) for more reuse?

Also, I believe the vm needs to be a param because of, most importantly, vm.reboot() unless there is an alternative that doesn't require the vm. So the session would need to be in addition to the vm, not instead of.

rh-jugraham:
@smitterl @dzhengfy Created new PR - https://github.com/avocado-framework/avocado-vt/pull/4087 - for moving update_boot_option that depends on this PR (for shared imports/helper functions)

dzhengfy:
> @smitterl @dzhengfy Created new PR - #4087 - for moving update_boot_option that depends on this PR (for shared imports/helper functions)

@rh-jugraham @smitterl I think more about this change. Moving update_boot_option() and this pr to utils_sys.py will make  __init__.py pretty clean, but seems we also need move update_boot_option_ubuntu(), __run_cmd_and_handle_error(),  check_kernel_cmdline() .  Without moving them, we will have utils_test import utils_sys , and utils_sys import  utils_test  as well.

And if we move update_boot_option() to utils_sys.py, we have to also update tp-libvirt and tp-qemu for their reference to this function or we need still keep an encapsulating function in __init__.py for ease like below
in __init__.py

```
from virttest import utils_sys

def update_boot_option( ...):
   utils_sys.update_boot_option(...)
```

And check_kernel_cmdline() has one reference in tp-qemu (tp-libvirt does not use it) as below:
`qemu/tests/boot_without_vectors.py:40:                    req_args = utils_test.check_kernel_cmdline(session,`

Based on above analysis, it is high cost to make __init__.py to be pretty, at least I do not expect this happen in your PR. Maybe we use separate PR to achieve this later. So let's keep your current implementation using __init.py this time. 
Please help close https://github.com/avocado-framework/avocado-vt/pull/4087 and thank you for submitting it anyway.

dzhengfy (approved):
LGTM

rh-jugraham:
@dzhengfy Moved back to utils_test as indeed, currently the two tests are too integrated with utils_test and reliant on other functions there so it makes sense for a different PR to figure out how to best seperate out the functions.

dzhengfy (approved):
LGTM

virttest/utils_test/__init__.py
```diff
@@ -251,6 +251,72 @@ def update_boot_option(
...
+        return
+    timeout = int(timeout) or int(vm.params.get("login_timeout"))
+    session = vm.wait_for_login(
+        timeout=timeout, serial=serial_login, restart_network=True
+    )
+    try:
+        if not utils_package.package_install("grubby", session=session):
+            raise exceptions.TestError("Failed to install grubby package")
+
+        cmd = "grubby --default-kernel"
```
smitterl, line 286: Grubby will use the kernel image path

```
# grubby --default-kernel
/boot/vmlinuz-6.13.5-200.fc41.x86_64
```
That makes me wonder which format you use for `kernel_version`, especially considering that there's some arch-specific info in that image file name.
From line 299 I assume you expect the full kernel path as there's no manipulation.
How would a user retrieve that value before calling this function?  Would it make sense for this function to construct part of the full path, e.g. expecting only `6.13.5-200.fc41`?

smitterl (changes requested):

virttest/utils_test/__init__.py
```diff
@@ -251,6 +251,72 @@ def update_boot_option(
             session.close()
 
 
+def update_vm_default_kernel(
+    vm,
+    kernel_version,
+    reboot=True,
```
smitterl, line 257: I don't think `reboot` should have a default value. I saw it where it's used and `update_vm_default_kernel`, honestly I didn't expect it to do more than the config-update initially. IMO the user of this function must select a value explicitly because the reviewer of the code where this function is used might not know it and `reboot=True` or `False` tells them that this function also reboots. Reboot is such an important and time consuming task that I think it shouldn't be done silently.

virttest/utils_test/__init__.py
```diff
@@ -251,6 +251,72 @@ def update_boot_option(
...
+    timeout="",
+    guest_arch_name="x86_64",
```
smitterl, line 259: This one should also not have a default value, s. comment above. This will avoid failures due to 'copy-paste' which we often use to discover new functions.

virttest/utils_test/__init__.py
rh-jugraham: Below is some experimentation with what --set-kernel can and can't take. So some variation is accepted. However, I wouldn't be sure how to take in "6.13.5-200.fc41" as, for aarch64, there is the 64k option, so wouldn't be sure how that would work, especially if multiple kernels match the same partial string - such as differentiating between "vmlinuz-5.14.0-570.4.1.el9_6.aarch64+64k" and "vmlinuz-5.14.0-570.4.1.el9_6.aarch64" if only "5.14.0-570.4.1.el9_6" is given.

I added "Can use the entire kernel path or just be the full kernel name." to hopefully clarify that it can be either the path or just the name, but I don't know how to use just part of the kernel name without overcomplicating the function and making incorrect assumptions.

```
[root@ampere-mtsnow-altramax-47 /]# grubby --default-kernel
/boot/vmlinuz-5.14.0-570.4.1.el9_6.aarch64+64k

# using full output from grubby
[root@ampere-mtsnow-altramax-47 /]# grubby --set-default="kernel="/boot/vmlinuz-5.14.0-570.4.1.el9_6.aarch64""
The default is /boot/loader/entries/0eae46187a3c4734ad721259b7d1c2b5-5.14.0-570.4.1.el9_6.aarch64.conf with index 1 and kernel /boot/vmlinuz-5.14.0-570.4.1.el9_6.aarch64
[root@ampere-mtsnow-altramax-47 /]# grubby --default-kernel
/boot/vmlinuz-5.14.0-570.4.1.el9_6.aarch64

# using full path
[root@ampere-mtsnow-altramax-47 /]# grubby --set-default="/boot/vmlinuz-5.14.0-570.4.1.el9_6.aarch64+64k"
The default is /boot/loader/entries/0eae46187a3c4734ad721259b7d1c2b5-5.14.0-570.4.1.el9_6.aarch64+64k.conf with index 0 and kernel /boot/vmlinuz-5.14.0-570.4.1.el9_6.aarch64+64k
[root@ampere-mtsnow-altramax-47 /]# grubby --default-kernel
/boot/vmlinuz-5.14.0-570.4.1.el9_6.aarch64+64k

# using just full name 
[root@ampere-mtsnow-altramax-47 /]# grubby --set-default="vmlinuz-5.14.0-570.4.1.el9_6.aarch64"
The default is /boot/loader/entries/0eae46187a3c4734ad721259b7d1c2b5-5.14.0-570.4.1.el9_6.aarch64.conf with index 1 and kernel /boot/vmlinuz-5.14.0-570.4.1.el9_6.aarch64
[root@ampere-mtsnow-altramax-47 /]# grubby --default-kernel
/boot/vmlinuz-5.14.0-570.4.1.el9_6.aarch64

# using partial name (not including vmlinuz)
[root@ampere-mtsnow-altramax-47 /]# grubby --set-default="5.14.0-570.4.1.el9_6.aarch64+64k"
The param 5.14.0-570.4.1.el9_6.aarch64+64k is incorrect
```

rh-jugraham:
@smitterl thanks for the suggestions! I addressed them all, except for 1 I had a question about.

virttest/utils_test/__init__.py
smitterl: To me it looks like `/boot` is not necessary but the full kernel image file name is.

In any case, I believe you don't have to worry about that, `grubby` can return all kernel paths and you can implement a function that helps select one and define it here right next to it and maybe even mention its and `update_vm_...` docstrings that it can be used to select the right kernel path:

```
def get_available_kernel(session, kernel_pattern):
    """
    Looks up all available kernels on the system and returns the first one whose image file name matches a pattern
    ...
    """
    s, output = utils_misc.cmd_status_output("grubby --info ALL", session=session ...)
    ...
    for line in output.split('\n'):
        if line.startswith("kernel=") and re.match(kernel_pattern, line):
            return line.split("=")[1]
    ...
```

rh-jugraham: Sounds good! I did make it so the function returns all of the kernels it found in a list in case there is some reason to find all --> perhaps might make the function more useful.

PR #4073, opened 2025-03-07
add check_boot_config function in libvirt_bios

nanli1:
for https://github.com/autotest/tp-libvirt/pull/6216


virttest/utils_libvirt/libvirt_bios.py
```diff
@@ -29,3 +29,27 @@ def remove_bootconfig_items_from_vmos(osxml):
...
+    """
+    Check /boot/config-$KVER file.
+
+    :param session: vm session.
+    :param test: test object.
+    :param check_list: checking list, default None.
+    :raises: test.fail if checking fails.
+    """
+    if not check_list:
+        check_list = [
```
dzhengfy, line 44: This specific checking list are not suitable in the common utility function as default values. They are better to be in the tp-libvirt case.

nanli1: Thanks, Updated

dzhengfy (approved):

PR #4074, opened 2025-03-07
Fix snapshot name with letter

liang-cong-red-hat:
For snapshot name with letter, func get_image_snapshot return nothing, so fix this scenario


dzhengfy (approved):
LGTM

PR #4076, opened 2025-03-11
avocado_vt/test: enable storage of VM logs

smitterl:
For libvirt type VMs, store the VM log into the test debug directory. The motivating use case is to confirm if a guest has crashed.

### DIFF
--- a/avocado_vt/test.py
+++ b/avocado_vt/test.py
@@ -18,6 +18,7 @@ Avocado VT plugin
 
 import os
 import shlex
+import shutil
 import sys
 
 from avocado.core import exceptions, test
@@ -153,6 +154,14 @@ class VirtTest(test.Test, utils.TestUtils):
             self.__exc_info = sys.exc_info()
             self.__status = self.__exc_info[1]
         finally:
+            if (
+                self.params.get("vm_type") == "libvirt"
+                and self.params.get("store_libvirt_vm_logs", "yes") == "yes"
+            ):
+                source_dir = "/var/log/libvirt/qemu"
+                file_names = os.listdir(source_dir)
+                for file_name in file_names:
+                    shutil.move(os.path.join(source_dir, file_name), self.debugdir)
             # Clean libvirtd debug logs if the test is not fail or error
             if self.params.get("libvirtd_log_cleanup", "no") == "yes":
                 if (
--- a/virttest/shared/cfg/base.cfg
+++ b/virttest/shared/cfg/base.cfg
@@ -827,6 +827,11 @@ virtinstall_extra_args = ""
 #realtime_mlock = on
 #keyboard_layout = en-us
 
+# If to move VM logs - usually located under /var/log/libvirt/qemu/
+# into the test debug directory. This can help confirm the qemu command
+# and crashes.
+store_libvirt_vm_logs = "yes"
+
 # Params for enable/disable Libvirtd debug logs by default it is
 # enabled and retrieve logs for fail and error tests, the logs
 # are saved in default debug dir


bgartzi:
I agree with both the use case and the implementation. Just a couple general questions:
- any reason why moving the file is preferred over copying it?
- This only affects libvirt tests, would by any chance make more sense to call the `store_vm_logs` parameter `libvirt_store_vm_logs` or something similar?

A side note that nobody asked for: it's true this could be implemented as a `Setuper` and register it in the `virttest.env_process._setup_manager`. In that case, I think it should be one of the latest (if not the last) cleanup steps to be executed. It would be also vital to ensure the logs are copied correctly into the test's `debugdir`. Currently, if any error happens during `env_process` `postprocess` and the error is not handled just there (there are some cases), the logs could not be properly copied into the `debugdir`. So until the steps in `postprocess` aren't refactored into proper `Setuper` subclasses to avoid that hypothetical issue, it doesn't make sense to write a `Setuper` for this case.

chloerh (approved):

smitterl:
> I agree with both the use case and the implementation. Just a couple general questions:
> 
> * any reason why moving the file is preferred over copying it?
No. Would you prefer to copy it instead?
> * This only affects libvirt tests, would by any chance make more sense to call the `store_vm_logs` parameter `libvirt_store_vm_logs` or something similar?
> 
You are right I'll update this part.
> A side note that nobody asked for: it's true this could be implemented as a `Setuper` and register it in the `virttest.env_process._setup_manager`. In that case, I think it should be one of the latest (if not the last) cleanup steps to be executed. It would be also vital to ensure the logs are copied correctly into the test's `debugdir`. Currently, if any error happens during `env_process` `postprocess` and the error is not handled just there (there are some cases), the logs could not be properly copied into the `debugdir`. So until the steps in `postprocess` aren't refactored into proper `Setuper` subclasses to avoid that hypothetical issue, it doesn't make sense to write a `Setuper` for this case.
As this is not super urgent, if you prefer this code be converted to follow a given code pattern like the `Setuper` please let me know.

avocado_vt/test.py
```diff
@@ -153,6 +154,14 @@ def setUp(self):
...
             self.__status = self.__exc_info[1]
         finally:
+            if (
+                self.params.get("vm_type") == "libvirt"
+                and self.params.get("store_libvirt_vm_logs", "yes") == "yes"
+            ):
+                source_dir = "/var/log/libvirt/qemu"
+                file_names = os.listdir(source_dir)
+                for file_name in file_names:
+                    shutil.move(os.path.join(source_dir, file_name), self.debugdir)
```
bgartzi, line 164: >> any reason why moving the file is preferred over copying it?
> 
>No. Would you prefer to copy it instead?

It's not that I'd prefer it or not. I just wanted to know if you considered copying it instead. I'm not sure if it would have sense to keep the original file, as long as it is properly put in the new location (which should be the case, and more handy to the framework user). I don't think this is problematic, specially being logs what we are moving, and doing it once the test is finished.

However, I didn't think about that before, but just came to my mind. Are the files under `/var/log/libvirt/qemu` cleaned up between test runs? For example, given test A and B:
- A is configured so `store_libvirt_vm_logs='no'`.
- B is configured so `store_libvirt_vm_logs='yes'`.
- We run B right after A in the same machine.
- Would logs from A be leaked into the debugdir from test B?

virttest/shared/cfg/base.cfg
```diff
@@ -827,6 +827,11 @@ virtinstall_extra_args = ""
 #realtime_mlock = on
 #keyboard_layout = en-us
 
+# If to move VM logs - usually located under /var/log/libvirt/qemu/
+# into the test debug directory. This can help confirm the qemu command
+# and crashes.
+store_libvirt_vm_logs = "yes"
```
bgartzi, line 833: Thanks for updating the configuration parameter name @smitterl. I'm adding this comment here just to mark it as "resolved" right after.

avocado_vt/test.py
smitterl: > However, I didn't think about that before, but just came to my mind. Are the files under `/var/log/libvirt/qemu` cleaned up between test runs? For example, given test A and B:
> 
> * A is configured so `store_libvirt_vm_logs='no'`.
> * B is configured so `store_libvirt_vm_logs='yes'`.
> * We run B right after A in the same machine.
> * Would logs from A be leaked into the debugdir from test B?

Yes, there could be some leak because for each start the VM log is appended to. There might be settings to avoid this but I think moving the file instead of copying can help avoid this.

bgartzi (approved):
Thanks @smitterl, it looks good

PR #4080, opened 2025-03-13
__init__.py: check if tool exist before install

aliang123:
Check if stress tool exist before install it.
id:3428


XueqiangWei:
@aliang123 Some checks were not successful, please resolve the issue. Thanks.

PR #4085, opened 2025-03-18
storage: Fix the external data file backup dir issue

ybduan:
The backup directory depends on the setting of the parameter backup_dir,
we didn't add backup_dir as a parameter of the external data file, then
the root_dir is used because backup_dir = "", so adding backup_dir param
for the external data file can fix this.

ID: 3516

### DIFF
--- a/virttest/storage.py
+++ b/virttest/storage.py
@@ -970,6 +970,7 @@ class QemuImg(object):
                 "image_format": "raw",
                 "image_size": image_size,
                 "image_raw_device": "yes",
+                "backup_dir": params.get("backup_dir", "")
             }
         )
         return cls(data_file_params, root_dir, "%s_data_file" % tag)


ybduan:
Could you please help review this PR? Thanks! @YongxueHong @XueqiangWei

YongxueHong:
Hi @ybduan 
Could provide more about this RP? I'm sorry, I didn't get your purpose clear.
Thanks.

YongxueHong:
Hi @zhencliu 
I would appreciate it if you could have a look, thanks.

zhencliu:
hi @ybduan , I am afraid this fix could not solve your problem, you can set the cartesian param 'backup_dir' to a dir with large free space, then your issue should be solved. 
root_dir here is not the backup dir, it's used for make a backup dir when the backup_dir is not set. So when the param backup_dir is not set, the variable backup_dir = os.path.join(root_dir, backup_dir), it still uses root_dir

YongxueHong:
> hi @ybduan , I am afraid this fix could not solve your problem, you can set the cartesian param 'backup_dir' to a dir with large free space, then your issue should be solved. root_dir here is not the backup dir, it's used for make a backup dir when the backup_dir is not set. So when the param backup_dir is not set, the variable backup_dir = os.path.join(root_dir, backup_dir), it still uses root_dir

Hi @ybduan 
Yeah, from my understanding of the changes, I did not see any impacts you want based on the commit message, so that is why I need more information.
Thanks.

ybduan:
I updated this PR according to our off-line discussion, please help review it again, thanks! @YongxueHong @zhencliu

zhencliu:
hi @ybduan , the code is good to me, would you change the commit head to something like "storage: Fix the external data file backup dir issue" and then explain a little bit on the solution in the commit body? The backup dir depends on the setting of the param backup_dir, we didn't add backup_dir as a param of the external data file, then the root_dir is used 
 because backup_dir = "", so adding backup_dir  param for the external data file can fix this

zhencliu (approved):

YongxueHong (approved):
LGTM.

PR #4087, opened 2025-03-19
utils_sys: moved update_boot_option from utils_test

rh-jugraham:
Depends on: https://github.com/avocado-framework/avocado-vt/pull/4070

Moved update_boot_option from utils_test to utils_sys so that __init__.py is not as long.


rh-jugraham:
Will return later, as this change will require more effort than initially suspected.

PR #4094, opened 2025-03-27
utils_net:Retry when not getting default gateway

chloerh:
Test result:

```
 (1/1) type_specific.local.virtual_network.connectivity_check.bridge_interface.linux_br.default: STARTED
 (1/1) type_specific.local.virtual_network.connectivity_check.bridge_interface.linux_br.default: PASS (68.59 s)
```

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -4059,7 +4059,12 @@ def get_host_iface():
 
 
 def get_default_gateway_json(
-    iface_name=False, session=None, ip_ver="ipv4", force_dhcp=False, target_iface=None
+    iface_name=False,
+    session=None,
+    ip_ver="ipv4",
+    force_dhcp=False,
+    target_iface=None,
+    timeout=30,
 ):
     """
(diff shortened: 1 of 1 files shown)


virttest/utils_net.py
```diff
@@ -4080,35 +4085,40 @@ def get_default_gateway_json(
...
+            non_multi = [x for x in default_route_list if x.get("dev") == target_iface]
+            multi = [
+                p
+                for x in default_route_list
+                if "nexthops" in x
+                for p in x["nexthops"]
+                if p.get("dev") == target_iface
+            ]
+            default_route_list = non_multi + multi
+        LOG.debug(f"default_route_list: {default_route_list}")
```
smitterl, line 4114: IIUC it's the first time we observe this so having to retry shouldn't be necessary but important here to stabilize our test runs. Therefore, would it make sense to LOG.warning or something similar whenever there wasn't a default route?

chloerh: Updated

PR #4097, opened 2025-04-01
utils_disk: fix the verbose value in umount function

hellohellenmao:
The available value for verbose in this fuction is a bool, but not 'verbose'

ID: 3635

### DIFF
--- a/virttest/utils_disk.py
+++ b/virttest/utils_disk.py
@@ -135,7 +135,7 @@ def mount(src, dst, fstype=None, options=None, verbose=False, session=None):
     return process.system(cmd, verbose=verbose) == 0
 
 
-def umount(src, dst, fstype=None, verbose=False, session=None):
+def umount(src, dst, fstype=None, verbose=True, session=None):
     """
     Umount src from dst, if src really mounted under dst.
 


virttest/utils_disk.py
```diff
@@ -146,7 +146,7 @@ def umount(src, dst, fstype=None, verbose=False, session=None):
 
     :return: if unmounted return True else return False
     """
-    mounted = is_mount(src, dst, fstype, verbose=verbose, session=session)
+    mounted = is_mount(src, dst, fstype, verbose=True, session=session)
```
zhencliu, line 149: hi @hellohellenmao , value of the verbose here is passed by the umount, so why not calling umount with verbose=True in your test case? Then you can get the same result.

Btw, does that matters if the umount function prints details?

hellohellenmao: Yes, I need it for debugging, and if the value is False, when umounting failed I just do not know the mounted details which is needed for my debugging, I need to do it manually to login to the guest to check the details.

YongxueHong: Hi @hellohellenmao 
I did not get your point, I think that  you could get the details by the way @zhencliu mentioned, but we set the `verbose=True` mandatory here. 
From my understanding, your solution is the same to pass the` verbose=True` with calling `umount(src, dst, fstype=None, verbose=False, session=None)`

hellohellenmao: Sorry that mixing you up, acctually I agree that to pass  verbose=True with calling umount, I am just keeping on failing with `ssh: connect to host github.com port 22: Connection timed out` while pushing the update.
And my last comment to zhenchao is for the answer to "Btw, does that matters if the umount function prints details?"

hellohellenmao:
@YongxueHong @zhencliu Updated with thanks, and could you please help to review again?

virttest/utils_disk.py
```diff
@@ -135,7 +135,7 @@ def mount(src, dst, fstype=None, options=None, verbose=False, session=None):
     return process.system(cmd, verbose=verbose) == 0
 
 
-def umount(src, dst, fstype=None, verbose=False, session=None):
```
YongxueHong, line 138: Hi @hellohellenmao 
I suggest that you could customize the argument for the test case, rather than change the default here since this change will affect the previous behaviors and we need to involve more reviewers to evaluate it.
In order to reduce the influence and satisfy your requirements, I suggest specifying `verbose=True` in your scenario.

hellohellenmao: Ha, make sense. Will close this PR, thanks.

PR #4100, opened 2025-04-04
Integrate with cartconf as standalone Cartesian config

pevogam:
This pull requests replaces the included somewhat standalone cartesian_config module with separately maintained and a completely standalone cartconf avocado varianter.


pevogam:
@clebergnu @richtja @harvey0100 Does anyone of you think they can assist me to establish a Fedora-friendly RPM for the [cartconf](https://github.com/avocado-framework/cartconf) varianter plugin? I already have a spec file but I assume something like that should be distributed in a similar way to aexpect in order to be defined as a proper installable dependency for integrating and testing with VT here.

PR #4090, opened 2025-03-21
env_process: Refactor EGD setup

bgartzi:
Replace the steps setting and cleaning EGD up into a Setuper that is registered into the setup_manager in env_process.

This is a patch from a larger patch series refactoring the env_process preprocess and postprocess functions. In each of these patches, a pre/post process step is identified and replaced with a Setuper subclass so the following can finally be met:
    - Only cleanup steps of successful setup steps are run to avoid possible environment corruption or hard to read errors.
    - Running setup/cleanup steps symmetrically during env pre/post process.
    - Reduce explicit pre/post process function code length.

This patch shouldn't introduce any regression or breaking change.
With `--testcase=boot --rng_type=virtio_rng.rng_egd`:
```
 (1/1) Host_RHEL.m9.u5.ovmf.virtio_rng.rng_egd.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.x86_64.io-github-autotest-qemu.boot.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.virtio_rng.rng_egd.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.x86_64.io-github-autotest-qemu.boot.q35:  PASS (73.16 s)
```

ID: 2941

PR #4103, opened 2025-04-11
Env process refactoring kill tail threads

bgartzi:
The order in which `aexpect.kill_tail_threads` was called during env_process.postprocess changed while we were refactoring it. Now, tail threads were killed before the IPSniffer or other processes using Tail objects were cleaned up.
Calling `aexpect.kill_tail_threads` when it was supposed to: right after the IPSniffer is cleaned up.

ID: 3698

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -11,7 +11,6 @@ import sys
 import threading
 import time
 
-import aexpect
 import six
 from aexpect import remote
 from avocado.core import exceptions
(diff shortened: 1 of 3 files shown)


bgartzi:
Hi @YongxueHong,

I'm sending yet another refactoring patch. The special thing about this one is that I should have sent it before, but I missed it. That's why it doesn't depend on the currently ongoing https://github.com/avocado-framework/avocado-vt/pull/4090.

I'm also adding some slight change to make the CI static checks happy.

Could you have a look?

Thanks! :)

Edit: let's see if I'm able to pass CI checks this time...

YongxueHong (approved):
LGTM.

PR #4105, opened 2025-04-14
Draft: DON'T MERGE env_process: add log info

smitterl:
1. Log all pre/postprocess function entries
2. Log moment before `undefine` is called directly.


smitterl:
avocado-vt gets stuck both in preprocess as well as postprocess, we believe as pointed out by @ldoktor we need to look into 

* session.close() (in finally)
* cleanup_serial_console()

for more information

PR #4116, opened 2025-04-25
utils_backup.py: use f-string formatting instead of % formatting

meinaLi:
It is recommended to use f-string (formatted string literal), because it is clearer, more intuitive, easier to read, and less error-prone than the old % formatting method. Or sometimes cases will get error "not all arguments converted during string formatting".


virttest/utils_backup.py
```diff
@@ -339,10 +339,11 @@ def dump_data_to_file(data_map, source_file_path, target_file_path, driver="qcow
     :param driver: The image format of the source file
     """
     with open(target_file_path, "w"):
-        cmd = "qemu-io -f %s -c \"read -vC %s %s\" %s -r -U  | sed '$d'"
         for entry in data_map:
-            cmd %= (driver, entry["start"], entry["length"], source_file_path)
-            cmd += ">> %s" % target_file_path
+            cmd = (
+                f"qemu-io -f {driver} -c 'read -vC {entry['start']} {entry['length']}'"
```
chloerh, line 344: Seems there're more than one pair of single quotes, but looks like it's working well, so it's ok to me.

chloerh (approved):

PR #4119, opened 2025-05-06
virttest/utils_zchannel: add function to select first dasd

smitterl:
Our test environments now use DASD only for passthrough tests. Therefore it's safe to select the first DASD in Use. Add function for convience.


smitterl:
Change covered in https://github.com/autotest/tp-libvirt/pull/6305 with test results there.

smitterl:
@liang-cong-red-hat please help review

virttest/utils_zchannels.py
```diff
@@ -109,6 +109,19 @@ def get_device(self, devid=None):
...
+
+        Requires get_info() to be called first.
+        :return: Device info
+        """
+        used = [x for x in self.devices if x[self.HEADER["Use"]] == "yes"]
+        for device in used:
+            if "3390" in device[self.HEADER["DevType"]]:
+                return device
+        return None
+
```
liang-cong-red-hat, line 124: I think we could do like below, that only has one loop:
for device in self.devices:
        if device[self.HEADER["Use"]] == "yes" and "3390" in device[self.HEADER["DevType"]]:
            return device
 return None

smitterl: > I think we could do like below, that only has one loop:
> 
> ```
> for device in self.devices:
>     if device[self.HEADER["Use"]] == "yes" and "3390" in device[self.HEADER["DevType"]]:
>         return device
> return None
> ```

OK, I assume you mean that's for better performance? Updated.

chunfuwen (approved):
lgtm

liang-cong-red-hat (approved):
LGTM

PR #4124, opened 2025-05-19
Remove virt-manager as dependence when bootstraping virttools

chunfuwen:
Remove virt-manager as dependence when bootstraping virttools
virt-manager is UI version of virt-install, and virt-tools test cases doesn't really use it


chunfuwen:
@smitterl  please help double check it

smitterl (approved):
FYI, we still have virt-manager-common in our repos. But I believe retaining virt-install, virt-clone and virt-xml explicilty makes sense because the actual commands. 

Thanks for this!

PR #4128, opened 2025-05-22
utils_net:Fix index error when trying to get defaut gateway

chloerh:
Set default value when trying to get metrix of a gateway to avoid clearing up default_route_list which leads to IndexError

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -4114,7 +4114,7 @@ def get_default_gateway_json(
         default_route_list = [
             x
             for x in default_route_list
-            if x.get("metric")
+            if x.get("metric", float("inf"))
             == min([y.get("metric", float("inf")) for y in default_route_list])
         ]
 


chloerh:
Before fix:

```
default_route_list = [{'gateway': 'fe80:b20',
  'dev': 'eno1',
  'weight': 1,
  'flags': []},
 {'gateway': 'fe80:2920',
  'dev': 'eno1',
  'weight': 1,
  'flags': []}]

      [
            x
            for x in default_route_list
            if x.get("metric")
            == min([y.get("metric", float("inf")) for y in default_route_list])
        ]

[]
```

After fix
```
      [
            x
            for x in default_route_list
            if x.get("metric", float("inf"))
            == min([y.get("metric", float("inf")) for y in default_route_list])
        ]

[{'gateway': 'fe80:b20', 'dev': 'eno1', 'weight': 1, 'flags': []},
 {'gateway': 'fe80:2920', 'dev': 'eno1', 'weight': 1, 'flags': []}]
```

dzhengfy:
Could you please provide the test result on x86_64?

dzhengfy:
Test on arm before the fix:
rhel.virtual_network.passt.attach_detach.ip_portfw.attach_detach.non_root_user
list index out of range

Test on arm afte the fix:
 (1/1) type_specific.io-github-autotest-libvirt.virtual_network.passt.attach_detach.ip_portfw.default.attach_detach.non_root_user: FAIL: Host default ipv6 gateway not consistent with vm. (41.06 s)

The list index issue disappears.

dzhengfy (approved):
LGTM

PR #4133, opened 2025-05-28
utils_package: Add '--transient' for image mode

qiankehan:
For the image mode of RHEL/CentOS/Fedora, option '--transient' is needed to install or uninstall a package.

### DIFF
--- a/virttest/utils_package.py
+++ b/virttest/utils_package.py
@@ -2,6 +2,7 @@
 Package utility for manage package operation on host
 """
 
+import os
 import logging
 
 import aexpect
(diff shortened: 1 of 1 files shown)


smitterl:
@qiankehan Is this PR ready for review? The title still says "[draft]". Also, Please fix the static checks.

smitterl:
> Since the guest may reboot during a testing, 'transient' cannot make the pkgs installed after reboot. This PR need to be updated.

Based on this comment I set the PR into Draft state.

qiankehan:
v2: Implement the package manager **rpm-ostree** for bootc system. I will update the test results TMR.

qiankehan:
For v2, the package remove don't work well if the package is in the base layer of bootc image. For package remove, these are some problems to overcome:
- How to make it work immediately (without reboot): `dnf remove --transient {pkg}` is a prossible way
- How to make it work after reboot:
  - How to make it work if the package is in the overlayer: `rpm-ostree uninstall -y {pkg} && rpm-ostree apply-live --allow-replacement` should work
  - How to make it work if the package is in the base layer:  I tried `rpm-ostree override remove <pkg> && reboot`. But it cannot work

qiankehan:
v3: Update the note for the package removal failure.

qiankehan:
@smitterl Hi Sebas, this PR is ready for review now.

PR #4149, opened 2025-06-23
utils_test: Decouple package check from installation

YongxueHong:
Previously, the logic within the Stress class for checking if the
`stress_package` exists was coupled with its installation logic.
This was done to support installing the package from a source file
if it wasn't found in the repository.

This commit decouples the existence check from the installation steps
for the `stress` package to make the code more robust, explicit, and
easier to maintain. The `install` method in the `Stress` class now
first checks if the package is already installed and exits if true.
Otherwise, it proceeds with repository or source installation attempts.

Note: utils_package.package_install still performs its own internal
check. Refactoring that function to fully separate concerns is out
of scope for this commit and will be addressed separately.
ID: 3428

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -2554,20 +2554,24 @@ class Stress(object):
                 raise exceptions.TestError(
                     "Installing dependency packages for" " %s failed" % self.stress_type
                 )
-        if self.stress_install_from_repo and self.stress_package:
-            # Install the stress package from existing repos
-            # If succeed, no need to download stress src any more
-            if not utils_package.package_install(
-                self.stress_package, session=self.session, timeout=60
-            ):
-                self.stress_install_from_repo = False
-                LOG.debug(
-                    "Fail to install stress tool via repo and "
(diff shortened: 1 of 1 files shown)


mcasquer:
@YongxueHong thanks for the patch, but either the commit or the description are too general and not mentioning that the changes apply to the stress package, IMHO I think these should be updated

YongxueHong:
> @YongxueHong thanks for the patch, but either the commit or the description are too general and not mentioning that the changes apply to the stress package, IMHO I think these should be updated

Hi @mcasquer it has been updated, pls help review it, thanks a lot.

PaulYuuu (approved):

YongxueHong:
Hi @aliang123 @fbq815 @cliping @YanhuiMa @Yingshun
Would you help review it? Thanks in advance.

YongxueHong:
Hi guys,
We have received approvals from more than two reviewers, so this now meets the requirements for merging.
Let's merge it. Please comment if you have any concerns in the future.
Thank you.

PR #4153, opened 2025-06-30
libvirt_memory: fix the incorrect parameter

dzhengfy:
The original parameter is used for remained memory, not for allocation. So I rename the parameter.


dzhengfy:
Some cases in tp-libvirt will also need to update according to this PR.  I will create them.

Yingshun (approved):

PR #4160, opened 2025-07-03
netkvm: add nic_index argument to set/get nic params

heywji:
Tell them which NIC to use (script picks NIC 0 if we don’t say). This patch uses that number instead of always 0 in the command and logs, and the old code still works fine.

ID: 4042
Signed-off-by: wji <wji@redhat.com>

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -4543,7 +4543,7 @@ def get_msis_and_queues_windows(params, vm, timeout=360):
     return _get_msis_queues_from_traceview_output(output)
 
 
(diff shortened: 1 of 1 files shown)


heywji:
@leidwang Hi Leidong, could you please help review this patch as well?

heywji:
@PaulYuuu Hi Yihuang, could you help review it as well?

heywji:
@leidwang @vivianQizhu Hi Leidong and Qianqian, Could you also help review this patch?

PR #4182, opened 2025-07-16
Reduction of Cirrus CI tasks

richtja:
This will reduce the number of CI task which we run in Cirrus CI. It is needed for avoiding the Cirrus CI's computation limit in the future. It removes the tests with avocado 92lts version, which is not used anymore. Also, it removes the `setup.py` installation method which is planned to be removed soon, because the `setup.py` has been deprecated in python.


PaulYuuu (changes requested):
Thank you @richtja, reduce the test matrix is perfect. However, I have a concern, cirrus CI still using fedora-35, which is EOL since 2022: https://docs.fedoraproject.org/en-US/releases/eol/

Maintain the Cirrus CI is not only for matrixes, but should also keeps the runner image up to date. We can use `quay.io/avocado-framework/avocado-ci-fedora-40` and install required pkgs in this yaml file, or build a new image from `contrib/containers/ci/fedora-35.docker`, update the dockerfile to use fedora40/41, what do you think?

richtja:
> Thank you @richtja, reduce the test matrix is perfect. However, I have a concern, cirrus CI still using fedora-35, which is EOL since 2022: https://docs.fedoraproject.org/en-US/releases/eol/
> 
> Maintain the Cirrus CI is not only for matrixes, but should also keeps the runner image up to date. We can use `quay.io/avocado-framework/avocado-ci-fedora-40` and install required pkgs in this yaml file, or build a new image from `contrib/containers/ci/fedora-35.docker`, update the dockerfile to use fedora40/41, what do you think?

Hi @PaulYuuu, great idea. I have built a new container image with fedora-40 and pushed it to https://quay.io/repository/avocado-framework/avocado-vt-ci-fedora-40. Also, I updated this PR accordingly, please take a look.

PaulYuuu (approved):

PR #4196, opened 2025-07-28
Extend root to support check root elememt attributes

liang-cong-red-hat:
Since the findall only support fiding child elements of root, so extend a new root to support finding old root attributes


smitterl (changes requested):
Please update header and message of your commit and PR according to https://github.com/avocado-framework/avocado/pull/6215/files

Please, add an example of what you're trying to solve in the commit message, too.

dzhengfy (approved):
LGTM

PR #4198, opened 2025-07-28
Add nfs4 type to nfs umount

liang-cong-red-hat:
Since https://github.com/avocado-framework/avocado-vt/pull/4098, using findmnt to find mount points. It is file type sensitive, so add nfs4 to support nfs4 type.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unmount operation to support both NFS and NFSv4 protocols, enhancing compatibility during unmounting.

### DIFF
--- a/virttest/nfs.py
+++ b/virttest/nfs.py
@@ -230,7 +230,7 @@ class Nfs(object):
         """
         Umount the given mount point.
         """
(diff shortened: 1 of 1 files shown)


dzhengfy (approved):
Thanks for fixing it

PR #4215, opened 2025-08-11
utils_sys: add utility to check image mode

dzhengfy:
## Summary by CodeRabbit

* **New Features**
  * Automatic detection of system "image mode" vs "package mode" to adapt behavior across system types.
  * Works with both local execution and remote/session connections for reliable checks.
  * Emits debug messages indicating the detected mode to aid troubleshooting.

### DIFF
--- a/virttest/utils_sys.py
+++ b/virttest/utils_sys.py
@@ -74,6 +74,37 @@ def get_host_bridge_id(session=None):
     return host_bridges if host_bridges else []
 
 
+def is_image_mode(session=None):
+    """
+    Check if current OS is in image mode or package mode
+
+    :param session: aexpect session
+    :return: boolean, True for image mode, otherwise False
+    """
+    check_command = "bootc status"
+    err_msg = "image:\\s+null"
+
+    def _check_output(status, output):
+        LOG.debug("status: %s\noutput:\n%s", status, output)
+        if not status:
+            result = re.search(err_msg, output) is None
+        else:
+            result = False
+        LOG.debug(
+            "Detected %s mode by command '%s'",
+            "image" if result else "package",
+            check_command,
+        )
+        return result
+
+    if session:
+        status, output = session.cmd_status_output(check_command)
+        return _check_output(status, output)
+    else:
+        ret = process.run(check_command, shell=True, verbose=True, ignore_status=True)
+        return _check_output(ret.exit_status, ret.stdout_text)
+
+
 def get_pids_for(process_names, sort_pids=True, session=None):
     """
     Given a list of names, retrieve the PIDs for


dzhengfy:
Python 3.12.11 (main, Jun 12 2025, 00:00:00) [GCC 14.2.1 20250110 (Red Hat 14.2.1-8)] on linux
>>> from virttest import utils_sys
No python imaging library installed. Screendump and Windows guest BSOD detection are disabled. In order to enable it, please install python-imaging or the equivalent for your distro.
>>> utils_sys.is_image_mode()
True
>>> 
# df -m
Filesystem                            1M-blocks  Used Available Use% Mounted on
/dev/mapper/rhel_dell--r640--036-root     71616 12615     59002  18% /sysroot
**composefs**                                    17    17         0 100% /
devtmpfs                                      4     0         4   0% /dev
tmpfs                                     96229     0     96229   0% /dev/shm
tmpfs                                     38492     2     38490   1% /run

virttest/utils_sys.py
```diff
@@ -74,6 +75,32 @@ def get_host_bridge_id(session=None):
...
+    """
+    Check if current OS is in image mode or package mode
+
+    :param check_command: str, the command to check the mode
+    :param session: aexpect session
+    :return: boolean, True for image mode, otherwise False
+    """
+    if session:
+        status = session.cmd_status(check_command)
+        ret = status == 0
```
liang-cong-red-hat, line 88: here only execute cmd rpm-ostree, the return code would be 1, the cmd itself could not run only itself. I suggest using "which rpm-ostree". ("rpm-ostree -v" will also report error)

liang-cong-red-hat (approved):
LGTM

virttest/utils_sys.py
dzhengfy: thanks for finding the mistake. Updated. @liang-cong-red-hat

PaulYuuu (changes requested):
Sorry, but I would say this mixed the definition between image mode and ostree-based system.

Even if there is a package mode system, it can install rpm-ostree package as well.
```bash
rpm-ostree status
error: This system was not booted via libostree; found container=podman environment variable.
Currently, most rpm-ostree commands only work on ostree-based host systems.
```

Bootc is bootc, ostree is ostree, fedora silverblue system is also a atomic system(ostree-based), does silverblue is image mode?

BTW, rpm-ostree may be dropped in the future in image mode, please refer to https://docs.fedoraproject.org/en-US/bootc/rpm-ostree/ and 

Image

So before making new image mode implementation, please know what's the correct solution from https://docs.fedoraproject.org/en-US/bootc/ and https://bootc-dev.github.io/bootc/

virttest/utils_sys.py
```diff
@@ -74,6 +75,33 @@ def get_host_bridge_id(session=None):
...
+    :param session: aexpect session
+    :return: boolean, True for image mode, otherwise False
+    """
+    if session:
+        check_command = "which %s" % check_command
+        status = session.cmd_status(check_command)
+        ret = status == 0
+    else:
+        try:
+            utils_path.find_command(check_command)
```
qiankehan, line 92: I don't think it is a reliable way to check if it is image mode because rpm-ostree could be installed in image mode as well. The better way is via bootc:
- No bootc cmd --> package mode
- With boot, then check `bootc status`
  - Return **System is not deployed via bootc** --> package mode
  - Else --> image mode

virttest/utils_sys.py
```diff
@@ -74,6 +75,33 @@ def get_host_bridge_id(session=None):
     return host_bridges if host_bridges else []
 
 
+def is_image_mode(check_command="rpm-ostree", session=None):
```
qiankehan, line 78: check_command parameter is not needed because the user don't need to know how to check it

cliping (approved):

liang-cong-red-hat (approved):
only one comment. others LGTM

virttest/utils_sys.py
```diff
@@ -74,6 +74,37 @@ def get_host_bridge_id(session=None):
...
+    :param session: aexpect session
+    :return: boolean, True for image mode, otherwise False
+    """
+    check_command = "bootc status"
+    err_msg = "image:\\s+null"
+
+    def _check_output(status, output):
+        LOG.debug("status: %s\noutput:\n%s", status, output)
+        if not status:
+            result = re.search(err_msg, output) == None
```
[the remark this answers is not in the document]
dzhengfy: Thanks,  I forgot to update this line. Updated

chunfuwen (approved):
lgtm

PR #4219, opened 2025-08-13
RFC: add pre-commit hooks and commit linting configuration

PaulYuuu:
Add comprehensive pre-commit hooks including security checks, code
formatting, and commit message validation. Includes gitleaks for secret
detection, black and isort for Python formatting, and commitlint for
conventional commit enforcement.

ID: 4267


PaulYuuu:
Usage:
- Install pre-commit: `pip install pre-commit`
- Install git hook: `cd avocado-vt; pre-commit install`
- Install optional commit-msg hook to enable commitlint: `pre-commit install -t commit-msg`
- Run on all files(How this PR did): pre-commit run --all-files
- Auto-run on commits: Hooks execute automatically on git commit, see below demo
```yaml
touch demo_file
echo "pre-comit   " >> demo_file
git add demo_file
git commit
check for added large files..............................................Passed
check python ast.....................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
detect private key.......................................................Passed
fix end of files.........................................................Passed
mixed line ending........................................................Passed
don't commit to branch...................................................Passed
trim trailing whitespace.................................................Failed
- hook id: trailing-whitespace
- exit code: 1
- files were modified by this hook

Fixing demo_file

Detect hardcoded secrets.................................................Passed
isort................................................(no files to check)Skipped
black................................................(no files to check)Skipped
codespell................................................................Passed

# the hook fixed it automatically, let's add the file again and commit
git add demo_file
git commit
check for added large files..............................................Passed
check python ast.....................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
detect private key.......................................................Passed
fix end of files.........................................................Passed
mixed line ending........................................................Passed
don't commit to branch...................................................Passed
trim trailing whitespace.................................................Passed
Detect hardcoded secrets.................................................Passed
isort................................................(no files to check)Skipped
black................................................(no files to check)Skipped
codespell................................................................Passed
commitlint...............................................................Failed
- hook id: commitlint
- exit code: 1

⧗   input: chore(pre-commit-demo): test if this commit message is match the commitlint rules

commit body for test. xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
✖   header must not be longer than 72 characters, current length is 81 [header-max-length]
✖   message must be signed off [signed-off-by]
✖   message must have `Signed-off-by:` trailer [trailer-exists]

✖   found 3 problems, 0 warnings
ⓘ   Get help: https://avocado-framework.readthedocs.io/en/latest/guides/contributor/chapters/styleguides.html#commit-style-guide
```

PaulYuuu:
Another topic is keep using black + isort, or using the popular tool [ruff](https://docs.astral.sh/ruff/), we can use this unique tool to manage style/lint.

richtja:
> > Hi @PaulYuuu, thank you for introducing this. I definatly see a value in having pre-commit hooks for our static checks. But IIUIC this removes the `avocado-static-checks` and uses the dedicated static checkers like black, isort, etc. in pre-commit without the current configuration. IMO this is not the right solution because the `avocado-static-checks` submodule was created to generalize the static checks for whole avocado organization and I see this as step back.
> 
> Hello @richtja, my initial plan is to enable pre-commit in avocado-vt, once it's stable, I will announce it to avocado-static-checks repo. pre-commit also supports local script integration, in tp-qemu, we have
> 
> ```yaml
> - repo: local
>   hooks:
>   - id: cfg-lint
>     entry: .ci/cfg-lint-check.py
>     files: (^|/)tests/cfg/.+\.cfg$
>     language: python
>     name: check cfg
>     require_serial: true
> ```
> 
> This means we can combine it with avocado-static-checks, but here I would explain my concern why I skipped it.
> 
>     * Local entry requires tools(black/isort) be installed in advance, for local dev environment, if the contributor installed other version black/isort, they cannot align the code quality between VT itself and their local env.

I don't see this as an problem, we have [requirements.txt](https://github.com/avocado-framework/avocado-static-checks/blob/main/requirements.txt) file to specify the needed versions.

> 
>     * black/isort pre-commit can pin versions, and pre-commit will install them in venv, so users can save time this part and easy to maintain their develop env.

To be honest, I don't see a big advantage in this when we talk only about static checkers which won't pollute develop env. But maybe I am wrong here.

> 
>     * Once this pre-commit config file in avocado-static-checks repo, we only need to maintain this unified file rather than maintain most static scripts.

Yes, that would be ideal solution, unfortunatly I am not shure if pre-commit config file has the same features as the static scripts. One of the very used feature on AAutils and Avocado is possiblity to have diferent configuration for diferent parts of codebase.

> 
>     * When we doing `git commit`, pre-commit hooks only monitor those modified files, avocado-static-checks will check all files in the repo.

I agree, this could be opimized.

> 
> 
> > I would prefer to use `avocado-static-checks` inside the pre-commit hook so we keep the checks aligned between different avocado projects. Or even better, introduce this in `avocado-static-checks` repo so it could be easily consumed by others. What do you think?
> 
> I totally agree maintain pre-commit config file in avocado-static-checks repo, we can link the config file from submodule to the repo root dir, a more sound approach is define our own `.pre-commit-hooks.yaml` in avocado-static-checks, and other avocado repos can use these hooks directly, but this will make the avocado-static-checks repo complex, because we need to [create new hooks](https://pre-commit.com/#new-hooks).
> 
> To summary:
> 
>     * Enable pre-commit as experimental function in VT, and move the config file to avocado-static-checks in the future

I wouldn't do this because it will again separate Avocado-vt checks from Avocado and Autils.

> 
>     * Or directly introduce this config file in static checks repo, VT link the config file from submodule to root dir

Yes, this is a possible solution, I just not sure If we can keep the `avocado-static-checks` features like having different `.pylintrc` files for different parts of code-base. Maybe for now we can introduce pre-commit to `avocado-static-checks` and run the `avocado-static-checks` scripts from pre-commint. What do you think?

> 
>     * Or Implement own `.pre-commit-hooks.yaml` in avocado-static-checks, but this requires the repo be a [real project](https://pre-commit.com/#python)

I don't know how difficult it would be, and I would prefer a simple solution for now which would reuse the current solution with advantage of pre-commit hook.

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "avocado-static-checks"]
 	path = avocado-static-checks
-	url = https://github.com/avocado-framework/avocado-static-checks
+	url = https://github.com/PaulYuuu/avocado-static-checks
+	branch = pre-commit
```
PaulYuuu, line 4: This is a revision, taking my personal `avocado-static-checks` branch as an example, we can store `.pre-commit-config.yaml` and/or `commitlint.config.mjs` in `avocado-static-checks`, then link them in the current root, and pre-commit GitHub Action can use it without error by adding 
```
with:
    submodules: recursive
```

PaulYuuu:
> I don't see this as an problem, we have [requirements.txt](https://github.com/avocado-framework/avocado-static-checks/blob/main/requirements.txt?rgh-link-date=2025-08-14T14%3A40%3A47Z) file to specify the needed versions.
> 
`requirements.txt` requires users handle by themselves, once the file is updated, them have to run `make requirements` before check. And another thing is these steps are not running in an isolate environment, if users have other projects using different isort/black etc. versions, This will pollute their environment.

> To be honest, I don't see a big advantage in this when we talk only about static checkers which won't pollute develop env. But maybe I am wrong here.
> 
As I mentioned above, using a isolate venv for style/lint check can avoid modifying the system, but if they ar e already using venv, then this is indeed safe.

> Yes, that would be ideal solution, unfortunatly I am not shure if pre-commit config file has the same features as the static scripts. One of the very used feature on AAutils and Avocado is possiblity to have diferent configuration for diferent parts of codebase.
>
Please see my above comment. Now I am using my `avocado-static-checks` repo as an example, if each repo has its configuration files, like sepll.ignore, .pylintrc, .isort.cfg, that is possible, we can store those files in project root, pre-commit can use them, for common config files, add them in `avocado-static-checks` and link to the project root.

> > * Or directly introduce this config file in static checks repo, VT link the config file from submodule to root dir
> 
> Yes, this is a possible solution, I just not sure If we can keep the `avocado-static-checks` features like having different `.pylintrc` files for different parts of code-base. Maybe for now we can introduce pre-commit to `avocado-static-checks` and run the `avocado-static-checks` scripts from pre-commint. What do you think?
>
That's definitely we can. in the `.pre-commit-config.yaml` files, we can add more than one `pylint` hooks, using different name and config file to distinguish them, and use like `files`, `exclude`, `types`, `exclude_types` filters.

> > ```
> > * Or Implement own `.pre-commit-hooks.yaml` in avocado-static-checks, but this requires the repo be a [real project](https://pre-commit.com/#python)
> > ```
> 
> I don't know how difficult it would be, and I would prefer a simple solution for now which would reuse the current solution with advantage of pre-commit hook.
Agree, this is a relatively complex implementation, we can ignore it.

richtja:
> > I don't see this as an problem, we have [requirements.txt](https://github.com/avocado-framework/avocado-static-checks/blob/main/requirements.txt?rgh-link-date=2025-08-14T14%3A40%3A47Z) file to specify the needed versions.
> 
> `requirements.txt` requires users handle by themselves, once the file is updated, them have to run `make requirements` before check. And another thing is these steps are not running in an isolate environment, if users have other projects using different isort/black etc. versions, This will pollute their environment.
> 
> > To be honest, I don't see a big advantage in this when we talk only about static checkers which won't pollute develop env. But maybe I am wrong here.
> 
> As I mentioned above, using a isolate venv for style/lint check can avoid modifying the system, but if they ar e already using venv, then this is indeed safe.

IMO when you develop multiple projects on your system it is good practice and developer responsibility to have some environment management like venv, therefore I think we don't need to supplement it. But I understand the additional value which an isolated venv for style/lint check will bring. 

> 
> > Yes, that would be ideal solution, unfortunatly I am not shure if pre-commit config file has the same features as the static scripts. One of the very used feature on AAutils and Avocado is possiblity to have diferent configuration for diferent parts of codebase.
> 
> Please see my above comment. Now I am using my `avocado-static-checks` repo as an example, if each repo has its configuration files, like sepll.ignore, .pylintrc, .isort.cfg, that is possible, we can store those files in project root, pre-commit can use them, for common config files, add them in `avocado-static-checks` and link to the project root.
> 
> > > * Or directly introduce this config file in static checks repo, VT link the config file from submodule to root dir
> > 
> > 
> > Yes, this is a possible solution, I just not sure If we can keep the `avocado-static-checks` features like having different `.pylintrc` files for different parts of code-base. Maybe for now we can introduce pre-commit to `avocado-static-checks` and run the `avocado-static-checks` scripts from pre-commint. What do you think?
> 
> That's definitely we can. in the `.pre-commit-config.yaml` files, we can add more than one `pylint` hooks, using different name and config file to distinguish them, and use like `files`, `exclude`, `types`, `exclude_types` filters.

If it is also possible to have some default configuration file in `avocado-static-checks` which would be used when no config is provided like [this](https://github.com/avocado-framework/avocado-static-checks/tree/main/default_configs), then I see this as the right solution, and it could replace the current scripts in `avocado-static-checks` repo.

selftests/unit/unittest_data/qemu-1.5.0__devices_help
```diff
@@ -132,4 +132,4 @@ name "virtio-blk-device", bus virtio-bus
 name "pc-testdev", bus ISA
 name "virtio-scsi-device", bus virtio-bus
 name "virtio-9p-device", bus virtio-bus
-name "virtio-net-pci", bus PCI, alias "virtio-net"
```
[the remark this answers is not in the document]
PaulYuuu: This makes sure files end in a newline.

zhencliu:
> Well, I really don't like using pylint to check typos, it always check username from docstring but for spell checker, should focus on code, the codespell/typos is better than pylint, they are typo check tool, and the performance is much better than pylint. image


I hit the same issue when I pushed my PRs, it said my name is a typo, I have to add my name into the spell.ignore to avoid this check

nickzhq:
Hello @PaulYuuu , sorry for the delayed response. The patch looks good to me.

selftests/checkall
```diff
@@ -11,4 +11,3 @@ run_rc() {
 run_rc './avocado-static-checks/run-static-checks'
```
YongxueHong:
```suggestion
run_rc './static-checks/run-static-checks'
```
I think that we need to update the path of `run-static-checks` since we have updated the configuration of the submodules `avocado-static-checks`: https://github.com/avocado-framework/avocado-vt/pull/4219/changes#diff-fe7afb5c9c916e521401d3fcfb4277d5071798c3baf83baf11d6071742823584
Please correct me if I was misunderstood, thanks.

BTW, could you elaborate more on why you update the submodule `avocado-static-checks`? Thanks.

YongxueHong (approved):
LGTM.

PR #4199, opened 2025-07-28
Fix the issue when var is empty but not None

leidwang:
Currently if var is empty but not None it will cause findmnt cmd run into error, so update the code logic to fix this issue. Also add some log output for the exception.

ID:4169

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling and logging for mount checks, providing clearer feedback and avoiding silent failures.
  * Broader detection of shell-command errors and immediate handling to reduce unreliable behavior.

* **Behavior Changes**
  * Mount option validation tightened — empty or meaningless options now produce an explicit error.
  * Mount checks prefer provided interactive sessions when available, improving reliability.

PR #4221, opened 2025-08-14
utils_disk: Fix findmnt error

fangge1212:
Before the fix, findmnt will succeed even if the device is not mounted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mount detection reliability by explicitly handling empty and None responses.
  * Prevented empty mount options from being generated, avoiding invalid command flags and potential errors.
  * Reduces false negatives when checking whether a path is mounted.
  * More stable behavior in environments with missing or empty configuration values.


fangge1212:
> Maybe [4199](https://github.com/avocado-framework/avocado-vt/pull/4199) can fix your issue.

yes, thanks

PR #4227, opened 2025-08-25
Increase timeout for lock acquisition on slower hardware

pevogam:
This is meant to help slower platforms that for some reason with Python 3.13 and other recent environment versions need a bit of extra time to acquire VT logs' lock in a timely manner.


ldoktor (approved):
LGTM, it's a busy loop so it shouldn't prolong the good-case scenarios and bad-case scenarios shouldn't happen (and when they do it's probably fine to let it run for a bit longer)

pevogam:
> **Actionable comments posted: 0**
> 🧹 Nitpick comments (4)
> 
> > virttest/utils_logfile.py (4)> `38-41`: **Tighten the docstring phrasing.**
> > > Minor clarity fix. Also lowercase “wait” and clarify units.
> > > ```diff
> > > -    :param lock: threading.RLock object
> > > -    :param timeout: time to Wait for the lock
> > > +    :param lock: threading.RLock (or compatible) instance
> > > +    :param timeout: seconds to wait for the lock
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `43-44`: **Use time.monotonic() for timeout measuring to avoid clock skew issues.**
> > > time.time() can move backward/forward if the system clock changes, altering the effective wait. time.monotonic() is the standard for measuring durations.
> > > ```diff
> > > -    end_time = time.time() + timeout
> > > -    while time.time() < end_time:
> > > +    end_time = time.monotonic() + timeout
> > > +    while time.monotonic() < end_time:
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `43-48`: **Prefer blocking acquire with timeout over spin + sleep.**
> > > threading.RLock supports acquire(timeout=...), which avoids a polling loop and wakes promptly. This also keeps behavior identical (returns bool), but reduces wakeups and CPU jitter. If you want periodic wakeups to perform other checks, Option B keeps a bounded wait per iteration.
> > > Option A (simplest):
> > > ```diff
> > > -    end_time = time.time() + timeout
> > > -    while time.time() < end_time:
> > > -        if lock.acquire(False):
> > > -            return True
> > > -        time.sleep(0.05)
> > > -    return False
> > > +    return lock.acquire(timeout=timeout)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Option B (bounded wait + monotonic):
> > > ```diff
> > > -    end_time = time.time() + timeout
> > > -    while time.time() < end_time:
> > > -        if lock.acquire(False):
> > > -            return True
> > > -        time.sleep(0.05)
> > > -    return False
> > > +    end_time = time.monotonic() + timeout
> > > +    while True:
> > > +        remaining = end_time - time.monotonic()
> > > +        if remaining <= 0:
> > > +            return False
> > > +        # wake periodically (<= 1s) to keep responsiveness if needed
> > > +        if lock.acquire(timeout=min(remaining, 1.0)):
> > > +            return True
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `33-33`: **Replace the magic number 30 with a module constant for discoverability (optional).**
> > > Keeps the default in one place and makes it obvious for future tuning. If you prefer runtime configurability, we can also back it by an env var; happy to follow project conventions.
> > > Apply within this range:
> > > ```diff
> > > -def _acquire_lock(lock, timeout=30):
> > > +def _acquire_lock(lock, timeout=LOG_LOCK_ACQUIRE_TIMEOUT):
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > And add this near the top of the module (example placement after _log_lock definition):
> > > ```python
> > > # Default acquire timeout for log operations (seconds)
> > > LOG_LOCK_ACQUIRE_TIMEOUT = 30
> > > ```
> 
> 📜 Review details

Just to add some replies to the rabbit bot (for all humans involved):

1) The literal value for the timeout is something standard done in most VT methods and I would approve using a constant only if this becomes more important function.
2) The use of monotonic time is definitely a good idea but it something that should be done throughout the code base in its own focused pull request or commit.
3) I am not sure if the `lock.acquire(timeout=timeout)` is true or hallucinated but from what I see above there is also a choice to have a full control over the acquisition, e.g. time to wait in between attempts and other things like these that will not be configurable otherwise.

Of course, I am willing to go for any of these directions if any of our reviewers here believe they are important to follow upon.

virttest/utils_logfile.py
```diff
@@ -30,7 +30,7 @@
 _open_log_files = {}  # pylint: disable=C0103
 
 
-def _acquire_lock(lock, timeout=10):
+def _acquire_lock(lock, timeout=30):
```
YongxueHong: Hi @pevogam,

I agree that this is a solid temporary workaround for performance issues on constrained hardware.

One thought: should we parameterize the timeout? If we update `log_line` to take a `timeout` argument (defaulting to 10) and pass it to `_acquire_lock`, it would make the system more flexible. That way, users on very slow hardware can increase it themselves, and we won't have to keep raising the default hard limit (e.g., to 30 or higher).

In any case, the current solution works for me. Thanks!

YongxueHong (approved):

PR #4233, opened 2025-09-04
Fix issue in libvirt_storage.py

yalzhang:
The help text of qemu-img has been refactored in latest version, which cause the output of "qemu-img -h" different from before. To fix it, update it to be more specific as "qemu-img info -h", which can always work as expected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of image lock support by changing how the image tool’s help output is queried, reducing false negatives across versions. No user-facing interface or configuration changes.

* **Refactor**
  * Centralized image lock detection so all components use a shared implementation, ensuring consistent behavior.

### DIFF
--- a/virttest/libvirt_storage.py
+++ b/virttest/libvirt_storage.py
@@ -575,6 +575,6 @@ def check_qemu_image_lock_support():
     except process.CmdError:
         raise process.CmdError(cmd, binary_path, "qemu-img command is not found")
     cmd_result = process.run(
-        binary_path + " -h", ignore_status=True, shell=True, verbose=False
+        binary_path + " info -h", ignore_status=True, shell=True, verbose=False
     )
     return "-U" in cmd_result.stdout_text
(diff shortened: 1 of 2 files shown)


yalzhang:
before the fix, on rhel 10.2, the case fail:

backingchain.blockcommit.conventional_chain.file_disk_qcow2.top_to_base
Command 'qemu-img info --backing-chain /var/lib/avocado/data/avocado-vt3' failed.&#10;stdout: b''&#10;stderr: b'qemu-img: Could not open \'/var/lib/avocado/data/avocado-vt3\': Failed to get shared "write" lock\nIs another process using the image [/var/lib/avocado/data/avocado-vt3]?\n'&#10;additional_info: None

after the fix, it run pass locally:
(1/1) type_specific.io-github-autotest-libvirt.backingchain.blockcommit.conventional_chain.file_disk_qcow2.top_to_base: PASS (39.16 s)

yalzhang:
The 2 function are the same, I tried to delete the one in utils_misc.py and use the one in libvirt_storage.py, but failed since it will cause circular import. Several files are using libvirt_storage.check_qemu_image_lock_support(). To make the changes smaller, I updated the cmd in these 2 places together.

cliping (approved):
LGTM

PR #4234, opened 2025-09-05
virttest/iscsi.py: create the non-existed dir before _get_create_command()

nickzhq:
1. Remove the _get_create_command() in Fileio init()
2. Create the non-existed dir before _get_create_command()
3. Call _get_create_command() before the `create file` command

ID: 4348

## Summary by CodeRabbit

* **New Features**
  * Image creation now determines the most efficient method at creation time (e.g., fallocate when available) and lazily recomputes the approach as needed.
  * Automatically ensures target directories exist before creating images.
  * Creation steps are logged for visibility.

* **Bug Fixes**
  * Recomputes creation method after image deletion/config save to avoid stale commands and ensure correct handling of on-disk size.

### DIFF
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -938,16 +938,12 @@ class Fileio(Backstore):
         block_size = emulated_size[self.unit][1]
         size = int(self.emulated_size) * emulated_size[self.unit][0]
         self.emulated_expect_size = block_size * size
-        dd_command = "dd if=/dev/zero of=%s count=%s bs=%sK" % (
+        self.create_cmd = "dd if=/dev/zero of=%s count=%s bs=%sK" % (
             self.emulated_image,
             size,
             block_size,
         )
-
-        # optimize the `create file` command
-        self.create_cmd = _get_create_command(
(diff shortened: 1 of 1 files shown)


nickzhq:
@zhencliu @YongxueHong @qingwangrh Please take a review, thanks!

virttest/iscsi.py
```diff
@@ -970,10 +965,20 @@ def create_backstore(self):
             # create image disk
             if not os.path.isfile(self.emulated_image):
                 process.system("mkdir -p %s" % os.path.dirname(self.emulated_image))
+                # optimize the `create file` command
+                self.create_cmd = _get_create_command(
+                    self.emulated_image, self.emulated_size + self.unit, self.create_cmd
+                )
                 process.system(self.create_cmd)
```
qingwangrh, line 972: Could you add print info for the command to create a file? That is easy to debug.

qingwangrh: Yes, i want to know what command will be executed.

nickzhq:
@YongxueHong 
Please take a review again, thanks!

PR #4235, opened 2025-09-05
Update the function exec_virsh_edit()

yalzhang:
Update the function exec_virsh_edit() in libvirt.py to test virsh managedsave-edit as well.


virttest/utils_test/libvirt.py
```diff
@@ -3543,28 +3543,58 @@ def update_vm_disk_source(
...
-        remote.handle_prompts(session, None, None, r"[\#\$]\s*$", debug=True)
+        # Capture output until the next shell prompt
+        prompt_pattern = r"[\#\$]\s*$"
+        match, output = session.read_until_last_line_matches(
+            [prompt_pattern], timeout=10
+        )
+
+        # Check for virsh errors in the captured output
+        if "error:" in output or "failed:" in output:
+            LOG.error("virsh edit command failed with output:\n%s", output)
```
smitterl: Does this and do other error messages now apply to 'managedsave-edit' now, too?

virttest/utils_test/libvirt.py
```diff
@@ -3543,28 +3543,58 @@ def update_vm_disk_source(
...
+        else:
+            session.sendline(
+                "virsh -c %s %s %s %s" % (connect_uri, command, source, virsh_opt)
+            )
         for cmd in edit_cmd:
             session.sendline(cmd)
         session.send("\x1b")
         session.send("ZZ")
-        remote.handle_prompts(session, None, None, r"[\#\$]\s*$", debug=True)
+        # Capture output until the next shell prompt
```
smitterl: Why did you substitute the call to `handle_prompts`, looks like a risky change that might introduce regression. Please elaborate in your commit message why you changed this and what you think the involved risk is and if necessary, what you did to mitigate that risk.

smitterl (changes requested):

virttest/utils_test/libvirt.py
yalzhang: The handle_prompts automates interaction with a shell to get it to a state where it's ready for the next command. With `session.read_until_last_line_matches([prompt_pattern], timeout=10)`, it directly waits for the shell prompt (# or $). It also captures all the output from the virsh edit command into the output variable. It will then inspect the output, which is more explicit and robust.

virttest/utils_test/libvirt.py
```diff
@@ -3543,28 +3543,58 @@ def update_vm_disk_source(
...
+            LOG.error("virsh edit command failed with output:\n%s", output)
+            session.close()
+            return False
+
         session.close()
         return True
     except Exception as e:
         session.close()
-        LOG.error("Error occurred: %s", e)
+        LOG.error("Error occurred during virsh edit session: %s", e)
```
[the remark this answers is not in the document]
yalzhang: Updated, Thanks!

virttest/utils_test/libvirt.py
chunfuwen: As Sebastian highlight here, it needs to do some testing validating whether  remote.handle_prompts can be replaced equally with session.read_until_last_line_matches.

PR #4238, opened 2025-09-09
env_process: allow creating the dir while it exists

hellohellenmao:
set exist_ok=True for os.makedirs function

ID: 3442

## Summary by CodeRabbit

* **Bug Fixes**
  * Setup now tolerates pre-existing directories for mounted filesystem sources, preventing unnecessary errors during environment preparation.
  * Avoids failures when the target directory already exists, improving reliability of repeated runs.
  * Reduces manual cleanup and reruns by making directory creation idempotent.
  * Improves stability for repeated and concurrent workflows while preserving existing error handling in other scenarios.

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -182,7 +182,7 @@ def preprocess_fs_source(test, params, fs_name, vm_process_status=None):
                 if os.path.exists(fs_source):
                     shutil.rmtree(fs_source, ignore_errors=True)
                 LOG.info("Create filesystem source %s." % fs_source)
-                os.makedirs(fs_source)
+                os.makedirs(fs_source, exist_ok=True)
     else:
         test.cancel('Unsupport the type of filesystem "%s"' % fs_type)
 


hellohellenmao:
> Hi @hellohellenmao Could you clarify more about this PR in the commit message, such as the motivation and approach? Thanks.
> 
> ```
> Write a good commit message, pointing motivation, issues that you're
> addressing. Usually you should try to explain 3 points in the commit message:
> motivation, approach and effects
> ```

Sure, updated. Review again, please.

YongxueHong (approved):
LGTM.

fbq815 (approved):
LGTM:
(1/3)unattended_install.cdrom.extra_cdrom_ks.default_install.aio_threads.s390-virtio: PASS (377.62 s)
 (2/3) virtio_fs_multi_vms.with_multi_fs_sources.s390-virtio: PASS (49.11 s)
 (3/3) virtio_fs_multi_vms.share_fs_source.s390-virtio: PASS (49.16 s)


re-run:

hellohellenmao:
> > If there are existing files in the test folder, they may affect the test results. Could you please take this into account? @hellohellenmao
> 
> Hi @xiagao , with the changes here, creating the dir will success even though there are files in the test folder. And I believe that if the user are calling `create_fs_source` function that their intention should be creating a clean new dir. so overwriting the original dir/files should be fine. What your points here ?

@xiagao Could you please help to check out here? Thanks

xiagao:
> > > If there are existing files in the test folder, they may affect the test results. Could you please take this into account? @hellohellenmao
> > 
> > 
> > Hi @xiagao , with the changes here, creating the dir will success even though there are files in the test folder. And I believe that if the user are calling `create_fs_source` function that their intention should be creating a clean new dir. so overwriting the original dir/files should be fine. What your points here ?
> 
> @xiagao Could you please help to check out here? Thanks


I mean if the directory already exists, its contents will remain unchanged. os.makedirs() never deletes or overwrites files.
In such situation, it would impact the test result due to the existed content in directory.

hellohellenmao:
> > > > If there are existing files in the test folder, they may affect the test results. Could you please take this into account? @hellohellenmao
> > > 
> > > 
> > > Hi @xiagao , with the changes here, creating the dir will success even though there are files in the test folder. And I believe that if the user are calling `create_fs_source` function that their intention should be creating a clean new dir. so overwriting the original dir/files should be fine. What your points here ?
> > 
> > 
> > @xiagao Could you please help to check out here? Thanks
> 
> I mean if the directory already exists, its contents will remain unchanged. os.makedirs() never deletes or overwrites files. In such situation, it would impact the test result due to the existed content in directory.

So my patch here is to resolve this issue, with  `exist_ok=True` the dir will be created again even though the directory already exists

xiagao:
> > Hi, tingting If the dir exist, `exist_ok=True` will skip creating the dir. The existed dir might be not clean which might lead to test fail. So if the dir existed, we should re-create it.
> 
> Yeah, you're right. Thanks for pointing this out. And I remember that `shutil.rmtree(fs_source)` fails in this case is because that the `fs_source` is mounted. So we might solve this at first, I am going to `umount` before `shutil.rmtree(fs_source)`. What your points on my new fix plan?

It looks good to me.  And I don't know why ignore_errors=True here, is there risk to remove it? 
            if create_fs_source:
                if os.path.exists(fs_source):
                    shutil.rmtree(fs_source, ignore_errors=True)
                LOG.info("Create filesystem source %s." % fs_source)

harvey0100:

PR #4258, opened 2025-10-30
Modify the nic's sub_option parameter to be compatible with the new format

yanglei-rh:
Because of the new version's virt-install --network help sub_option's format already from driver_queues to driver.queues, driver_queues as an alias, is still supportted by virt-install command, so modified sub_option to support both them.

ID: LIBVIRTAT-22112
Signed-off-by: Lei Yang leiyang@redhat.com

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of driver-related network options to recognize variant option name formats (dot or underscore).
  * Deferred compatibility validation to the underlying virtualization provider to avoid spurious errors or warnings for unavailable driver features.
  * Preserved existing behavior for applying driver parameters when network type and NIC conditions are met.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -937,10 +937,10 @@ class VM(virt_vm.BaseVM):
                 if (
                     nettype
                     and nic_queues
-                    and has_sub_option("network", "driver_queues")
+                    and has_sub_option("network", "driver[._]queues")
                 ):
                     result += ",driver_queues=%s" % nic_queues
-                    if nic_driver and has_sub_option("network", "driver_name"):
+                    if nic_driver and has_sub_option("network", "driver[._]name"):
                         result += ",driver_name=%s" % nic_driver
                 elif mac:  # possible to specify --mac w/o --network
                     result += " --mac=%s" % mac


yanglei-rh:
Hi @richtja Please also help delegate a maintainer to help review this patch which about re-write add_nic function support mq and driver_name's logic. Thanks a lot.

richtja:
Hi @smitterl, can you please look into this or find someone suitable for the review? Thank you.

chunfuwen (approved):
lgtm

virttest/libvirt_vm.py
```diff
@@ -934,14 +934,21 @@ def add_nic(help_text, nic_params):
...
-                if (
-                    nettype
-                    and nic_queues
-                    and has_sub_option("network", "driver_queues")
-                ):
-                    result += ",driver_queues=%s" % nic_queues
-                    if nic_driver and has_sub_option("network", "driver_name"):
-                        result += ",driver_name=%s" % nic_driver
+                if nettype and nic_queues:
+                    if has_sub_option("network", "driver_queues"):
```
YongxueHong: Hi @yanglei-rh 
From my understanding of your code changes, I think it does not change the logic, just prints the log warning if `has_sub_option` is `False`.
And, I checked the function `has_sub_option` and figured out that it may not work for the latest `virt-install`
https://github.com/avocado-framework/avocado-vt/blob/92402c95d97d4bbc0686fcc32b5efd5857d4c8ed/virttest/libvirt_vm.py#L461-L465

For example:
The `has_sub_option("network", "driver_queues")`, we can not get the `driver_queues` from the output of `help` , but it is `driver.queues`.
```
# virt-install --network help
--network options:
  ...
  bridge
  driver.ats
  driver.iommu
  driver.name
  driver.packed
  driver.page_per_vq
  driver.queues
  ...
# virt-install --version
5.0.0
```
There is a changelog of the `virt-install`: https://github.com/virt-manager/virt-manager/compare/v2.1.0...v2.2.0 
I think we may update the `has_sub_option` to support this change of `virt-install`
The above is just from my investigation. Please correct me if I was wrong.
Thanks.

virttest/libvirt_vm.py
```diff
@@ -937,10 +937,10 @@ def add_nic(help_text, nic_params):
                 if (
                     nettype
                     and nic_queues
-                    and has_sub_option("network", "driver_queues")
+                    and has_sub_option("network", r"driver[._]queues")
```
YongxueHong:
```suggestion
                    and has_sub_option("network", "driver[._]queues")
```
I think that we do not need to pass the Python’s raw string notation for regular expression patterns since there already been: https://github.com/avocado-framework/avocado-vt/blob/92402c95d97d4bbc0686fcc32b5efd5857d4c8ed/virttest/libvirt_vm.py#L465

Thanks.

virttest/libvirt_vm.py
```diff
@@ -937,10 +937,10 @@ def add_nic(help_text, nic_params):
...
                 ):
                     result += ",driver_queues=%s" % nic_queues
-                    if nic_driver and has_sub_option("network", "driver_name"):
+                    if nic_driver and has_sub_option("network", r"driver[._]name"):
```
YongxueHong: Same as the above. Thanks.

yanglei-rh: Fixed. please help review them again, thanks a lot.

YongxueHong (approved):

yanglei-rh:
Before this patch, driver_queues could not be passed correctly, the test result please refer to following command line:
`--network=bridge=switch,model=virtio,mac=52:54:00:8b:3c:42 --mac=52:54:00:8b:3c:42
`
After apply this patch, driver_queues can be passed correctly with the same test command: 
`--network=bridge=switch,model=virtio,mac=52:54:00:34:ff:64,driver_queues=4,driver_name=vhost`

So I think we fixed this problem, and doesn't introduced any regression issues.

PR #4259, opened 2025-10-30
utils_test/libvirt: wait for iscsi disk to become available

smitterl:
After creating the iSCSI disk, the device file under /dev might not be immediately shown. In our environments with multipath and 64 SCSI block devices, the test would sometimes when trying to partition the disk.

Wait for up to 10 seconds for the device to show up and raise a TestError if it doesn't.

Also apply usual `if...else` idiom instead of verbose `if..not if`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved iSCSI device detection after login by verifying the device is present both at the kernel/device layer and on the filesystem. Added a 10-second timeout for the device path check and clearer error handling and cleanup when the device does not appear, reducing false positives and improving reliability of subsequent operations.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -587,15 +587,27 @@ def setup_or_cleanup_iscsi(
     if is_setup:
         if is_login:
             _iscsi.login()
-            # The device doesn't necessarily appear instantaneously, so give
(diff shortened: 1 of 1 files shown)


fbq815 (approved):
LGTM

PR #4262, opened 2025-11-07
Add delvm to loadvm and savevm for complete internal qcow2 snapshot ops

pevogam:
## Summary by CodeRabbit

* **New Features**
  * Added snapshot deletion: can delete a VM snapshot only when the VM is paused; performs pre- and post-operation state checks and logs the action. Logging messages now refer to "snapshot" in save/load operations.

* **Chores**
  * Updated tooling configuration to recognize the new deletion keyword.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -5766,7 +5766,7 @@ class VM(virt_vm.BaseVM):
         Override BaseVM savevm method
         """
         self.verify_status("paused")  # Throws exception if not
-        LOG.debug("Saving VM %s to %s" % (self.name, tag_name))
+        LOG.debug("Saving snapshot %s from VM %s", tag_name, self.name)
         self.monitor.send_args_cmd("savevm id=%s" % tag_name)
         self.monitor.cmd("system_reset")
         self.verify_status("paused")  # Throws exception if not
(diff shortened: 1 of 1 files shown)


luckyh (approved):
LGTM, thanks.

virttest/qemu_vm.py
```diff
@@ -5776,10 +5776,19 @@ def loadvm(self, tag_name):
...
         """
         self.verify_status("paused")  # Throws exception if not
-        LOG.debug("Loading VM %s from %s" % (self.name, tag_name))
+        LOG.debug("Loading snapshot %s from VM %s", tag_name, self.name)
         self.monitor.send_args_cmd("loadvm id=%s" % tag_name)
         self.verify_status("paused")  # Throws exception if not
 
+    def delvm(self, tag_name):
+        """
+        Override BaseVM delvm method
```
YongxueHong: According to https://github.com/avocado-framework/avocado-vt/blob/017ae5cc073e8d8110300d43eafa00dfe370c631/virttest/virt_vm.py#L1789-L1804
The `BaseVM` class just has the `savevm` and `loadvm` methods, not the delvm.
For this docsting, I suggest that we could update it without the `Override` string, or we can add the abstract method in the `BaseVM`.
Both approaches are okay for me, thanks.

luckyh (approved):

PR #4274, opened 2025-11-24
Perform vm alive checks in the right order

pevogam:
Some of the subchecks rely on pre-existing vm process which is only validated and established during the base class check.

## Summary by CodeRabbit

* **Refactor**
  * Modified the order of VM health verification checks to execute after initial connection attempts, refining error detection timing during VM status verification processes without altering overall functionality.


luckyh (approved):
Good catch, thanks!

PR #4285, opened 2025-12-13
migration: Retrieve QEMU logs after migration failure

dzhengfy:
## Summary

Add functionality to capture QEMU logs from both source and destination hosts when migration fails unexpectedly. This helps with debugging migration failures.

## Changes

- Add `get_local_qemu_log()` in `utils_sys.py` to retrieve QEMU log from local host
- Add `get_remote_qemu_log()` in `utils_sys.py` to retrieve QEMU log from remote host via SSH
- Update `check_result()` in `migration.py` to call these functions when migration fails unexpectedly (status_error=no but migration failed)
- Add test parameter to `do_migration()` to pass test object

## Details

The functions retrieve the last 10 lines of QEMU logs by default, which is configurable via the `log_lines` parameter.

When migration fails unexpectedly:
- Local QEMU log is read from `test.debugdir/vm.name.log`
- Remote QEMU log is read from `/var/log/libvirt/qemu/vm.name.log` via SSH

This provides valuable debugging information in the test logs when investigating migration failures.

🤖 Assisted by [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Collect QEMU logs from local and remote hosts (configurable) to capture diagnostics during migrations.
  * Migration flow now includes VM context in result validation for richer diagnostics.

* **Bug Fixes**
  * Improved error handling on migration failures to gather relevant logs before reporting errors, aiding faster debugging.

### DIFF
--- a/virttest/migration.py
+++ b/virttest/migration.py
@@ -18,6 +18,7 @@ from virttest import (
     utils_iptables,
     utils_misc,
     utils_net,
+    utils_sys,
     utils_test,
 )
 
@@ -372,6 +373,7 @@ class MigrationTest(object):
                                     {"func": <function domjobabort at 0x7f5835cd08c8>,
                                      "before_pause": "no"}
                                    ]
+        :param virsh_opt: str, virsh options
         :param args: dictionary used by func,
                      'func_param' is mandatory for func parameter.
                      If no real func_param, none is requested.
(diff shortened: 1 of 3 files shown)


hholoubk:
Hi @dzhengfy 
I would consider to get also the qemu-log from both in case, there is the "No migration result" ...
Not sure if in other cases of failure ... it can be also usefull and probably better to have it in the test finally part before recovery?  

Otherwise the method and overall implementation looks great for me. Just thinking about better usage

Yingshun (approved):
LGTM

PR #4299, opened 2026-01-07
Function to filter NUMA node ids with allocated memory!

Anushree-Mathur:
Added a function to identify a nodeset consisting of only those numa node ids which has memory allocated to it.

This function is then used to retrieve the filtered NUMA nodeset, which is subsequently employed for hugepage memory backing. This ensures that hugepages are allocated only from nodes with actual memory resources. For example:
Host has 2 numa nodes with id as 0 and 4, but only numa node 4 has memory so that will be used to back the memory being used as a nodeset.
Signed-off-by: Anushree-Mathur <anushree.mathur@linux.ibm.com>

## Release Notes

* **Bug Fixes**
  * Improved guest NUMA node detection to accurately identify memory-enabled nodes, enabling more intelligent memory pinning support when hugepage functionality is enabled.
  * Enhanced fallback behavior for memory pinning configuration, providing smarter defaults when user-specified settings are unavailable.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -14,6 +14,7 @@ import platform
 import re
 import shutil
 import string
+import subprocess as sp
 import tempfile
 import time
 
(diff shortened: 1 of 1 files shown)


harihare:
@Anushree-Mathur fix the AI review comments and 2 failing checks.

PR #4314, opened 2026-01-29
add "dump memory file from guest" feature

nickzhq:
- Add guest_dump_file_manager.py
- Add dump_file.py
- Update env_process.py

Add "dump memory file" step in postprocess_vm function The guest_dump_file_manager.py implements the main function The dump_file.py implements "prepare the environment"

ID: KVMAUTOMA-1304

## Summary by CodeRabbit

* **New Features**
  * Automatic extraction of guest crash/dump files from VM disk images after tests, supporting QEMU and libvirt workflows and Windows/Linux guests.
  * Offline extraction integrates into the test lifecycle when enabled, registering necessary setup steps automatically.

* **Chores**
  * Test setup now ensures host kernel support for disk access (nbd), attempts to load it, logs failures without stopping tests, and performs conditional cleanup.


virttest/test_setup/dump_file.py
```diff
@@ -0,0 +1,46 @@
...
+
+from avocado.utils import linux_modules
+
+from virttest import utils_package
+from virttest.test_setup.core import Setuper
+
+LOG = logging.getLogger(__name__)
+
+
+class DumpFileSetup(Setuper):
```
YongxueHong: Actually, we already have a similar `test_setup` module to handle the kernel modules, which is `kernel`:
https://github.com/avocado-framework/avocado-vt/blob/0103e092833db7dcf1b92a3c6d34279452c9c73c/virttest/test_setup/kernel.py

From my understanding,  the `DumpFileSetup` aims to load the related kernel module `nbd` for getting the dump file by the `qemu-nbd`, and it is not specific to the `DumpFileSetup` , maybe other cases need this kernel module.

I suggest naming it commonly in the `virttest/test_setup/kernel.py`, for example: 
```
class NBDSetup(Setuper):
    def setup(self):
       ...
    def cleanup(self):
       ...
```
Thanks.

virttest/guest_dump_file_manager.py
```diff
@@ -0,0 +1,332 @@
...
+                self._nbd_dev = row.split()[0]
+                break
+        LOG.debug("Find the available nbd device: %s", self._nbd_dev)
+
+        try:
+            qemu_nbd_cmd = "qemu-nbd -c /dev/%s %s" % (
+                self._nbd_dev,
+                image_path,
+            )
+            process.run(qemu_nbd_cmd)
```
zhencliu: move this to nbd.py?

virttest/guest_dump_file_manager.py
```diff
@@ -0,0 +1,506 @@
...
+        mounting the guest filesystem.
+
+        :param res_dir: Directory where results should be stored
+        """
+        _dirs = []
+        for item in self._dmp_dirs_scaned[self._guest_type]:
+            if os.path.isdir(item):
+                _dirs.append(item)
+        if _dirs:
+            _dirs = " ".join(_dirs)
```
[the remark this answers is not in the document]
leidwang: Perhaps we could add some log here so that users can clearly see what happened.For example, whether the DMP file wasn't found, or if the copy operation failed.

virttest/test_setup/dump_file.py
YongxueHong: Hi @nickzhq 
Thanks for the clarification, it is also good for me.

virttest/shared/cfg/base.cfg
```diff
@@ -1256,3 +1256,49 @@ uuid_dimm = ""
 # enable_igvm = no
 # igvm_path = /usr/share/coconut-svsm
 # igvm_filename = coconut-qemu.igvm
+
+# =============== vm_memory_dump_method parameter ===============
```
YongxueHong:
```suggestion
# VM memory dump method params
```
Maybe no need the extra symbol `=` to make it consistent with the others, thanks.

virttest/shared/cfg/base.cfg
```diff
@@ -1256,3 +1256,49 @@ uuid_dimm = ""
...
+#                     yet supported.
+#
+#   "qmp"           - (Not yet implemented) Online extraction via QMP/guest-agent.
+#                     Intended for collecting dump files from a running VM
+#                     through the QEMU guest agent. Requires the VM to be
+#                     running with guest-agent installed.
+#
+#   "backup_image"  - (Not yet implemented) Offline extraction from a backup
+#                     image via NBD. Similar to "nbd" but operates on a
+#                     separate backup copy of the disk image.
```
YongxueHong:
```suggestion
#   "nbd"           - Offline extraction via QEMU NBD.
#                     Connects the guest disk image to an NBD device, detects
#                     partitions and filesystem type, mounts the appropriate
#                     partition, and copies .dmp files to the results directory.
#                     Requires the VM to be stopped. Supported guest filesystems:
#                     NTFS (read-write via ntfs-3g). LVM2 is detected but not
#                     yet supported.
```
Just keep the implemented method in order to avoid the noise, thanks.

virttest/shared/cfg/base.cfg
```diff
@@ -1256,3 +1256,49 @@ uuid_dimm = ""
...
+#
+# Works with
+#   vm_type: qemu and libvirt (libvirt methods not yet implemented).
+#   os_type: windows and linux ( linux methods not yet implemented ).
+#
+# Usage examples:
+#
+#   # Enable NBD-based dump extraction for a Windows guest:
+#   vm_memory_dump_method = nbd
```
YongxueHong:
```suggestion
#vm_memory_dump_method = nbd
```

virttest/test_setup/dump_file.py
```diff
@@ -0,0 +1,51 @@
...
+    _nbd_is_loaded_by_vt = False
+
+    def setup(self):
```
YongxueHong, line 28: For the test_setup `dump_file`, I think it is just necessary that the test case requests to dump the memory for the guest with the specific condition, not for all the scenarios.  We could refer to https://github.com/avocado-framework/avocado-vt/blob/master/virttest/test_setup/rng_egd.py

harvey0100:

PR #4322, opened 2026-02-12
qemu_monitor: Replace select with DefaultSelector to avoid FD limit

YongxueHong:
Use DefaultSelector instead of select() in Monitor._data_available() to
avoid the 1024 file descriptor limit that can cause issues with systems
having many open file descriptors.

ID: 5075

### DIFF
--- a/virttest/qemu_monitor.py
+++ b/virttest/qemu_monitor.py
@@ -11,7 +11,7 @@ import json
 import logging
 import os
 import re
-import select
+import selectors
 import socket
 import threading
 import time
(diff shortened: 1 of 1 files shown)


zhencliu:
The code change LGTM.

Just one question, as we are replacing select here, would it be possible to use selector(https://docs.python.org/3/library/selectors.html) , which is high level and built on top of select?

zhencliu (approved):

YongxueHong:
Hi @zixi-chen 
Thanks for your review, pls help approve this pr if you do not have other comments. I will merge it later. Thanks a lot!

PR #4323, opened 2026-02-13
utils_net: Fix uninitialized variable 'restart_cmd'

qiankehan:
The variable 'restart_cmd' is not initialized in the else branch. It will cause the error like:
cannot access local variable 'restart_cmd' where it is not associated with a value

Fixes: 785a92daa (utils_net: fix guest net restart)


qiankehan:
@smitterl @yanglei-rh @nanli1 It is a tiny fix. Please help to review it. Thanks

dzhengfy (approved):
LGTM

PR #4345, opened 2026-03-20
Handle escape character syntax warnings via raw strings

pevogam:


YvanY0 (approved):
Thank you for the fix.

YongxueHong (approved):
LGTM.
Thanks for pointing it out and fixing it.

PR #4358, opened 2026-05-05
update ovmf firmware path

kraxel:
The firmware images are in /usr/share/edk2/ovmf these days, the old
locaation in /usr/share/OVMF has only some backward compatibility
symlinks only.  Update the configuration accordingly.

Signed-off-by: Gerd Hoffmann <kraxel@redhat.com>

Also (second commit added later):

set ovmf filenames only if not present
This allows test variants override the firmware filenames.

### DIFF
--- a/virttest/shared/cfg/guest-hw.cfg
+++ b/virttest/shared/cfg/guest-hw.cfg
@@ -462,7 +462,7 @@ variants:
         no i386 ppc64 ppc64le s390x
         no Win2003 WinXP WinVista
         # Please update this based on your test env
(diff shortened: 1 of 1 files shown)


pevogam (approved):
Hi @kraxel indeed some of us already have the new path in various customized configs for the virtualization hardware. Regarding the second commit, I can already see that there are some lazy set operations like `required_qemu` in the same config so this should be sane and it properly motivated.

pevogam:
Since this concerns no particular styling I will afford merging without the pre-commit hook.

PR #4359, opened 2026-05-06
Undefine persistent VM before virt-install in create()

hholoubk:
When create_vm_libvirt=yes triggers vm.create(), the method only calls self.destroy() before running virt-install. If a previous test left the VM defined (e.g. via bk_xml.sync() in its cleanup), the VM is still persistent and virt-install fails with:

  "Disk ... is already in use by other guests ['avocado-vt-vm1']"

Add self.undefine() after destroy so that any stale VM definition is removed before virt-install attempts to create the VM fresh.


harvey0100:

Issue #4266, opened 2025-11-18
TypeError: cannot pickle 'itertools.count' object on Fedora 43 with Python 3.14


hpaluch-pil:
On latest stable Fedora 43 with Python 3.14 (and 2 other workarounds applied - see below) I get fatal error when running trivial example:
```shell
$ avocado run --vt-guest-os Guest.Linux.JeOS.27.x86_64  type_specific.io-github-autotest-qemu.hello_world  
JOB ID     : a5b8286439a60a739c1363249b97be9fbea0b47d                                                                             
JOB LOG    : /home/ansible/avocado/job-results/job-2025-11-18T08.37-a5b8286/job.log                                               
 (1/1) type_specific.io-github-autotest-qemu.hello_world: STARTED                                                                 
 (1/1) type_specific.io-github-autotest-qemu.hello_world:  ERROR: cannot pickle 'itertools.count' object (16.04 s)                
RESULTS    : PASS 0 | ERROR 1 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0                                                 
JOB TIME   : 16.90 s                                                     
```

Full error details from `~/avocado/job-results/latest/test-results/1-type_specific.io-github-autotest-qemu.hello_
world/stderr`:
```
Traceback (most recent call last):
  File "/usr/lib/python3.14/site-packages/avocado_vt/utils.py", line 140, in _safe_env_save
    env.save()
    ~~~~~~~~^^
  File "/usr/lib/python3.14/site-packages/virttest/utils_env.py", line 108, in save
    cPickle.dump(self.data, f, protocol=0)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.14/copyreg.py", line 75, in _reduce_ex
    raise TypeError(f"cannot pickle {cls.__name__!r} object")
TypeError: cannot pickle 'count' object
when serializing dict item '_index'
when serializing virttest.vt_iothread.PredefinedManager state
when serializing virttest.vt_iothread.PredefinedManager object
when serializing dict item '_DevContainer__iothread_manager'
when serializing virttest.qemu_devices.qcontainer.DevContainer state
when serializing virttest.qemu_devices.qcontainer.DevContainer object
when serializing dict item 'devices'
when serializing virttest.qemu_vm.VM state
when serializing virttest.qemu_vm.VM object
[...]

PR #4360, opened 2026-05-08
Update  support Python 3.14

Awall3:
Changed two files to fix issue with pickling changes in 3.14

Add a proxy function for calling tests with multiprocessing to avoid pickling the entire VirtTest class

'multiprocessing' must pickle the 'target' and 'args' when using the 'forkserver' (default on Python 3.14+) or 'spawn' start methods. A bound method like 'VirtTest.runTest' would pickle the whole test instance, which includes unpicklable thread locks from Avocado's base test class. Building :class:`VirtTest` inside the child keeps the parent-to-child payload to 'SimpleQueue' and :class:`Runnable` only.

Remove self._index from IOThreadManagerBase as itertools count is not pickleable to address #4266

Can now run hello world in Python 3.14

Assisted-by: Claude

### DIFF
--- a/avocado_vt/plugins/vt_runner.py
+++ b/avocado_vt/plugins/vt_runner.py
@@ -112,6 +112,12 @@ class VirtTest(test.VirtTest):
                 self.queue.put(messages.FinishedMessage.get(status, fail_reason))
 
 
+def _vt_test_subprocess_main(queue, runnable):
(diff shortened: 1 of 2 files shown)


pevogam:
Hi @Awall3, could you rebase to see if any of this is still relevant? We already have passing runs with Python 3.14 on the main branch.

Awall3:
Fixed by #4375

PR #4365, opened 2026-06-08
Fix TypeError when updating NUMA cells with existing config!

Anushree-Mathur:
When updating NUMA cell configuration for VMs with existing cells, the code was passing NumaCellXML objects containing LibvirtXMLProperty attributes to dicts_to_cells(), which expects only str/int values.

Fix by using fetch_attrs() to convert NumaCellXML to plain dict before updating attributes. This ensures type validation passes.

Signed-off-by: Anushree-Mathur <anushree.mathur@linux.ibm.com>


pevogam (commented):
This looks good, @Anushree-Mathur could you confirm that you have tested your change manually and everything passes so that we can merge this?

PR #4368, opened 2026-06-11
Fix/restart guest network rhel10 nmcli

BulaYoungR:


harihare (changes requested):
Request to explain the patch in detail in the comment or first commet in PR. 

Please fix all auto review comments.

PR #4370, opened 2026-06-16
utils_net: Use dhcpcd for OVS bridge setup on RHEL 10

hholoubk:
## Summary
- Fix `create_ovs_bridge()` / `delete_ovs_bridge()` hardcoding `dhcp-client`/`dhclient`, which fails on RHEL 10 where only `dhcpcd` is available
- Detect RHEL major version per target host (local or remote SSH session) via `rpm -E '%{rhel}'`
- Reuse and fix `get_dhcp_client()` to choose the DHCP client based on the **session host**, not the local runner (fixes 9.9 worker → 10.3 resource cross-migration)
- Use `dhcpcd -n <iface>` for one-shot lease acquisition on OVS bridges

## Problem
`virsh.migrate_network.ovs_br.without_postcopy` fails on aarch64 cross-migration (RHEL 9.9 ↔ 10.3) when creating the OVS bridge on the RHEL 10 resource host:

```
yum install -y dhcp-client
No match for argument: dhcp-client
```

## Test plan
- [ ] RHEL 9.9 host: `create_ovs_bridge()` still uses `dhclient` (unchanged)
- [ ] RHEL 10.3 host: `create_ovs_bridge()` installs/uses `dhcpcd`
- [ ] Cross-migration: `virsh.migrate_network.ovs_br.without_postcopy` (9.9 worker → 10.3 resource)
- [ ] Regression: single-host OVS bridge tests on x86_64 / RHEL 9

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -1710,6 +1710,47 @@ def _deactivate_ipv4ll_dhcpcd(session):
         raise exceptions.TestError("Couldn't deactivate IPv4 local link setting.")
 
 
+def _get_rhel_major_version(session=None):
+    """
+    Return the RHEL major version for the target host.
+
+    :param session: guest/remote session, or None for local host
+    :return: major version int, or None if not RHEL/unknown
+    """
+    if session is None:
+        if distro.detect().name == "rhel":
+            return int(distro.detect().version_parts()[0])
+        return None
+
+    status, output = utils_misc.cmd_status_output(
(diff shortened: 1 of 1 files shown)


aniket-sahu-ibmx:
Hi @hholoubk

Can you please attach some test logs if you're still actively working on this?

pevogam:
Also if you address the AI review, rebase, and push again that would be great so we can retest everything with the extended recent CI.

PR #4372, opened 2026-06-19
virttest: add ipv6 portal check for iscsi

aniket-sahu-ibmx:
The portal check only looked for IPv4 format (0.0.0.0:3260) and failed to detect IPv6 default portals ([::]:3260 or [::0]:3260), causing unnecessary portal creation attempts in IPv6 environments.

Now checks for both IPv4 and IPv6 default portal formats before attempting to create a new portal.

### DIFF
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -726,10 +726,15 @@ class IscsiLIO(_IscsiComm):
 
             check_portal = "targetcli /iscsi/%s/tpg1/portals ls" % self.target
             portal_info = process.run(check_portal).stdout_text
-            if "0.0.0.0:3260" not in portal_info:
+            # Check for both IPv4 (0.0.0.0:3260) and IPv6 ([::]:3260 or [::0]:3260) default portals
+            if not any(
+                portal in portal_info
+                for portal in ("0.0.0.0:3260", "[::]:3260", "[::0]:3260")
+            ):
                 # Create portal
-                # 0.0.0.0 means binding to INADDR_ANY
-                # and using default IP port 3260
+                # 0.0.0.0 means binding to INADDR_ANY (IPv4)
+                # [::] or [::0] means binding to in6addr_any (IPv6)
+                # using default IP port 3260
                 portal_cmd = "targetcli /iscsi/%s/tpg1/portals/ create %s" % (
                     self.target,
                     "0.0.0.0",


virttest/iscsi.py
```diff
@@ -726,10 +726,12 @@ def export_target(self):
...
             check_portal = "targetcli /iscsi/%s/tpg1/portals ls" % self.target
             portal_info = process.run(check_portal).stdout_text
-            if "0.0.0.0:3260" not in portal_info:
+            # Check for both IPv4 (0.0.0.0:3260) and IPv6 ([::]:3260 or [::0]:3260) default portals
+            if "0.0.0.0:3260" not in portal_info and "[::]:3260" not in portal_info and "[::0]:3260" not in portal_info:
                 # Create portal
-                # 0.0.0.0 means binding to INADDR_ANY
-                # and using default IP port 3260
+                # 0.0.0.0 means binding to INADDR_ANY (IPv4)
+                # [::] or [::0] means binding to in6addr_any (IPv6)
```
kraxel, line 736: Where does the '[::0]' come from?  Looks bogus to me.

aniket-sahu-ibmx: When using targetcli on SLES16.1 closed beta, the iscsi default portal is represented as below:

```
ltcblue9yr-lp2:~ # targetcli ls
<redacted>
  o- iscsi ............................................................................................................ [Targets: 1]
  | o- iqn.2026-06.com.example:avocado.target ............................................................................ [TPGs: 1]
  |   o- tpg1 .................................................................................................. [gen-acls, no-auth]
  |     o- acls .......................................................................................................... [ACLs: 0]
  |     o- luns .......................................................................................................... [LUNs: 1]
  |     | o- lun0 .......................................... [fileio/device.avocado (/var/tmp/avocado_iscsi.img) (default_tg_pt_gp)]
  |     o- portals .................................................................................................... [Portals: 1]
  |       o- [::0]:3260 ....................................................................................................... [OK]
  o- loopback ......................................................................................................... [Targets: 0]
  o- vhost ............................................................................................................ [Targets: 0]
  o- xen-pvscsi ....................................................................................................... [Targets: 0]
```

I'm not sure if this is the correct way but that's how it's represented in the targetcli output

kraxel: Ok, seems to be some `targetcli` thing.  Guess we have to deal with it then.

kraxel (approved):

harihare (approved):
LGTM

PR #4373, opened 2026-06-23
libvirt_xml: add VMOSXML support for uefi-vars varstore element

hholoubk:
## Summary

Libvirt 12.1+ enlarged the domain `<os>` XML for the new uefi-vars firmware
model on aarch64 (and optionally x86): besides `<loader>` and `<nvram>`,
guests can now use a `<varstore template="..." path="..."/>` element for
UEFI variable storage (e.g. Secure Boot with `vars.secboot.json`).

This PR extends avocado-vt's `VMOSXML` bindings so tests can build and
verify that XML via `setup_attrs()` / `varstore_attrs`, matching the
existing `nvram_attrs` pattern.

Changes:
- Add `varstore_attrs` accessor for `<varstore template="..." path="..."/>`
- Add `loader_format` accessor for the loader `format` attribute
- Update `VMXML.undefine()` to pass `--keep-nvram` when varstore is present
- Update `remove_bootconfig_items_from_vmos()` to remove stale `<varstore>`
  before applying new firmware XML
- Document enlarged `<os>` elements in `VMOSXML` docstring

## Test plan

- [ ] Used locally with tp-libvirt `guest_os_booting.ovmf_firmware_feature`
      aarch64 secure-boot / custom varstore template variants
- [ ] `varstore_attrs` round-trip: `setup_attrs(os={...})` produces
      `<varstore template='...'/>` in dumpxml
- [ ] Existing nvram-based OVMF tests unaffected

### DIFF
--- a/virttest/libvirt_xml/vm_xml.py
+++ b/virttest/libvirt_xml/vm_xml.py
@@ -826,8 +826,12 @@ class VMXML(VMXMLBase):
     def undefine(self, options=None, virsh_instance=base.virsh):
         """Undefine this VM with libvirt retaining XML in instance"""
         os_attrs = self.os.fetch_attrs()
-        nvram = any([os_attrs.get("os_firmware") == "efi", os_attrs.get("nvram")])
-        if nvram:
+        has_uefi_vars = any([
+            os_attrs.get("os_firmware") == "efi",
+            os_attrs.get("nvram"),
+            os_attrs.get("varstore_attrs"),
+        ])
+        if has_uefi_vars:
             nvram_option = "--keep-nvram"
             if options is None:
                 options = nvram_option
@@ -3528,7 +3532,7 @@ class VMOSXML(base.LibvirtXMLBase):
     Elements:
         os:           list attributes - firmware
         type:         text attributes - arch, machine
-        loader:       path
+        loader:       path  attributes - readonly, type, format, stateless, secure
         boots:        list attributes - dev
         bootmenu:          attributes - enable, timeout
         smbios:            attributes - mode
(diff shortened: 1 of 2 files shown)


kraxel:
Looks like the patch also (correctly) adds attributes for `loader` and `nvram` which are older than uefivars support added in libvirt 12.1.  Specifically the `secure` attribute for `loader` exists more than 10 years already I think.  The `format` attributes already exist a few years too.  The commit message should note this correctly.

kraxel:
@andreabolognani

hholoubk:
and for some reason there is (again) missing the Sign-off and also the Assisted-by: as I heavily involved the AI to resolve this. @kraxel  thank you for pointing it out. I will do the update. 
I was going to check if the suggestion of Gemini shouldn't be applied also elsewhere where this approach can be historically missing.

hholoubk:
hm .. I am now not sure if it will work ... converting in draft (which it should be at the beginning ... 

two points .. for me it looks like it is mixin nvram element with varstore_attrs

(and for the secure attribute .. it was already there ... the update just moved it togehter for better readability.

I am going to test it proper way.

kraxel:
If this turns into some larger cleanup it makes sense to also split this into multiple patches.

hholoubk:
it looks like it is not ... :) I just didin't fully understood all of it ... but still I want to run it properly again

andreabolognani:
I'm not sure why the `loader_format` attribute was added. It doesn't seem to be used anywhere. Unless the goal is to have 100% coverage of the XML domain format, I would leave it out.

Other than that, the changes make semantic sense to me. I'm not familiar enough with the codebase to comment on the finer details of the implementation.

PR #4385, opened 2026-07-08
virttest: add wrapper for await in virsh.py

aniket-sahu-ibmx:
await is a feature added in libvirt 11.7.0, as a way to pause scripts until a certain state change is achieved by the VM. It currently supports only two states, domain-inactive and guest-agent-available.


pevogam (approved):
LGTM

PR #4388, opened 2026-07-23
Extend Avocado VT with support for multiple mutually networked vms

pevogam:
This introduces all functionality previously available in the Avocado I2N into Avocado VT plugin since the former merely extends the latter with more sophisticated scheduler and multi-vm VT tests of arbitrary network topology (incl. vm state derivation).

From the Avocado I2N description:

Plugins for avocado extending avocado-vt with automated vm state setup, inheritance, and traversal


pevogam:
Ok, after one and a half month of being open for all humans and multiple reviews by various AI-s (some external to the pipeline here) this PR is finally ready for merging!

PR #4390, opened 2026-07-27
cpu: remove vnmi from EPYC-Genoa CPU_TYPES_RE

imammedo:
EPYC-Genoa does not include vnmi in its QEMU CPU model definition.
Remove it from CPU_TYPES_RE so cpu_info_check doesn't fail trying
to match a nonexistent property.

Signed-off-by: Igor Mammedov <imammedo@redhat.com>


imammedo:
Duplicate of #4387 which is already merged.

PR #4391, opened 2026-07-27
cpu: use alternative QEMU property names for AMD ssbd/stibp/ibrs

imammedo:
QEMU uses 'amd-ssbd', 'amd-stibp', 'amd-ibrs' property names for AMD
CPU models, while Intel uses 'ssbd', 'stibp', 'ibrs'. The CPU_TYPES_RE
entries for EPYC-Milan and EPYC-Rome used the Intel names, causing
cpu_info_check to fail when query-cpu-model-expansion returns the AMD
property names.

Use the pipe alternative syntax (already supported by get_patterns) to
accept both naming conventions.

Tested on AMD EPYC 9335 (Turin) and EPYC 7302P (Rome) with RHEL 9.9.

Signed-off-by: Igor Mammedov <imammedo@redhat.com>


kraxel:
Why support both?  The amd cpus should never have the intel names, no?

imammedo:
@kraxel while they look similar, they are different cpuid bits/leafs form hw/qemu POV.
kernel when displaying them in /proc/cpuinfo conflates them into the one flag. And to make matters fun, kvm exposes both intel and amd cpuid bits/leaf regardless of vendor.

That's said, the patch is mere fix up to make test work and put out fires. All these definitions probably belong to qemu-tp instead of avocado-vt, but that's for future refactoring.

pevogam:
@imammedo Can I rely that you would open an issue here regarding the suggested refactoring and separation of concerns between this and qemu-tp and would follow up with the suggested refactoring resolving that issue? That would be great to lower more technical debt.

imammedo:
@pevogam, I guess I will do that eventually when dealing with cpu models regression testing, at which time this will become obsolete so I'll clean it up then.

pevogam:
> @pevogam, I guess I will do that eventually when dealing with cpu models regression testing, at which time this will become obsolete so I'll clean it up then.

Yes my point is if you could open an issue here so we can keep track of the added technical debt or need of clean up. Maybe you could even self-assign so that we can both keep track of this conversation here.

pevogam (approved):
LGTM, follow up in https://github.com/avocado-framework/avocado-vt/issues/4393.

PR #4394, opened 2026-07-31
vt_utils: add CPU codename detection

bssrikanth:
Add a new utility function that maps CPU family and model to an AMD platform codename (milan, genoa, bergamo, turin, venice). Model ranges are sourced from arch/x86/kernel/cpu/amd.c in the upstream Linux kernel.

## Summary by CodeRabbit

* **New Features**
  * Added automatic CPU model detection from system information.
  * Added AMD CPU codename identification for Milan, Genoa, Bergamo, Turin, and Venice generations.
  * Added validation to confirm systems meet a required minimum CPU generation.
  * Added clear error reporting for unsupported vendors, unrecognized CPU models, invalid codenames, and systems below the required generation.

### DIFF
--- a/virttest/vt_utils/cpu.py
+++ b/virttest/vt_utils/cpu.py
@@ -130,6 +130,101 @@ def get_cpu_family():
     return cpu_family
 
 
+def get_cpu_id():
+    """
+    Return the cpu model id.
+
+    :return: the id of the cpu model in string
+    :rtype: string
+    :raises: An OSError will be raised if it's NOT available on the platform
+    """
+    cpu_model_re = r"(?m)^model\s+:\s+(\d+)$"
+    with open("/proc/cpuinfo") as fd:
+        cpu_info = fd.read()
+    cpu_model = re.search(cpu_model_re, cpu_info)
+    if not cpu_model:
+        raise OSError("The cpu model id was NOT found!")
+    cpu_model_id = cpu_model.groups()[0]
+    return cpu_model_id
+
+
+# AMD CPU codename mappings ordered from oldest to newest.
+# Each entry is (codename, family, [(model_min, model_max), ...], generation).
+# Codenames that share a generation (e.g. genoa and bergamo) are treated as
+# equivalent for minimum-generation checks.  Ranges sourced from
+# arch/x86/kernel/cpu/amd.c (upstream Linux kernel).
+_AMD_CPU_CODENAMES = [
+    ("milan", 25, [(0, 15)], 3),
+    ("genoa", 25, [(16, 31)], 4),
+    ("bergamo", 25, [(160, 175)], 4),
+    ("turin", 26, [(0, 47), (64, 79), (96, 127), (208, 215)], 5),
+    ("venice", 26, [(80, 95), (128, 175), (192, 207), (216, 239)], 6),
+]
+
+
+def get_cpu_codename(vendor=None):
+    """
+    Return the microarchitecture codename for the host CPU.
+
+    :param vendor: optional vendor_id override (defaults to get_cpu_vendor_id())
+    :type vendor: str or None
+    :return: detected CPU codename
+    :rtype: str
+    :raises NotImplementedError: if codename detection is not implemented for
+        the vendor
+    :raises OSError: if family/model is unrecognised
+    """
+    vendor = vendor or get_cpu_vendor_id()
+    if vendor != "AuthenticAMD":
+        raise NotImplementedError(
+            f"CPU codename detection is not implemented for vendor '{vendor}'."
+        )
+
+    family = int(get_cpu_family())
+    model = int(get_cpu_id())
+    for codename, fam, ranges, _generation in _AMD_CPU_CODENAMES:
+        if fam == family and any(lo <= model <= hi for lo, hi in ranges):
+            return codename
+    raise OSError(f"Unrecognised AMD CPU: family={family}, model={model}.")
+
+
+def verify_min_cpu_codename(min_codename, vendor=None):
+    """
+    Verify the host CPU codename meets a minimum generation requirement.
+
+    :param min_codename: minimum required codename (e.g. "milan")
+    :param vendor: optional vendor_id override (defaults to get_cpu_vendor_id())
+    :return: detected CPU codename when the requirement is met
+    :raises NotImplementedError: if codename detection is not implemented for
+        the vendor
+    :raises ValueError: if min_codename is not a recognised codename for the
+        vendor
+    :raises OSError: if family/model is unrecognised or the host CPU is older
+        than min_codename
+    """
+    min_codename = min_codename.lower()
+    generations = {
+        codename: generation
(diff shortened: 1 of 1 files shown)


pevogam (commented):
Hi @bssrikanth, I wonder about how this related to the pre-existing functionality. For instance we don't have a separate `get_intel_platform` and all other functions are particular to CPU. So why "platform" for this function? Could it be integrated better into the pre-existing API?

PraveenPenguin:
+1 to @pevogam's concern. The existing API is strictly CPU attribute getters (family, model, stepping, vendor_id, flags). This function does something qualitatively different: it combines multiple attributes and applies
vendor-specific business logic to return a codename.

How about this as a generalise slightly to get_cpu_codename(vendor=None) that dispatches on vendor_id, so Intel equivalents can be added later without a parallel function explosion.

bssrikanth:
@pevogam  @PraveenPenguin  Thank you for the review comments. I have tried to implement the changes in line with your feedback. Could you please review this revision and let me know if it meets your expectations? Thank you.

virttest/vt_utils/cpu.py
```diff
@@ -130,6 +130,108 @@ def get_cpu_family():
     return cpu_family
 
 
+def get_cpu_model():
```
pevogam: What is the difference between this and `get_cpu_model_name()`?

virttest/vt_utils/cpu.py
```diff
@@ -130,6 +130,108 @@ def get_cpu_family():
...
+    with open("/proc/cpuinfo") as fd:
+        cpu_info = fd.read()
+    cpu_model = re.search(cpu_model_re, cpu_info)
+    if not cpu_model:
+        raise OSError("The cpu model was NOT found!")
+    cpu_model = cpu_model.groups()[0]
+    return cpu_model
+
+
+def _amd_cpu_codenames():
```
pevogam: Maybe this can be stored as list + dictionary or any JSON structure instead? Not sure why it has to be stored as a function instead.

virttest/vt_utils/cpu.py
```diff
@@ -130,6 +130,108 @@ def get_cpu_family():
...
+    :param vendor: optional vendor_id override (defaults to get_cpu_vendor_id())
+    :type vendor: str or None
+    :return: detected CPU codename
+    :rtype: str
+    :raises OSError: if the vendor is unsupported or family/model is unrecognised
+    """
+    vendor = vendor or get_cpu_vendor_id()
+    if vendor == "AuthenticAMD":
+        return _lookup_amd_cpu_codename(int(get_cpu_family()), int(get_cpu_model()))
+    raise OSError(f"CPU codename detection is not supported for vendor '{vendor}'.")
```
pevogam: How about NotImplementedError instead?

virttest/vt_utils/cpu.py
```diff
@@ -130,6 +130,108 @@ def get_cpu_family():
...
+    :return: detected CPU codename when the requirement is met
+    :raises ValueError: if min_codename is not a recognised codename for the
+        vendor
+    :raises OSError: if the vendor is unsupported, family/model is
+        unrecognised, or the host CPU is older than min_codename
+    """
+    min_codename = min_codename.lower()
+    generations = {
+        codename: generation
+        for codename, _fam, _ranges, generation in _amd_cpu_codenames()
```
pevogam: This still doesn't generalize to other vendors, please throw a NotImplementedError in all other cases.

pevogam (changes requested):
Hi @bssrikanth, this is a step in the right direction but still needs some work at least from a brief look. My comments are included below.

virttest/vt_utils/cpu.py
bssrikanth: get_cpu_model() returns the model [numeric ID] which is then used with  get_cpu_family() for codename mapping where as get_cpu_model_name returns the Model name which is the human-readable name.

virttest/vt_utils/cpu.py
bssrikanth: Sure will take care in the next revision

virttest/vt_utils/cpu.py
bssrikanth: Sure will take care in the next revision

virttest/vt_utils/cpu.py
bssrikanth: Sure will take care in the next revision

bssrikanth:
@pevogam please let me know if the revised version is in line with your review comments.

virttest/vt_utils/cpu.py
pevogam: I think it wasn't clear that I am not just asking this to know but also because the current naming is not good enough. So please rename this to `get_cpu_id` instead.

virttest/vt_utils/cpu.py
```diff
@@ -130,6 +130,116 @@ def get_cpu_family():
...
+_AMD_CPU_CODENAMES = [
+    ("milan", 25, [(0, 15)], 3),
+    ("genoa", 25, [(16, 31)], 4),
+    ("bergamo", 25, [(160, 175)], 4),
+    ("turin", 26, [(0, 47), (64, 79), (96, 127), (208, 215)], 5),
+    ("venice", 26, [(80, 95), (128, 175), (192, 207), (216, 239)], 6),
+]
+
+
+def _lookup_amd_cpu_codename(family, model):
```
pevogam: We don't need this for a few lines that could be added back into the get_cpu_codename implementation.

virttest/vt_utils/cpu.py
```diff
@@ -130,6 +130,116 @@ def get_cpu_family():
...
+    :return: detected CPU codename when the requirement is met
+    :raises NotImplementedError: if codename detection is not implemented for
+        the vendor
+    :raises ValueError: if min_codename is not a recognised codename for the
+        vendor
+    :raises OSError: if family/model is unrecognised or the host CPU is older
+        than min_codename
+    """
+    vendor = vendor or get_cpu_vendor_id()
+    if vendor != "AuthenticAMD":
```
pevogam: You could use similar style of raising the NotImplementedError above to this one. Meaning you could just check the CPU model at the beginning of `get_cpu_codename` and raise in a similar way to this, then move on in the same way towards the regular logic for the function.

pevogam (commented):
Hi @bssrikanth, just a few minor comments now + the one that was not yet addressed above.

bssrikanth:
> Hi @bssrikanth, just a few minor comments now + the one that was not yet addressed above.

Thanks for your time, @pevogam  I have addressed all the review comments now. Please let me know if this version is okay with you.

pevogam (approved):
Ok, LGTM! In the future I would still advise you if possible to instead contribute such code directly to [aautils](https://github.com/avocado-framework/aautils/blob/main/autils/system/cpu.py) if possible or even help us migrate more of the functionality in this util there since this is the ultimate destination for such functions.

PR #4396, opened 2026-08-01
Handle syntax warnings detectable in the newest Python 3.14

pevogam:
Add some isolation (unit) tests to verify the misc util functions affected return correct results on different exceptions.


pevogam:
@PraveenPenguin This handles all newly detected syntax warnings by python 31.4 while also adding additional unit test coverage for the modified cases.

PraveenPenguin (approved):
LGTM , Thanks @pevogam  for the patch

